PR #10, opened 2026-02-12
Feature/high dpi awareness

### LINKED ISSUE #9: [FEATURE] Make the application "High DPI" aware
## Problem Statement

When I open the GUI and navigate to any tab with a chart, this is what I see.

Lite:
Image

Dashboard:

Image

It's unreadable.
This happens due to a lack of "High DPI" awareness.

## Proposed Solution

Add native High DPI awareness. 
Recognises per monitor.

NOTE: I will submit a PR.

## Use Case
Better GUI without the need to set extra options.

Expected after the change:

Image

## Alternatives Considered

A possible workaround is editing the properties of the executable.
Go to `Compatibility` -> `Change high DPI settings` -> `Override high DPI scaling behavior`

Image

## Additional Context

- Is this related to a specific SQL Server version?
No
- Would this require schema changes?
No
- How frequently would you use this feature?
Always!

## Scope

Please indicate which component(s) this affects:

- [x] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

**IMPORTANT: If you're going to contribute code, please read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md) first.**

ClaudioESSilva:
## What does this PR do?

Fixes #9 

## Claude summary
I used the help of my cousin Claude.

Here is the summary of the changes made.

> ### Changes Made:
> 1. Created `app.manifest`
> 
> - Added PerMonitorV2 DPI awareness declaration
> - Configured the application to be DPI-aware at the manifest level
> - Set long path awareness for compatibility
> 
> 
> 2. Updated `PerformanceMonitorLite.csproj`
> Added `<ApplicationManifest>app.manifest</ApplicationManifest>` to include the manifest in the build
> - Updated `App.xaml.cs`
> - Added P/Invoke declarations for Windows DPI APIs
> - Created `EnableDpiAwareness()]` method that programmatically sets DPI awareness
> - Added call to `EnableDpiAwareness()` at the start of `OnStartup()` before any UI is created
> - Tries PerMonitorV2 mode first (Windows 10 1703+) for best quality
> - Falls back to PerMonitor mode (Windows 8.1+) if V2 is not available
> 
> ### How It Works:
> The application now properly declares itself as per-monitor DPI aware in two ways:
> 
> 1. Manifest-level declaration for system-wide DPI handling
> 2. Programmatic DPI awareness setting for runtime configuration
> 
> This means:
> 
> ✅ Charts and text will automatically scale correctly on high DPI displays
> ✅ No need to manually adjust executable properties anymore
> ✅ Works properly on multi-monitor setups with different DPI settings
> ✅ The application adapts when moved between monitors with different DPI
> ✅ Font sizes remain readable at any DPI scale (100%, 125%, 150%, 200%, etc.)
> 

- [x] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

### Lite

Before:
Image

After:
image

### Dashboard

Before:
Image

After:
Image

## Checklist

### DIFF
--- a/Dashboard/App.xaml.cs
+++ b/Dashboard/App.xaml.cs
@@ -8,7 +8,6 @@
 
 using System;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
@@ -22,33 +21,8 @@ namespace PerformanceMonitorDashboard
         private const string MutexName = "PerformanceMonitorDashboard_SingleInstance";
         private Mutex? _singleInstanceMutex;
 
-        // DPI awareness for proper scaling on high DPI displays
-        private enum PROCESS_DPI_AWARENESS
-        {
-            Process_DPI_Unaware = 0,
-            Process_System_DPI_Aware = 1,
-            Process_Per_Monitor_DPI_Aware = 2
-        }
-
-        private enum DPI_AWARENESS_CONTEXT
-        {
-            Unaware = -1,
-            SystemAware = -2,
-            PerMonitorAware = -3,
-            PerMonitorAwareV2 = -4
-        }
-
-        [DllImport("SHCore.dll", SetLastError = true)]
-        private static extern bool SetProcessDpiAwareness(PROCESS_DPI_AWARENESS awareness);
-
-        [DllImport("user32.dll", SetLastError = true)]
-        private static extern bool SetProcessDpiAwarenessContext(int dpiFlag);
-
         protected override void OnStartup(StartupEventArgs e)
         {
-            // Enable per-monitor DPI awareness for proper scaling on high DPI displays
-            EnableDpiAwareness();
-
             // Check for existing instance
             _singleInstanceMutex = new Mutex(true, MutexName, out bool isNewInstance);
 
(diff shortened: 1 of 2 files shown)


erikdarlingdata (commented):
Thanks for the PR and for filing the issue — the High DPI fix is a real improvement and the before/after screenshots demonstrate the problem clearly.

**The manifest files are great.** The `app.manifest` approach with `<dpiAwareness>PerMonitorV2</dpiAwareness>` is the [Microsoft-recommended way](https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process) to declare DPI awareness for WPF apps. The `.csproj` references are correct. This is the right solution.

**The P/Invoke code in both `App.xaml.cs` files should be removed.** Here's why:

1. The Windows loader processes the manifest during `CreateProcessInternalW`, *before* the CLR initializes or any managed code runs. By the time `OnStartup()` calls `EnableDpiAwareness()`, PerMonitorV2 awareness is already set.

2. Calling `SetProcessDpiAwareness()` after the manifest has already set DPI awareness returns `E_ACCESSDENIED` ([documented here](https://learn.microsoft.com/en-us/windows/win32/api/shellscalingapi/nf-shellscalingapi-setprocessdpiawareness)): *"The DPI awareness is already set, either by calling this API previously or through the application (.exe) manifest."* Similarly, `SetProcessDpiAwarenessContext()` returns `FALSE` with `ERROR_ACCESS_DENIED` ([documented here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiawarenesscontext)).

3. The `try/catch` blocks won't catch these failures — P/Invoke calls return error codes, they don't throw .NET exceptions. The catches would only fire for `DllNotFoundException` or `EntryPointNotFoundException`.

4. There's also a subtle marshaling issue: `SetProcessDpiAwareness` returns `HRESULT` (not `BOOL`), but the P/Invoke declares it as `bool`. `E_ACCESSDENIED` (0x80070005) would marshal as `true`, so even if you checked the return value, it would look like success.

**In short:** the manifests do all the work. The ~60 lines of P/Invoke per app (enums, DllImports, `EnableDpiAwareness()`) are dead code that silently fails every time. Removing them keeps the fix clean and correct.

If you'd like, I can trim the P/Invoke code on my end after merge — or you can update the PR to include only the manifest changes. Either way works. The manifests themselves are good to go.

ClaudioESSilva:
Done!
Compiled, tested, and it's good to go 😄

PR #291, opened 2026-02-25
Improve experience when adding server - Save tests the connection

### LINKED ISSUE #29: [FEATURE] Behaviour when saving a connection
## Problem Statement

Currently, it's possible to "Save" a connection that isn't working. 
Example: Type a wrong password or the password box is empty

## Proposed Solution

I would like it when I hit the "Save" button that:
- The connection is tested; 
- If it doesn't work, the regular error message should be shown and an additional text saying something like "Because we can't connect, if you want to save it, you need to uncheck the 'Enable data collection for this server' option." - I may want to come back later and assume it won't be working for now.

Pro:
- This will make it possible to save a connection only when it is working.
- Otherwise, we need to uncheck the box, and the dot color will be "red" (not collecting).

## Scope

Please indicate which component(s) this affects:

- [x] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

**IMPORTANT: If you're going to contribute code, please read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md) first.**

ClaudioESSilva:
## What does this PR do?

Fixes #29 

- [x] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

## How was this tested?

Tested on the Dashboard version by using the wrong password and hit "Save" button:
image

For the Lite version, this is how it will show
image

Now, on both versions, when testing the connection, the "Save" button also becomes disabled.

## Checklist

>[!Important]
>It has some warnings `CA1873`, but that isn't related to anything I've changed.
> For reference, I forked and updated from a recent `dev` branch

### DIFF
--- a/Dashboard/AddServerDialog.xaml
+++ b/Dashboard/AddServerDialog.xaml
@@ -51,29 +51,40 @@
                 <!-- Authentication Type -->
                 <TextBlock Text="Authentication:" FontWeight="Bold" Margin="0,0,0,5"
                            Foreground="{DynamicResource ForegroundBrush}"/>
-                <StackPanel Orientation="Horizontal" Margin="0,0,0,15">
+                <StackPanel Margin="0,0,0,15">
                     <RadioButton x:Name="WindowsAuthRadio" Content="Windows Authentication"
                                  GroupName="Auth" IsChecked="True"
                                  Foreground="{DynamicResource ForegroundBrush}"
-                                 Checked="AuthType_Changed" Margin="0,0,20,0"/>
+                                 Checked="AuthType_Changed" Margin="0,0,0,4"/>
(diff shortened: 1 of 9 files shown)


ClaudioESSilva:
@erikdarlingdata, for "Dashboard should support Entra MFA authentication". 

Should we test the connection against the `InitialCatalog = "PerformanceMonitor"`? 
Otherwise, it can give a false positive because it can reach the server, but then it won't be able to access the specific database and will show a red status.

image

If you agree with it, that is the current state of this PR.
If not, let me know and I can adjust.

PR #329, opened 2026-02-26
Add Server List Default Sort to Full Dashboard

HannahVernon:
## What does this PR do?

This PR adds default sorting for the list of servers in the Full Dashboard.  Servers will automatically be sorted by the Display Name property.

- [✔] Full Dashboard

## How was this tested?

I compiled the Full Dashboard, added a bunch of servers with varying display names, and noticed the list of servers is automatically sorted by the display name.

## Checklist

- [✔] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md)
- [✔] My code builds with zero warnings (`dotnet build -c Debug`)
- [✔] I have tested my changes against at least one SQL Server version
- [✔] I have not introduced any hardcoded credentials or server names

### DIFF
--- a/Dashboard/Controls/LandingPage.xaml.cs
+++ b/Dashboard/Controls/LandingPage.xaml.cs
@@ -82,6 +82,9 @@ namespace PerformanceMonitorDashboard.Controls
                 _serverHealthStatuses.Add(new ServerHealthStatus(server));
             }
 
+            // Sort servers alphabetically by display name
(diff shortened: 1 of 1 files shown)


Dashboard/MainWindow.xaml.cs
```diff
@@ -24,6 +24,8 @@
 using PerformanceMonitorDashboard.Controls;
 using PerformanceMonitorDashboard.Helpers;
 using PerformanceMonitorDashboard.Services;
+using System.ComponentModel;
```
HannahVernon: Visual Studio 2022 automatically added these two `Using` statements - presumably they are needed by the `OrderBy(...)` code I added below.

PR #362, opened 2026-02-27
Added exclusions for sp_server_diagnostics and WAITFOR wait types to GetLongRunningQueriesAsync() method.

HannahVernon:
## What does this PR do?

Adds exclusions to `GetLongRunningQueriesAsync()` to exclude `sp_server_diagnostics` and `WAITFOR` wait types.  Adds parameters to `GetLongRunningQueriesAsync()` for future use to allow precise control over exclusion of those wait types, along with an optional parameter to control the maximum number of long running queries returned to the caller.

- [✔] Full Dashboard

## How was this tested?

Tested against SQL Server 2022 with sp_server_diagnostics wait types caused by Windows Server Failover Cluster monitoring, and via a manual query with `WAITFOR DELAY '00:30:00';`

## Checklist

- [✔] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md)
- [✔] My code builds with zero warnings (`dotnet build -c Debug`)
- [✔] I have tested my changes against at least one SQL Server version
- [✔] I have not introduced any hardcoded credentials or server names

### DIFF
--- a/Dashboard/Controls/LandingPage.xaml.cs
+++ b/Dashboard/Controls/LandingPage.xaml.cs
@@ -7,7 +7,7 @@
  */
 
 using System;
-//using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
--- a/Dashboard/Services/DatabaseService.NocHealth.cs
+++ b/Dashboard/Services/DatabaseService.NocHealth.cs
@@ -603,29 +603,24 @@ namespace PerformanceMonitorDashboard.Services
         /// Gets currently running queries that exceed the duration threshold.
         /// Uses live DMV data (sys.dm_exec_requests) for immediate detection.
         /// </summary>
-        private async Task<List<LongRunningQueryInfo>> GetLongRunningQueriesAsync(SqlConnection connection, int thresholdMinutes, bool excludeSpServerDiagnostics = true, bool excludeWaitFor = true, int maxLongRunningQueryCount = 5, bool excludeBackupWaits = true)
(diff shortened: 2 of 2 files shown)


erikdarlingdata (changes requested):
Thanks for this — filtering by `r.wait_type` instead of query text is a much better approach, and the sp_server_diagnostics exclusion is a real need for WSFC/AG environments.

A few things to address:

### WAITFOR RECEIVE regression
The old code filtered `t.text NOT LIKE N'%waitfor receive%'` which caught Service Broker waits. The new `r.wait_type <> N'WAITFOR'` doesn't cover that — Service Broker uses `BROKER_RECEIVE_WAITFOR` as its wait type. Please add that to the exclusions so we don't regress.

### Add XE_LIVE_TARGET_TVF exclusion
Please also exclude `XE_LIVE_TARGET_TVF` — Extended Events live target reads show up as long-running and aren't actionable.

### Unused parameters
The `excludeSpServerDiagnostics`, `excludeWaitFor`, `excludeBackupWaits`, and `maxLongRunningQueryCount` parameters aren't used by any caller. Let's keep the method signature simple — just apply the exclusions unconditionally. If we need configurability later, we can add it then.

### Minor code issues
- `if (maxLongRunningQueryCount <= 5)` clamps to minimum 5 — if the intent is to prevent invalid values, this should be `<= 0` or `< 1`
- `string LongRunningQueryCount` — PascalCase for a local variable (should be camelCase per C# conventions)
- Trailing `};` after the if block (semicolon unnecessary)
- The `//using System.Collections.Generic;` comment-out in `LandingPage.xaml.cs` is unrelated to this PR — please revert that file

Overall the direction is good, just needs some cleanup. Thanks!

erikdarlingdata:
@HannahVernon Fine as-is, thank you again for the contribution! You will be remembered forever as an Open Source Heroine.

PR #408, opened 2026-03-04
Add items to the Settings Window for Long Running Queries alerts

HannahVernon:
Adds config options to control various aspects of GetLongRunningQueriesAsync() method.
Which component(s) does this affect?

    [✅] Full Dashboard
    [✅] Lite Dashboard

How was this tested?

Tested against SQL Server 2019 and SQL Server 2022 standalone instances.

Checklist

    [✅] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md)
    [✅] My code builds with zero warnings (dotnet build -c Debug)
    [✅] I have tested my changes against at least one SQL Server version
    [✅] I have not introduced any hardcoded credentials or server names

### DIFF
--- a/Lite/Services/LocalDataService.WaitStats.cs
+++ b/Lite/Services/LocalDataService.WaitStats.cs
@@ -220,6 +220,7 @@ LIMIT 3";
         string miscWaitsFilter = excludeMiscWaits
             ? "AND r.wait_type NOT IN (N'XE_LIVE_TARGET_TVF')" : "";
 
+
         command.CommandText = @$"
                 SELECT
                     r.session_id,


erikdarlingdata (commented):
Nice work Hannah — clean implementation, both apps stay in sync, and defaults preserve backward compatibility.

A few notes:

### 1. `maxResults` clamp is a no-op

`Math.Clamp(maxResults, 1, int.MaxValue)` doesn't do anything since an `int` can never exceed `int.MaxValue`. Same with the Dashboard validation check `lrqMaxResults <= int.MaxValue` — always true. Should probably cap at something reasonable like 100 or 1000 so someone doesn't accidentally type 999999999 and blow up their alert check.

### 2. `maxResults` could be parameterized

The value is interpolated directly into the query string (`SELECT TOP({maxResults})` in Dashboard, `LIMIT {maxResults}` in Lite). It's an `int` so it's safe in practice, but parameterizing it would be more consistent with how `thresholdMinutes` is handled.

### 3. PR branch

Head and base are both `dev` — this will need to come from a feature branch to actually merge.

### 4. Minor

Extra blank line in `LocalDataService.WaitStats.cs` after the filter assignments (around line 226).

PR #427, opened 2026-03-04
Feature/add alert filter by database

HannahVernon:
## What does this PR do?

Adds the ability to filter certain alert features by database name:
 - Deadlocks
 - Long Running Queries

- [✅] Full Dashboard
- [✅] Lite Dashboard

## How was this tested?

Tested deadlock alerts with a classic "lock resources in opposite order" deadlock test in a dedicated database.  With the database name included in the exclusion list, the alert is not fired when a deadlock occurs.  With the database not present in the exclusion list, the deadlock alert *is* fired as expected.

Tested long running queries with the execution-database  both in the list and not in the list - exclusion works as expected.

## Checklist

- [✅] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md)
- [✅] My code builds with zero warnings (`dotnet build -c Debug`)
- [✅] I have tested my changes against at least one SQL Server version
- [✅] I have not introduced any hardcoded credentials or server names

### DIFF
--- a/Dashboard/SettingsWindow.xaml.cs
+++ b/Dashboard/SettingsWindow.xaml.cs
@@ -313,6 +313,7 @@ namespace PerformanceMonitorDashboard
             LongRunningQueryThresholdTextBox.Text = "30";
             TempDbSpaceThresholdTextBox.Text = "80";
             LongRunningJobMultiplierTextBox.Text = "3";
+            AlertExcludedDatabasesTextBox.Text = "";
             UpdateAlertPreviewText();
         }
 
--- a/Lite/Windows/SettingsWindow.xaml.cs
+++ b/Lite/Windows/SettingsWindow.xaml.cs
@@ -551,6 +551,7 @@ public partial class SettingsWindow : Window
         AlertLongRunningQueryThresholdBox.Text = "30";
         AlertTempDbSpaceThresholdBox.Text = "80";
         AlertLongRunningJobMultiplierBox.Text = "3";
+        AlertExcludedDatabasesBox.Text = "";
         UpdateAlertPreviewText();
     }
 


HannahVernon:
@erikdarlingdata - here's a summary of the changes I made today:

Four issues addressed across three commits:

  ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  1. Dashboard deadlock alert firing on unfiltered count (f41d087)

  `GetDeadlockCountAsync` reads from `sys.dm_os_performance_counters`, which is server-wide and cannot be filtered by database. Added a new `GetFilteredDeadlockCountAsync` method that instead queries `collect.blocking_deadlock_stats` — summing
  deadlock_count_delta over a 5-minute window (matching AlertCooldown), excluding configured databases. When excluded databases are configured, `EvaluateAlertConditionsAsync` uses this filtered count (effectiveDeadlockDelta) for the
  threshold check, tray notification, RecordAlert, and email currentValue. The raw perf counter tracking (_previousDeadlockCounts) is left untouched.

  ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  2. Lite missing deadlock and blocking filtering (5e41b5d)

  `CheckPerformanceAlerts` was using summary.BlockingCount and summary.DeadlockCount (both unfiltered server-wide counts) directly against thresholds. Fixed to mirror the existing long-running query pattern:

   - Blocking: when excluded databases are configured and the raw count is at or above threshold, fetches `GetRecentBlockedProcessReportsAsync` and filters in-memory by DatabaseName. Uses effectiveBlockingCount for the threshold check,
  tray notification, and email value.
   - Deadlocks: same approach using `GetRecentDeadlocksAsync`, filtered via a new IsDeadlockExcluded helper that parses each deadlock's XML and excludes a deadlock only when every process node's currentdbname is in the excluded list
  (cross-database deadlocks still fire).
   - BuildBlockingContextAsync and BuildDeadlockContextAsync also now filter by excluded databases so email alert bodies are consistent with tray notifications.

  ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  3. Restore safety cap on alert_long_running_query_max_results (746591a)

  The Math.Clamp upper bound had been accidentally changed from 1000 to int.MaxValue. Restored to 1000.

  ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

  4. Settings help text updated (781069f)

  Both Dashboard and Lite `SettingsWindow.xaml` hint text updated from "applies to long-running query and blocking alerts" to "applies to blocking, deadlock, and long-running query alerts" to accurately reflect all three alert types.

erikdarlingdata (commented):
## Review

Nice feature — the parameterized queries, the conservative cross-database deadlock XML logic, and the null fallback in `GetFilteredDeadlockCountAsync` are all well done.

### Bug: Dashboard `BuildDeadlockContextAsync` missing empty check after filtering

Dashboard's `BuildBlockingContextAsync` correctly returns `null` when all events are filtered out:
```csharp
if (events.Count == 0) return null;
```

But `BuildDeadlockContextAsync` does **not** check after filtering — it just proceeds with the (potentially empty) list:
```csharp
deadlocks = deadlocks
    .Where(d => !IsDeadlockExcluded(d, excludedDatabases))
    .ToList();
// Missing: if (deadlocks.Count == 0) return null;
```

The Lite version has this check. Without it, a timing gap between `GetFilteredDeadlockCountAsync` (count > 0) and `GetDeadlocksAsync` (all detail rows filtered out) could send an email with empty deadlock context. **Please add the guard before merging.**

### Minor notes (non-blocking)

1. **Double-fetch in Lite alert path** — when excluded databases are configured, blocking/deadlock rows are fetched once for the effective count, then fetched *again* in `BuildBlockingContextAsync`/`BuildDeadlockContextAsync`. Not a real problem since DuckDB is local, but the filtered list could be passed down to avoid the second trip.

2. **Repeated filtering pattern** — the case-insensitive exclusion check (`!excludedDatabases.Any(e => string.Equals(e, q.DatabaseName, ...))`) appears in ~6 places. A small helper like `IsDatabaseExcluded(string? dbName, List<string> excluded)` would reduce duplication.

3. **No Trim on Lite JSON load** — `alert_excluded_databases` values are taken as-is from settings.json. If someone hand-edits the JSON with spaces (`"  msdb "`), the comparison would fail. The save path trims, but the load path doesn't — adding `.Trim()` on load would close the gap:
   ```csharp
   if (!string.IsNullOrWhiteSpace(db)) AlertExcludedDatabases.Add(db.Trim());
   ```

HannahVernon:
As far as I can tell, `BuildDeadlockContextAsync()` *does* check for `deadlocks.Count == 0` at line 1468 (the last line below):

```
            try
            {
                var deadlocks = await databaseService.GetDeadlocksAsync(hoursBack: 1);
                if (deadlocks == null || deadlocks.Count == 0) return null;
```

So, if deadlocks are null or the count of items in the deadlocks collection is 0, return null.  I checked the branch code at https://github.com/HannahVernon/PerformanceMonitor/blob/feature/add-alert-filter-by-database/Dashboard/MainWindow.xaml.cs - it looks like the change is there.

I'm out of town this weekend so won't get a chance to look at this until next week.

PR #483, opened 2026-03-09
Fix PerfMon first-value spike that skews graphs

### LINKED ISSUE #482: [BUG] Lite - PerfMon first values spike skews graph
### Component

Lite

2.1.0

### SQL Server Version

SQL Server 2022 (RTM-CU16)

### Windows Version

Windows 11 23H2

### Describe the Bug

1. Under the Server Tab -> Perfmon the initial/first values that load creates a spike on the graph, which causes a skew on the graph. 
    - This seems incorrect as subsequent values are significantly smaller than the initial values.
    - It seems to be the aggregation of values shown for the first timestamp. 
2. The initial graph's X-axis (timestamp) initializes towards the right, whereas I was expecting it to load more towards the origin axis.
    - Most other graphs starts the timestamp on the origin (to the left) on the X-axis.

### Steps to Reproduce

1. Open Lite Edition
2. Open Server Tab
3. Select Perfmon Tab
4. Wait for collection to complete
**- First values loaded seems incorrectly aggregated on first timestamp**

7. Select different Counter Group (i.e. Memory Pressure, CPU / Compilation, etc.)
**- First values loaded seems incorrectly aggregated on first timestamp**

### Expected Behavior

1. Initial / first values loaded for the graph should only be for the timestamp at which it ran, not an aggregation of some kind for that initial timestamp.
2. Initial timestamp on the X-axis load close to \ on origin to the left, like most other graphs.

### Actual Behavior

1. Initial / first values loaded for the graph seems to be an aggregation of some kind that is shown on the initial timestamp.
2. Initial timestamp on the X-axis loads towards the end (right) of the X-axis.

```text
No error messages.
```

### Screenshots

**_All Counters:_**

Image

**_Memory Pressure:_**

Image

**_CPU / Compilation:_**

Image

**_IO Pressure:_**

Image

**_TempDB Pressure:_**

Image

**_Lock / Blocking:_**

Image

### Additional Context

I'm only using Lite at the moment, so not sure whether this would apply to the Full edition as well.

erikdarlingdata:
## Summary
- Fixes #482
- Added `baselineOnly` parameter to `DeltaCalculator.CalculateDelta()` — when `true`, the first read stores the baseline and returns `0` instead of the raw cumulative value
- Applied to all cumulative-counter collectors: PerfMon, WaitStats, FileIO, MemoryGrants
- Query/proc stats unchanged (single-execution queries still surface in top-N views)
- X-axis anchoring issue was a visual side effect of the spike dominating the Y scale

## Test plan
- [ ] Start Lite, open Server > PerfMon tab
- [ ] Verify first data point does not create a spike for any counter group
- [ ] Verify chart scales normally after 2+ collections
- [ ] Verify WaitStats, FileIO, MemoryGrants tabs also have no first-value spike
- [ ] Verify Top Queries / Top Procs still show single-execution entries

### DIFF
--- a/Lite/Services/DeltaCalculator.cs
+++ b/Lite/Services/DeltaCalculator.cs
@@ -65,7 +65,7 @@ public class DeltaCalculator
     /// Counter reset (value decreased): returns 0 to avoid inflated deltas from plan cache churn.
     /// Thread-safe via atomic AddOrUpdate.
     /// </summary>
-    public long CalculateDelta(int serverId, string collectorName, string key, long currentValue)
+    public long CalculateDelta(int serverId, string collectorName, string key, long currentValue, bool baselineOnly = false)
     {
         var serverCache = _cache.GetOrAdd(serverId, _ => new ConcurrentDictionary<string, ConcurrentDictionary<string, long>>());
         var collectorCache = serverCache.GetOrAdd(collectorName, _ => new ConcurrentDictionary<string, long>());
@@ -74,11 +74,12 @@ public class DeltaCalculator
 
         collectorCache.AddOrUpdate(
             key,
-            /* Add: first time seeing this key — use current value as delta
-               so queries that execute once still surface in top-N views */
+            /* Add: first time seeing this key.
+               baselineOnly = true: store baseline only, return 0 (for cumulative counters like perfmon).
+               baselineOnly = false: use current value as delta so single-execution queries surface. */
             _ =>
             {
-                delta = currentValue;
+                delta = baselineOnly ? 0 : currentValue;
                 return currentValue;
             },
             /* Update: compute delta atomically */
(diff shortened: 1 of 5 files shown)


dphugo:
Hi @erikdarlingdata. Wow, what a rapid response! 🚀
Not sure if I should respond here or on the original Issue I raised (482) ?:

I noticed the scope was expanded to the following files, which is great:
image


However, I've picked up that _deltaCalculator.CalculateDelta is also used in: 
- Services\RemoteCollectorService.ProdecureStats.cs 
- Services\RemoteCollectorService.QueryStats.cs 
image

These are currently not included in the commit and (for graphing) it does affect:
- "Server Tab -> Queries -> Performance Trends" as can be seen in this screenshot:
image

As the scope was expanded by this commit, I figured it would makes sense to include it on this branch?
Or, because it is closed/merged, should I raise a separate bug for the above increased scope?

Kind regards,
Daniel

PR #679, opened 2026-03-24
Fix Azure SQL DB server_id collision (#677)

erikdarlingdata:
Fixes #677 — Azure SQL DB connections to different databases on the same logical server shared the same server_id, causing CPU and other database-scoped DMV data to interleave.

GetServerNameForStorage() now appends /DatabaseName when set, so myserver.database.windows.net/Db1 and myserver.database.windows.net/Db2 hash to different server_ids.

Existing data under the old hash becomes orphaned (acceptable — it was mixed/incorrect anyway).

## Test plan
- Builds clean (no new warnings)
- On-prem connections without DatabaseName unaffected (empty/null check)
- ReadOnlyIntent suffix still appended correctly after database name
- MCP ServerResolver still resolves by ServerName/DisplayName (unaffected)

PR #680, opened 2026-03-24
Fix CPU metrics mixed for different Azure SQL databases on same server

### LINKED ISSUE #677: [BUG] CPU metrics mixed for different Azure SQL databases on the same logical server
### Component

Lite

2.4.1

### SQL Server Version

Microsoft SQL Azure

### Windows Version

Windows 11 Enterprise

## Description

When two Azure SQL Database connections share the same logical server (e.g. `myserver.database.windows.net`), their CPU utilization data is stored under the same `server_id` because `GetServerNameForStorage()` only uses `ServerName` for hashing.

This means CPU data from different databases (e.g. Db1 and Db2) gets interleaved.

  ## Root Cause

  `GetServerNameForStorage()` in `RemoteCollectorService.cs` returns only `server.ServerName`, ignoring `server.DatabaseName`. For Azure SQL DB,
  `sys.dm_db_resource_stats` is database-scoped, but both connections hash to the same `server_id`.

  ## Expected Behavior

  Each Azure SQL Database connection should produce a unique `server_id` so CPU data is tracked per-database.

### Steps to Reproduce

1. Add two Azure SQL Database connections on the same logical server with different databases
2. Connect to both to start collecting CPU data

### Expected Behavior

Under FinOps > Utilization, the two Azure SQL Databases should have different CPU stats (Avg, P95, Max) 

### Actual Behavior

Under FinOps > Utilization, the two Azure SQL Databases have the same CPU stats (Avg, P95, Max) 

```text

```

### Screenshots

_No response_

### Additional Context

_No response_

tnguyen7s:
## Summary   
                                                                                                                                                    
- `GetServerNameForStorage()` used only the server hostname to generate storage keys, so multiple Azure SQL databases on the same logical server produced identical `server_id`s and their metrics were conflated
- Include `DatabaseName` in the storage key when set, following the existing pattern used for `:RO` (ReadOnlyIntent) connections

Fixes #677

## Test plan
  - [x] Build compiles without warnings (there are 3 warnings, CS8602: Dereference of a possibly null reference, that appear to happen before the changes)
  - [x] Connect two Azure SQL databases on the same logical server with different database names
  - [x] Wait for initial collection (5 minutes)
  - [x] Verify FinOps > Utilization shows separate CPU stats (Average, P95, Maximum) for each database
  - [x] Code follows the style guidelines 
  - [x] No hardcoded paths or credentials
  - [x] Changes are documented (comments, README if applicable)


erikdarlingdata:
Great fix and thorough testing! We had independently landed on the same approach on our side (#679). One minor suggestion — would you mind switching the separator from `:` to `/` so the storage key reads `server/database:RO` instead of `server:database:RO`? Keeps the `:RO` suffix unambiguous. Either way, thanks for the contribution and the real-world Azure SQL DB validation!

Issue #712, opened 2026-03-26
[BUG] config.InstallationHistory - [sql_server_version] column length


jakemorgangit:
### Component

Installer (GUI)

### Performance Monitor Version

2.4.1

### SQL Server Version

Microsoft SQL Server 2019 (RTM-CU16-GDR) (KB5014353) - 15.0.4236.7 (X64)  	May 29 2022 15:55:47  	Copyright (C) 2019 Microsoft Corporation 	Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor)

### Windows Version

Windows Server 2019

### Describe the Bug

The [sql_server_version] column in [config].[installation_history] is currently limited to NVARCHAR(255).

However, some SQL Server version strings exceed this length, which results in the value being truncated or not inserted at all. This causes the dashboard to show no version history.

 Example
=========

In my case, the version string is ~267 characters, exceeding the current limit of 255:

```
Microsoft SQL Server 2019 (RTM-CU16-GDR) (KB5014353) - 15.0.4236.7 (X64) 
    May 29 2022 15:55:47 
    Copyright (C) 2019 Microsoft Corporation
    Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor)
```

Impact
=========
Version information is not stored in [config].[installation_history]
Installer runs roll up scripts from v1
Dashboard thinks my version is out of date and required an update.

### Steps to Reproduce

Install via PerformanceMonitor Installer, deployed against a SQL Server version with version info > 255 chars

### Expected Behavior

Installer succeeds without errors, adding version info to [config].[installation_history] 

### Actual Behavior

Character limit is exceeded, meaning version info can't be written to  [config].[installation_history]  

### Error Messages / Log Output

```text

```

erikdarlingdata:
@jakemorgangit ooh nice catch! Did you want to submit a fix for this?

erikdarlingdata:
Fixed in PR #722. Widened sql_server_version and sql_server_edition to nvarchar(512). Upgrade script in 2.4.0-to-2.5.0 folder.

PR #716, opened 2026-03-26
Widen installation_history sql_server_versioncolumn 

jakemorgangit:
Increase sql_server_version to nvarchar(512) to prevent truncation of long SQL Server version strings

## What does this PR do?

Fixes - https://github.com/erikdarlingdata/PerformanceMonitor/issues/712

Updates config.installation_history to increase sql_server_version from nvarchar(255) to nvarchar(512).

Some SQL Server version and edition strings can exceed 255 characters. This change prevents truncation and ensures full version information is retained.

- [ ] Full Dashboard
- [ ] Lite Dashboard
- [ ] Lite Tests
- [ ] SQL collection scripts
- [X] CLI Installer
- [X] GUI Installer
- [ ] Documentation

## How was this tested?
Tested against SQL Server 2019.

Steps:

1. Applied updated schema via GUI installer and CLI installer script
2. Inserted values using @@VERSION and SERVERPROPERTY('Edition')
3. Verified no truncation occurs for long version strings
4. Confirmed existing data migrates successfully during table recreation

## Checklist

- [X] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md)
- [X] My code builds with zero warnings (`dotnet build -c Debug`)
- [X] I have tested my changes against at least one SQL Server version
- [X] I have not introduced any hardcoded credentials or server names


erikdarlingdata (changes requested):
Thanks for tackling this — the upgrade script and schema change are solid. A couple of things to address before merge:

**1. Remove the ALTER TABLE block from `01_install_database.sql` (lines 644–672)**

This project's convention is that schema changes for existing installations go in `upgrades/` scripts only, not in the install script. Your `CREATE TABLE` on line 623 already has `nvarchar(512)`, so new installs are covered. The upgrade script at `upgrades/2.4.1-to-2.4.2/` handles existing installs. The ALTER TABLE block in the install script is redundant and goes against the project pattern — please remove it.

**2. Branch hygiene**

The PR head ref is `main`, which means this was committed directly to your fork's main branch. That pulls in unrelated upstream merge commits (there are 4 in the commit list). For future PRs, creating a feature branch (e.g. `fix/widen-version-column`) keeps the history clean. For this one, we can squash-merge, but a rebase onto a feature branch would be ideal.

**3. Minor: commit message vs. actual change**

The first commit message says "Increase sql_server_version **and sql_server_edition** to nvarchar(512)" but the diff only widens `sql_server_version`. `sql_server_edition` stays at `nvarchar(255)`. That's probably fine (edition strings are short), but the commit message is misleading.

Everything else looks good — the upgrade script is properly idempotent with correct SET options, the `upgrade.txt` is in place, and the test helper is updated. Just need item #1 fixed and this is ready to go.

Issue #718, opened 2026-03-26
[QUESTION]


vpnicholls:
### Which component is your question about?

Lite

### Performance Monitor Version

2.4.1, 2.3.0

### Is your question about how it works, or the results?

How the tool works

### What's your question?

Just wondering if anyone has noticed this behaviour. I've scanned through other questions and issues but haven't seen anything.

After installing 2.4.1, I could dismiss alerts, either select alert or all alerts. At some stage there seems to be a tipping point where I neither of the dismiss buttons works. If I click the dismiss all button, then the Dismiss All Alerts dialog box appears. I click OK, then nothing.

I can't figure out the pattern. The underlying cause for each has since resolved. The alerts are for two different instances (no Azure SQL Databases) and two different Alert Types (Blocking Detected and Long-Running Query). I can't recall if there were any other types previously. When it was working, there were fewer alerts, about 4 or 5. When I first noticed it not working there were about 20-30. Right now, I have 35 alerts over 30 hours that I can't dismiss. 

I've checked the logs, but I can't see anything around the relevant times when I attempt to dismiss the alerts.

This is happening for me in v2.4.1, but I first noticed this in 2.3.0.

erikdarlingdata:
@vpnicholls there may be some tie-in to this and the work done by @HannahVernon on the alert muting in #512 and #642. This code path is somewhat hard for me to test thoroughly with my limited home environment set up.

HannahVernon:
I'm investigating.

vpnicholls:
Thanks @erikdarlingdata and @HannahVernon. I'll have a read of #512 and #642 .

HannahVernon:
@erikdarlingdata  - I've added alert dismissal / mute logging into a new fix branch which I'll push to you shortly for review.  Before I do that, I wanted to present some options we could implement to make Alerts more feature rich and diagnosable, both for the user and for those supporting the product.  Take a gander below, and let me know if you think we should implement everything, or just certain things, or nothing at all  ☘️🦈🚀

# Issue #718 — Alert Dismiss Stops Working: Root Cause Analysis & Suggested Fixes

## Root Cause

The Lite dismiss flow has a **read/write mismatch** between the view and the live table:

- **`GetAlertHistoryAsync`** reads from `v_config_alert_log` — a view that unions the live `config_alert_log` table with archived parquet files via `UNION ALL BY NAME SELECT * FROM read_parquet(glob)`
[...]

PR #730, opened 2026-03-27
Add dismissed_archive_alerts sidecar table for parquet-aware dismissal (#718)

HannahVernon:
## What does this PR do?

Addresses issue #718 where alert dismissal silently fails for archived alerts. The dismiss flow targets the live `config_alert_log` table, but alerts archived to parquet files are read-only  `UPDATE` returns 0 affected rows with no user feedback.

This PR adds a **sidecar table** (`dismissed_archive_alerts`) that tracks which archived alerts have been dismissed, and modifies the `v_config_alert_log` view to filter them out.

### Changes

| File | Change |
|------|--------|
| `Lite/Database/Schema.cs` | New `dismissed_archive_alerts` table (alert_time, server_id, metric_name, dismissed_at) with composite index |
| `Lite/Database/DuckDbInitializer.cs` | Schema version 2223 with migration; `v_config_alert_log` parquet half now excludes rows present in sidecar via `NOT EXISTS` |
| `Lite/Services/LocalDataService.AlertHistory.cs` | `DismissAlertsAsync`: when UPDATE returns 0 rows, falls back to sidecar INSERT with duplicate prevention. `DismissAllVisibleAlertsAsync`: after live UPDATE, inserts remaining undismissed archived alerts into sidecar |
| `Lite.Tests/DuckDbSchemaTests.cs` | Updated table count assertion from 28 to 29 |
| `Lite.Tests/Helpers/TestAlertDataHelper.cs` | Reusable test utility for inserting live alerts and creating archived parquet files |
| `Lite.Tests/DismissedArchiveSidecarTests.cs` | 7 new xUnit tests |

### How it works

**Before:** Archived alerts appear in the grid (via the view's parquet union) but `UPDATE config_alert_log` can never reach them. Clicking Dismiss silently does nothing.

**After:** When `DismissAlertsAsync` detects a 0-rows-affected UPDATE, it INSERTs the alert's key into `dismissed_archive_alerts`. The view's parquet half uses `NOT EXISTS` against this sidecar table to filter them out. The archived alert disappears from the grid on the next refresh.

### Design decisions

- **No ALTER TABLE**  the sidecar is a new table, not a column addition. No DuckDB NULL-on-upgrade issues.
- **Not in ArchivableTables**  the sidecar is a small metadata table (one row per dismissed archive alert), not time-series data. It does not need its own archival cycle.
- **Duplicate prevention**  all INSERT paths use `WHERE NOT EXISTS` to prevent duplicate sidecar entries.
[...]

PR #733, opened 2026-03-27
fix: dismiss reliability - write lock, batched UPDATE, transaction, timeout (#718)

HannahVernon:
## What does this PR do?

Addresses write lock, transaction wrapping, and batch updates from the issue #718 improvement list. This makes alert dismissal reliable under concurrent archival and prevents UI freezes.

**Key changes:**

- **Write lock for dismiss operations** — `DismissAlertsAsync` and `DismissAllVisibleAlertsAsync` now acquire an exclusive write lock instead of a read lock, preventing race conditions with archival/compaction
- **Batched UPDATE** — `DismissAlertsAsync` sends a single `UPDATE ... WHERE (alert_time, server_id, metric_name) IN (VALUES ...)` instead of looping, reducing round-trips and lock hold time
- **Transaction wrapping** — `DismissAlertsAsync` uses `BEGIN/COMMIT/ROLLBACK` for all-or-nothing semantics
- **Timeout on write lock** — `AcquireWriteLock(TimeSpan?)` now supports an optional timeout (5 seconds for UI paths) to prevent indefinite UI freeze when archival holds the lock
- **User-friendly error handling** — both dismiss button handlers catch `TimeoutException` and show a "database busy, try again" message instead of silently failing
- **LockedConnection cleanup** — updated class documentation and field naming to reflect that it handles both read and write locks

- [ ] Full Dashboard
- [x] Lite Dashboard
- [x] Lite Tests
- [ ] SQL collection scripts
- [ ] CLI Installer
- [ ] GUI Installer
- [ ] Documentation

## How was this tested?

- 7 new xUnit v3 tests in `DismissReliabilityTests.cs` covering:
  - Batched UPDATE dismisses multiple alerts in a single statement
  - Correct affected count when some alerts are already dismissed
  - Transaction rollback restores undismissed state
  - Transaction commit persists dismissed state
  - Write lock exclusivity blocks concurrent acquisition
  - Timeout throws `TimeoutException` instead of blocking indefinitely
  - DismissAll UPDATE works correctly under write lock
- Full test suite: **225 tests, all passing**
- Build: **0 errors, 0 new warnings** (pre-existing CS8602 only)

## Checklist


erikdarlingdata:
Hey Hannah — we merged #729, #730, and #734 into dev. This PR (#733) needs a rebase onto current dev to reconcile the batched UPDATE/write lock/transaction changes with the sidecar fallback from #730 and the structured telemetry from #734. The dismiss methods in `LocalDataService.AlertHistory.cs` have changed significantly since this branch was last synced.

Specifically, `DismissAlertsAsync` now has the sidecar INSERT fallback and structured logging — the batched UPDATE rewrite here needs to preserve both of those. Same for `DismissAllVisibleAlertsAsync` with the sidecar bulk INSERT.

No rush — the existing dismiss flow works correctly, this PR adds reliability improvements on top. Thanks for the great work on this series!

HannahVernon:
@erikdarlingdata this is now rebased

HannahVernon:
> Batched UPDATE — DismissAlertsAsync sends a single UPDATE ... WHERE (alert_time, server_id, metric_name) IN (VALUES ...)

In SQL Server, I'd frown on using an `IN (...)` construct, but I'm not sure what options we have for DuckDB; perhaps batching in stages if there are more than 'x' number of alerts being dismissed.  Not sure what 'x' should be; perhaps that should be tested.

erikdarlingdata:
Agreed in principle, but in practice the dismiss count is going to be small — there's a low chance anyone will dismiss more than 100 alerts in a go. If it ever becomes a concern we can chunk it, but not worth the complexity right now.

Issue #735, opened 2026-03-27
[BUG] Lite - FinOps CPU right-sizing uses node-level cpu_count instead of per-database vCores on Azure SQL Database


kontrago:
### Component

Lite

### Performance Monitor Version

2.3.0

### SQL Server Version

Microsoft SQL Azure (Hyperscale)

### Windows Version

Windows 11

### Describe the Bug

On Azure SQL Database, the FinOps CPU right-sizing recommendation incorrectly reports databases as overprovisioned because it reads `cpu_count` from `sys.dm_os_sys_info`. On Azure SQL DB, this DMV returns the CPU count of the **underlying compute node**, not the per-database vCore allocation.

For example, three databases on the same logical server with vCore allocations of 6, 14, and 14 all report `cpu_count = 20` because they share the same compute node. The FinOps module then uses 20 as the baseline for all three, producing misleading "cut down on vCores" recommendations.

This does **not** affect on-premises SQL Server or Azure SQL Managed Instance, where `sys.dm_os_sys_info.cpu_count` accurately reflects the instance's CPU allocation.

### Root Cause

`sys.dm_os_sys_info.cpu_count` on Azure SQL Database reflects the compute replica's total hardware capacity, which is shared across all databases on the same logical server. The actual per-database vCore provisioning is set via service objectives (e.g., `HS_PRMS_6`, `HS_PRMS_14`) but is not exposed through `sys.dm_os_sys_info`.

The correct source for per-database vCore count on Azure SQL Database is `sys.database_service_objectives`, where the `service_objective` column encodes the vCore count (e.g., `HS_PRMS_14` = 14 vCores).

### Proposed Fix

1. When `engine_edition = 5` (Azure SQL Database), query `sys.database_service_objectives` to get the `service_objective` string
2. Parse the vCore count from the service objective name (e.g., extract `14` from `HS_PRMS_14`)
3. Store the parsed value in a new `vcore_count` column in the `server_properties` DuckDB table
4. In FinOps queries, use `COALESCE(vcore_count, cpu_count)` so Azure SQL DB uses the correct per-database value while on-prem continues using `cpu_count` unchanged

### Steps to Reproduce

1. Add two or more Azure SQL Database connections on the same logical server with different vCore allocations
2. Wait for server properties collection
3. Navigate to FinOps tab
4. Observe that all databases show the same cpu_count (the compute node total) and may receive incorrect CPU right-sizing recommendations

### Expected Behavior

[...]

PR #738, opened 2026-03-27
Fix: skip tempdb_stats collector on Azure SQL DB

### LINKED ISSUE #737: Lite: tempdb_stats collector fails on Azure SQL DB with permission denied
## Description

The `tempdb_stats` collector in Performance Monitor Lite causes recurring collector errors when connected to Azure SQL Database instances.

**Error:**
```
VIEW DATABASE PERFORMANCE STATE permission denied in database 'tempdb'
```

## Root Cause

The `tempdb_stats` collector queries `tempdb.sys.dm_db_file_space_usage` and `sys.dm_db_session_space_usage`, which require cross-database access to `tempdb`. Azure SQL DB (engine edition 5) does not allow cross-database access to `tempdb`, so these queries fail with a permission error.

The collector is not listed in the Azure SQL DB blocklist in `IsCollectorSupported`, so it runs unconditionally on all platforms including Azure SQL DB.

## Affected Edition

**Lite only** — the Full Edition uses SQL Agent jobs installed on the server and is not affected.

## Proposed Fix

Add `"tempdb_stats"` to the Azure SQL DB edition gate in `IsCollectorSupported` (`Lite/Services/RemoteCollectorService.cs`), following the same pattern already used for `server_config`, `trace_flags`, and `running_jobs`.

This is a known Azure SQL DB limitation with no workaround — the DMVs simply aren't accessible.

kontrago:
## Summary

- Adds `tempdb_stats` to the Azure SQL DB edition gate in `IsCollectorSupported`, preventing the collector from running on Azure SQL DB (engine edition 5)
- Azure SQL DB does not allow cross-database access to `tempdb`, causing `VIEW DATABASE PERFORMANCE STATE permission denied in database 'tempdb'` errors on every collection cycle
- Follows the existing blocklist pattern already used for `server_config`, `trace_flags`, and `running_jobs`

Closes #737

## Test plan

- [x] `dotnet build Lite/PerformanceMonitorLite.csproj` — compiles with 0 errors
- [x] `dotnet test Lite.Tests/` — 225 tests passed, 0 failed
- [x] Live tested against Azure SQL Database — confirmed tempdb_stats collector no longer runs and the permission denied error is gone

### DIFF
--- a/Lite/Services/RemoteCollectorService.cs
+++ b/Lite/Services/RemoteCollectorService.cs
@@ -810,6 +810,7 @@ WHERE server_id = $3";
                 case "server_config":     /* sys.configurations not available */
                 case "trace_flags":       /* DBCC TRACESTATUS not available */
                 case "running_jobs":      /* msdb.dbo.sysjobs not available */
+                case "tempdb_stats":      /* cross-database tempdb access not available */
                     return false;
             }
         }


erikdarlingdata:
Thanks for the report and fix — the permission error is real and we were able to reproduce it on Azure SQL DB with non-admin users.

However, we don't think the blocklist approach is the right fix here. The existing error handling in `RemoteCollectorService.cs` (lines 426-431) already catches SQL errors 229/297/300 and marks the collector as `PERMISSIONS` status — which logs a warning but doesn't count as a failure. So the tempdb collector already handles this gracefully for non-admin users without any code changes.

The issue with the blocklist is that it also prevents server admin users on Azure SQL DB from getting tempdb stats — and we confirmed that `tempdb.sys.dm_db_file_space_usage` works fine for admin users (engine edition 5, GP Gen5 provisioned). Blanket-blocking the collector removes functionality that works for some users.

We're going to close this in favor of a different approach that checks permissions at runtime and collects what it can, rather than skipping the entire collector based on edition.

Thanks again for filing #735 and #737 — the Azure SQL DB coverage improvements are really valuable!

PR #858, opened 2026-04-17
Fall back to single-DB mode when Azure master is inaccessible (#857)

### LINKED ISSUE #857: [FEATURE] Review/lower required permission for the Azure server to user database level(no master)

- [ ] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

### Problem Statement

Thanks for the tool. I tried to run it (Lite version) for Microsoft Dynamics 365 FO, and currently, it seems the tool requires a full connection to the server itself rather than just to the database.
The issue is that Microsoft provides only a connection to the user database(with full admin rights)
This connection(that they provide) allows viewing TOP queries, deadlocks etc.. if used in the Management Studio 

They allow only exec **sp_set_database_firewall_rule** for the user DB , but not for the master database

But if run your Lite version, it gives that error 
Image

and the screen is empty

Image

But I can view TOP queries in Management Studio without master DB access using the same computer/credentials

Image

 

### Proposed Solution

Review queries that are really needed for master DB access and that require only user DB access. 
Execute some queries and provide results without master DB access

### Use Case

Monitor Databases without require full access to the server. Mostly related to Microsoft D365FO ERP.

I can help in testing if needed

### Alternatives Considered

_No response_

### Additional Context

_No response_

erikdarlingdata:
## Summary
- Fixes #857 — Lite failed and showed an empty screen when connecting to an Azure SQL DB login that lacks master access (common with Microsoft Dynamics 365 FO and other single-tenant logins).
- `GetAzureDatabaseListAsync` now catches known access-denied/login-failed SQL errors from the master connection, caches the per-server decision, and returns the connection's `InitialCatalog` as a single-element list.
- The three callers (`query_stats`, `database_size_stats`, `file_io_stats`) already iterate per-database — single-DB mode works without further changes.
- Users with master access retain today's full-server enumeration behavior.

## Behavior matrix
| InitialCatalog | Master accessible | Result |
|---|---|---|
| empty | yes | enumerate all DBs (unchanged) |
| empty | no | nothing collected, warning logged |
| set to a DB | yes | enumerate all DBs (unchanged) |
| set to a DB | no | collect from that one DB (new, fixes #857) |

## Test plan
- [ ] Connect Lite to an Azure SQL DB login with master access — enumeration still works, all DBs appear.
- [ ] Connect Lite to an Azure SQL DB login without master access and a specified target DB — no error popup, data collected for target DB only, info log line appears.
- [ ] Connect Lite to an Azure SQL DB login without master access and no target DB — warning log line appears, no crash.
- [ ] Confirm the fallback decision is cached: subsequent cycles don't retry master (check logs don't show repeated fallback messages).
- [ ] Boxed SQL Server and Azure MI unaffected (they don't call `GetAzureDatabaseListAsync`).

### DIFF
--- a/Lite/Services/RemoteCollectorService.cs
+++ b/Lite/Services/RemoteCollectorService.cs
@@ -102,6 +102,14 @@ public partial class RemoteCollectorService
     private int _logInsertFailures;
     private string? _lastLogInsertError;
 
+    /// <summary>
+    /// Per-server flag indicating that master DB enumeration has failed with an access-denied
+    /// error and should not be retried. Used on Azure SQL DB where per-database logins may not
+    /// have master access (e.g. Microsoft Dynamics 365 FO). See issue #857.
+    /// </summary>
+    private readonly Dictionary<int, bool> _azureMasterInaccessible = new();
+    private readonly object _azureMasterInaccessibleLock = new();
+
     public RemoteCollectorService(
         DuckDbInitializer duckDb,
         ServerManager serverManager,
@@ -561,10 +569,30 @@ WHERE server_id = $3";
     /// Enumerates online databases on an Azure SQL DB logical server.
     /// HAS_DBACCESS() returns false for user databases from master on Azure SQL DB,
     /// so we skip that filter — inaccessible databases should be handled by callers via try/catch.
+    ///
+    /// On Azure SQL DB, logins are sometimes granted access only to a specific user database and
+    /// not to master (e.g. Microsoft Dynamics 365 FO). In that case, master enumeration fails with
+    /// an access/login error; we fall back to returning the connection's initial catalog as a
+    /// single-database list, and cache that decision per server so we don't retry master each cycle.
+    /// See issue #857.
     /// </summary>
     protected async Task<List<string>> GetAzureDatabaseListAsync(ServerConnection server, CancellationToken cancellationToken)
     {
+        var serverId = GetServerId(server);
         var baseConnStr = server.GetConnectionString(_serverManager.CredentialService);
+        var targetDb = new SqlConnectionStringBuilder(baseConnStr).InitialCatalog;
+
+        bool knownInaccessible;
+        lock (_azureMasterInaccessibleLock)
+        {
+            _azureMasterInaccessible.TryGetValue(serverId, out knownInaccessible);
+        }
+
+        if (knownInaccessible)
+        {
+            return SingleDbOrEmpty(targetDb);
+        }
+
         var connStr = new SqlConnectionStringBuilder(baseConnStr)
         {
             ConnectTimeout = ConnectionTimeoutSeconds,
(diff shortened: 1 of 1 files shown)


Lite/Services/RemoteCollectorService.cs
```diff
@@ -102,6 +102,14 @@ public partial class RemoteCollectorService
     private int _logInsertFailures;
     private string? _lastLogInsertError;
 
+    /// <summary>
+    /// Per-server flag indicating that master DB enumeration has failed with an access-denied
+    /// error and should not be retried. Used on Azure SQL DB where per-database logins may not
+    /// have master access (e.g. Microsoft Dynamics 365 FO). See issue #857.
```
erikdarlingdata, line 108: Nit: the value is always `true` when the key is present — a `HashSet<int>` would express intent more clearly and avoid the misleading `TryGetValue(..., out knownInaccessible)` pattern where the `false` branch actually means "not yet determined" rather than "accessible". Not blocking.

---

Lite/Services/RemoteCollectorService.cs
```diff
@@ -561,27 +569,94 @@ INSERT INTO collection_log (log_id, server_id, server_name, collector_name, coll
...
+    private static bool IsMasterAccessDeniedError(SqlException ex)
+    {
+        return ex.Number switch
+        {
+            229   => true, // Permission denied on object
+            230   => true, // Permission denied on column
+            916   => true, // Server principal is not able to access the database under the current security context
+            4060  => true, // Cannot open database requested by the login
+            18456 => true, // Login failed for user
+            40613 => true, // Database 'master' on server is not currently available
```
erikdarlingdata, line 656: Two of these aren't really "master inaccessible" signals:

- **18456** (Login failed for user) — the login is wrong/disabled/locked. The per-DB connection opened later by `OpenAzureDatabaseConnectionAsync` will fail for the exact same reason.
- **40615** (Cannot open server — login denied, firewall/auth) — usually a firewall IP rule or tenant-auth issue. Again, per-DB connections will also fail.

Treating these as "fallback to single DB" masks a broken connection as "master inaccessible (SQL error 18456) — collecting from '…' only", and then every per-DB collector fails on the next line. Consider letting 18456/40615 bubble up as real connection errors, and reserving the fallback for errors that genuinely indicate master-only lockout (229/230/916/4060/40613).

Also worth noting: the error list only ever grows the inaccessibility cache; if a login is later granted master access, this process won't pick it up until restart. Intentional per the PR description, but worth a one-line doc note on `_azureMasterInaccessible`.

---

Lite/Services/RemoteCollectorService.cs
```diff
@@ -561,27 +569,94 @@ INSERT INTO collection_log (log_id, server_id, server_name, collector_name, coll
...
+            }
+            return fallback;
+        }
+    }
+
+    private static List<string> SingleDbOrEmpty(string? targetDb)
+    {
+        if (string.IsNullOrEmpty(targetDb) || string.Equals(targetDb, "master", StringComparison.OrdinalIgnoreCase))
+            return new List<string>();
+        return new List<string> { targetDb };
```
erikdarlingdata, line 640: Minor: comparing `targetDb` against `"master"` case-insensitively is good, but `tempdb`/`msdb`/`model` would also be meaningless fallback targets on Azure SQL DB (not that they'd normally appear in `InitialCatalog`, but defensive). More practically: what if `targetDb` is quoted/bracketed like `[mydb]`? `SqlConnectionStringBuilder.InitialCatalog` strips the brackets so this should be fine — just noting the assumption.

---

Lite/Services/RemoteCollectorService.cs
```diff
@@ -561,27 +569,94 @@ INSERT INTO collection_log (log_id, server_id, server_name, collector_name, coll
...
+            while (await reader.ReadAsync(cancellationToken))
+                databases.Add(reader.GetString(0));
+            return databases;
+        }
+        catch (SqlException ex) when (IsMasterAccessDeniedError(ex))
+        {
+            lock (_azureMasterInaccessibleLock)
+            {
+                _azureMasterInaccessible[serverId] = true;
+            }
```
erikdarlingdata, line 621: Two concurrent cycles racing here both open master, both fail, both grab the lock in turn, both set `[serverId] = true`, and both emit the `AppLogger.Info/Warn` line — so the first cycle after permission loss can double-log. Not a correctness problem, just cosmetic noise in the log on the transition. If you want to suppress the second log, check-and-set under a single lock and only log when the flag flips from unset to set.

---

erikdarlingdata (commented):
## Review: #858 — Azure master fallback

### What this does
On Azure SQL DB, `GetAzureDatabaseListAsync` now catches specific SQL errors indicating master is inaccessible, caches that decision per server under a lock, and returns the connection's `InitialCatalog` as a single-element list. The three per-database callers (`query_stats`, `database_size_stats`, `file_io_stats`) already iterate per DB, so they work unchanged. Server-level collectors (which don't call this method) are unaffected.

### What's good
- Scope is right: Lite-only change, single file, narrowly targeted at the issue.
- Base branch is `dev`. No schema, install, upgrade, or workflow changes.
- No PlanAnalyzer or Dashboard/Lite-duplicated code touched — sync rule N/A.
- Exception filter is specific (`when (IsMasterAccessDeniedError(ex))`), so unrelated `SqlException`s still bubble up rather than being swallowed.
- `using` disposal for `SqlConnection`/`SqlCommand`/`SqlDataReader` is correct.
- Cache is behind a lock; no `async void`; no cross-thread UI work.
- Log message distinguishes the info-level "collecting from X only" case from the warn-level "nothing to collect" case.

### What needs attention
1. **Error list is too broad** (`IsMasterAccessDeniedError`, [RemoteCollectorService.cs#L649-L656](https://github.com/erikdarlingdata/PerformanceMonitor/blob/3e950a5c68c8014ead31be3847e5332e9a02a4f4/Lite/Services/RemoteCollectorService.cs#L649-L656)): 18456 and 40615 are generic auth/firewall failures, not "master-only" lockouts. When the connection is outright broken, the fallback cascades into failed per-DB opens but logs a misleading "master inaccessible" message first. Consider narrowing to 229/230/916/4060/40613. Inline comment has detail.
2. **Cache never clears** ([RemoteCollectorService.cs#L105-L108](https://github.com/erikdarlingdata/PerformanceMonitor/blob/3e950a5c68c8014ead31be3847e5332e9a02a4f4/Lite/Services/RemoteCollectorService.cs#L105-L108)): Regranting master requires a service restart to pick up. Intentional per the PR description; worth a one-line XML-doc note so it isn't a surprise later.
3. **`Dictionary<int, bool>` reads as tri-state** but isn't — value is always `true` when present. `HashSet<int>` would be clearer.
4. **Double-log on concurrent first failure** — two racing cycles will each emit the Info/Warn line once. Cosmetic, but the fix is a single check-and-set under the lock.
5. **No tests added.** `Lite.Tests` has no `RemoteCollectorService` coverage at all, so this is consistent with existing patterns rather than a regression, but the fallback logic (error-code dispatch, `SingleDbOrEmpty`, cache behavior) is pure and testable without a live Azure endpoint. Worth adding at least unit coverage for `IsMasterAccessDeniedError` and `SingleDbOrEmpty` if you want a safety net for future tuning of the error list.

### Test plan check
The manual test plan in the PR description covers the right matrix (master accessible / not × target DB set / not) and explicitly checks the cache, which is the non-obvious part. No automated coverage to run.

Nothing blocking from my side — maintainer's call.

---

PR #864, opened 2026-04-19
Fix Overview crosshair disappearing after tab switches / layout passes

erikdarlingdata:
## Summary
Crosshair on the Overview tab's correlated timeline lanes was working on first open but disappearing after some combination of auto-refreshes, tab switches, or layout events. Diagnosed via logging: the manager's lane list was going from 5 to 0 unexpectedly.

**Root cause:** the control had `Unloaded += ...Dispose()` on the crosshair manager. WPF fires `Unloaded` for transient reasons (tab virtualization, layout rebuilds) — not just when the control is truly going away. `Dispose()` clears the lane list, after which `ReattachVLines` runs over an empty list and the crosshair is permanently gone.

## Changes
- **Remove the `Unloaded → Dispose()` handler** in both Lite and Dashboard copies. The manager holds only managed state (a Popup + lane references) — GC will clean it up with the control.
- **Remove the redundant `_isRefreshing` flag** from `CorrelatedCrosshairManager`. The `lane.VLine == null` check in `OnMouseMove` is a sufficient "not ready" guard and is self-healing once VLines are recreated.
- **Wrap `ReattachVLines` in a try/finally** on the control side, with a new idempotent `EnsureVLinesAttached()` safety net that only creates VLines for lanes where they're still null.
- **Per-lane exception handling** in `CreateVLine` so one failing chart can't prevent the others from recovering.

## Test plan
- [x] Builds clean (Lite + Dashboard)
- [x] Crosshair survives multiple auto-refresh cycles on Lite (verified with logging before commit)
- [ ] Crosshair survives tab switches (Overview → Queries → back to Overview)
- [ ] Crosshair works on manual time range changes

### DIFF
--- a/Dashboard/Controls/CorrelatedTimelineLanesControl.xaml.cs
+++ b/Dashboard/Controls/CorrelatedTimelineLanesControl.xaml.cs
@@ -31,7 +31,12 @@ public partial class CorrelatedTimelineLanesControl : UserControl
     public CorrelatedTimelineLanesControl()
     {
         InitializeComponent();
-        Unloaded += (_, _) => _crosshairManager?.Dispose();
+        /* No Unloaded → Dispose() handler: WPF fires Unloaded for transient
+           reasons (tab virtualization, layout rebuilds) and Dispose() clears
+           the crosshair manager's lane list, permanently breaking the crosshair
+           until the ServerTab is rebuilt. The manager holds only managed state
+           (a Popup + lane references) — letting GC clean it up with the control
+           is fine. */
     }
 
     /// <summary>
@@ -68,6 +73,9 @@ public partial class CorrelatedTimelineLanesControl : UserControl
         if (_dataService == null) return;
 
         _crosshairManager?.PrepareForRefresh();
+
+        try
+        {
 
         var cpuTask = _dataService.GetCpuUtilizationAsync(hoursBack, fromDate, toDate);
         var waitTask = _dataService.GetTotalWaitStatsTrendAsync(hoursBack, fromDate, toDate);
@@ -225,8 +233,18 @@ public partial class CorrelatedTimelineLanesControl : UserControl
             _crosshairManager?.SetComparisonLabel(ComparisonLabel(comparisonRange.Value, fromDate, hoursBack));
         }
 
+        /* VLines must be re-attached before SyncXAxes so they're part of
+           the render set when the chart refreshes. */
         _crosshairManager?.ReattachVLines();
         SyncXAxes(hoursBack, fromDate, toDate);
+        }
+        finally
+        {
+            /* Safety net: if something threw between PrepareForRefresh() and the
+               ReattachVLines() call above, VLines are still null. EnsureVLinesAttached
+               creates them only for lanes where VLine is null, so it's idempotent. */
+            _crosshairManager?.EnsureVLinesAttached();
+        }
     }
 
     /// <summary>
--- a/Dashboard/Helpers/CorrelatedCrosshairManager.cs
+++ b/Dashboard/Helpers/CorrelatedCrosshairManager.cs
@@ -33,7 +33,6 @@ internal sealed class CorrelatedCrosshairManager : IDisposable
     private readonly Popup _tooltip;
     private readonly TextBlock _tooltipText;
     private DateTime _lastUpdate;
-    private bool _isRefreshing;
 
     public CorrelatedCrosshairManager()
     {
@@ -144,10 +143,11 @@ internal sealed class CorrelatedCrosshairManager : IDisposable
 
     /// <summary>
     /// Clears data and VLines. Call before re-populating charts.
+    /// The OnMouseMove guard relies on lane.VLine == null to detect "not ready",
+    /// so this is self-healing: once ReattachVLines runs, crosshairs resume.
     /// </summary>
     public void PrepareForRefresh()
     {
-        _isRefreshing = true;
         _tooltip.IsOpen = false;
         _comparisonLabel = null;
         foreach (var lane in _lanes)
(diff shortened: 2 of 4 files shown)


Dashboard/Helpers/CorrelatedCrosshairManager.cs
```diff
@@ -162,25 +162,54 @@ public void PrepareForRefresh()
...
             vline.LinePattern = ScottPlot.LinePattern.Dashed;
             vline.IsVisible = false;
-            lane.VLine = vline;
+            return vline;
+        }
+        catch
+        {
+            /* If attach fails, return null so OnMouseMove skips this lane.
+               Next refresh will try again. */
+            return null;
```
erikdarlingdata, line 206: Silent catch-all. Everywhere else in `CorrelatedTimelineLanesControl.RefreshAsync` a `Debug.WriteLine(...)` accompanies `catch (Exception ex)` (e.g. line 116, 161, 198). If `Plot.Add.VerticalLine` ever starts throwing, the only signal will be a silently-missing crosshair on one lane — exactly the class of bug this PR exists to diagnose. At minimum `catch (Exception ex) { Debug.WriteLine($"CorrelatedCrosshairManager: VLine attach failed: {ex.Message}"); return null; }`.

---

Lite/Helpers/CorrelatedCrosshairManager.cs
```diff
@@ -162,25 +162,54 @@ public void PrepareForRefresh()
...
             vline.LinePattern = ScottPlot.LinePattern.Dashed;
             vline.IsVisible = false;
-            lane.VLine = vline;
+            return vline;
+        }
+        catch
+        {
+            /* If attach fails, return null so OnMouseMove skips this lane.
+               Next refresh will try again. */
+            return null;
```
erikdarlingdata, line 206: Same silent-catch issue as the Dashboard copy. Kept in sync, but fix belongs in both — log the exception instead of swallowing it. A VLine attach failure here is invisible except as a missing crosshair.

---

Dashboard/Controls/CorrelatedTimelineLanesControl.xaml.cs
```diff
@@ -225,8 +233,18 @@ public async Task RefreshAsync(int hoursBack, DateTime? fromDate, DateTime? toDa
             _crosshairManager?.SetComparisonLabel(ComparisonLabel(comparisonRange.Value, fromDate, hoursBack));
         }
 
+        /* VLines must be re-attached before SyncXAxes so they're part of
+           the render set when the chart refreshes. */
         _crosshairManager?.ReattachVLines();
         SyncXAxes(hoursBack, fromDate, toDate);
+        }
```
erikdarlingdata, line 240: The body of this `try` (lines 80-239) is not re-indented and there's a stray blank line at 79. Compiles fine, but the raw file now has ~160 lines sitting at outer-method indent inside a `try` block, which will confuse anyone reading the file without a diff view. I get that leaving it un-indented minimizes this diff — worth a follow-up tidy-up commit to re-indent the block so future readers don't do a double-take.

---

Dashboard/Helpers/CorrelatedCrosshairManager.cs
```diff
@@ -162,25 +162,54 @@ public void PrepareForRefresh()
 
     /// <summary>
     /// Creates fresh VLine plottables on each lane's chart.
-    /// Must be called AFTER chart data is populated.
+    /// Must be called AFTER chart data is populated. Safe to call in a finally
+    /// block — if chart state is invalid, a failure on one lane won't prevent
+    /// the others from recovering.
     /// </summary>
     public void ReattachVLines()
```
erikdarlingdata, line 169: Heads-up: after removing the `Unloaded → Dispose()` handler, nothing in the codebase calls `CorrelatedCrosshairManager.Dispose()` anymore (the class still implements `IDisposable` and keeps a `Dispose()` at line 371). That's now dead code. Two options for a follow-up: (a) wire `Dispose()` to a real teardown point such as `ServerTab` close so `_lanes.Clear()` actually runs when the tab goes away, or (b) drop `IDisposable` + `Dispose()` entirely and let GC handle it (matches the comment rationale on the constructor). Leaving `IDisposable` declared with no caller sets a false expectation for the next reader. Mirror applies in `Lite/Helpers/CorrelatedCrosshairManager.cs`.

---

erikdarlingdata (commented):
Review comments inline. Summary to follow as an issue comment.

---

erikdarlingdata:
## Review summary

**Note:** PR was already merged by the time this review ran — comments are for follow-up consideration, not blocking.

### What this PR does
Fixes Overview-tab crosshair disappearing after tab switches / layout passes. Root cause: `Unloaded += _crosshairManager.Dispose()` fires on transient layout events (tab virtualization), clears the lane list, and subsequent `ReattachVLines` iterates an empty list. Fix removes the Unloaded handler, drops the now-redundant `_isRefreshing` flag (the `lane.VLine == null` guard in `OnMouseMove` is sufficient), wraps refresh in try/finally with an idempotent `EnsureVLinesAttached()` recovery path, and isolates per-lane VLine creation with try/catch.

### What's good
- Fix applied to **both** `Dashboard/Helpers/CorrelatedCrosshairManager.cs` and `Lite/Helpers/CorrelatedCrosshairManager.cs` — copies stay in sync.
- Base branch is `dev`. ✓
- Diagnosis in the PR body is solid and the `EnsureVLinesAttached()` idempotency reasoning is correct.
- No schema / install-script / workflow changes, so upgrade-path and SignPath concerns don't apply.

### What needs attention
1. **Silent catch in `CreateVLine`** (`Dashboard/Helpers/CorrelatedCrosshairManager.cs:202-207`, mirror in Lite). Everywhere else in the control's refresh flow, `catch` is paired with `Debug.WriteLine`. Swallowing here hides exactly the class of failure this PR exists to diagnose. Log it.
2. **Dead `Dispose()`** (`CorrelatedCrosshairManager.cs:371` in both copies). After removing the Unloaded handler, nothing calls `Dispose()` anymore, but the class still implements `IDisposable`. Either wire it to a real teardown point (ServerTab close) or drop `IDisposable`. Leaving it declared misleads future readers.
3. **Un-reindented `try` body** (`Dashboard/Controls/CorrelatedTimelineLanesControl.xaml.cs:77-240`). The try wraps ~160 lines but the body stays at the outer indent. Compiles fine — diff-minimizing but ugly in the raw file. Lite's copy was already structured with try/finally so it didn't have this problem. Worth a cosmetic follow-up.
4. **Test plan** — the crosshair-survives-tab-switches item (the exact scenario this PR claims to fix) is unchecked. Manual validation still outstanding per the PR description.
5. **Lite-first**: this fix landed in Lite + Dashboard simultaneously rather than Lite-first-then-port. Defensible for a parity bugfix where both copies had the same bug, but noting it against the stated ordering.

No tests were added — UI-timing behavior in WPF/Avalonia is hard to cover with the existing harness; calling it out rather than flagging as blocking.

---

PR #868, opened 2026-04-20
Fix blocked process report plan lookup (#867)

### LINKED ISSUE #867: [BUG] Not clear how to work with blocking monitor
### Component

Lite

latest nightly build

### SQL Server Version

Azure SQL

### Windows Version

Windows server 2022

### Describe the Bug

Not clear how **Blocked process report** should work

I see this picture 

Image

Hit Refresh

When I press Right Click - View plan, the following error occurs

Image

When I press Get Actual plan, this error

Image

When I press Save on the line, it saves the XML like this (SQL truncated): 

Image

### Steps to Reproduce

Select Blocking-Blocked process reports and press Refresh

### Expected Behavior

Expected behaviour:
There should be some functions that allow:
1. See the blocking/blocked SQL text (without SQL text truncate)
2. See plans for these blocking 

### Actual Behavior

Text is truncating, and plan selection generates an error. (maybe it is related to #857, but blocking should not require master DB access)

```text

```

### Screenshots

_No response_

### Additional Context

_No response_

erikdarlingdata:
## Summary
- Right-click on a Blocked Process Reports row now offers **View Blocked Plan** and **View Blocking Plan**; shared context menu previously fell through with no handler case
- Removed **Get Actual Plan** from this grid — re-executing a query that was mid-transaction holding locks is a foot-gun
- Added `FetchPlanBySqlHandleAsync` keyed on `sql_handle` + statement offsets against `sys.dm_exec_query_stats` (query_hash isn't in the XE event payload, so the existing hash-based path couldn't reach these plans)
- Frame extraction enumerates every `<frame>` in `executionStack`, filters the 42-byte all-zero placeholder handle (dynamic SQL / system contexts), and defaults `stmtstart=0` / `stmtend=-1` per the `dm_exec_text_query_plan` convention — matches sp_HumanEventsBlockViewer's XPath and join shape

Fixes #867.

Not addressed here: the truncated SQL text the reporter noted — that's SQL Server's own cap on `<inputbuf>` inside the `blocked_process_report` XE event, nothing we can do from the collector side. Related master-DB concern is tracked separately on #857.

## Test plan
- [x] Build clean (0 errors)
- [x] Manual smoke test — right-click a BPR row, both View Blocked/Blocking Plan resolve when the plan is still in cache
- [ ] Verify "no longer in cache" message when the plan has been evicted
- [ ] Verify the all-zero-handle path (dynamic SQL blocker) shows the "no resolvable sql_handle" message instead of churning round-trips

### DIFF
--- a/Lite/Controls/ServerTab.xaml
+++ b/Lite/Controls/ServerTab.xaml
@@ -32,12 +32,38 @@
                 </MenuItem>
             </ContextMenu>
 
+            <!-- Context menu for Blocked Process Reports: separate Blocked/Blocking plan
+                 actions; no Get Actual Plan because re-executing a blocked query is a foot-gun. -->
+            <ContextMenu x:Key="BlockedProcessContextMenu">
+                <MenuItem Header="Copy Cell" Click="CopyCell_Click">
+                    <MenuItem.Icon><TextBlock Text="&#x1F4CB;"/></MenuItem.Icon>
+                </MenuItem>
+                <MenuItem Header="Copy Row" Click="CopyRow_Click">
+                    <MenuItem.Icon><TextBlock Text="&#x1F4C4;"/></MenuItem.Icon>
+                </MenuItem>
+                <MenuItem Header="Copy All Rows" Click="CopyAllRows_Click">
+                    <MenuItem.Icon><TextBlock Text="&#x1F4D1;"/></MenuItem.Icon>
+                </MenuItem>
+                <Separator/>
+                <MenuItem Header="Export to CSV..." Click="ExportToCsv_Click">
+                    <MenuItem.Icon><TextBlock Text="&#x1F4CA;"/></MenuItem.Icon>
+                </MenuItem>
+                <Separator/>
+                <MenuItem Header="View Blocked Plan" Click="ViewBlockedSidePlan_Click">
+                    <MenuItem.Icon><TextBlock Text="&#x1F50D;"/></MenuItem.Icon>
+                </MenuItem>
+                <MenuItem Header="View Blocking Plan" Click="ViewBlockingSidePlan_Click">
+                    <MenuItem.Icon><TextBlock Text="&#x1F50D;"/></MenuItem.Icon>
+                </MenuItem>
+            </ContextMenu>
+
             <Style x:Key="GridRowStyle" TargetType="DataGridRow">
                 <Setter Property="ContextMenu" Value="{StaticResource DataGridContextMenu}"/>
             </Style>
 
             <!-- Row style for blocking grid - highlight long-duration blocks -->
-            <Style x:Key="BlockingRowStyle" TargetType="DataGridRow" BasedOn="{StaticResource GridRowStyle}">
+            <Style x:Key="BlockingRowStyle" TargetType="DataGridRow">
+                <Setter Property="ContextMenu" Value="{StaticResource BlockedProcessContextMenu}"/>
                 <Style.Triggers>
                     <DataTrigger Binding="{Binding IsLongBlock}" Value="True">
                         <Setter Property="Background" Value="#33FF6B6B"/>
--- a/Lite/Controls/ServerTab.xaml.cs
+++ b/Lite/Controls/ServerTab.xaml.cs
@@ -4861,6 +4861,104 @@ public partial class ServerTab : UserControl
         catch { return null; }
     }
 
+    // ── Blocked Process Report plan lookup ──
+
+    /* SQL Server writes this 42-byte all-zero handle into executionStack frames
+       for dynamic SQL / system contexts where no persistent sql_handle exists.
+       Filter matches sp_HumanEventsBlockViewer's XPath exclusion. */
+    private static readonly string ZeroSqlHandle = "0x" + new string('0', 84);
+
+    private async void ViewBlockedSidePlan_Click(object sender, RoutedEventArgs e)
+        => await ShowBlockedProcessPlanAsync(sender, blockingSide: false);
+
+    private async void ViewBlockingSidePlan_Click(object sender, RoutedEventArgs e)
+        => await ShowBlockedProcessPlanAsync(sender, blockingSide: true);
+
+    private async System.Threading.Tasks.Task ShowBlockedProcessPlanAsync(object sender, bool blockingSide)
+    {
+        if (sender is not MenuItem menuItem) return;
+        var grid = FindParentDataGrid(menuItem);
+        if (grid?.CurrentItem is not BlockedProcessReportRow row) return;
+
+        var sideLabel = blockingSide ? "Blocking" : "Blocked";
+        var spid = blockingSide ? row.BlockingSpid : row.BlockedSpid;
+        var queryText = blockingSide ? row.BlockingSqlText : row.BlockedSqlText;
+        var label = $"Est Plan - {sideLabel} SPID {spid}";
+
+        var frames = ExtractBlockedProcessFrames(row.BlockedProcessReportXml, blockingSide);
+        if (frames.Count == 0)
+        {
+            MessageBox.Show(
+                $"The {sideLabel.ToLowerInvariant()} process report has no resolvable sql_handle. " +
+                "This usually means the query ran as dynamic SQL or a system context — " +
+                "SQL Server records a zero handle in that case and the plan can't be recovered.",
+                "No Plan Available", MessageBoxButton.OK, MessageBoxImage.Information);
+            return;
+        }
+
+        string? planXml = null;
+        try
+        {
+            var connStr = _server.GetConnectionString(_credentialService);
+            foreach (var f in frames)
+            {
+                planXml = await LocalDataService.FetchPlanBySqlHandleAsync(
+                    connStr, row.DatabaseName, f.SqlHandle, f.StmtStart, f.StmtEnd);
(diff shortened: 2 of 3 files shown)


Lite/Controls/ServerTab.xaml.cs
```diff
@@ -4861,6 +4861,104 @@ private async void GetActualPlan_Click(object sender, RoutedEventArgs e)
...
+        {
+            var connStr = _server.GetConnectionString(_credentialService);
+            foreach (var f in frames)
+            {
+                planXml = await LocalDataService.FetchPlanBySqlHandleAsync(
+                    connStr, row.DatabaseName, f.SqlHandle, f.StmtStart, f.StmtEnd);
+                if (!string.IsNullOrEmpty(planXml)) break;
+            }
+        }
+        catch { }
```
erikdarlingdata, line 4910: Swallowed exception. If `SqlConnection.Open` fails (network, auth, firewall) or the SQL itself throws (permissions on DMVs, login disabled), control falls through to the `else` branch and the user is told *"The plan ... is no longer in the plan cache"* — a misleading diagnosis for a connection-layer failure. At minimum, distinguish connection/permission errors from "not found" so the dialog doesn't lie about the cause.

---

Lite/Controls/ServerTab.xaml.cs
```diff
@@ -4861,6 +4861,104 @@ private async void GetActualPlan_Click(object sender, RoutedEventArgs e)
...
+        var label = $"Est Plan - {sideLabel} SPID {spid}";
+
+        var frames = ExtractBlockedProcessFrames(row.BlockedProcessReportXml, blockingSide);
+        if (frames.Count == 0)
+        {
+            MessageBox.Show(
+                $"The {sideLabel.ToLowerInvariant()} process report has no resolvable sql_handle. " +
+                "This usually means the query ran as dynamic SQL or a system context — " +
+                "SQL Server records a zero handle in that case and the plan can't be recovered.",
+                "No Plan Available", MessageBoxButton.OK, MessageBoxImage.Information);
```
erikdarlingdata, line 4895: `ExtractBlockedProcessFrames` returns an empty list on *any* failure path: XML parse error, missing `blocked-process`/`blocking-process` element, missing `executionStack`, or all frames filtered out as zero handles. All of those land here and the user is told "no resolvable sql_handle ... dynamic SQL or system context", even when the real cause is malformed XML or a collector bug. Consider differentiating "no frames in stack" from "parse failed" from "all handles zero" — at least so support triage doesn't get wild-goose-chased.

---

Lite/Services/LocalDataService.QueryStats.cs
```diff
@@ -569,6 +569,72 @@ ps.total_elapsed_time DESC
...
+    {
+        if (string.IsNullOrWhiteSpace(sqlHandleHex)) return null;
+        var handleBytes = HexStringToBytes(sqlHandleHex);
+        if (handleBytes == null || handleBytes.Length == 0) return null;
+
+        using var connection = new SqlConnection(connectionString);
+        await connection.OpenAsync();
+
+        var quotedDbName = await GetValidatedDatabaseNameAsync(connection, databaseName)
+                           ?? "[master]";
```
erikdarlingdata, line 592: `?? "[master]"` silently swaps to master when the database from the BPR XML isn't on the server (dropped/renamed, or the caller lacks VIEW ANY DEFINITION). The sibling `FetchProcedurePlanOnDemandAsync` at L537-540 returns `null` when validation fails — this one deviates without explanation.

Functionally the lookup works because `sys.dm_exec_query_stats` is server-scoped, but the caller must have rights to execute `sp_executesql` in master, which isn't guaranteed in hardened environments. I'd either (a) document the fallback, or (b) match the sibling's contract and return `null` so the caller's messaging stays honest.

---

Lite/Services/LocalDataService.QueryStats.cs
```diff
@@ -569,6 +569,72 @@ ps.total_elapsed_time DESC
...
+AND   qs.statement_end_offset = @stmt_end
+AND   tqp.query_plan IS NOT NULL
+ORDER BY
+    qs.last_execution_time DESC
+OPTION(RECOMPILE);',
+    N'@h varbinary(64), @stmt_start int, @stmt_end int',
+    @h, @stmt_start, @stmt_end;";
+
+        using var command = new SqlCommand(query, connection) { CommandTimeout = 30 };
+        command.Parameters.Add(new SqlParameter("@h", SqlDbType.VarBinary, 64) { Value = handleBytes });
```
erikdarlingdata, line 612: Worth confirming: the `sqlhandle` written into BPR `<frame>` elements and `sys.dm_exec_query_stats.sql_handle` are the same 64-byte binary identity for the same ad-hoc/proc statement. In practice they match, but there are historical edge cases where BPR frames carry a shorter handle (e.g., 42 bytes for certain system/dynamic contexts) that won't equate to a 64-byte query_stats handle. A smoke test against a known repro (deliberate key-range block, BPR captured, confirm plan resolves) would close the loop — I don't see a test asserting this.

---

Lite/Services/LocalDataService.QueryStats.cs
```diff
@@ -569,6 +569,72 @@ ps.total_elapsed_time DESC
...
+            if (!byte.TryParse(hex.AsSpan(start + i * 2, 2),
+                               System.Globalization.NumberStyles.HexNumber,
+                               System.Globalization.CultureInfo.InvariantCulture,
+                               out bytes[i]))
+            {
+                return null;
+            }
+        }
+        return bytes;
+    }
```
erikdarlingdata, line 636: `HexStringToBytes` and `ExtractBlockedProcessFrames` are pure, static, and trivially testable — yet Lite.Tests gains nothing in this PR. Minimal coverage worth adding: zero handle → filtered; odd-length hex → null; missing `stmtend` attribute → defaults applied; malformed XML → empty list; multi-frame stack picks frames in document order.

---

Lite/Controls/ServerTab.xaml
```diff
@@ -32,12 +32,38 @@
...
+            </ContextMenu>
+
             <Style x:Key="GridRowStyle" TargetType="DataGridRow">
                 <Setter Property="ContextMenu" Value="{StaticResource DataGridContextMenu}"/>
             </Style>
 
             <!-- Row style for blocking grid - highlight long-duration blocks -->
-            <Style x:Key="BlockingRowStyle" TargetType="DataGridRow" BasedOn="{StaticResource GridRowStyle}">
+            <Style x:Key="BlockingRowStyle" TargetType="DataGridRow">
+                <Setter Property="ContextMenu" Value="{StaticResource BlockedProcessContextMenu}"/>
```
erikdarlingdata, line 66: Dropping `BasedOn="{StaticResource GridRowStyle}"` is fine today because `GridRowStyle` only sets `ContextMenu`, which this style now overrides. Flagging so the next setter added to `GridRowStyle` isn't silently skipped here — easy landmine.

---

erikdarlingdata (commented):
## What this PR does
Fixes #867. On the **Blocked Process Reports** grid in Lite, right-click now routes through a dedicated `BlockedProcessContextMenu` (previously it fell through the shared `DataGridContextMenu` with no handler). Two new entries — **View Blocked Plan** / **View Blocking Plan** — extract `sqlhandle` + `stmtstart`/`stmtend` from the BPR `executionStack` frames and look the plan up against `sys.dm_exec_query_stats` on the source server. **Get Actual Plan** is intentionally omitted here (re-executing a mid-transaction blocked query is a foot-gun — correct call).

## What's good
- Targets `dev`. ✓
- Lite-only; Dashboard's BPR grid currently only offers "Save XML" (Dashboard/ServerTab.xaml.cs:1947), so this is a genuine Lite-first feature, not a skipped port.
- `PlanAnalyzer` not touched — no sync concern.
- No schema changes — no install/upgrade concern.
- SQL follows the existing Erik style in this file: `EXECUTE db.sys.sp_executesql`, column-alias-left form, `OPTION(RECOMPILE)`, no `NOLOCK`, parameterized. Three-part naming for Azure SQL DB compatibility is preserved.
- `HexStringToBytes` uses `Span`+`InvariantCulture` — correct.
- `async void` only on Click handlers — fine.
- `BlockingRowStyle` is applied only to `BlockedProcessReportGrid` (ServerTab.xaml:1317), so the menu change is targeted correctly.

## Needs attention
1. **Swallowed exception at ServerTab.xaml.cs:4910** produces a misleading "plan evicted" dialog when the real failure is connection/auth/permission. Distinguish.
2. **"No resolvable sql_handle" message at ServerTab.xaml.cs:4891** is also shown when XML parsing fails or the stack element is missing — wrong explanation for the user.
3. **`?? "[master]"` fallback at LocalDataService.QueryStats.cs:592** diverges from `FetchProcedurePlanOnDemandAsync`'s null-on-fail contract (same file, L537-540). Pick one.
4. **BPR `sqlhandle` ↔ `dm_exec_query_stats.sql_handle` equivalence** is assumed, not verified. A repro-based smoke test would be cheap insurance.
5. **No tests** for `ExtractBlockedProcessFrames` / `HexStringToBytes` despite Lite.Tests existing — both are pure, static, trivially testable.
6. **`BlockingRowStyle` lost its `BasedOn`** — harmless today, latent footgun later.

Inline comments on each.

No approval, no changes requested — maintainer call.

---

PR #869, opened 2026-04-20
Pre-filter query snapshots into #temp on Azure SQL DB (#857)

erikdarlingdata:
## Summary

Follow-up to #861. @TrudAX confirmed the fix didn't land — the Live Snapshot button still errored with SQL 300 (`VIEW SERVER PERFORMANCE STATE permission was denied on object 'server', database 'master'`) on his D365FO Azure SQL Database.

Root cause: the `DB_ID()` predicate in the WHERE clause wasn't sufficient. The `OUTER APPLY`s to `sys.dm_exec_sql_text` and `sys.dm_exec_text_query_plan` were being evaluated against master-scoped rows from `sys.dm_exec_requests` *before* the filter applied, so the DMFs got called with master-scoped `sql_handle` values — which a DB-scoped (contained) login isn't allowed to resolve.

TrudAX confirmed the DMFs themselves work fine when he passes handles from his own database in SSMS. The problem was purely query shape / evaluation order.

A CTE or derived table wouldn't fix this — the optimizer is free to push the `OUTER APPLY` above the filter. Only materialisation guarantees order, so I pre-filter the request rows into a `#req` temp table first and drive the DMFs off that.

- New `QuerySnapshotsAzureBase` template selects filtered `sys.dm_exec_requests` rows `INTO #req` scoped to `DB_ID()`, then does the `OUTER APPLY`s against `#req`
- Boxed SQL Server and Azure MI paths (`isAzureSqlDatabase = false`) are unchanged — still use the original inline template
- Azure DB_ID predicate removed from the shared template (it's now in the Azure-only template's pre-filter)

## Test plan

- [x] Build: `dotnet build Lite/PerformanceMonitorLite.csproj -c Debug` — 0 warnings, 0 errors
- [x] Boxed SQL Server smoke test (SQL2022): non-Azure template unchanged, still returns rows normally
- [x] Azure template syntax validated on SQL2022 via sqlcmd — `#req` materialises, final SELECT runs, no errors
- [ ] D365FO environment confirmation (deferred to @TrudAX when the nightly lands)

The memory / tempdb / waiting_tasks collector failures called out in the issue are separate — those need their own graceful-degradation handling and I'll tackle them in a follow-up PR.

### DIFF
--- a/Lite/Services/RemoteCollectorService.QuerySnapshots.cs
+++ b/Lite/Services/RemoteCollectorService.QuerySnapshots.cs
@@ -21,7 +21,7 @@ namespace PerformanceMonitorLite.Services;
 public partial class RemoteCollectorService
 {
     private const string QuerySnapshotsBase = """
-        
+
 SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
 SET LOCK_TIMEOUT 1000;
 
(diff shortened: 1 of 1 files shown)


Lite/Services/RemoteCollectorService.QuerySnapshots.cs
```diff
@@ -81,17 +81,124 @@ WHERE der.session_id <> @@SPID
...
+LEFT JOIN sys.databases AS d
+  ON d.database_id = der.database_id
+{1}
+WHERE dest.text IS NOT NULL
+ORDER BY der.cpu_time DESC, der.parallel_worker_count DESC
+OPTION(MAXDOP 1, RECOMPILE);
+
+DROP TABLE #req;
+""";
+    private readonly static CompositeFormat QuerySnapshotsAzureBaseFormat = CompositeFormat.Parse(QuerySnapshotsAzureBase);
```
erikdarlingdata, line 192: No unit test covers the new Azure template branch. `BuildQuerySnapshotsQuery(_, isAzureSqlDatabase: true)` with both `supportsLiveQueryPlan` values is now a meaningfully different code path (different template, different placeholder count) and would regress silently if `{0}`/`{1}` drift. A couple of golden-string/contains assertions in `Lite.Tests` — e.g. "contains `INTO #req`", "contains `DB_ID()`", "does not contain `{2}`" — would be cheap insurance.

---

Lite/Services/RemoteCollectorService.QuerySnapshots.cs
```diff
@@ -81,17 +81,124 @@ WHERE der.session_id <> @@SPID
...
+            WHEN 5 THEN 'Snapshot'
+            ELSE '???'
+        END,
+    der.dop,
+    der.parallel_worker_count,
+    des.login_name,
+    des.host_name,
+    des.program_name,
+    des.open_transaction_count,
+    der.percent_complete
```
erikdarlingdata, line 177: The final projection here is a near byte-for-byte copy of the projection in `QuerySnapshotsBase` (lines 28–71). Any future column add/rename/type change now has to land in two templates, and the Lite collector's `AppendValue` ordinal reads (lines 257–281) are driven off both. Worth a TODO or a comment pointer between the two so a drift is at least obvious. Not a blocker for this PR.

---

Lite/Services/RemoteCollectorService.QuerySnapshots.cs
```diff
@@ -81,17 +81,124 @@ WHERE der.session_id <> @@SPID
...
+    der.transaction_isolation_level,
+    der.dop,
+    der.parallel_worker_count,
+    der.percent_complete
+INTO #req
+FROM sys.dm_exec_requests AS der
+WHERE der.session_id <> @@SPID
+AND   der.session_id >= 50
+AND   der.database_id = DB_ID()
+AND   der.database_id <> ISNULL(DB_ID(N'PerformanceMonitor'), 0);
```
erikdarlingdata, line 132: Minor: `der.database_id <> ISNULL(DB_ID(N'PerformanceMonitor'), 0)` is vestigial in the Lite path — Lite targets don't host a `PerformanceMonitor` database (Lite uses local DuckDB), so `DB_ID('PerformanceMonitor')` is always NULL and the predicate reduces to `database_id <> 0`. Combined with `der.database_id = DB_ID()` one line up, it's effectively a no-op. Harmless, but you could drop it to keep the Azure pre-filter tight. Same applies to line 83 in the boxed template, though that's out of scope here.

---

Lite/Services/RemoteCollectorService.QuerySnapshots.cs
```diff
@@ -81,17 +81,124 @@ WHERE der.session_id <> @@SPID
...
+     * PERFORMANCE STATE denied). Materialising the filtered request rows
+     * into a #temp table first guarantees the DMFs only see handles from
+     * the current database. See #857.
+     */
+    private const string QuerySnapshotsAzureBase = """
+
+SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;
+SET LOCK_TIMEOUT 1000;
+
+IF OBJECT_ID(N'tempdb..#req') IS NOT NULL DROP TABLE #req;
```
erikdarlingdata, line 103: Good call keeping the defensive pre-drop — `SqlClient` pools connections, so a prior execution that threw before line 190 would leave `#req` on a pooled session. The paired `DROP TABLE #req;` at the tail is good hygiene too. If you want belt-and-braces, wrapping the body in `BEGIN TRY / BEGIN CATCH` with a cleanup `DROP` in the catch would prevent the next call on the same pooled connection from paying the pre-drop cost, but it isn't necessary.

---

erikdarlingdata (commented):
## What this PR does
Follow-up to #861 for issue #857. Adds a dedicated `QuerySnapshotsAzureBase` template used only when `isAzureSqlDatabase == true`. The template pre-filters `sys.dm_exec_requests` into `#req` scoped by `DB_ID()` before the `OUTER APPLY`s to `sys.dm_exec_sql_text` / `sys.dm_exec_text_query_plan`, so a DB-scoped (contained) login on Azure SQL DB — e.g. D365FO — never asks those DMFs to resolve master-scoped handles. Boxed SQL and Azure MI paths are unchanged. `{2}` placeholder for the old Azure predicate is removed and the format args drop from 3 to 2.

## What's good
- Root-cause analysis in the PR body is correct: `OUTER APPLY` evaluation order vs. WHERE predicate pushdown isn't guaranteed, and materialisation is the only reliable fix. A CTE/derived table wouldn't have helped.
- Azure-only branch — boxed/MI keep the simpler shape, no extra tempdb churn for non-Azure targets.
- Defensive pre-drop of `#req` (line 103) + tail drop (line 190) handle pooled-connection reuse.
- Filter set in the pre-filter (`@@SPID`, `session_id >= 50`, `DB_ID()`, PerformanceMonitor exclusion) matches what the final WHERE used to enforce.
- Targets `dev`. No PlanAnalyzer touch (irrelevant here). No schema change. No install/upgrade script churn. No workflow / SignPath change.
- Lite-first ordering satisfied — the Dashboard equivalent lives in SQL Agent (`install/11_collect_query_snapshots.sql`) and uses `sp_WhoIsActive`, which sidesteps this issue entirely.

## What needs attention
- **Test coverage (minor).** The new Azure branch in `BuildQuerySnapshotsQuery` is reachable and structurally different but has no unit test. Inline comment at `RemoteCollectorService.QuerySnapshots.cs:192` with suggested assertions.
- **Duplication risk (minor).** The 40-line final projection is copy-pasted between `QuerySnapshotsBase` and `QuerySnapshotsAzureBase`, and the collector's ordinal `AppendValue` reads depend on it. Any future column change must touch both. Inline at 134–177.
- **Vestigial predicate (nit).** `database_id <> ISNULL(DB_ID(N'PerformanceMonitor'), 0)` is a no-op in the Lite path. Inline at 132.
- **Defensive cleanup (nit).** Optional `TRY/CATCH` cleanup in the Azure template. Inline at 103.

Nothing blocking. Maintainer's call on merge.

---

PR #870, opened 2026-04-20
Skip retrying collectors after permission denial (#857)

erikdarlingdata:
## Summary

Follow-up to #857. @TrudAX's Collection Health screenshot showed five collectors (`memory_clerks`, `memory_stats`, `tempdb_stats`, `query_snapshots`, `waiting_tasks`) all stuck at "10 runs, 0 success" — each re-running every collection interval, hitting the same SQL error 300 (`VIEW SERVER PERFORMANCE STATE permission was denied`) every time, and logging a fresh identical denial row into the collection log.

The denial isn't transient. A DB-scoped login (D365FO) will never grow server-level permission mid-session. Retrying just churns the log.

This PR flags the collector on first denial and short-circuits `RunCollectorAsync` so we skip the round-trip and the log entry entirely.

- New `CollectorHealthEntry.IsPermissionRestricted` bool set when `RecordCollectorResult` sees status `PERMISSIONS`
- New `IsCollectorPermissionRestricted(serverId, collectorName)` helper read under the existing `_healthLock`
- Early-return in `RunCollectorAsync` right after the existing MFA-cancelled skip, logging at Debug level
- Flag is in-memory per (server, collector) — app restart retries once. If perms are still missing the flag re-applies after the first attempt

Collection Health's `NO_PERMISSIONS` status logic is unchanged — it still renders correctly from the single recorded denial row.

Per-collector DB-scoped fallback queries (e.g. `sys.dm_db_resource_stats` for memory_stats) were considered but deliberately out of scope — their semantics differ from the boxed DMVs and each is its own decision.

## Test plan

- [ ] Boxed SQL Server smoke test: no behavioural change expected (no permission errors normally)
- [ ] D365FO confirmation (deferred to @TrudAX once the nightly lands): collectors should hit denial once then stop; Collection Health shows `NO_PERMISSIONS` instead of `10 runs, 0 success` growing unbounded

### DIFF
--- a/Lite/Services/RemoteCollectorService.cs
+++ b/Lite/Services/RemoteCollectorService.cs
@@ -39,6 +39,15 @@ public class CollectorHealthEntry
     public int ConsecutiveErrors { get; set; }
     public int TotalErrors { get; set; }
     public int TotalSuccesses { get; set; }
+
+    /*
+     * Set when a collector hits a non-transient permission error
+     * (SQL errors 229 / 297 / 300). The scheduler skips the collector
+     * for the rest of the app session so we don't churn the log with
+     * identical denials every interval. Cleared on app restart — if
+     * permissions get granted later, the next launch retries once.
+     */
+    public bool IsPermissionRestricted { get; set; }
 }
 
 /// <summary>
@@ -225,9 +234,12 @@ public partial class RemoteCollectorService
             {
                 /* Permission errors are not transient — don't count as failures
                    (which would show FAILING) but don't count as success either.
-                   Record the error message so the user can see what's wrong. */
+                   Record the error message so the user can see what's wrong,
+                   and flag the collector so the scheduler stops retrying for
+                   the rest of the app session. */
                 entry.LastErrorTime = DateTime.UtcNow;
                 entry.LastErrorMessage = errorMessage;
+                entry.IsPermissionRestricted = true;
             }
             else
             {
@@ -236,6 +248,19 @@ public partial class RemoteCollectorService
                 entry.ConsecutiveErrors++;
                 entry.TotalErrors++;
             }
+        }
+    }
+
+    /// <summary>
+    /// Returns true if a collector has hit a permission denial this session
+    /// and should be skipped without re-running. See <see cref="CollectorHealthEntry.IsPermissionRestricted"/>.
+    /// </summary>
+    private bool IsCollectorPermissionRestricted(int serverId, string collectorName)
+    {
+        lock (_healthLock)
+        {
+            return _collectorHealth.TryGetValue((serverId, collectorName), out var entry)
+                && entry.IsPermissionRestricted;
         }
     }
 
@@ -376,6 +401,15 @@ public partial class RemoteCollectorService
             {
                 AppLogger.Info("Collector", $"  [{server.DisplayName}] {collectorName} SKIPPED - MFA authentication cancelled by user");
                 _logger?.LogDebug("Skipping collector '{Collector}' for server '{Server}' - user cancelled MFA",
+                    collectorName, server.DisplayName);
+                return;
+            }
+
+            // Skip collectors that have already hit a non-transient permission denial
+            // this session. Flag is in-memory — next app start retries once (see #857).
+            if (IsCollectorPermissionRestricted(GetServerId(server), collectorName))
+            {
+                _logger?.LogDebug("Skipping collector '{Collector}' for server '{Server}' - permission denied this session",
                     collectorName, server.DisplayName);
                 return;
             }


Lite/Services/RemoteCollectorService.cs
```diff
@@ -39,6 +39,15 @@ public class CollectorHealthEntry
...
     public int TotalSuccesses { get; set; }
+
+    /*
+     * Set when a collector hits a non-transient permission error
+     * (SQL errors 229 / 297 / 300). The scheduler skips the collector
+     * for the rest of the app session so we don't churn the log with
+     * identical denials every interval. Cleared on app restart — if
+     * permissions get granted later, the next launch retries once.
+     */
+    public bool IsPermissionRestricted { get; set; }
```
erikdarlingdata, line 50: Flag is never cleared mid-session. `ClearHealthForServer` is only invoked from `ServerContextMenu_Remove_Click` (`Lite/MainWindow.xaml.cs:1144`); the Edit-server success path at `Lite/MainWindow.xaml.cs:1124` just calls `RefreshServerList()`. So if a user hits a permission denial, fixes perms on the SQL side (or edits the server entry to use a higher-privilege login) without restarting the app, the collector stays skipped until restart even though the denial is no longer valid.

Cheap fix: call `_collectorService?.ClearHealthForServer(...)` in the `dialog.ShowDialog() == true` branch of `ServerContextMenu_Edit_Click`, or expose a narrower `ClearPermissionFlagsForServer` that only resets `IsPermissionRestricted` so success/error counts aren't lost.

---

Lite/Services/RemoteCollectorService.cs
```diff
@@ -225,9 +234,12 @@ private void RecordCollectorResult(int serverId, string collectorName, string st
             {
                 /* Permission errors are not transient — don't count as failures
                    (which would show FAILING) but don't count as success either.
-                   Record the error message so the user can see what's wrong. */
+                   Record the error message so the user can see what's wrong,
+                   and flag the collector so the scheduler stops retrying for
+                   the rest of the app session. */
                 entry.LastErrorTime = DateTime.UtcNow;
                 entry.LastErrorMessage = errorMessage;
+                entry.IsPermissionRestricted = true;
```
erikdarlingdata, line 242: `IsPermissionRestricted` is latched to `true` but never flipped back to `false`. If `RecordCollectorResult` is later called with status `SUCCESS` (e.g. after the Edit-server fix suggested above, or if perms get granted and the flag is cleared by some future path), the success branch at line 227 doesn't reset this flag. Worth resetting `entry.IsPermissionRestricted = false;` in the `SUCCESS` branch so the state is self-healing even without an explicit clear call.

---

Lite/Services/RemoteCollectorService.cs
```diff
@@ -380,6 +405,15 @@ public async Task RunCollectorAsync(ServerConnection server, string collectorNam
...
             }
 
+            // Skip collectors that have already hit a non-transient permission denial
+            // this session. Flag is in-memory — next app start retries once (see #857).
+            if (IsCollectorPermissionRestricted(GetServerId(server), collectorName))
+            {
+                _logger?.LogDebug("Skipping collector '{Collector}' for server '{Server}' - permission denied this session",
+                    collectorName, server.DisplayName);
+                return;
+            }
```
erikdarlingdata, line 415: Minor: the skip happens after `GetConnectionStatus`, `IsCollectorSupported`, and the MFA gate — all cheap, but it means every skipped collector still runs three dictionary lookups and a method call before bailing. Not worth restructuring, but if `RunAllCollectorsForServerAsync` ever gets called on a server where most collectors are permission-restricted, the cumulative cost is noticeable. Consider hoisting the restricted check into the scheduler (`GetDueCollectorsForServer`) so denied collectors aren't even enumerated. Not a blocker.

---

Lite/Services/RemoteCollectorService.cs
```diff
@@ -239,6 +251,19 @@ Record the error message so the user can see what's wrong. */
...
+    /// and should be skipped without re-running. See <see cref="CollectorHealthEntry.IsPermissionRestricted"/>.
+    /// </summary>
+    private bool IsCollectorPermissionRestricted(int serverId, string collectorName)
+    {
+        lock (_healthLock)
+        {
+            return _collectorHealth.TryGetValue((serverId, collectorName), out var entry)
+                && entry.IsPermissionRestricted;
+        }
+    }
```
erikdarlingdata, line 265: No test coverage in `Lite.Tests/`. The behavior here is pure state logic and easily unit-testable: instantiate `RemoteCollectorService`, call `RecordCollectorResult(..., "PERMISSIONS", ...)` via a test-only accessor (or refactor the entry dictionary behind a small testable surface), then assert `IsCollectorPermissionRestricted` returns true and that a subsequent `SUCCESS` resets it (see the other comment). Worth adding alongside `FactCollectorTests.cs` or as a new `CollectorHealthTests.cs` — the D365FO smoke test in the PR description is deferred, so unit coverage is the only thing that catches a regression before the next nightly.

---

erikdarlingdata (commented):
## What this does
Follow-up to #857. Adds an in-memory `IsPermissionRestricted` flag on `CollectorHealthEntry`, set in `RecordCollectorResult` when a collector hits SQL error 229/297/300, and short-circuits `RunCollectorAsync` on the next scheduled run so we stop churning identical denials into the collection log. Scope is correctly limited to Lite (issue originated with D365FO on Azure SQL).

## Good
- Lite-only change, base branch `dev` — correct per repo conventions.
- Flag is read/written under the existing `_healthLock`; no new lock and no lock-ordering risk (`IsCollectorPermissionRestricted` takes a single lock and returns).
- Decision to leave per-collector DB-scoped fallback queries out of scope is the right call — those have different semantics from the boxed DMVs and each deserves its own decision.
- `NO_PERMISSIONS` render path in `Lite/Services/LocalDataService.CollectionHealth.cs` is unchanged — the single recorded denial row still drives the UI state correctly.
- No PlanAnalyzer changes (N/A), no SQL changes (N/A), no workflow/signing changes (N/A), no schema upgrade concerns (N/A).

## Needs attention
1. **Flag never clears mid-session.** `ClearHealthForServer` is only wired to server removal, not server edit (`Lite/MainWindow.xaml.cs:1118`). Fixing perms or swapping in a higher-privilege login via the Edit dialog won't recover the collector until app restart. See inline on line 50.
2. **SUCCESS branch doesn't reset the flag.** `RecordCollectorResult` latches `IsPermissionRestricted = true` but never sets it back to `false`. Self-healing would be one line. See inline on line 242.
3. **No test coverage added.** The PR description explicitly defers smoke testing to the nightly + @TrudAX; unit tests on the flag state machine are the only thing catching regressions in the meantime. See inline on line 265.
4. **Minor** — skip check sits downstream of several cheap gates in `RunCollectorAsync`; could live in the scheduler instead. Not a blocker, just noted for future cleanup. See inline on line 415.

No blockers. (1) and (2) together would close the "user fixes perms without restart" gap cleanly.

---

PR #871, opened 2026-04-21
Skip live query plans on Azure SQL DB (#857)

erikdarlingdata:
## Summary

Third follow-up to #857. @TrudAX confirmed the `#temp` pre-filter from #869 didn't resolve the Live Snapshot error for his D365FO database — same SQL 300 (`VIEW SERVER PERFORMANCE STATE permission was denied`).

His working SSMS query uses `sys.dm_exec_sql_text` and `sys.dm_exec_query_plan` but **not** `sys.dm_exec_query_statistics_xml` — that's the signal. Per MS docs, `sys.dm_exec_query_statistics_xml` requires `VIEW SERVER PERFORMANCE STATE` on Azure SQL Database regardless of service tier or scope. Even with the `#temp` pre-filter only feeding it current-DB session IDs, the DMF does a server-level permission check before returning, so it fails for DB-scoped logins.

Fix: disable live query plans on Azure SQL Database in both the collector and the Live Snapshot button. Users lose the animated in-progress query plan on Azure SQL DB but get working current-activity visibility — which they don't have at all today. Boxed SQL Server and Azure Managed Instance (edition 8) still get live plans.

- `CollectQuerySnapshotsAsync`: `supportsLiveQueryPlan = !isAzureSqlDatabase && (version >= 13 || version == 0 || edition == 8)`
- `LiveSnapshot_Click`: pass `!_isAzureSqlDatabase` instead of hardcoded `true`
- No SQL template changes — when `supportsLiveQueryPlan` is false the existing `BuildQuerySnapshotsQuery` already emits `live_query_plan = CONVERT(xml, NULL),` and omits the OUTER APPLY

## Test plan

- [x] Boxed SQL Server smoke test (SQL2022): live plans still emitted, behaviour unchanged
- [ ] D365FO confirmation (deferred to @TrudAX): Live Snapshot returns rows without the permission error; `live_query_plan` column will be NULL on Azure SQL DB

### DIFF
--- a/Lite/Controls/ServerTab.xaml.cs
+++ b/Lite/Controls/ServerTab.xaml.cs
@@ -5365,7 +5365,9 @@ public partial class ServerTab : UserControl
                 ConnectTimeout = 15
             };
 
-            var query = RemoteCollectorService.BuildQuerySnapshotsQuery(supportsLiveQueryPlan: true, isAzureSqlDatabase: _isAzureSqlDatabase);
+            // Live query plans require VIEW SERVER PERFORMANCE STATE on Azure SQL DB,
+            // which DB-scoped logins don't have — skip them there. See #857.
+            var query = RemoteCollectorService.BuildQuerySnapshotsQuery(supportsLiveQueryPlan: !_isAzureSqlDatabase, isAzureSqlDatabase: _isAzureSqlDatabase);
 
             await using var connection = new SqlConnection(builder.ConnectionString);
             await connection.OpenAsync();
--- a/Lite/Services/RemoteCollectorService.QuerySnapshots.cs
+++ b/Lite/Services/RemoteCollectorService.QuerySnapshots.cs
@@ -217,11 +217,18 @@ DROP TABLE #req;
     /// </summary>
     private async Task<int> CollectQuerySnapshotsAsync(ServerConnection server, CancellationToken cancellationToken)
     {
-        // dm_exec_query_statistics_xml requires SQL Server 2016 SP1+ (version 13)
+        /*
+         * sys.dm_exec_query_statistics_xml requires SQL Server 2016 SP1+ (version 13)
+         * on boxed, and VIEW SERVER PERFORMANCE STATE on Azure SQL Database regardless
+         * of tier. DB-scoped logins (e.g. D365FO) don't have the server-level grant
+         * and the DMF raises error 300 even when only called for current-DB sessions,
+         * so we disable live query plans on Azure SQL DB entirely. See #857.
+         */
         var serverStatus = _serverManager.GetConnectionStatus(server.Id);
-        var supportsLiveQueryPlan = serverStatus.SqlMajorVersion >= 13 || serverStatus.SqlMajorVersion == 0
-            || serverStatus.SqlEngineEdition == 5 || serverStatus.SqlEngineEdition == 8;
         var isAzureSqlDatabase = serverStatus.SqlEngineEdition == 5;
+        var supportsLiveQueryPlan = !isAzureSqlDatabase
+            && (serverStatus.SqlMajorVersion >= 13 || serverStatus.SqlMajorVersion == 0
+                || serverStatus.SqlEngineEdition == 8);
 
         var query = BuildQuerySnapshotsQuery(supportsLiveQueryPlan, isAzureSqlDatabase);
 


Lite/Services/RemoteCollectorService.QuerySnapshots.cs
```diff
@@ -217,11 +217,18 @@ internal static string BuildQuerySnapshotsQuery(bool supportsLiveQueryPlan, bool
...
+         * and the DMF raises error 300 even when only called for current-DB sessions,
+         * so we disable live query plans on Azure SQL DB entirely. See #857.
+         */
         var serverStatus = _serverManager.GetConnectionStatus(server.Id);
-        var supportsLiveQueryPlan = serverStatus.SqlMajorVersion >= 13 || serverStatus.SqlMajorVersion == 0
-            || serverStatus.SqlEngineEdition == 5 || serverStatus.SqlEngineEdition == 8;
         var isAzureSqlDatabase = serverStatus.SqlEngineEdition == 5;
+        var supportsLiveQueryPlan = !isAzureSqlDatabase
+            && (serverStatus.SqlMajorVersion >= 13 || serverStatus.SqlMajorVersion == 0
+                || serverStatus.SqlEngineEdition == 8);
```
erikdarlingdata, line 231: Behaviour change worth calling out: previously `SqlEngineEdition == 5` (Azure SQL DB) explicitly *enabled* live plans (the old expression OR'd `EngineEdition == 5`). After this change Azure SQL DB is unconditionally excluded — including service-tier/login combos where `VIEW SERVER PERFORMANCE STATE` *is* held (some elastic admin / server-admin logins can grant it). That's almost certainly the right tradeoff given #857, but it removes the capability for the small set of Azure SQL DB users who weren't broken before. The PR description's deferred test on D365FO is the right gate.

---

Lite/Controls/ServerTab.xaml.cs
```diff
@@ -5365,7 +5365,9 @@ private async void LiveSnapshot_Click(object sender, RoutedEventArgs e)
                 ConnectTimeout = 15
             };
 
-            var query = RemoteCollectorService.BuildQuerySnapshotsQuery(supportsLiveQueryPlan: true, isAzureSqlDatabase: _isAzureSqlDatabase);
+            // Live query plans require VIEW SERVER PERFORMANCE STATE on Azure SQL DB,
+            // which DB-scoped logins don't have — skip them there. See #857.
+            var query = RemoteCollectorService.BuildQuerySnapshotsQuery(supportsLiveQueryPlan: !_isAzureSqlDatabase, isAzureSqlDatabase: _isAzureSqlDatabase);
```
erikdarlingdata, line 5370: Gating logic now lives in two places — here and `RemoteCollectorService.QuerySnapshots.cs:231`. They agree today (`!isAzureSqlDatabase`), but the collector also factors version/edition (`>= 13 || == 0 || == 8`) while this caller only checks Azure. If the collector's predicate gets tighter later (e.g. excludes some other engine edition), this site won't follow. Consider exposing a single `RemoteCollectorService.SupportsLiveQueryPlan(serverStatus)` helper so both sites share the rule.

---

erikdarlingdata (commented):
## What this PR does

Disables `sys.dm_exec_query_statistics_xml` (the live in-progress plan DMF) on Azure SQL Database in the Lite collector and the Live Snapshot button. On Azure SQL DB the DMF requires `VIEW SERVER PERFORMANCE STATE`, which DB-scoped logins (D365FO and similar) don't hold; the DMF performs a server-level permission check before honoring the per-session filter, so the `#temp` pre-filter from #869 didn't help. The fallback path in `BuildQuerySnapshotsQuery` already emits `live_query_plan = CONVERT(xml, NULL)` and drops the OUTER APPLY, so no SQL template change is needed.

## Good

- Targets `dev` ✓
- No `PlanAnalyzer.cs` changes — Dashboard/Lite sync not at risk
- No SQL/install/upgrade changes — no upgrade-script concern
- Lite-only fix (Dashboard's collectors are SQL Agent jobs, not affected)
- No `.github/workflows/build.yml` changes — SignPath policy unchanged
- Comment block at `Lite/Services/RemoteCollectorService.QuerySnapshots.cs:220-227` explains the *why* with issue reference; that's the kind of comment that earns its keep
- `LiveSnapshot_Click` is `async void` but it's a routed event handler — fine
- Boxed (version >= 13), unknown version (== 0), and Azure Managed Instance (edition 8) still get live plans; correctly preserved

## Needs attention

- **Duplicated gating predicate.** See inline at `Lite/Controls/ServerTab.xaml.cs:5370` — the click handler and the collector each compute `supportsLiveQueryPlan` independently. They agree today but are easy to drift. A `RemoteCollectorService.SupportsLiveQueryPlan(ConnectionStatus)` helper would keep them honest.
- **Behaviour change vs. previous code** at `Lite/Services/RemoteCollectorService.QuerySnapshots.cs:231` — the old expression had `|| SqlEngineEdition == 5`, i.e. Azure SQL DB was explicitly *enabled*. Now unconditionally disabled. Right call for the #857 cohort, but removes the capability for any Azure SQL DB login that *did* hold `VIEW SERVER PERFORMANCE STATE`. Worth confirming with @TrudAX (deferred test in PR description) before this lands.
- **No test coverage.** `Lite.Tests` has no tests covering `BuildQuerySnapshotsQuery` or the gating predicate. This logic has now changed three times for #857 (#869, #870, #871) — a small unit test that asserts the column/APPLY emission for `(supportsLiveQueryPlan, isAzureSqlDatabase)` matrix would have caught any regression and is cheap to add.
- **Related, not in scope:** Dashboard's `Dashboard/Services/DatabaseService.QueryPerformance.cs:3556` issues `sys.dm_exec_query_statistics_xml` unconditionally for the live-queries view, with no Azure-DB awareness. Dashboard normally connects to a centralized PerformanceMonitor DB on boxed SQL, so this likely isn't hit in practice — but if a Dashboard install ever points at an Azure SQL DB instance with a DB-scoped login, it will reproduce #857 there. Worth a follow-up issue, not a blocker for this PR.

---

Issue #972, opened 2026-05-21
[BUG] Fix #951 — clean up Monitor_LongQueries_*.trc files in data_retention not working and job failing


CADelete:
### Component

Full Dashboard

### Performance Monitor Version

2.11.0

### SQL Server Version

SQL Server 2019

### Windows Version

Windows Server 2019

### Describe the Bug

Data Retention job failing on new version.  Output from job log

Found 7 Monitor LongQueries_* file(s) in I:\Logs\MSSQLSERVER\; cutoff=2026-05-11T07:38.22.239 [SQLSTATE 010000]
Msg 22049, Sev16, Sate 1: xp_delete_files() returned error 2, 'The system cannot find the file specified.' [SQLSTATE 42000]

### Steps to Reproduce

Happening on all instances following upgrade to 2.11

### Expected Behavior

Should complete and remove data

### Actual Behavior

Job fails instantly

### Error Messages / Log Output

```text
Found 7 Monitor LongQueries_* file(s) in I:\Logs\MSSQLSERVER\; cutoff=2026-05-11T07:38.22.239 [SQLSTATE 010000]
Msg 22049, Sev16, Sate 1: xp_delete_files() returned error 2, 'The system cannot find the file specified.' [SQLSTATE 42000]
```

### Screenshots

<img width="500" height="137" alt="Image" src="https://github.com/user-attachments/assets/6e821bb3-fcd4-4b6d-b61f-7f23993ac51b" />

### Additional Context

Worked in previous version

CADelete:
The 7 files from the message are not within the cut off period, error message suggests that it cannot find the file which would be correct as there are no files to delete

<img width="332" height="79" alt="Image" src="https://github.com/user-attachments/assets/7a33999c-e54f-4864-9ff3-5f6862be4eda" />

This also matches the code, however it still fails

<img width="274" height="140" alt="Image" src="https://github.com/user-attachments/assets/e8165219-659d-4584-9472-bce332cf01de" />

erikdarlingdata:
Thanks for the detailed report — and for the follow-up screenshots, they pointed straight at the cause.

## Root cause

The trace-file cleanup added in v2.11.0 was broken in two ways:

[...]

Issue #951, opened 2026-05-11
[BUG] Trace files never get cleaned up


mike-hodgson-icon:
### Component

Full Dashboard

### Performance Monitor Version

2.10

### SQL Server Version

SQL Server 2019 CU32

### Windows Version

Windows Server 2022 Standard Edition

### Describe the Bug

Trace files (N'Monitor_LongQueries_*.trc') get created in the MSSQL errorlog directory by [collect].[trace_management_collector] and are dissected by [collect].[trace_analysis_collector] to grab info for long running queries, but they are never cleaned up (i.e. deleted) by [config].[data_retention].  The default retention for the trace files is 30 days.

<img width="957" height="329" alt="Image" src="https://github.com/user-attachments/assets/43152afb-8f54-430b-a3fe-20b66f2782c9" />

### Steps to Reproduce

- Open dashboard
- Add server
- Let that run for > retention days for its Trace Management collector schedule
- Check MSSQL errorlog directory for "Monitor_LongQueries_*.trc"

### Expected Behavior

"Monitor_LongQueries_*.trc" should get deleted by [config].[data_retention] when they're older than retention_days from config.collection_schedule where collector_name = N'trace_management_collector'.  My suggestion would be to use xp_delete_file, even though it's undocumented, to delete files older than the retention date with something like:

```
exec master.dbo.xp_delete_file
	1,                                /* 1 = delete files older than cutoff date  */
	N'C:\Program Files\Microsoft SQL Server\MSSQL15.MSSQLSERVER\MSSQL\Log\Monitor_LongQueries_*.trc',
	N'trc',
	'2026-04-12T08:00:00.000';             /* delete .trc files older than this date   */
```
...assuming you derive the path (there's code at the start of [collect].[trace_management_collector] to do that - easy copy/paste) and the appropriate retention date.

### Actual Behavior

Trace files accumulate indefinitely, as far as I can see, and will eventually fill the disk where the errorlog files sit.  One of my servers currently has 106 Monitor_LongQueries_*.trc files accounting for almost 7GB of disk, going back almost 2 months.  When the server is particularly busy a single trace can rollover to many trace files, with a max rollover file size of 200MB theoretically (although my earlier trace files were mostly rolling over at about 110-180MB).

<img width="521" height="284" alt="Image" src="https://github.com/user-attachments/assets/f5d21939-95f0-47ea-a16c-dbd83fbbe3f9" />

### Error Messages / Log Output

```text

```

[...]

PR #973, opened 2026-05-21
Fix trace-file accumulation and Data Retention job failure (#972)

erikdarlingdata:
## Problem

The trace-file cleanup added in v2.11.0 (#951) failed the `PerformanceMonitor - Data Retention` Agent job on every run with `Msg 22049` once any `Monitor_LongQueries_*.trc` files existed.

Root causes found while investigating #972:

1. **`xp_delete_file` cannot delete `.trc` files at all** — it only accepts SQL Server backup files and Maintenance Plan report files, and validates the file header. The #951 cleanup could never have worked.
2. The call passed a **wildcard path** as the folder argument, which raised an **uncatchable** `Msg 22049` (extended-proc errors bypass `TRY...CATCH`) that failed the whole Agent job step.
3. The trace files accumulate because `scheduled_master_collector` issued `RESTART` every cycle (tearing down the trace and spawning a fresh timestamped one), and the trace was created with **no rollover file-count cap**.

## Changes

- **`config.data_retention`** — removed the broken `xp_delete_file` block (the crash).
- **`collect.trace_management_collector`** — new `@max_files` parameter (default 5) → `sp_trace_create @filecount`, so SQL Server prunes old `.trc` files itself as the trace rolls. `START` now also replaces an *unbounded* trace left by an older version, so the fix self-heals without waiting for a SQL Server restart.
- **`scheduled_master_collector`** — calls `START` instead of `RESTART`; keeps one bounded trace running instead of orphaning files every cycle.
- **`tools/Remove-OrphanedTraceFiles.ps1`** — new one-time cleanup for `.trc` files left on disk by versions ≤ 2.11.0; referenced from the README troubleshooting section.

No version bump (release-time step). `CHANGELOG.md` updated under `[Unreleased]`.

## Test plan

Tested live on SQL Server 2016 and 2019:

- Collector: `@max_files = 1` → validation error; `START` → trace created with `max_files = 5` and rollover on; `START` against an *unbounded* trace → replaced with a bounded one; `START` against a bounded trace → idempotent no-op; `RESTART` / `STATUS` / `STOP` all work.
- `config.data_retention` runs `SUCCESS — Cleaned 51 tables`, no `Msg 22049`.
- `Remove-OrphanedTraceFiles.ps1`: `-WhatIf` preview + real run deleted 181 (2019) / 280 (2016) orphaned files, correctly skipping the running trace's file and locked files.
- `@filecount` rollover-delete behavior verified against MS Learn `sp_trace_create` docs.
- All three existing callers of `trace_management_collector` use named parameters, so the new parameter is safe.


mike-hodgson-icon:
I've been doing some troubleshooting today and was about to add similar comments (to yours) to #951 (but you beat me to it on #972).  Sorry about suggesting xp_delete_file; I didn't know it was hobbled to only work with SQL backup files (you learn something new every day).

I did upgrade the dashboard (full) to 2.11.0 a few days ago and then used the dashboard to upgrade a few of my targeted SQL servers (two SQL 2019 and two SQL 2022 servers) but none of them got the new version of the config.data_retention proc.  The PerformanceMonitor database on each of them still has the older version of the proc with the xp_delete_file code.

erikdarlingdata:
@
Quick heads-up on the workflow: the Dashboard **binary** update (via Velopack auto-update, or a full re-install of the Dashboard app) is **separate** from the **script** update that runs against each monitored SQL Server. Velopack only swaps the .exe/.dll files on your workstation — it does not re-run the install scripts against your targets. To push the new procs out to a server you have to trigger the per-server upgrade explicitly (Manage Servers → pick the server → Install/Upgrade).

That likely explains what youre seeing: the Dashboard binary is current but the procs on your instances are still whatever the script workflow last installed.

A few things that would help pin it down:

1. What does the Dashboard report as its version (title bar / Help → About) — `2.11.0` or something like `2.11.0-nightly.YYYYMMDD`? The PR #973 fix is currently only in `dev` / nightly, not in the v2.11.0 release tag, so even a correctly script-upgraded server on v2.11.0 will still have the `xp_delete_file` block.
2. When you "used the dashboard to upgrade", did you open Manage Servers and click Install/Upgrade on each of the four targets, or was it just the Dashboard app self-updating?
3. On one of the affected servers, can you run this and share the top few rows?
   ```sql
   SELECT TOP 5
       installer_version,
       installer_info_version,
       installation_date,
       installation_type,
       installation_status
   FROM PerformanceMonitor.config.installation_history
   ORDER BY installation_date DESC;
   ```
   That will show exactly which installer version last touched the database and when.
@

mike-hodgson-icon:
Sorry, I didn't explain myself very well.  When I said I "used the dashboard to upgrade", what I meant was that I did: Manage Servers | picked a server | Check for Updates | Upgrade Now.
image

The dashboard binaries that I have installed are v2.11.0 (not a nightly revision).  That will be why `config.data_retention` is still the old version (with the #951 code).  My bad.  I'll just wait for v2.12.0 (or perhaps grab a nightly build).

For completeness, this is the recent install history on one of the affected SQL instances:
image

Coincidentally, v2.11.0 was the first time I used the `PerformanceMonitorDashboard-dashboard-Setup.exe` to install the dashboard (previously I just unzipped directly from `PerformanceMonitorDashboard-<version>.zip` and ran the exe straight from the unzipped dir).  I briefly wondered if PerformanceMonitorDashboard-dashboard-Setup.exe was missing the new code, until I read your explanation above.  I was also considering using PerformanceMonitorInstaller-2.11.0 to upgrade the procs on the SQL instances, but I checked the 43_data_retention.sql script and the upgrade scripts and could quickly see that wouldn't work.

All good, I'll just be patient.  🙂
Thanks for taking the time to reply (and for all the work you're putting into this tool - it really is excellent).

erikdarlingdata:
@
No apology needed at all — your diagnosis is spot on. The per-server upgrade (Manage Servers → Check for Updates → Upgrade Now) did exactly what it should have; the only reason `config.data_retention` is still the old version is that the #973 fix is currently sitting in `dev`/nightly and hasn't made it into a tagged release yet. A correctly-upgraded server on v2.11.0 will still have the `xp_delete_file` block until v2.12.0 ships.

Your instincts were right on both counts, too: `PerformanceMonitorDashboard-dashboard-Setup.exe` isn't missing anything (it's just the Velopack-friendly installer wrapper around the same binaries), and the `PerformanceMonitorInstaller-2.11.0` package wouldn't have helped either — the fixed `43_data_retention.sql` and the new upgrade script only exist on `dev` right now.

So waiting for v2.12.0 is the cleanest path, but if the failing Data Retention job is noisy in the meantime, grabbing a nightly and re-running the per-server upgrade will get you the fix immediately. The `tools/Remove-OrphanedTraceFiles.ps1` script in the PR will also clean up the `.trc` files already on disk.

Thanks for the careful investigation and the kind words — genuinely appreciated. 🙂
@

PR #1183, opened 2026-06-20
Refresh Active Queries sub-tab on view; guard the drill-down race

erikdarlingdata:
## What

Make the **Active Queries** sub-tab re-run its live snapshot fetch each time it becomes the active sub-tab, so the grid is fresh on view instead of showing stale data until the next timer tick. Both apps.

## Why

- **Dashboard** had no auto-refresh on the Active Queries sub-tab — `SubTabControl.SelectionChanged` only reset the drill-down flag and updated the Compare dropdown. Switching to the tab showed whatever was last loaded.
- **Lite** already auto-refreshes the visible sub-tab via `MainTabControl_SelectionChanged`, but its Active-Queries drill-downs raced that auto-refresh (double-load + the "Drill-down: …" indicator could get overwritten with "" depending on async ordering).

## The race

The wait/chart/heatmap drill-downs select Active Queries **programmatically** and then load **filtered** snapshot data. WPF raises `SelectionChanged` *synchronously* inside the `SelectedIndex` setter, so a refresh fired from that handler would run concurrently with (and clobber) the filtered load.

## The fix

A `_suppressActiveQueriesAutoRefresh` flag, set inside a new `SelectActiveQueriesForDrillDown()` helper (try/finally around the tab switch) and honored by the SelectionChanged auto-refresh. Because the suppressed handler returns before its first `await`, the set → (synchronous handler) → finally-clear sequence is airtight; the flag can't leak across an await or get stuck when the target tab is already selected.

**Dashboard** (`QueryPerformanceContent.xaml.cs`, `ServerTab.DrillDown.cs`)
- `SelectionChanged` now runs `RefreshActiveQueriesAsync()` when Active Queries (index 1) becomes selected and the flag is clear.
- Heatmap drill-down, `ShowActiveQueriesForRange`, and the parent `OnQueryDrillDown`/`OnChildChartDrillDown` route through the helper. The now-unused `SelectSubTab(int)` is removed.

**Lite** (`ServerTab.xaml.cs`, `ServerTab.DrillDown.cs`)
- The Cpu/Memory/TempDb/Heatmap drill-downs route through the helper (it spans both `MainTabControl` and `QueriesSubTabControl` index assignments, so both handler firings on a cross-main-tab drill are suppressed).
- `MainTabControl_SelectionChanged` returns early when the flag is set — *after* `UpdateCompareDropdownState()` so the Compare dropdown still updates.

## Out of scope

Lite's blocking/deadlock drill-downs navigate to a different main tab (Blocking) and were left unchanged — see PR comment.

## Test plan (manual)

Both apps, against a server with recent query-snapshot data:
1. **Fresh on view** — switch to another sub-tab, run some queries on the monitored server, then click **Active Queries**: the grid reloads with the latest snapshots (no need to wait for the timer / hit Refresh).
2. **Drill-down not clobbered** — on the Query Heatmap (or a CPU/wait chart in Lite), drill into a point: Active Queries shows the **filtered** set for that window (Dashboard keeps drill-down mode; Lite shows the "Drill-down: …" indicator), *not* a full refresh.
3. Drill in, switch away, click back to Active Queries: it refreshes normally (no stuck-suppression).

Builds clean (both apps, net10.0-windows). Independent code review confirmed the race handling and that no other site programmatically selects Active Queries + loads filtered data.

### DIFF
--- a/Dashboard/Controls/QueryPerformanceContent.xaml.cs
+++ b/Dashboard/Controls/QueryPerformanceContent.xaml.cs
@@ -70,6 +70,10 @@ namespace PerformanceMonitorDashboard.Controls
         private DateTime? _activeQueriesFromDate;
         private DateTime? _activeQueriesToDate;
         private bool _isDrillDownActive;
+        // Guards the Active Queries auto-refresh: SelectActiveQueriesForDrillDown() sets this before
+        // a drill-down flips the sub-tab to Active Queries, so the SelectionChanged handler skips its
+        // refresh and doesn't clobber the filtered snapshot the drill-down loads next (async race).
+        private bool _suppressActiveQueriesAutoRefresh;
 
         // Query Stats state
         private int _queryStatsHoursBack = 24;
@@ -126,12 +130,20 @@ namespace PerformanceMonitorDashboard.Controls
             SetupChartSaveMenus();
             Loaded += OnLoaded;
             Unloaded += OnUnloaded;
-            SubTabControl.SelectionChanged += (s, e) =>
+            SubTabControl.SelectionChanged += async (s, e) =>
             {
-                if (e.Source == SubTabControl)
+                if (e.Source != SubTabControl) return;
+
+                _isDrillDownActive = false;
+                SubTabChanged?.Invoke();
+
+                // Re-run Active Queries whenever it becomes the active sub-tab so the live snapshot
+                // grid is fresh on view. Drill-downs route through SelectActiveQueriesForDrillDown(),
+                // which sets _suppressActiveQueriesAutoRefresh first so this doesn't clobber the
+                // filtered snapshot they load via an async race.
+                if (SubTabControl.SelectedIndex == 1 && !_suppressActiveQueriesAutoRefresh)
                 {
-                    _isDrillDownActive = false;
-                    SubTabChanged?.Invoke();
+                    await RefreshActiveQueriesAsync();
                 }
             };
             ThemeManager.ThemeChanged += OnThemeChanged;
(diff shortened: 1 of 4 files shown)


erikdarlingdata:
**Scope note — Lite blocking/deadlock drill-downs left unchanged (deliberate, want your call).**

`OnBlockingDrillDown`/`OnDeadlockDrillDown` (`Lite/Controls/ServerTab.DrillDown.cs`) navigate to `MainTabControl.SelectedIndex = 8` (Blocking) and then load filtered data, so they have the *same shape* of latent race against Lite's existing `MainTabControl_SelectionChanged` auto-refresh. I did **not** apply the suppress helper to them, for three reasons:

1. **Different feature, out of the task's "Active Queries" scope.** This PR is about the Active Queries sub-tab being fresh-on-view.
2. **It would create a NEW Lite-vs-Dashboard divergence, not parity.** This PR brings Active Queries to Dashboard↔Lite parity. The Dashboard's blocking/deadlock drills go through a different control (Locking) and get **no** auto-refresh change here, so suppressing only Lite's blocking auto-refresh would make the two apps diverge — the opposite of what we want.
3. **Lower severity + unverifiable blast radius.** Blocking has no `LiveSnapshotIndicator`, so the indicator-overwrite bug that motivated the Active Queries fix doesn't apply — the worst case is a redundant double-load of the same range. And the blocking auto-refresh (`RefreshBlockingAsync(subTabOnly:true)`) may load more than the explicit drill load, so suppressing it isn't obviously safe without runtime testing.

If you'd rather I close that race too, the right move is a separate change that handles **both** apps' blocking drills consistently. Happy to do it — just say the word.

PR #1205, opened 2026-06-22
Share PlanViewerControl across both apps via PerformanceMonitor.Ui

erikdarlingdata:
## What

The query-plan viewer control existed as two near-identical 6-file copies (`Dashboard\Controls\PlanViewerControl.*` and `Lite\Controls\PlanViewerControl.*`). The parsing / layout / analysis already lives in `PerformanceMonitor.PlanAnalysis`; only the WPF shell was duplicated. This moves that shell into **one shared control** at `PerformanceMonitor.Ui/PlanViewerControl.*`, used by both apps, so it can no longer drift.

- Added a `.Ui -> .PlanAnalysis` project reference (cycle-free).
- Deleted all 12 old per-app files; repointed every call site in both apps (standalone `PlanViewerWindow`, the MainWindow plan-viewer sub-tabs, and the per-server `ServerTab` plan tabs).

## :warning: Intentional UX change to Lite -- please sign off

**Lite's statement selector changes from a simple ComboBox to Dashboard's richer sortable DataGrid** (the `Statements` panel: #, Query, CPU / Elapsed / UDF or Est. Cost, Critical, Warnings columns; click a row to render that statement; sortable columns; right-click -> Copy Query Text). This is the one user-visible behavior change to Lite. It is easy to veto -- say the word and I'll keep Lite on the ComboBox instead.

## Other unifications (no UI change intended)

- **Async LoadPlan.** The control now parses / analyzes off the UI thread (`await Task.Run(...)`), so a multi-MB showplan no longer freezes the window. Every caller awaits it (or is `async void`). The redundant up-front `XDocument.Parse` validation in Lite's hosts is dropped in favor of the off-thread parse's `XmlException` (same "not valid" dialog, just sourced from the real parse).
- **Cleanup() lifecycle.** Unified on Lite's safe pattern: subscribe to `ThemeManager.ThemeChanged` in the ctor, **never** unsubscribe on `Unloaded` (a `TabControl` raises `Unloaded` on tab switch, which would permanently detach the handler -- a latent Dashboard bug where background plan tabs stopped re-theming), and expose `Cleanup()`. Both apps now call `Cleanup()` at every genuine teardown: plan-window close, plan-tab close, server-tab close, MainWindow sub-tab close, MainWindow plan-viewer close, and the malformed-XML error path.

## Two divergences found while merging (folded in)

- **Row-count badge** now divides actual rows by execution count before comparing to the per-execution estimate -- this was Lite's (correct) behavior and matches `PlanAnalyzer`'s own idiom. Dashboard previously compared total-across-executions actual rows to a per-execution estimate, mis-reporting accuracy for multi-execution operators (e.g. a seek on a nested-loops inner side). Dashboard's badge will now read differently (correctly) for those operators.
- **Operator property values** stay in a read-only **selectable** `TextBox` (Dashboard's form); Lite previously used a non-selectable `TextBlock`, so Lite gains selectable property text.

## Build

Both apps build clean (0 errors): `Dashboard/Dashboard.csproj` and `Lite/PerformanceMonitorLite.csproj`. There are no unit tests for this WPF control, so a clean build is necessary but not sufficient -- **visual verification is the real gate.**

## Visual test checklist (please run in BOTH Dashboard and Lite)

- [ ] Open a single-statement plan and a multi-statement plan; confirm the **statements DataGrid** appears, sorts, and selecting a row re-renders the plan.
- [ ] Zoom in / out / fit, mouse-wheel + Ctrl-zoom, drag-pan, scrollbar behavior.
- [ ] Click a node -> properties panel; hover -> tooltips; right-click -> context menu (copy operator / object / predicate).
- [ ] Ctrl+V paste a plan; paste invalid XML (expect the "not valid" dialog, no crash).
- [ ] Save `.sqlplan`.
- [ ] Switch theme while a plan is open (re-renders, keeps selection); confirm no stale / wrong colors.
- [ ] Open and close the viewer repeatedly (Dashboard window/tab and Lite plan tab) to exercise `Cleanup()` -- no crash, no leak, theme-change still works after reopening.

:robot: Generated with [Claude Code](https://claude.com/claude-code)


erikdarlingdata:
Decision confirmed: keep the sortable **DataGrid** statement selector in both apps (Lite's ComboBox -> DataGrid is intentional and signed off). Rationale: this PR's goal is a single shared control, so splitting selectors would re-introduce per-app divergence inside the shared code; the DataGrid is the strict superset; and the panel only appears for multi-statement plans (single-statement plans are unchanged). No further code change -- ready for visual verification.

PR #1217, opened 2026-06-23
Dashboard apex-identity parity: parse blocker login/host/app

erikdarlingdata:
## Dashboard apex-identity parity — parse blocker login / host / app

Follow-up to the block-chain viewer PRs (#1207, #1215). A lead-blocker (apex) node in the **Dashboard** block-chain viewer now shows **login / host / client app**, like Lite already does. Today it degrades to SPID + db + SQL because Dashboard's `collect.blocking_BlockedProcessReport` only parsed the blocker's *status / tran / SQL* from the report XML — never its identity. Dashboard-only (Lite denormalizes both sides already; the `BlockingPairRow.Blocking{Login,Host,App}` fields exist from #1215, just unpopulated on Dashboard).

### Changes
- **Schema (fresh installs)** — `install/02_create_tables.sql`: add `blocking_login_name`, `blocking_host_name`, `blocking_client_app` (`nvarchar(256) NULL`) to `collect.blocking_BlockedProcessReport`, matching the blocked-side identity column types.
- **Proc parse** — `install/23_process_blocked_process_xml.sql`: extract the three blocking-side identity attributes from `<blocking-process><process @loginname @hostname @clientapp>` in the same `UPDATE` that already types `blocking_status` (exact mirror — same `.value()` pattern, descendant axis, N-prefixed literals).
- **Upgrade (existing installs)** — `upgrades/3.0.0-to-3.1.0/01_add_blocking_identity_columns.sql` (+ `upgrade.txt`): idempotent guarded `ALTER ... ADD` per column (separate `GO` batches) + a one-pass backfill of existing `activity='blocked'` rows from their stored XML, guarded on `blocking_login_name IS NULL`. Mirrors the `2.11.0-to-3.0.0` column-extend upgrade.
- **Query** — Dashboard `BlockingPairRowQuery.Sql`: SELECT the 3 new columns and map them into `BlockingPairRow.Blocking{Login,Host,App}` at **appended ordinals 14/15/16** — existing ordinals (fed to the alert engine via the shared `Read`) are unchanged.
- **Version** — bumped `Version`/`AssemblyVersion`/`FileVersion`/`InformationalVersion` `3.0.0 → 3.1.0` in **both** installer csprojs (`Installer/PerformanceMonitorInstaller.csproj`, `Installer.Core/Installer.Core.csproj`) so the upgrade path triggers.

### Upgrade folder / version used
- New folder: **`upgrades/3.0.0-to-3.1.0/`** (current installer version is 3.0.0; latest existing folder was `2.11.0-to-3.0.0`).
- New installer version: **3.1.0**.
- The installer auto-embeds `upgrades/**/*` + `install/**/*.sql` (Installer.Core.csproj), so the new folder is packaged with no csproj glob change; `targetVersion` comes from the assembly version, so the bump drives `GetApplicableUpgrades(installed, 3.1.0)`.

### XML attribute names — verified against MS Learn
`loginname` / `hostname` / `clientapp` (all lowercase, no underscores), identical on the `<process>` node across the blocked-process-report and deadlock-graph shapes, and across SQL Server 2016+, Azure SQL DB, and Managed Instance. The BPR nests `<process>` directly under `<blocking-process>` (no `<process-list>` wrapper), so the descendant-axis path is correct. Note: the Azure `blocked_process_report_filtered` variant returns these values as the literal `"filtered"` (masked value, attribute still present) — the parse still works.

### Build
Dashboard, Lite, and Installer all build green (0 errors). T-SQL style reviewed against the guide — clean (mirrors the proven scripts).

### Not tested here / your gate
Per instruction, I did **not** run the Installer CLI against the shared test servers — live **fresh + upgrade** install testing (verifying the columns are added, the proc parses, and the backfill populates existing rows) is the pre-merge gate you'll coordinate. Idempotency was written to the established pattern but not exercised against a live DB by me.

**Do NOT merge** — pending your verification + install test.

### DIFF
--- a/Dashboard/Analysis/BlockingPairRowQuery.cs
+++ b/Dashboard/Analysis/BlockingPairRowQuery.cs
@@ -33,7 +33,10 @@ SELECT TOP (5000)
     blocking_sql_text,
     login_name,
     host_name,
-    client_app
+    client_app,
+    blocking_login_name,
+    blocking_host_name,
+    blocking_client_app
 FROM collect.blocking_BlockedProcessReport
 WHERE collection_time >= @collectionWindow
 AND   event_time >= @startTime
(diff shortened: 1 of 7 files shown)


erikdarlingdata:
Closing without merge. The premise was wrong: this PR claimed the block-chain viewer already shows blocker login/host/app in Lite and only Dashboard lacked it. Verified against dev -- neither app's viewer shows identity (the shared BlockingChainControl + BlockingChainNode carry no identity fields, and neither pair-row query selects it), so there is no Dashboard-only gap, and the schema columns this PR adds are read by nothing. Replacing it with an app-side change that adds login/host/app to the shared viewer for BOTH apps, sourced from existing per-process report rows (Dashboard) and the already-denormalized columns (Lite) -- no schema change, no upgrade folder, no version bump.

PR #1242, opened 2026-06-27
Add chart click-to-isolate on legend keys and series lines

erikdarlingdata:
## Chart click-to-isolate

Clicking a multi-series trend chart's built-in **legend key** — or a series' **line** — now dims every other series and **auto-fits the Y axis** to the clicked one, so a wait type (or metric) that sits flat against the axis under the big lines becomes readable. Clicking it again, clicking a different series, or double-clicking (autoscale) restores the full view.

It is a **transient view toggle only**: it never changes the picker selection, never refetches or deletes data, and any re-render (picker/time-range change or a background poll) resets it.

### Why legend-click is the critical path
A wait type shows up in the legend but its line is flat against the axis / hidden under the big lines — there is no visible line to click. The legend key is the only way to grab it, and the Y-auto-fit is what makes it appear.

### Design
The whole mechanic lives in the shared `PerformanceMonitor.Ui/ChartHoverHelper` that every dynamic-legend chart in both apps already routes its series through, so Lite and Dashboard **can't drift**:

- **Identity capture (`Add`)** — captures each series' identity color from `scatter.MarkerStyle.FillColor` (not `scatter.Color`, which `StyleScatter` has already mutated to alpha-215).
- **Click handlers** — `PreviewMouseLeftButtonDown` records the press; `MouseLeftButtonUp` treats it as a click only if movement < 5px (a drag is a pan) and `ClickCount == 1` (so a double-click doesn't re-isolate). Never sets `e.Handled`, so pan/zoom/right-click keep working.
- **Legend hit-test** — discovers the bottom `LegendPanel` from `LastRender.Layout`, reconstructs its rect via `GetPanelRect`, aligns the tight legend layout with `AlignedInside(rect, panel.Alignment)`, and maps a hit to a series by `LegendItem.Plottable` **reference** equality (labels are truncated in the legend, so never matched by string). Implemented exactly per the review-hardened recipe; every member verified by the compiler against the shipped ScottPlot 5.1.58 reference assembly.
- **Isolate / Restore** — dims non-target series via their own identity color `WithAlpha(40)` + drops the gradient fill ribbon (theme-safe across Dark / Light / CoolBreeze); the target is restored to full and re-`StyleScatter`'d. The Y-fit transiently **clears and restores Dashboard's `LockedVertical` axis rule** so `SetLimitsY` actually sticks (Lite installs no rule, so it is a no-op there). Switching directly A→B keeps A's original limits/rules snapshot, so toggling off later restores the true pre-isolate view.
- **AutoFitY** — fits the isolated series' Y over the **current visible X-range** (with a whole-series fallback), padded ~5% each side, with the degenerate-flat guard (`max <= min` → `min..min+1`). Deliberately **not** anchored to zero, so a high-baseline series' own variation is revealed.
- **Reset** — `Clear()` (already called before every re-plot) resets isolate; the rebuild re-installs the `LockedVertical` rule itself.
- **Cross-app autoscale coordination** — a static `ConditionalWeakTable<WpfPlot, ChartHoverHelper>` + `TryGetForChart` lets the per-app handlers find the helper. Both the **"Revert (Autoscale)" menu item** and the **double-click** handler in `Dashboard/Helpers/TabHelpers.cs` and `Lite/Helpers/ContextMenuHelper.cs` call `Restore()` before `AutoScale()` — 4 sites, symmetric across the two apps.

### Files
| File | Change |
|------|--------|
| `PerformanceMonitor.Ui/ChartHoverHelper.cs` | The entire shared mechanic + pure helpers (the existing hover loops now iterate an enriched `_series` list that also carries each series' identity color). |
| `Dashboard/Helpers/TabHelpers.cs` | 2 hooks (Revert menu item + double-click). |
| `Lite/Helpers/ContextMenuHelper.cs` | 2 hooks (Revert menu item + double-click). |
| `Dashboard.Tests/ChartClickIsolateTests.cs` | 20 unit tests. |
| `Lite.Tests/ChartClickIsolateTests.cs` | 20 unit tests (parity mirror). |
| `CHANGELOG.md` | `[Unreleased]` entry. |

### Tested
- **Both apps build green** — `dotnet build Dashboard/Dashboard.csproj` and `dotnet build Lite/PerformanceMonitorLite.csproj`, 0 errors.
- **Both suites pass** — Dashboard.Tests **687 / 687**, Lite.Tests **628 / 628** (includes 20 new isolate tests each).
- Unit coverage (pure logic, no live `WpfPlot`): toggle transitions (`NextIsolate`); dim-vs-restore decision + `FillY` flag (`ResolveSeriesVisual` / `IsolateVisual` / `DimAlpha`); AutoFitY range math (`ComputeIsolateYLimits`) incl. the degenerate-flat guard, visible-X filtering with whole-series fallback, NaN/Infinity handling, and the deliberate no-zero-anchor; axis-rules save/clear/restore bookkeeping (`SaveAndClearRules` / `RestoreAxisRules`).
- A `code-reviewer` pass found two interactive state-machine bugs — an A→B switch corrupting the restore snapshot, and a double-click re-isolating after the double-click handler restored — **both fixed** and re-verified green.

### Needs your runtime / visual sign-off (cannot be unit-tested)
The legend hit-test and Y-fit-with-rule-clear are UI-thread behaviors. I did **not** launch the apps (single-instance would evict your running apps and kill the MCP), so these need your eyes on **Lite + Dashboard wait stats**:

1. **Legend mapping end-to-end** — click a tiny wait's legend key → it isolates, the others dim (fill gone), the Y rescales so it's visible (verify on **Dashboard** specifically, where the `LockedVertical` rule lives).
2. **DPI** — the legend hit selects the correct key at **125% and 150%** display scaling (the bottom legend sits outside `DataRect`, so scale errors show there first).
3. **Restore fidelity** — click again → exact restore (colors, fill, markers, Y limits, axis rule); repeat via **line-click**; **isolate A then click B** then toggle off → restores the original full view; **double-click while isolated** → clean restore; pan / zoom / right-click drill-down still work; switch **theme** while isolated; change **picker** → isolate resets; click a memory-pressure **bar** legend key → no-op (stacked bars deferred). Spot-check memory clerks + perfmon + a file-I/O chart.

**Merge is gated on your visual sign-off — I stopped before merging.**


erikdarlingdata:
## Review + fixes (verified, not relayed)

Re-verified this rather than trusting the implementation report: re-ran both builds and both suites locally, and ran an independent code review + a second-round verification. Found and fixed two real issues the self-review/green-tests had missed — pushed in `f9768e3`.

**Fixed:**
1. **Restore was unfaithful for line-only charts.** `CollectorDuration` and the trend charts build line-only (MarkerSize 0, no fill) and never call `StyleScatter`, but Isolate/Restore re-ran `StyleScatter` on every series — sprouting density markers + a gradient fill they never had. Now `Add` snapshots each series' full visual state and `RestoreSeriesVisual` writes it back, re-running `StyleScatter` only for fill charts (it regenerates the gradient from the unchanged data). Faithful for fill, line-only, and flat-StyleScatter'd series. **+2 headless regression tests per suite.**
2. **Double-click determinism.** Replaced reliance on `e.ClickCount` on the terminal mouse-up (uncertain WPF semantics) with a `MouseDoubleClick` suppress flag, consumed at the top of `OnLeftButtonUp` — verified against decompiled `Control.HandleDoubleClick` that the 2nd-down's `PreviewMouseLeftButtonDown` is marked Handled (so `_leftPressed` is already false there, and consuming the flag late would leave it stuck and swallow the next genuine click).

Plus a marker-color safety tweak (set via `Color`, not `MarkerStyle.FillColor`).

**Verified state:** both apps build green; **Dashboard.Tests 689/0, Lite.Tests 630/0.**

**Known minors (acceptable, not blocking):** mixed-chart event-marker overlays that aren't registered with the helper aren't dimmed; "click empty area to restore" is unreliable inside a dense data region (re-clicking the key/line and double-click restore reliably).

**Still gated on visual sign-off** (UI-thread, not unit-testable):
- Legend-key click selects the right series at 100/125/150% DPI on a wrapped bottom legend, and the Y rescales so a tiny series becomes visible (**check on Dashboard** — the `LockedVertical` rule lives there).
- Restore fidelity on a **line-only chart (CollectorDuration)** — no phantom markers/fill after isolate→restore.
- Double-click-while-isolated → clean restore; theme switch while isolated; picker change resets; pan/zoom/right-click drill-down still work; memory-pressure **bar** legend key is a no-op.
- Spot-check memory clerks + perfmon + a file-I/O chart.

PR #1274, opened 2026-06-30
Surface dropped columns in Active Queries + Wait drilldown (both apps)

erikdarlingdata:
**Drilldown-completeness batch, item 2** — the clearest "pay to collect, then drop" case.

Both surfaces read sp_WhoIsActive's `query_snapshots`, and the collector turns on `@get_memory_info=1` + `@get_task_info=2` — then the read queries select none of it.

**Dashboard** — surface `requested_memory`, `granted_memory`, `max_used_memory`, `tasks`, `physical_io` (live column names confirmed on the view; appended to all 4 snapshot SELECTs, reader ordinals 28–32, comma-stripped like the existing `used_memory`/`reads`). Both the Active Queries grid and the Wait drilldown get those 5 columns. The **Wait drilldown** additionally gains: an **Additional Info** column (the model already carried it; the grid had no column — Active Queries shows it), and a **Query Plan download button** (on-demand fetch via `GetQuerySnapshotPlanAsync`, mirroring Active Queries — its context-menu View Plan already worked, but there was no visible button).

**Lite** — surface `query_hash` (collected but dropped from the snapshot SELECT/model/grid) so a live session can be joined to its `query_stats` / Query Store history. Added to `GetLatestQuerySnapshotsAsync` + both wait-snapshot reads + both grids. (Lite's `query_snapshots` collects no extra memory-grant columns beyond the already-shown `granted_query_memory_gb`.)

## ⚠ Decision needed — a pre-existing units bug this surfaced
The memory columns show sp_WhoIsActive's **raw 8KB-page count under a "(MB)" label**. The *already-shipped* `Used Mem (MB)` column has the same mislabel — `14,307` reads as 14 GB but is really ~112 MB. I kept the **new** columns consistent with the existing one (rather than silently diverging). Fixing it accurately means dividing **all four** memory columns by 128 (pages → MB) — a display change to a shipped column. **Want me to apply the accurate MB conversion across all four?** (I can verify the exact sp_WhoIsActive unit first.)

## Verification
- Live column names confirmed on `report.query_snapshots`. Dashboard.Tests 732/732; both apps build clean; rebased onto current dev (merged cleanly with the tempdb edits in the same files).

Held for your review.

### DIFF
--- a/Dashboard/Controls/QueryPerformanceContent.xaml
+++ b/Dashboard/Controls/QueryPerformanceContent.xaml
@@ -179,6 +179,14 @@
                             </StackPanel>
                         </DataGridTextColumn.Header>
                     </DataGridTextColumn>
+                    <DataGridTextColumn Binding="{Binding PhysicalIo, StringFormat='{}{0:N0}'}" ElementStyle="{StaticResource NumericCell}" Width="100">
+                        <DataGridTextColumn.Header>
+                            <StackPanel Orientation="Horizontal">
+                                <Button Style="{DynamicResource ColumnFilterButtonStyle}" Tag="PhysicalIo" Click="ActiveQueriesFilter_Click" Margin="0,0,4,0"/>
+                                <TextBlock Text="Physical I/O" FontWeight="Bold" VerticalAlignment="Center"/>
+                            </StackPanel>
+                        </DataGridTextColumn.Header>
+                    </DataGridTextColumn>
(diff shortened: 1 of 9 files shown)


erikdarlingdata:
Units **fixed** in the latest commit: the four memory columns (Used / Requested / Granted / Max Used) now divide sp_WhoIsActive's 8KB-page count by 128 to show true MB, so the existing `Used Mem (MB)` label is finally accurate (14,307 pages -> ~112 MB, not 14 GB). Dashboard-only; Lite's memory is already GB. No remaining decision on this PR.

Issue #1321, opened 2026-07-02
[QUESTION] Help with Architecture and Security clearance for the tool


rvsuresh19:
### Which component is your question about?

Lite

### Performance Monitor Version

3.1.0

### Is your question about how it works, or the results?

How the tool works

### What's your question?

I was initially very hesitant to post this question because I felt it might be too much to ask. However, I thought I would at least give it a try, so I would appreciate any guidance you can provide.
Since Managed Identity support (#1036) is now available, our Database Service Owner has requested that I present the tool to our Architecture Board and Security Team to obtain approval before onboarding several hundred Azure databases within our tenant.

Presenting solutions to architecture and security review boards is not something I do regularly, so I am a little nervous about securing the necessary clearance. I am currently preparing the documentation using the content available in the README and Contributing sections, covering the architecture overview, collectors, and configuration components.

From a security perspective, the documentation mainly discusses responsible vulnerability disclosure and not publicly sharing newly discovered vulnerabilities. Would it be possible to share any additional security-related insights, documentation, or recommendations that could help during a security review, even if only in a high-level or "light" version suitable for an architecture/security audience?
For example:

- How are vulnerabilities in DuckDB version used by the tool tracked, assessed, and remediated?
- The executable is always flagged as unsafe when downloaded, despite multiple submissions to Microsoft feedback channels indicating that it is a legitimate and safe tool. Is there any documentation that explains this behavior or provides guidance for security reviewers?
-
Are there any other security and operational considerations that would help architecture and security teams evaluate and approve the tool?

### Additional Context

Azure SQL

erikdarlingdata:
Hi Suresh,

Great to see #1038 paid off and you're far enough along to be taking this in front of an architecture/security board. I dug into the actual code, the build pipeline, and the repo's GitHub security settings rather than just repeating the README back to you - and in the process of verifying my own answer, found a real bug, so this reply covers that too rather than waiting until it's fully resolved.

## DuckDB vulnerability tracking

Lite pulls DuckDB through two NuGet packages, `DuckDB.NET.Data` and `DuckDB.NET.Bindings.Full`, currently pinned at 1.5.2. DuckDB.NET's version number tracks the native DuckDB core version it bundles 1:1, so 1.5.2 means native DuckDB core 1.5.2, not an old core wrapped in a newer shim.

[...]

PR #1327, opened 2026-07-02
Sign the Velopack (Setup.exe) build output -- currently ships unsigned

erikdarlingdata:
## The problem

**Setup.exe -- the install method the README recommends first -- ships completely unsigned**, and has since Velopack packaging was introduced (v2.4.0, 2026-03-23), including every release since Setup.exe became the *recommended* path (v2.11.0, 2026-05-19). Verified directly against the live v3.1.0 release, not just by reading the workflow:

```
PS> Get-AuthenticodeSignature .\PerformanceMonitorDashboard-dashboard-Setup.exe | fl Status
Status : NotSigned

PS> Get-AuthenticodeSignature .\explicit-zip\PerformanceMonitorDashboard.exe | fl Status
Status : Valid   (chains to CN=SignPath Foundation, issued by GlobalSign)
```

`PerformanceMonitorDashboard-dashboard-Portable.zip` (Velopack's own portable-zip output) and its bundled `Update.exe` are also unsigned. Only the legacy `PerformanceMonitorDashboard-3.1.0.zip` / `PerformanceMonitorLite-3.1.0.zip` / `PerformanceMonitorInstaller-3.1.0.zip` -- built and signed via the workflow's own explicit `Compress-Archive` + SignPath steps -- are actually signed.

**Root cause:** the Velopack path publishes a separate self-contained build (`publish/Dashboard-velopack`, `publish/Lite-velopack`) that `vpk pack` consumed directly. Only the framework-dependent `publish/Dashboard` / `publish/Lite` / `publish/Installer` trees (used for the ZIPs) were ever uploaded to SignPath. SignPath signing was wired up in v2.6.0, when the ZIP was still the primary artifact; when Setup.exe was promoted to primary in v2.11.0, signing was never extended to cover it.

This means every "unsafe/flagged" download report against Setup.exe (see #1321) has a real, mundane cause -- not just SmartScreen reputation lag, an actually-missing signature that no amount of resubmitting to Microsoft's feedback portal will fix.

## The fix

Mirrors the exact upload/sign pattern already used 3x in this file for Dashboard/Lite/Installer: upload the two velopack publish trees for signing right after they're built, submit to SignPath, then point `vpk pack -p ...` at the signed output (`signed/Dashboard-Velopack`, `signed/Lite-Velopack`) instead of the raw `publish/*-velopack` directory.

## Open questions before merging -- please review carefully

1. **SignPath artifact-configuration may need adjustment.** The self-contained publish is a much larger tree (bundled .NET runtime DLLs) than anything the existing `Dashboard`/`Lite` artifact-configuration slugs have signed before (which only ever saw the framework-dependent tree). I reused the same slugs as a first attempt, but I can't verify from the repo whether their file-matching glob correctly targets just `PerformanceMonitor*.exe` (what actually needs a fresh signature) versus trying to touch bundled runtime DLLs (already Microsoft-signed, don't need it, and might not even be glob-matched the same way). You may need to check/adjust this in the SignPath web dashboard, or this PR's first real run will surface it.
2. **This doubles SignPath signing requests per release** (3 -> 5: Dashboard, Lite, Installer, Dashboard-Velopack, Lite-Velopack). If SignPath's approval flow is manual per request, this means more manual clicks per release, not just more CI time.
3. **Cannot be verified outside a real `release` event.** `pull_request`-triggered runs skip every step in this diff (all gated on `github.event_name == 'release'`), so this PR's own CI run will build/test green but prove nothing about signing. The only way to validate this is a real tag/release (or a throwaway pre-release) that actually calls SignPath with real credentials.
4. **Possible cleaner alternative for later, out of scope here:** unify the ZIP and Velopack publishes into one self-contained build so both derive from the same signed tree instead of building (and now signing) two separate trees per app. That's a bigger change with real tradeoffs (ZIP size, air-gapped-user runtime assumptions) and a product decision, not a quick fix -- flagging it, not doing it here.

Given the above, this is a **draft PR** -- recommend testing on a patch release before merging to dev, not merging straight through.

## Test plan

- [x] YAML validated (`python -c "import yaml; yaml.safe_load(...)"`)
- [x] Diff mirrors the existing signing pattern exactly (same action, same secrets, same `wait-for-completion`/`output-artifact-directory` shape)
- [ ] Real `release` event run confirms SignPath accepts the Velopack artifact-configuration and returns a validly signed `PerformanceMonitorDashboard.exe`/`PerformanceMonitorLite.exe` inside Setup.exe
- [ ] `Get-AuthenticodeSignature` on the resulting Setup.exe from that test release comes back `Valid`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -235,6 +235,56 @@ jobs:
           Remove-Item "releases/PerformanceMonitorInstaller-$version.zip" -ErrorAction SilentlyContinue
           Compress-Archive -Path 'signed/Installer/*' -DestinationPath "releases/PerformanceMonitorInstaller-$version.zip" -Force
 
+      # The Velopack (Setup.exe) path publishes a SEPARATE self-contained build
+      # (publish/Dashboard-velopack, publish/Lite-velopack -- see the "self-contained
+      # for Velopack" steps above) that previously went straight into `vpk pack`
+      # without ever being uploaded to SignPath. Only the framework-dependent trees
+      # used for the legacy ZIPs were signed, so every Setup.exe shipped unsigned
+      # since Velopack packaging was introduced. These steps close that gap by
+      # mirroring the exact upload/sign pattern used for Dashboard/Lite/Installer
+      # above, and vpk pack below now reads from the signed output.
+      - name: Upload Dashboard (Velopack) for signing
+        if: github.event_name == 'release'
+        id: upload-dashboard-velopack
+        uses: actions/upload-artifact@v6
+        with:
+          name: Dashboard-Velopack-unsigned
+          path: publish/Dashboard-velopack/
+
+      - name: Upload Lite (Velopack) for signing
+        if: github.event_name == 'release'
+        id: upload-lite-velopack
+        uses: actions/upload-artifact@v6
+        with:
+          name: Lite-Velopack-unsigned
+          path: publish/Lite-velopack/
+
+      - name: Sign Dashboard (Velopack)
+        if: github.event_name == 'release'
+        uses: signpath/github-action-submit-signing-request@v2
+        with:
+          api-token: '${{ secrets.SIGNPATH_API_TOKEN }}'
+          organization-id: '7969f8b6-d946-4a74-9bac-a55856d8b8e0'
+          project-slug: 'PerformanceMonitor'
+          signing-policy-slug: 'release-signing'
+          artifact-configuration-slug: 'Dashboard'
+          github-artifact-id: '${{ steps.upload-dashboard-velopack.outputs.artifact-id }}'
+          wait-for-completion: true
+          output-artifact-directory: 'signed/Dashboard-Velopack'
+
+      - name: Sign Lite (Velopack)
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Merging into dev at Erik's request. Flagging for the record since this is the one caveat that matters:

Merging to dev is safe on its own -- the new signing steps are gated on `github.event_name == 'release'`, so nothing here runs (or touches SignPath) on a dev push, only build+test like any other change. That's exactly what the green check above confirms, and nothing more -- it does NOT validate that SignPath actually accepts a signing request for the larger self-contained Velopack tree, or that the `Dashboard`/`Lite` artifact-configuration slugs handle it correctly.

**The real test happens at the next actual release cut.** Whoever cuts it should watch the "Sign Dashboard (Velopack)" / "Sign Lite (Velopack)" steps specifically, and after publish, verify with `Get-AuthenticodeSignature` on the resulting Setup.exe before calling it done -- the same way this gap was originally found. If SignPath rejects the artifact-configuration, that release's build will fail loudly at the sign step rather than silently shipping unsigned again, which is a safe failure mode, but worth being ready for.

Issue #1262, opened 2026-06-30
[FEATURE] Add a headless Windows Service collector mode for Lite (gMSA-compatible), with the existing UI as a read-only viewer


gotqn:
### Which component(s) does this affect?

- [ ] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

### Problem Statement

Lite is a WPF GUI app whose collectors only run inside the GUI (CollectionBackgroundService is started from MainWindow.Loaded; Program.Main always calls app.Run()). Unattended collection therefore requires a dedicated user with autologon and a persistent interactive desktop session — a security/compliance concern. There's no way to run Lite collection under a service identity, including a gMSA with auto-rotating credentials. Full edition's headless option (SQL Agent jobs) requires deploying objects into each monitored instance, which is exactly what Lite's agentless model avoids.

### Proposed Solution

1. Extract the collection core (CollectionBackgroundService, RemoteCollectorService, DuckDbInitializer, ServerManager, ScheduleManager, archival/retention) into a shared library + a new headless host (.NET Worker) registered as a Windows Service via UseWindowsService(), runnable under LocalSystem / a service account / a gMSA.
2. Keep the WPF app as a viewer: remove its in-process collector startup; have it read the store. Resolve the DuckDB single-writer constraint by opening read-only (custom connection factory + WAL reader isolation) or by querying the service's MCP/localhost endpoint.
3. Ship as an optional service package so the current single-process Lite stays supported (backward compatible).

### Use Case

**Acceptance criteria**: service installs and runs under LocalSystem and a gMSA, starts at boot with no logon, survives reboot; reads servers.json + Credential Manager under the service identity (ACLs verified); the viewer renders dashboards against a store the service is actively writing, with no corruption; schedule/retention edits are honored by the service; existing desktop mode unchanged.

### Alternatives Considered

Currently, my setup is:

- dedicated account running Lite
- additional settings for surviving restart and auto-start
- additional settings (ssh) to expose the MCP server to remote users 
- if more then one dev wants to see the dashboard, a team call can be made as only one can log into the dedicated account 

### Additional Context

**Change size**: Medium — new host project + refactor to a shared collection assembly; the hard parts are cross-process DuckDB access and credential/ACL access under a service account, both needing careful testing.

[...]

PR #1370, opened 2026-07-04
Darling security Phase 1: collect/config schema split + least-privilege roles (#1262)

erikdarlingdata:
Implements **Phase 1** of the Darling security-hardening design (`plans/darling-security-hardening.md`). Do not merge — pre-release hardening, gated on Erik's call on the open items below. Phases 2 (LAN/TLS) and 3 (per-user auth) are out of scope.

## The problem
The bundled-managed Postgres store kept everything in one flat `public` schema, and **every** consumer — including the interactive Viewer, which only ever needs to read plus write three small config tables — connected as the `darling` cluster superuser. A loopback socket handing out a superuser credential is exactly what an enterprise scan flags, and it means a buggy/compromised Viewer can `DROP` the whole history.

## The schema/role/grant model
Two schemas, table names unchanged (only the schema moves):
- **`collect`** — 26 collector hypertables + service-written/user-read metadata (`servers`, `collection_log`, `analysis_findings`, `darling_schema_version`, the 24 `v_*` views). Read-only to users. *(`analysis_findings` → `collect`, per the confirmed fork #1: it's service-written / viewer-read-only.)*
- **`config`** — exactly the operator-writable tables (`config_mute_rules`, `config_alert_log`, `config_edge_trigger_watermarks`, `analysis_muted`).

Three roles: `darling` (owner/superuser, the service) unchanged; **`admin`** (`LOGIN NOSUPERUSER`, SELECT both schemas + INSERT/UPDATE/DELETE on `config` only — the Viewer's default); **`viewer`** (SELECT only). Grants ride `ALTER DEFAULT PRIVILEGES FOR ROLE darling` so new collector tables auto-inherit SELECT. Public hardening (`REVOKE CREATE ON SCHEMA public`, `REVOKE ALL ON DATABASE`) with a **`GRANT CONNECT ... TO admin, viewer`** re-grant — a correctness fix over the design's literal DDL, since `REVOKE ALL FROM PUBLIC` drops the roles' implicit CONNECT (validated live: `admin` connects after the REVOKE only because of the re-grant).

## search_path, not re-qualify
The decisive constraint: Darling's SQL is byte-identical to Lite's DuckDB SQL (a twinning contract pinned by tests). Re-qualifying ~50 sites with `collect.`/`config.` would fork that twin permanently. Instead the bare names resolve through **`search_path = collect, config, public`**, established three ways: `MigrateAsync` best-effort sets it as the **database default** (a least-privilege BYO login that can't `ALTER DATABASE` is warned, not failed); the managed connection strings carry `Search Path`; and the migrate session sets it so the post-V8 `darling_schema_version` stamp (that table moves to `collect`) resolves. **Zero query files edited** — proven live (see below).

## V8 migration
`StorageVersion` 7→8. `CREATE SCHEMA` + `ALTER … SET SCHEMA` generated from the collector catalog (`PgSchemaGenerator.GenerateV8Move`), so a future collector moves for free. Runs on a **fresh** store (V1–V7 make plain tables in `public` → V8 moves them → TimescaleDB then converts the now-in-`collect` tables) and on an **existing** hypertabled store (`ALTER TABLE <hypertable> SET SCHEMA` moves the hypertable + propagates to chunks). The `ALTER DATABASE search_path` is deliberately **outside** the V8 transaction (best-effort), so a BYO privilege gap can't roll back the moves.

## Compressed-hypertable move — worked, no fallback needed
The design's flagged validation item. Tested live on **TimescaleDB 2.28.1**: build a hypertable, insert 40 days, compress a chunk, `GRANT SELECT` to the least-privilege role, `ALTER TABLE … SET SCHEMA collect` — the compressed hypertable moves and the least-privilege role reads all 40 rows through compression. The decompress→move→recompress and fixture-sacrifice fallbacks were **not** needed.

## Per-role DPAPI credentials + Viewer degradation
`DarlingManagedRoles` (managed-only, after migrate, before Timescale) reads-or-generates `pg-admin-credential.dpapi` / `pg-viewer-credential.dpapi` beside the data dir (same DPAPI-LocalMachine posture as the owner's), then runs the idempotent, self-healing provisioning DDL. The Viewer drops to **`admin` by default** via `darling.json` `postgres.connectAs` (`admin`|`viewer`). A `viewer` deployment degrades gracefully: a `has_table_privilege` capability probe drives `ViewerDataService.IsReadOnly`, which **hides/disables** the mute-rule Add/Edit/Toggle/Delete/Purge buttons and the alert Dismiss/Dismiss All buttons; a reactive `42501` catch turns any write that slips through into a clear "read-only connection" message. *(Finding mute/unmute has no viewer UI yet and already degrades — `PgFindingStore` swallows the write; noted for whoever adds that surface.)*

## BYO posture (first-class)
The schema split runs everywhere (it's a migration). Role provisioning + DPAPI are managed-only; BYO ships `Darling/tools/provision-roles.sql` (the same DDL, password placeholders) and a documented `ALTER DATABASE … SET search_path` step for a least-privilege login.

## Per-file migration touch list
- **Storage**: `StorageVersion` 7→8; `PgSchemaGenerator.GenerateV8Move()` + `CollectSchema`/`ConfigSchema`/`SearchPath`/metadata+config+view lists; `PgMigrations` V8 entry, session `SET search_path`, `MigrateAsync` logger overload + best-effort `TrySetDatabaseSearchPathAsync`.
- **Service**: new `DarlingManagedRoles`; `DarlingManagedPostgres` credential path helpers + `Search Path` on `BuildConnectionString`; `DarlingWorker` provisioning hook; `DarlingConfig.PostgresConfig.ConnectAs`.
- **Viewer**: `ViewerSettings` role-aware derivation + `Search Path`; `ViewerDataService` `IsReadOnly` probe + `ExecuteWriteAsync`/`ViewerReadOnlyException` (mute-rule + alert-dismiss writes routed through it); `MainWindow` probe; `MuteRulesWindow` + `AlertsHistoryTab` proactive gating.
- **Query SQL: none re-qualified.**

## Test evidence
- **Ungated**: 480 Darling tests + 864 Lite tests green. V8 shape from the catalog, provisioning DDL shape + alphanumeric-password guard, `connectAs` derivation (admin/viewer/unknown), read-only probe + exception, `BuildConnectionString`/`ViewerSettings` search path, version pins 7→8.
- **Gated live** (stood up a scratch PG + TimescaleDB 2.28.1 on port 55632): **555 Darling tests green** — my 3 new `DarlingSecuritySplitLiveTests` (tables move, `admin` writes / `viewer` denied 42501 / new `collect` table auto-grants, compressed-hypertable move) **plus every pre-existing gated test**, which stay green post-move and prove the search_path approach needs zero query edits. The exact production provisioning DDL runs clean + idempotent.

## Open for Erik
1. Live-validated on a throwaway scratch instance; a run against your managed dev fixture (`DARLING_TEST_PG`) would confirm on real collected data.
2. Visual sign-off on the read-only Viewer degradation (button hide/disable) — I don't launch the worktree apps (single-instance would evict your running Dashboard/Lite).


erikdarlingdata:
## Review fixes pushed (commit `3005c19`)

Addresses the pre-merge review. **FINDING 2 (role naming) left untouched** pending Erik's call; **FINDING 4** already mitigated by `RequireAlphanumeric`.

### FINDING 1 (HIGH) — credential-file ACLs — the principal model to review
DPAPI LocalMachine scope is correct and unchanged (service writes, a *different* interactive user's Viewer reads); the fix is the **file ACL**, since a machine-bound blob + in-source entropy means "can read the file" == "can decrypt it". New `DarlingFileSecurity` strips the inherited world-read and re-grants narrowly:

| File(s) | Readable by | Rationale |
|---|---|---|
| `pg-credential.dpapi` (**superuser**) + transient init pwfile | SYSTEM, Administrators, service account | The RCE vector — locked hardest. The Viewer no longer needs it (it connects as admin/viewer post-split). |
| `pg-admin-credential.dpapi`, `pg-viewer-credential.dpapi` | above **+ `NT AUTHORITY\INTERACTIVE`** | The Viewer runs as the interactive operator, whose identity is unknown at provisioning time. INTERACTIVE reads with zero config and **excludes** the non-interactive local attack primitives the design's threat model names (services, SSRF/sandboxed sockets, scheduled tasks). |

**Assumption (please sanity-check this specifically):** the single-operator VM this edition targets, where `INTERACTIVE` == the operator. On a shared multi-user machine where untrusted users log on interactively, they could read the admin/viewer credentials (never the superuser one) — the documented tightening path is a configured operator account ACL'd by hand (a future knob, not built for the Phase 1 default). This is in `DarlingFileSecurity`'s XML doc, the README security section, and the commit message.

**Also:** a pre-plant guard (`IsTrustedOwner`) refuses/regenerates a credential file not owned by SYSTEM/Administrators/the service account (closes the seed-a-known-password vector). Hardening is best-effort + loud `LogError` (never bricks the service) and re-applied every start; it covers the pre-existing owner credential too.

### FINDING 3 (LOW) — sequence default privileges
`GRANT USAGE, SELECT ON SEQUENCES TO admin` added to both `BuildProvisioningSql` and `provision-roles.sql`.

### Validation (live, TimescaleDB 2.28.1)
- 484 ungated Darling tests green (incl. 4 new `DarlingFileSecurityTests`: protected DACL, no world-read ACE, per-file-kind grants, trusted-owner).
- **Managed bootstrap E2E green WITH hardening** — initdb/start/credential-read all unaffected by the ACLs.
- Full gated suite (559) green.
- Updated provisioning DDL runs clean, and `admin` successfully `INSERT`s into a `serial`-backed `config` table — proving the sequence grant (it fails without it).

PR #1461, opened 2026-07-09
Darling network endpoints Phase 2: MCP host network exposure

erikdarlingdata:
## Darling network endpoints — Phase 2 (MCP host)

Phase 2 of the locked `darling-network-endpoints` plan (D3): opt-in, secure-by-default LAN exposure for the embedded MCP server. Depends on Phase 1 (#1459, merged) — uses `McpConfig.Network`, `DarlingNetwork.IsExposedListenAddress`, `DarlingManagedPostgres.TryBuildMcpConnectionStringFromStoredCredential` / `McpCredentialPathFor` / `BuildFirewall*Command`, and the `mcp` role. **Only `Mcp/DarlingMcpHostService.cs` + its tests** — disjoint from Phase 3.

Default (no `mcp.network` block) = byte-for-byte today's loopback-only, tokenless HTTP; existing local clients are unaffected.

### Implemented (mapped to scope)
1. **`ResolveMcpBind` pure fn -> `(McpBindMode Mode, McpBindReason Reason)`** — `NetworkAndLoopback` only when the listen is non-loopback (Phase-1 classifier, so `127.0.0.1` -> loopback single-bind, no collision) AND `postgres.managed` AND a token is present AND `allowFrom` is a valid CIDR; else `LoopbackOnly` + the specific reason. Not a bare enum; `LogBindReason` (the caller) maps reason -> severity so tests assert `(Mode, Reason)` with no logger.
2. **Host-side enforcement (not `Validate()`)** — gated on `ResolveMcpBind` at the bind site. Missing token/allowFrom -> loopback + `LogCritical`; BYO (`managed==false`) with any `network.*` set -> loopback + `LogWarning` (D-BYO). A token-decrypt failure also fail-closes to loopback + `LogCritical`.
3. **Token middleware** — installed **only** in `NetworkAndLoopback` mode; constant-time (SHA-256 + `FixedTimeEquals`, no length leak), 401 on empty/missing/mismatch, **no loopback exemption**. `LoopbackOnly` stays tokenless.
4. **In-app CIDR middleware** — 403 for `RemoteIpAddress` outside `allowFrom`; loopback (`127.0.0.0/8`, `::1`, IPv4-mapped) always allowed (Round-4 #2). Enforced in-app so off-CIDR refusal holds regardless of the firewall.
5. **Bind** — `options.Listen(IPAddress.Parse(listen), port)` (specific family) + both loopback families, skipping the loopback `Listen`(s) when the listen is loopback or `0.0.0.0`/`::` (collision guard). Real bind logged (fixes the `http://localhost` log); port precheck probes the actual bind address.
6. **Store pool = the `mcp` role** via `TryBuildMcpConnectionStringFromStoredCredential` (not the superuser owner); credential poll extended to ~300s (60x5s) for the post-migration `mcp` cred; fail-closed on timeout.
7. **Firewall** — best-effort, symmetric (add when exposed / remove when not), scoped, port-specific named rule reusing Phase 1's tested command builders; never fatal, logs the scoped command on failure.
8. **Fixed the stale `:27` doc** ("six analysis MCP tools" -> the true ~60-tool surface + the network-exposure posture).

### Tests (same PR)
`DarlingMcpHostTests.cs`, 57 cases: `ResolveMcpBind` `(Mode, Reason)` matrix (exposed+token+allowFrom+managed -> NetworkAndLoopback for IPv4/IPv6/wildcard; token missing -> critical; allowFrom missing/`not-a-cidr`/`/33`/no-prefix -> critical; managed=false -> BYO; listen=127.0.0.1 & no-network -> loopback default; encryptedToken counts as present; BYO-configured-but-not-exposed -> BYO); `ShouldAddLoopbackListeners` (specific IP -> add, loopback/wildcard -> skip); `IsRemoteAddressAllowed` (in/out CIDR, loopback-exempt v4/v6/mapped, null fail-closed, IPv6 CIDR); `IsBearerTokenAuthorized` / `ExtractBearerToken`. Config parse / `ResolveToken` / `Validate()` are already pinned by Phase 1's `DarlingConfigTests`.

### Testing
- `dotnet build Darling/PerformanceMonitor.Darling.Service/...csproj -c Debug` -> **Build succeeded**, 0 errors, 0 warnings from the edited file.
- `dotnet test Darling/Darling.Tests/...csproj -c Debug` -> **Passed! Failed: 0, Passed: 1794, Skipped: 127** (127 = the live/gated tests needing a server; includes the 57 new tests). The known-flaky `ViewerWave2DisplayTests.BlockedRow_EventTimeLocal_...` passed.
- Live enable/disable/off-CIDR/token round-trip on DARLING01 is the reviewer's pre-merge step per the plan's rollout.

### FLAGS (per the "flag, never silently drop" contract)
1. **`mcp`-role store pool is applied UNCONDITIONALLY (both loopback and network mode), not only when exposed.** D3-d's prose sits under "when the listen is non-loopback," but Phase 2 scope item 6 ("Store pool = the mcp role ... NOT the superuser owner"), the plan's Files line ("point `WaitForManagedConnectionStringAsync` at it" — the single store-pool path), and least-privilege intent are all unconditional, so I made it unconditional. **Consequence:** the DEFAULT loopback MCP now connects as `mcp` instead of the superuser owner. This relies on the Phase-1 `mcp` role covering the full read surface + the `analysis_findings`/`analysis_muted` INSERTs the ~60 tools + `analyze_server` + `mute` need — which is already on the rollout live-validation checklist ("`analyze_server` persists (mcp grant) not just mute"). If a tool turns out to need a grant the `mcp` role lacks, the fix is the grant (Phase 1), per D3-role ("fix the tool, don't un-carve").
2. **Middleware order = token first, then CIDR.** My task said the token is "FIRST in the pipeline"; D3-c said CIDR "before/with the token check." Token-first with CIDR adjacent-and-before-`MapMcp` satisfies both, so I went token-first. Both gates must pass and CIDR never authorizes (only rejects/passes-through), so order creates no bypass; trivial to swap to CIDR-first if preferred.
3. **In exposed mode, loopback clients on the box now also require the token** (the token middleware has no loopback exemption — Round-4 #6). Operators enabling `mcp.network` must add the bearer token to their *local* MCP clients too. Documented in the class summary; the README/sample copy is Phase 3.
4. **`RunPowerShellAsync` is duplicated** from `DarlingManagedPostgres` (private there, on a `[SupportedOSPlatform("windows")]` class). I reused the security-sensitive part (the tested pure `BuildFirewall*Command` builders) and duplicated only the mechanical process runner, to honor the "don't touch Phase 1 files except to call" boundary. A shared extraction (e.g. widening it to `internal`, or a Common helper) is a clean future consolidation.
5. **Informational — plan/Phase-1 comment nit:** `System.Net.IPNetwork.TryParse` does not reject host-bits-set CIDRs; it **normalizes** them (`192.168.1.5/24` -> `192.168.1.0/24`). So an `allowFrom` with host bits is accepted and canonicalized (the firewall `RemoteAddress` and `IPNetwork.Contains` both get the canonical form). Fail-safe, and Phase 1's store side is unaffected (it already emits `BaseAddress/Prefix`). The "IPNetwork requires zeroed host bits" wording in the plan / a Phase-1 comment is slightly off, but no code change is needed.

### DIFF
--- a/Darling/Darling.Tests/DarlingMcpHostTests.cs
+++ b/Darling/Darling.Tests/DarlingMcpHostTests.cs
@@ -7,6 +7,7 @@
  */
 
 using System.Net;
+using Microsoft.Extensions.Logging;
 using PerformanceMonitor.Darling.Service;
 using Xunit;
 using Host = PerformanceMonitor.Darling.Service.Mcp.DarlingMcpHostService;
@@ -83,6 +84,34 @@ public sealed class DarlingMcpHostTests
     public void ResolveMcpBind_Exposed_Managed_Token_BadAllowFrom_IsLoopbackOnly_AllowFromInvalid(string? allowFrom)
     {
         var decision = Host.ResolveMcpBind(Mcp("192.168.1.205", allowFrom, token: "s3cr3t"), managed: true);
+
+        Assert.Equal(Host.McpBindMode.LoopbackOnly, decision.Mode);
+        Assert.Equal(Host.McpBindReason.AllowFromInvalid, decision.Reason);
+    }
+
+    [Theory]
+    [InlineData("localhost")]  // a name, not an IP (a plausible "I meant loopback" typo)
+    [InlineData("myhost")]     // a hostname
+    [InlineData("db.lan")]
+    [InlineData("*")]          // the postgres listen wildcard, not a Kestrel-bindable IP
+    public void ResolveMcpBind_Exposed_Managed_NonIpListen_IsLoopbackOnly_ListenInvalid(string listen)
+    {
+        /* A non-IP "exposed" listen must DEGRADE, not throw and take the whole host down (D-validate) — the
+           host's IPAddress.Parse would otherwise FormatException into the generic catch and exit MCP entirely. */
+        var decision = Host.ResolveMcpBind(Mcp(listen, "192.168.1.0/24", token: "s3cr3t"), managed: true);
+
+        Assert.Equal(Host.McpBindMode.LoopbackOnly, decision.Mode);
+        Assert.Equal(Host.McpBindReason.ListenInvalid, decision.Reason);
+    }
+
+    [Theory]
+    [InlineData("192.168.1.205", "2001:db8::/32")]  // IPv4 listen, IPv6 CIDR
+    [InlineData("2001:db8::5", "192.168.1.0/24")]   // IPv6 listen, IPv4 CIDR
+    public void ResolveMcpBind_Exposed_Managed_AllowFromFamilyMismatch_IsLoopbackOnly_AllowFromInvalid(string listen, string allowFrom)
+    {
+        /* A family mismatch binds one family while the CIDR check rejects the other -> every client 403s
+           (fail-closed but silently non-functional); degrade with a reason instead (mirrors the store's D4). */
+        var decision = Host.ResolveMcpBind(Mcp(listen, allowFrom, token: "s3cr3t"), managed: true);
 
         Assert.Equal(Host.McpBindMode.LoopbackOnly, decision.Mode);
         Assert.Equal(Host.McpBindReason.AllowFromInvalid, decision.Reason);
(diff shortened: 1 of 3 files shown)


erikdarlingdata:
### Review fixes landed (commit 8872e3ee)

All four targeted fixes from review, on this same branch — no re-architecture:

1. **Non-IP `mcp.network.listen` now degrades, not crashes.** `IsExposedListenAddress` treats `localhost`/hostname/`*` as exposed, so `IPAddress.Parse` at the bind site would `FormatException` → generic catch → the whole MCP host exits. `ResolveMcpBind` now requires `IPAddress.TryParse` before `NetworkExposed`; a non-IP listen → new **`ListenInvalid`** reason → `LoopbackOnly` + `LogCritical`. The `~:115` "parses cannot throw" comment corrected (now true for listen AND allowFrom). Tests: `localhost` / `myhost` / `db.lan` / `*` → `ListenInvalid`.
2. **`RunPowerShellAsync` extracted, not duplicated.** Widened `DarlingManagedPostgres.RunPowerShellAsync` `private` → `internal` (one-word change to the same Windows-gated class that already exposes the firewall command builders I call) and deleted my copy. This also kills the silent divergence my dup introduced (hard-pinned `TimeSpan.FromSeconds(30)` vs the store's `s_statusTimeout`).
3. **listen/allowFrom address-family agreement** (mirrors the store's D4): a mismatched family binds one family while the CIDR check rejects the other → every client 403s. Now → `LoopbackOnly` + `AllowFromInvalid` + `LogCritical`. Tests: IPv4 listen + IPv6 CIDR, and vice versa.
4. **Severity mapping is now single-source + tested.** `MapBindReasonSeverity` (pure) is the one source of truth for the log level; `LogBindReason` emits off it, so level + message can't diverge. Tests assert `ListenInvalid`/`TokenMissing`/`AllowFromInvalid` → Critical, `ManagedModeRequired` → Warning, `NetworkExposed`/`LoopbackByDefault` → silent.

Build succeeded; full `Darling.Tests`: **Passed 1802, Skipped 127, Failed 0** (+8 new). Middleware wiring stays on the DARLING01 live-validation checklist. Still stopped before merge.

PR #1565, opened 2026-07-18
query_store: bound the self-exclusion scan (75% of read cost), per-database log lines, rdsadmin default exclusion

erikdarlingdata:
## Summary

Field follow-up with an actual plan (103k-row burst, 14.9s): **75% of elapsed sat in the self-exclusion Filter** — `query_sql_text NOT LIKE N'%PerformanceMonitorLite%'` scanning full nvarchar(max) text per row to exclude ~2% of rows (A/B: 15.0s → 3.5s, 4.3×). Their diagnosis also cleared the `LOOP JOIN` hint (DOP-1 replicas) and cost-estimate misdirection (the 2.27M-row internal-table scan was 426ms actual).

## Changes

1. **Bounded self-exclusion**: `LEFT(query_sql_text, 100) NOT LIKE ...` — every collector query leads with the marker in its first ~40 chars, so the semantics hold at ~zero cost. A user query merely mentioning the marker deep in text is now included (harmless noise). Not mirrored to install/09 — the Dashboard proc never had this filter (its statements are proc-attributed).
2. **Per-database log lines** (reporter item 3 — their inference that calls are already per-database is correct): the shared driver's completion hook now carries per-item SQL/storage timings; both apps log `  [server] {collector} [Database] => N rows (sql:…ms, …)` for non-empty batches. Quiet databases stay silent; the per-server roll-up and collection_log are unchanged.
3. **`rdsadmin` default-excluded** from the box/RDS enumeration (AWS's management DB, on every RDS instance, no customer workload). Org-specific utility DBs (their DBAUtils) are what per-server `excludedDatabases` is for — deliberately not hardcoded.
4. **Flush-interval cadence documented** (reporter item 2 — correct, not a bug): collector comment + README troubleshooting entry so nobody "fixes" the 900s `DATA_FLUSH_INTERVAL_SECONDS` bursts by narrowing the poll.

## Field posture

Production load can't be reproduced locally (HammerDB ≠ their fleet) — these are the conservative reads of the reporter's own plan/A-B evidence, and the per-database lines exist precisely so the next soak measures the effect directly.

Also confirmed by the same report: the #1557 byte budget survived its first production soak (61 clean minutes; the previously-fatal database hit the budget gracefully).

## Testing

- Pins: bounded-filter form (both capture modes), rdsadmin exclusion in the box enumeration, 4-arg driver hook across all control-flow tests.
- Full suites, Release: **Lite 1369/0**, **Darling 2170/0**.


erikdarlingdata:
Design revision after review: the bounded-LEFT filter is gone. No SQL-side text predicate survives — the per-item query is now byte-identical to the no-filter form the field A/B measured (3.5s vs 15.0s), and the self-exclusion happens client-side in ReadItemAsync where the text is already materialized (dropped before batch entry: never stored, never counted against the byte budget). Sargability was never achievable on that column in any form; shipping a function-wrapped residual scan was the wrong call and Erik was right to reject it.

PR #1601, opened 2026-07-21
Darling: --enable/disable-mcp/web CLI verbs (headless endpoint bring-up + firewall)

erikdarlingdata:
## What & why

Four one-shot CLI verbs let a **headless** Darling operator bring an endpoint up (or down) in one elevated action: `--enable-mcp`, `--disable-mcp`, `--enable-web`, `--disable-web`. They close two gaps that only exist on a box without the WPF Viewer:

1. **No supported way to enable/disable an endpoint.** `mcp.enabled` / `web.enabled` in `darling.json` are only a first-run **seed**; after first run the store (`config.config_service.mcp_enabled` / `web_enabled`) is authoritative and is normally toggled only by the Viewer's Settings — which a headless box does not have. Each verb writes the store directly with a **targeted `UPDATE`** on the single `id=1` row.
2. **No way to open the endpoint's firewall.** The service runs as a virtual service account (`NT SERVICE\PerformanceMonitor Darling`) that **cannot modify Windows Firewall**, so its best-effort self-reconcile silently fails.

## How it behaves

- **Store write (live, no restart):** a direct `config_service` `UPDATE` self-bumps `config_version` via the existing **BEFORE-UPDATE trigger**, so the worker hot-reloads within one collection sweep. The verbs deliberately never set `config_version` themselves, and never touch `paused` or the *other* endpoint's flag. 0 rows affected ⇒ the store isn't seeded yet (clear message, exit 1).
- **Firewall (only when LAN-exposed):** exposure is decided the way the service does — a network block with a non-loopback `listen` (shared `DarlingNetwork.IsExposedListenAddress`). When exposed:
  - **Elevated** ⇒ runs the **same scoped, idempotent-by-DisplayName** rule the host self-reconciles (the two `McpFirewallRuleName`/`WebFirewallRuleName` builders are now `internal`, so the CLI and host open/close the *exact same* rule) via the shared `BuildFirewallEnable/DisableCommand` builders.
  - **Not elevated** ⇒ prints the exact elevated command as a **handoff**. The store toggle already succeeded, so a non-elevated shell is never a failure (exit 0). A firewall failure is likewise non-fatal.
  - **Loopback-only** ⇒ a note that it binds loopback only and that LAN exposure needs `--configure-network`; no firewall action.
- **UX:** output states the store change, the live-apply note, the firewall outcome/handoff, and a one-line reminder that `darling.json`'s `mcp.enabled`/`web.enabled` is only the seed — the store is the live switch.
- **Guards:** managed-mode only (BYO governs its own `config_service` + exposure) and Windows-only (DPAPI credential decrypt + `WindowsPrincipal` + firewall), the same guard shape as `--print-viewer-connection`. The verbs are wired into `IsKnownVerb`, `UsageText()`, and the `Program` dispatch (allow-list and dispatch kept in sync, per the existing anti-drift comments).

## Test coverage (Darling.Tests)

- **Pure SQL pins** for all four store-writes: correct `*_enabled = TRUE/FALSE`, `updated_by = 'cli'`, `WHERE id = 1`, and **never** `config_version` / `paused` / the other endpoint's flag.
- **Pure verb recognition + wiring:** `IsKnownVerb`/`ClassifyStartupArgs` recognize the four verbs case-insensitively, reject bogus ones, each predicate is exclusive, and `UsageText` documents them.
- **Pure firewall-step classifier** (`ClassifyFirewallPlan`): the exposed × elevated truth table.
- **Shared rule-name pin:** the CLI and hosts reconcile the same DisplayName strings.
- **One gated-live test** (`..._AgainstDevPostgres`, gated on `DARLING_TEST_PG`, transaction-rolled-back): seeds a `config_service` row, runs the exact enable-mcp then disable-mcp store-writes against a throwaway Postgres, and asserts the flag flips **and** `config_version` self-bumps each time (and the mcp toggle leaves `web_enabled` untouched). It **skips** in the normal unit run, exactly like the other gated-live tests; CI's `darling-pg` job runs it live.

## Verification

- `dotnet build Darling/PerformanceMonitor.Darling.Service/...` — **0 errors**; my new code adds **0 warnings** (the two Service warnings are pre-existing, only line-shifted).
- `dotnet test Darling/Darling.Tests` with `DARLING_TEST_PG` cleared — **2789 passed / 0 failed / 144 skipped** (the 144 include this new gated-live test).

No store schema change; no auto-generated bundle touched; nothing deployed to any live server.


erikdarlingdata:
Hard-reviewed the implementation directly (not just the summary) — it's solid. Verified the load-bearing parts:

- **Store write** relies on the BEFORE-UPDATE self-bump trigger and never touches `config_version`/`paused`/the other endpoint's flag; the gated-live test proves the flip **and** the bump against real Postgres, transaction-rolled-back.
- **Firewall** reuses the shared `DarlingNetwork.IsExposedListenAddress`, the `McpFirewallRuleName`/`WebFirewallRuleName` builders, and `BuildFirewall*Command` — no reinvented logic. The empty-`allowFrom` guard correctly bails rather than emit a malformed `New-NetFirewallRule` (which is what makes the later `allowFrom!` safe). Firewall failures are non-fatal since the toggle already succeeded.
- Both judgment calls are good: making the rule-name builders `internal` is the correct single-source-of-truth move (pinned by a test), and the README subsection is apt.

Two minor, **non-blocking** notes:
1. It reads the raw `darling.json` `allowFrom` rather than the service's validating resolver — a malformed-but-nonempty CIDR would fail gracefully (caught, handoff printed) but differs from the service's fail-close-to-loopback for a bad CIDR. The empty case is already guarded. Optional hardening: validate the CIDR (or route through the same resolver) for full parity.
2. The CHANGELOG entry will conflict with #1602's at merge (both prepend under `### Added`) — trivial, keep both.

LGTM once CI is green.

PR #1608, opened 2026-07-21
Darling MCP: alert-tuning write tools (update_alert_settings, create/delete_mute_rule)

erikdarlingdata:
Adds three Darling-only MCP write tools to `DarlingMcpAlertTools` so an MCP client (an LLM assistant) can tune a user's alerting conversationally — the direct sibling of the #1600 Custom Views MCP write tools (read tools existed → add write tools reusing the existing write authority + the least-privilege `mcp` role).

## The two gaps closed
- **Tune thresholds** — `update_alert_settings` is a PARTIAL update of the singleton `config.config_alert_settings` row (id=1). The agent reads via `get_alert_settings`, changes fields, and sends only those back in the SAME nested shape (e.g. `{"cpu":{"threshold_percent":90},"cooldown_minutes":10}`), so read → modify → write round-trips.
- **Manage mute rules** — `create_mute_rule` / `delete_mute_rule` add and remove the alert mute rules the delivery paths honor.

## Reused write authorities (no divergent second implementation)
- `update_alert_settings` validates EVERY provided field against the SAME ranges/enums the Viewer's Settings window enforces (`SettingsWindow.BuildAlertRowFromControls`) — thresholds in range, `cpu.mode` `sql`/`total`, `delivery.mode` `Summary`/`PerEvent`, counts within bounds. Validation runs BEFORE any write: an out-of-range value or an unknown field (top-level or nested) returns `{status:"invalid"}` and writes nothing. It then applies ONLY the provided columns via a targeted parameterized `UPDATE ... WHERE id = 1` and re-reads the merged state.
- `create_mute_rule` / `delete_mute_rule` reuse the SAME `PgMuteRuleStore` `get_mute_rules` reads through, with the same GUID id-generation the Viewer's mute-create path (`MuteRuleEditDialog`) uses.
- SMTP/webhook delivery credentials are out of scope (the `mcp` role cannot read or write the secret columns).

## config_version self-bump reliance
A `config_alert_settings` write self-bumps `config_service.config_version` via the existing config-table trigger (`trg_bump_alert_settings` → `config_bump_version`), so the running service hot-reloads the change within one collection sweep. The tool never writes `config_version` itself.

## Scoped grant — and one non-obvious grant flagged for your review
In `DarlingManagedRoles.BuildProvisioningSql` (role provisioning, NOT a migration), the `mcp` role is granted `INSERT/UPDATE/DELETE` on `config.config_mute_rules` and `UPDATE` on the singleton `config.config_alert_settings` (UPDATE only — no INSERT/DELETE), mirroring exactly how #1600 granted `config.custom_views`.

**Deviation from the literal spec, called out:** the `config_bump_version` trigger is **SECURITY INVOKER** and UPDATEs `config_service` AS the writing role. The `mcp` role had no `config_service` write, so every `update_alert_settings` write would have failed **42501 in production** — and the superuser-run gated-live tests would never catch it (they connect as the owner). Fix: a **COLUMN-level** `UPDATE (config_version, updated_at)` grant on `config.config_service` — mcp can bump the reload beacon but cannot flip `paused` / `capture_plans` / `mcp_enabled` / `mcp_port`. The live security test now proves this end-to-end as the real `mcp` role (a `config_alert_settings` UPDATE succeeds and fires the beacon; a `paused` UPDATE still 42501s). This is the "fix the latent defect in the same change" call; happy to revisit the shape (e.g. making the trigger `SECURITY DEFINER` instead) if you'd prefer.

## Also updated
- MCP instructions: document the three tools + the read-modify-write flow, and recount the tool prose (82 → 85).
- Cross-app pin (`CrossAppMcpToolInventoryPinTests`): the three names join the `KnownLiteMissingMcpTools` ratchet (Darling-only — Lite has no central alert store). **This is a `Lite.Tests` check the Darling-only build does not run locally; it bit #1602, so it's covered here.**
- The three write tools join the `/api/read` exclusion set (`DarlingWebEndpoints.ExcludedToolNames`) — writes with no read endpoint, like `mute_analysis_finding` (the parity reflection test caught this).
- README MCP blast-radius + roles table.

## Test coverage
- **Ungated** (`DarlingMcpAlertToolsSurfaceAndSqlTests`): six-tool surface pin (all static, `Task<string>`); Gemini-clean schema + required-params; and pure validate-before-write — 16 bad/unknown partial-update inputs return `{status:"invalid"}` WITHOUT opening a connection (dead-datasource trick), plus blank `rule_id` / bad `expires_at`.
- **Grant pin** (`DarlingManagedRolesTests`): the new mcp alert grants are narrow (single-table + column-level, no schema-wide, no ADP).
- **Live security** (`DarlingSecuritySplitLiveTests`, gated): the mcp role can now CRUD `config_mute_rules` and UPDATE `config_alert_settings` (the beacon trigger fires as mcp), but is still denied `config_command` and a table-wide `config_service` write.
- **One gated-live round-trip** (`DARLING_TEST_PG`, own-scoped + restored): `update_alert_settings` flips a threshold AND self-bumps `config_version`, and `create_mute_rule` → `get_mute_rules` → `delete_mute_rule` round-trips. Skips in the normal unit run.

## Verification
- Darling service + `Darling.Tests` build clean (0 errors; 0 warnings in the changed source).
- `Darling.Tests`: **2821 passed / 0 failed / 145 gated-live skipped** (with `DARLING_TEST_PG` cleared).
- `Lite.Tests` `CrossAppMcpToolInventoryPinTests`: **2 passed**.

The gated-live tests run in CI's `darling-pg` job; I did not stand up a throwaway Postgres locally to avoid touching this box's managed store.


erikdarlingdata:
Hard-reviewed, and I verified the load-bearing parts against the actual code rather than trusting the report:

- **The config_service column-grant is a genuine catch, verified.** `config.config_bump_version()` (PgMigrations.cs:644) is SECURITY INVOKER (no DEFINER clause; the comment says so) and does `UPDATE config.config_service ... WHERE id=1`; the `trg_bump_alert_settings` AFTER trigger fires it, so an `mcp` `UPDATE config_alert_settings` runs that beacon write **as mcp** → 42501 in prod, which superuser-run gated-live tests would never surface. The fix is correctly scoped: **column-level** `UPDATE (config_version, updated_at)` only — a test asserts it is NOT a full-table grant, and the live security test proves a `paused` write still 42501s. Better than making the trigger SECURITY DEFINER.
- **Validation is real:** typed range checks + enum whitelists matching `SettingsWindow.BuildAlertRowFromControls`, first-bad-value short-circuit that writes nothing, and the dynamic UPDATE's column names are compile-time constants (injection-safe). The dead-store test proves a bad/unknown field returns `invalid` without opening a connection.
- **Cross-app pin updated** (the #1602 trap), grants mirror #1600, `/api/read` exclusion + the security-test target switch (to `config_command`) are sound.

LGTM once CI is green. I'll merge, deploy to DARLING01, and dogfood the tools live.

PR #1609, opened 2026-07-21
Darling MCP: bulk add/remove monitored servers (add_servers, remove_server)

erikdarlingdata:
## The gap

The Darling MCP server exposed read tools plus the #1600 Custom Views and #1608 alert-tuning WRITE tools, but **no way for an MCP client (or Claude, over MCP) to add or remove the monitored servers themselves** — onboarding was WPF-Viewer-only (the Add / Manage Servers dialogs). Two new tools close that: **agent-driven BULK fleet onboarding** — "monitor these twenty servers with this login" stands up fleet monitoring in one call.

- **`add_servers`** (BULK) — a JSON **array** of server objects, processed **sequentially** (mirroring #1549's Darling bulk probe, serial to avoid a probe storm). For each: validate → case-folded **dedupe** → in-process **connection probe** → **DPAPI-encrypt** the SQL password → **INSERT**. Per-server `status` is `added` / `duplicate` (a case-variant or exact dup of an existing or earlier-in-batch server, skipped without a probe) / `connection_failed` (recorded — the batch **continues**) / `invalid` (a bad field, or Entra/MFA/Service-Principal/Managed-Identity auth — nonsensical headless, the #1549 belt). The whole call returns `{added, skipped, failed, results:[...]}`.
- **`remove_server`** — removes a monitored server by name (resolved the same way the read tools resolve `server_name`); DELETEs its `config_monitored_servers` row. Returns `{status:"removed", server}` or `{status:"not_found"}`.

## Reused authorities (no divergent second implementation)

- **`DarlingServerConnector.ProbeAsync` run IN-PROCESS** — the MCP host lives inside the service, which holds the network path + credentials, so (unlike the Viewer's dialogs, which enqueue a `test_connect` command for the service to run) no command plane is needed. Reports edition/version/msdb like `--test-connection`.
- **`ServerIdHelper.BuildStorageName` case-folded dedupe gate** — a `HashSet<string>(OrdinalIgnoreCase)` seeded from the existing rows, first-occurrence-wins, exactly as the #1549 bulk dialog.
- **`DarlingSecrets.Protect`** — DPAPI-encrypts the SQL password as the SAME service identity that decrypts it during collection (so it round-trips); Windows-auth servers store no secret; the plaintext is never logged or echoed.
- **`SeedMonitoredServersAsync`'s INSERT shape** — same column set + `server_id = GetDeterministicHashCode(StorageName)` identity, so a tool-written row JOINs the collected data and the service's reconcile matches it (picked up within one sweep, no restart).

## Scoped grant + the #1608 beacon already covers the trigger

The `mcp` role is granted (**provisioning, not a migration**) `INSERT/UPDATE/DELETE ON config.config_monitored_servers` — a single non-secret-KEY table. Its `encrypted_password` column **stays in the fail-closed secret carve**, so `mcp` can WRITE a credential blob (onboarding) but can never READ one back; it still cannot reach the `config_command` service-credential pivot or a schema-wide config write. A `config_monitored_servers` write fires the existing `trg_bump_monitored_servers → config_bump_version` trigger (SECURITY INVOKER, UPDATEs `config_service.config_version` AS `mcp`), and **#1608 already granted `mcp` UPDATE on the two `config_service` beacon columns** — verified, so **no new `config_service` grant is added** (the gated-live security test's monitored-servers INSERT doubles as proof the beacon grant covers this write).

## Encrypt-mode + cert-trust exposure, and the credential on the wire

Per Erik, `encrypt_mode` (`Optional`/`Mandatory`/`Strict`) and `trust_server_certificate` are **exposed** so a headless caller sets the connection posture explicitly (optional, fail-closed `Mandatory` / no-trust defaults). The SQL `password` **transits the MCP endpoint** in `add_servers`' request JSON (DPAPI-encrypted at rest, never returned by any read tool); the README's MCP blast-radius section now states this and recommends fronting a LAN deployment with the documented **TLS reverse proxy** (and preferring Windows/integrated auth where possible).

## Also updated

MCP server instructions (now **eighty-seven** tools, +2), the cross-app tool-inventory ratchet (Lite is a single-instance app with no central monitored-server store → Darling-only), the `/api/read` write-exclusion set, and the Darling README.

## Verification

- Darling service + `Darling.Tests` build **clean** (0 errors, no new warnings in the changed source).
- Full `Darling.Tests` **2847 passed / 0 failed / 146 gated-live skipped** (with `DARLING_TEST_PG` cleared).
- `Lite.Tests` `CrossAppMcpToolInventoryPinTests` **2 passed**.
- Ungated pins cover the two-tool surface, the Gemini-clean schema + required-params, validate-before-write (a malformed payload or a bad/MFA entry returns `invalid` **without probing or opening a connection**), and the pure case-folded dedupe partition.
- **One** gated-live test (`DARLING_TEST_PG`, own-scoped + cleaned up, the **SQL probe stubbed to success** — no live SQL Server touched in CI) proves the store INSERT (the SQL secret is DPAPI-encrypted at rest and round-trips; the Windows-auth server is secret-free), the duplicate skip, the `config_version` self-bump, and `remove_server` (removed, then not_found).

The **real end-to-end probe is what the human dogfoods** (remove sql2016, re-add via MCP) — no live SQL Server connection is attempted in CI.

**Do NOT merge** — hard-review first, then deploy to DARLING01 and dogfood the sql2016 remove/re-add.


erikdarlingdata:
Hard-reviewed against the code, focused on the credential surface since that's the real risk here:

- **Write-a-credential-but-never-read-one-back is proven LIVE** (not just a static SQL pin): `McpRole_DeniedSecretColumns_...` connects as the real mcp role and asserts **42501** on the secret columns (incl. `config_monitored_servers.encrypted_password`), while `HasPriv` + an actual INSERT confirm the write grant. The provisioning pins the `REVOKE SELECT` carve. So a token can onboard a server credential but cannot exfiltrate stored ones.
- **No credential leakage:** `add_servers` returns `{added,skipped,failed,results:[{server,status,detail}]}` — the plaintext password is used only for the in-process probe connect, DPAPI-encrypted for storage, and never logged or echoed.
- **Reuse, not reinvention:** in-process `ProbeAsync` (validate before store), case-folded `BuildStorageName` dedupe, `SeedMonitoredServersAsync` INSERT shape (parameterized), MFA/Entra rejected headless. The #1608 `config_service` beacon grant already covers the bump trigger — correctly verified and not re-added.
- Cross-app pin, `/api/read` exclusion, and README blast-radius (credential-on-the-wire → TLS reverse proxy) all updated.

LGTM once CI is green. I'll merge, deploy to DARLING01, and dogfood via the sql2016 remove/re-add.

Issue #1048, opened 2026-06-02
[BUG] Linux Host CPU is shown at 100% all the time (SQL Server CPU consumption probably correct)


bogdanben2000:
### Component

Full Dashboard

### Performance Monitor Version

2.11.0

### SQL Server Version

SQL Server 2019 CU12

### Windows Version

Linux (OpenSuse)

### Describe the Bug

<img width="382" height="277" alt="Image" src="https://github.com/user-attachments/assets/507ba59f-ea0a-4b4f-ac93-a52efa12b9a2" />

Result of Top command:

<img width="841" height="679" alt="Image" src="https://github.com/user-attachments/assets/efde759f-502e-4098-b44e-d7841ee727f3" />

### Steps to Reproduce

.

### Expected Behavior

.

### Actual Behavior

.

### Error Messages / Log Output

```text
.
```

### Screenshots

.

### Additional Context

Graph with Host CPU (not just SQL Server CPU consumption) would help

erikdarlingdata:
Thanks for the clear report — the `top` output alongside the Dashboard screenshot made the root cause easy to pin down.

**What's happening:** the host CPU figure is derived from the `RING_BUFFER_SCHEDULER_MONITOR` ring buffer as `100 - SystemIdle - SQLProcessUtilization`. On SQL Server on Linux, `SystemIdle` is **always reported as 0** — it's a documented platform limitation (Microsoft's own sample query even carries the comment `-- SystemIdle on Linux will be 0`). So the math collapses to `100 - 0 - sqlcpu`, and the host total pins at 100% no matter what the box is actually doing. Your SQL Server CPU number was correct the whole time; only the host/"other process" figure was fabricated. Unfortunately there's no DMV that exposes true host CPU on Linux.

**The fix** (just merged to `dev` in #1049): the collector now detects Linux and stores the host figure as "unavailable" instead of a bogus value, and the Dashboard falls back to showing the correct SQL-Server-only CPU rather than a flat 100%.

**Could you help verify on the nightly build?** I don't have a Linux SQL host to reproduce on, so a confirmation from your environment would be hugely helpful. The fix lands in the next nightly (auto-built from `dev` around 06:00 UTC daily):

[...]

PR #1630, opened 2026-07-24
Bugfix/issue - #1629 cpu other linux oversuppression

### LINKED ISSUE #1629: [BUG] `other_process_cpu_utilization` suppressed on Linux SQL Server 2025
### Component

SQL collection scripts

3.2.0

### SQL Server Version

Microsoft SQL Server 2025 (RTM-CU7) (KB5096981) - 17.0.4065.4 (X64) Jul 8 2026 23:26:08 Copyright (C) 2025 Microsoft Corporation Enterprise Developer Edition (64-bit) on Linux (Ubuntu 24.04.4 LTS) 

### Windows Version

Ubuntu 24.04 LTS 

### Describe the Bug

Hey Erik - found a spot where the Linux CPU guard from #1048 is now suppressing good data on SQL Server 2025.

The root cause appears to be `sys.dm_os_ring_buffers` reporting `SystemIdle` differently depending on platform and version:

| Platform | `SystemIdle` value |
|---|---|
| Linux SQL Server 2022 and older | always 0, regardless of actual load |
| Linux SQL Server 2025 | real % idle again |

The guard you added for #1048 works for 2022 and older versions:

```sql
WHEN @is_linux = 1 THEN NULL   -- blanket suppress on all Linux
```

On SQL Server 2025 Linux, `SystemIdle` is a real value again (same as Windows), but the above guard still fires and throws it away - `other_process_cpu_utilization` comes back permanently NULL on any Linux host running SQL 2025.

The fix is fairly simple:

```sql
WHEN @is_linux = 1 AND x.system_idle = 0 THEN NULL
```

### Steps to Reproduce

```bash
docker pull mcr.microsoft.com/mssql/server:2025-latest

docker run -e ACCEPT_EULA=Y -e MSSQL_SA_PASSWORD='Okay!Whatever123!' \
  --name sqltest -p 1433:1433 -d mcr.microsoft.com/mssql/server:2025-latest

sleep 15
```

Check the behavior of `SystemIdle` on SQL 2025 Linux:

```bash
docker exec -it sqltest /opt/mssql-tools18/bin/sqlcmd \
  -S localhost -U sa -P 'Okay!Whatever123!' -C -Q \
  "SET QUOTED_IDENTIFIER ON;
   SELECT TOP 5
     sql_server_version =
       CONVERT(nvarchar(20), SERVERPROPERTY('ProductMajorVersion')),
     process_utilization =
       CONVERT(xml, dorb.record).value(
         '(Record/SchedulerMonitorEvent/SystemHealth/ProcessUtilization)[1]', 'int'),
     system_idle =
       CONVERT(xml, dorb.record).value(
         '(Record/SchedulerMonitorEvent/SystemHealth/SystemIdle)[1]', 'int')
   FROM sys.dm_os_ring_buffers AS dorb
   WHERE dorb.ring_buffer_type = N'RING_BUFFER_SCHEDULER_MONITOR'
   ORDER BY dorb.timestamp DESC;"
```

Then run both guards side by side - 2022 behavior is simulated with a literal SystemIdle=0 row since we only have a 2025 container here:

```bash
docker exec -it sqltest /opt/mssql-tools18/bin/sqlcmd \
  -S localhost -U sa -P 'Okay!Whatever123!' -C -Q \
  "DECLARE @is_linux bit = 1;

   SELECT
     scenario = v.label,
     system_idle = v.system_idle,
     process_util = v.process_util,
     original_guard =
       CASE WHEN @is_linux = 1
            THEN NULL
            WHEN (100 - v.system_idle - v.process_util) < 0 THEN 0
            ELSE 100 - v.system_idle - v.process_util END,
     fixed_guard =
       CASE WHEN @is_linux = 1 AND v.system_idle = 0
            THEN NULL
            WHEN (100 - v.system_idle - v.process_util) < 0 THEN 0
            ELSE 100 - v.system_idle - v.process_util END
   FROM (VALUES
     ('SQL 2022 Linux  system_idle=0',  0,  2),
     ('SQL 2025 Linux  system_idle=95', 95,  3)
   ) AS v(label, system_idle, process_util);"
```

Cleanup:

```bash
docker rm -f sqltest
```

### Expected Behavior

`SystemIdle` on SQL 2025 Linux should come back as a real value:

```
sql_server_version   process_utilization  system_idle
-------------------  -------------------  -----------
17                   2                    97
17                   2                    97
17                   3                    96
17                   2                    97
17                   2                    98
```

And the guard comparison should look like this:

```
scenario                                   system_idle  process_util  original_guard  fixed_guard
------------------------------------------ -----------  ------------  --------------  -----------
SQL 2022 Linux  system_idle=0 (broken DMV)   0            2             NULL            NULL
SQL 2025 Linux  system_idle=95 (fixed DMV)  95            3             NULL            2
```

`fixed_guard`: 2022 stays NULL (SystemIdle=0, no usable host figure). 2025 returns 2%
(`100 - 95 - 3`), which is the real other-process CPU.

### Actual Behavior

`original_guard`: both rows come back NULL. 2025 data is silently lost - `other_process_cpu_utilization` is permanently NULL on any Linux host running SQL 2025, even though `SystemIdle` is reporting real numbers again.

```text
NA
```

### Screenshots

NA

### Additional Context

This is a regression from the Linux CPU guard added for #1048, which blanket-suppressed `other_process_cpu_utilization` on all Linux hosts to work around SQL Server 2022 and older always reporting `SystemIdle = 0`. SQL Server 2025 on Linux fixed that DMV to report real idle percentages, so the blanket guard is now too broad.

argpna:
## What does this PR do?

Fixes #1629. The Linux CPU guard added for #1048 blanket-suppressed `other_process_cpu_utilization` on every Linux host (`WHEN @is_linux = 1 THEN NULL`), to work around `SystemIdle` always reporting 0 in the SCHEDULER_MONITOR ring buffer on SQL Server 2022 and older on Linux.

SQL Server 2025 on Linux reports real `SystemIdle` values again, so the blanket guard now discards a derivable figure on any Linux host running 2025 - `other_process_cpu_utilization` comes back permanently NULL even though the data is available.

Narrows the guard to `WHEN @is_linux = 1 AND system_idle = 0 THEN NULL`, so suppression only happens when the value is actually non-derivable, matching the real condition instead of the platform. Applied consistently to:

- `PerformanceMonitor.Collectors/CpuUtilizationCollector.cs`
- `install/18_collect_cpu_utilization_stats.sql`
- `install/02_create_tables.sql` (table comment)
- `Dashboard/Services/DatabaseService.NocHealth.cs`

- [ ] Full Dashboard
- [x] Lite Dashboard
- [x] Lite Tests
- [x] SQL collection scripts
- [ ] CLI Installer
- [ ] Documentation

## How was this tested?

- Added `BuildQuery_RingBuffer_SuppressesOtherCpu_OnlyWhenSystemIdleIsZero` to
  `Lite.Tests/CpuUtilizationCollectorDefinitionTests.cs`, asserting the generated query text uses
  the narrowed guard and no longer contains the old unconditional `WHEN @is_linux = 1` branch.
- `Build` and `SQL Validation` GitHub Actions workflows passed against this branch.
- Verified the guard algebraically against the two scenarios from the bug report:
  - SQL 2022 Linux (`system_idle = 0`, `process_util = 2`) -> stays NULL (not derivable)
  - SQL 2025 Linux (`system_idle = 95`, `process_util = 3`) -> returns `2` (`100 - 95 - 3`)

## Checklist

- [ ] I have tested my changes against at least one SQL Server version

### DIFF
--- a/Lite.Tests/CpuUtilizationCollectorDefinitionTests.cs
+++ b/Lite.Tests/CpuUtilizationCollectorDefinitionTests.cs
@@ -37,6 +37,19 @@ public sealed class CpuUtilizationCollectorDefinitionTests
     }
 
     [Fact]
+    public void BuildQuery_RingBuffer_SuppressesOtherCpu_OnlyWhenSystemIdleIsZero()
+    {
+        /* Some Linux/SQL Server version combos report a non-zero SystemIdle, so
+           unconditionally nulling other-process CPU on @is_linux = 1 over-suppressed a
+           derivable value. The guard must require BOTH @is_linux = 1 AND system_idle = 0. */
+        var plan = CpuUtilizationCollector.Instance.BuildQuery(
+            CollectorTestContext.Make(s_deltas, isAzureSqlDb: false));
+
+        Assert.Contains("WHEN @is_linux = 1 AND x.system_idle = 0", plan.Text, StringComparison.Ordinal);
+        Assert.DoesNotContain("WHEN @is_linux = 1\n", plan.Text, StringComparison.Ordinal);
+    }
+
+    [Fact]
     public void BuildQuery_Azure_NoWatermark_UsesTop60()
     {
         var plan = CpuUtilizationCollector.Instance.BuildQuery(
--- a/PerformanceMonitor.Collectors/CpuUtilizationCollector.cs
+++ b/PerformanceMonitor.Collectors/CpuUtilizationCollector.cs
@@ -17,8 +17,8 @@ namespace PerformanceMonitor.Collectors;
 /// <summary>
 /// CPU utilization — ring buffer (on-prem, MI, RDS) or sys.dm_db_resource_stats (Azure SQL DB).
 /// Extracted verbatim from Lite's RemoteCollectorService.Cpu.cs, including the platform rules
-/// that are the parity contract: Linux stores NULL other-process CPU because SystemIdle is
-/// always 0 there (#1048); incomplete SystemHealth ring-buffer records are skipped (#989);
+/// that are the parity contract: Linux stores NULL other-process CPU when SystemIdle reports
+/// 0 there (#1048); incomplete SystemHealth ring-buffer records are skipped (#989);
 /// Azure filters server-side on the watermark while the ring buffer dedups client-side
 /// (its sample_time is computed and cannot be filtered in SQL).
 /// </summary>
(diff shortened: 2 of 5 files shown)


erikdarlingdata:
Merged, thanks for the clean fix and the test. Keying the guard on the actual condition (system_idle = 0) instead of the platform was the right call, it works regardless of which version/CU starts reporting real SystemIdle. Ships in the next release.

PR #1664, opened 2026-07-26
Route the viewer's built-in tabs to the retention rollups (#1661)

### LINKED ISSUE #1661: Darling viewer: built-in tabs silently lose history past 4 days - CAGG read-routing only covers Custom Views
Found by the issue audit and verified directly. **Not yet in a release — it is dev-only, so it can be fixed before 3.3.0 ships.** That is the good news; the bad news is that it is a silent data-loss regression, not an error anyone will see.

## What happens

The three-tier retention from #1623 applies automatically. `EnsureRetentionPoliciesAsync` is called unconditionally on every service start (`DarlingWorker.cs:651`), and raw retention is **4 days** (`TimescaleSupport.cs:585`, `RawRetentionInterval = "4 days"`).

The CAGG read-routing that preserves older windows (#1625/#1626/#1627) lives **only** in the Custom Views v2 compose path — `Compose/ComposeSourceRouter.cs` and `ComposeCompiler.cs`.

The viewer's built-in tabs never go through that path. They read `v_query_stats`, which is a plain passthrough over the raw hypertable:

```sql
CREATE OR REPLACE VIEW v_query_stats AS SELECT * FROM query_stats;
```
(`PgMigrations.cs:233`)

Confirmed production readers, all taking a caller-supplied window:

- `ViewerDataService.DailySummary.cs:64`
- `ViewerDataService.FinOps.Inventory.cs:80`
- `ViewerDataService.FinOps.Storage.cs:144`
- `ViewerDataService.FinOps.Workload.cs:37, :202, :279, :435`

So a user asking Daily Summary or FinOps for the last 30 days now silently gets 4 days of query data. Before the retention tiers those tables held ~30 days. Nothing errors; the numbers are just quietly wrong, which is the worst failure mode for a cost/workload view whose entire purpose is trend over time.

FinOps is the sharpest case — spend and workload attribution over a 4-day sample presented as a month is actively misleading.

## Why it slipped through

#1623 (write side: drop raw at 4 days) and #1625-#1627 (read side: route old windows to the rollups) were designed as a pair, and the read side was scoped to the composer because that was the surface being built at the time. The built-in tabs predate Custom Views and were never revisited. Each PR was correct in isolation.

Worth noting for the record: I had this recorded as "nothing falls back to raw," which was true **of the compose path** and which I over-generalized. The built-in tabs are exactly the thing that still falls back to raw.

## Options

1. **Route the built-in readers through the same age-based logic** the composer uses — most correct, and reuses the proven `ComposeSourceRouter` rules rather than a second implementation.
2. **Make `v_query_stats` a UNION view** over raw + hourly + daily with dedup. Fixes every reader at once with no C# change, but the CAGG columns are pre-aggregated (`x_sum`, `sample_count`) and do not have the same shape as raw rows, so this is likely not expressible without lossy assumptions.
3. **Raise the raw horizon** so built-in tabs keep working. Cheapest, but gives back the storage win that motivated #1623.

Option 1 looks right; option 3 is the safe stopgap if 3.3.0 is close.

## Also verify before shipping

The same passthrough pattern exists for the other tiered tables — `procedure_stats` and `query_store_stats` both have raw retention and hourly/daily CAGGs. Any built-in reader over their `v_*` views has the identical problem and should be swept in the same change.

erikdarlingdata:
Closes #1661. Grew from a foundation PR into the full fix; the commits tell the story in order.

## The bug

The three-tier retention from #1623 drops raw `query_stats` at **4 days** and runs automatically on every service start (`DarlingWorker.cs:651`). The CAGG read-routing from #1625/#1626/#1627 only ever covered the **Custom Views composer**. The viewer's built-in tabs read `v_query_stats` — a plain passthrough over the raw hypertable.

So the Performance Calendar reported **zero distinct queries** for every day older than ~4 days, and the FinOps panels aggregated a 4-day sample while the picker said 30. Nothing errored; the numbers were quietly wrong. For a cost view whose entire purpose is trend over time, that is the worst failure mode available.

Each prior PR was correct in isolation — the write side and read side were designed as a pair, and the read side was scoped to the surface being built at the time.

**Never released.** Caught in `dev` before 3.3.0.

## One tier decision, one definition

`RetentionTierRouter` lands in **Storage** because both consumers need it and neither can see the other (composer in the service, tabs in the viewer; both reference Storage). `ComposeSourceRouter` delegates to it and aliases its constants — behaviour unchanged, 209 composer tests still green, and no second copy of the thresholds.

An invariant test parses `TimescaleSupport`'s retention intervals and asserts each route threshold stays strictly **inside** the horizon it protects. Change a retention interval without moving the margin and the build fails — otherwise you reintroduce this exact bug one tier up.

## Readers, by what they can actually be served

**Routed** — Daily Summary's query count, the `get_daily_health` MCP reader, and three FinOps aggregates. These sum additively over dimensions the rollups materialize.

**New rollup** — FinOps' database-grain view sums `delta_logical_reads`/`delta_physical_reads`/`delta_logical_writes`, which **no** aggregate carried (the others were built to the composer's measure set, which never exposed I/O). It gets a new `query_stats_db_hourly`/`_daily` pair rather than a widened existing one:

> TimescaleDB cannot ALTER columns into a continuous aggregate, so widening means DROP + recreate. With retention live that rebuilds from 4 days of raw and **permanently destroys** the 21-day hourly and indefinite daily history the tiers exist to preserve. Unrecoverable on the field box. Additive costs nothing existing; its history simply starts from deploy.

Grouped by `database_name` and deliberately not `query_hash`, so it stays small despite carrying more columns.

**Clamped and labelled** — the expensive-queries panel projects `query_text`/`query_plan_xml`, which no rollup carries at any tier. It clamps to the text horizon and says so, rather than presenting a few days as a month. Without the note it would look inconsistent against the aggregate panels, which now *do* reach further back.

## Two things routing exposed

**A second copy of the daily-summary query** lived in the MCP health reader, documented as the viewer's "verbatim" with nothing enforcing it. They differed only in stripped comments, so unifying was behaviour-preserving — but routing one and not the other would have had the calendar and the MCP tool disagree about the same day. Now one definition in Storage. The proof the move was faithful: `ViewerDailyHealthTests` and `DarlingMcpHealthToolsTests` pin the shape independently from opposite sides, and **neither needed editing**.

**`ItemHistory` is not a text reader.** An earlier pass called it one; it projects `query_plan_hash` — a fixed-size hash — and no `query_text` at all. It is not routable either (it selects spills/dop/grant columns no rollup carries), but as a drill into one query's recent behaviour the raw window is a reasonable scope, so it is deliberately left alone rather than clamped for a limit it does not have.

## Test plan

Every substitution throws if its raw fragment no longer matches the SQL. **A guard that throws only protects you if something calls it**, so the tests exercise every tier of every routed reader — otherwise a drifted fragment surfaces as an exception in front of a user instead of a failed build. That guard earned its keep during development: the first extracted CTE had a mismatched raw-string margin and failed the test rather than silently leaving the calendar on raw, which is exactly how the original bug hid.

Raw is asserted to return each constant byte-for-byte, so the recent-window path is provably unchanged.

- **Darling.Tests: 3087 passed**, 0 failed (147 gated-live skipped)
- Solution builds **0 warnings, 0 errors**


erikdarlingdata:
Took this over after the gated-live CI failure. Root cause: the tier routing chose a relation by window age alone, but the CAGGs are runtime TimescaleDB setup — CI's dev PG (and any BYO plain-PostgreSQL store) doesn't have them, so old-window reads threw 42P01. Fix: a to_regclass availability probe (TimescaleSupport.DetectRollupsAsync, cached in the viewer) gating RetentionTierRouter — plain PG routes to raw, which is complete there since retention policies also need the extension. Partial builds degrade one tier at a time, mirroring the composer's DailyView-null rule. New live test at -10 days makes the failure deterministic (the -3d test straddles the route margin by time-of-day). Note the composer has the same pre-existing exposure from #1625 — filed separately.

Issue #1666, opened 2026-07-26
[BUG] `report.daily_summary_v2.metric_value` renders `worst_query_hash` as garbled text


argpna:
### Component

SQL collection scripts

### Performance Monitor Version

3.2.0

### SQL Server Version

SQL Server 2022 CU (docker image `mcr.microsoft.com/mssql/server:2022-latest`)

### Windows Version

N/A - reproduced via Docker. The bug is independent of the client/host OS

### Describe the Bug

`report.daily_summary_v2` pivots mixed-type metrics into a single `metric_value sql_variant` column so callers can `SELECT sort_order, metric_name, metric_value FROM report.daily_summary_v2 ORDER BY sort_order`. The `worst_query_hash` row boxes the raw `binary(8)` `query_hash` (`collect.query_stats.query_hash` is declared `binary(8)`, not `varbinary`) directly:

```sql
metric_name = N'worst_query_hash',
metric_value = CONVERT(sql_variant, wq.query_hash)
```

Any caller that follows the view's documented usage and converts `metric_value` straight to a string type gets unreadable output instead of a hash, because SQL Server's binary-to-char conversion reinterprets the raw bytes as UTF-16 code units.

Every other row in the view is a scalar (date/int/nvarchar) and converts to a sane string. Only the `worst_query_hash` row is affected, since it's the one row carrying a binary payload.

### Steps to Reproduce

```bash
docker pull mcr.microsoft.com/mssql/server:2022-latest

docker run -e ACCEPT_EULA=Y -e MSSQL_SA_PASSWORD='NotStrong!Passw0rd' \
  --name sqltest -p 1433:1433 -d mcr.microsoft.com/mssql/server:2022-latest

# short pause so database service can start up
sleep 15
```

Reproduce the view's own pattern with a literal hash:

```bash
docker exec -it sqltest /opt/mssql-tools18/bin/sqlcmd -S localhost -U sa \
  -P 'NotStrong!Passw0rd' -C -Q "
DECLARE @query_hash binary(8) = 0x1F7A7FB76757088B;
DECLARE @metric_value sql_variant = CONVERT(sql_variant, @query_hash);
SELECT metric_value = CONVERT(nvarchar(500), @metric_value);"
```

Confirm the base type SQL Server sees:

```bash
docker exec -it sqltest /opt/mssql-tools18/bin/sqlcmd -S localhost -U sa \
  -P 'NotStrong!Passw0rd' -C -Q "
[...]

Issue #1635, opened 2026-07-24
[BUG] `report.trace_flag_changes` fails with bit-to-nvarchar conversion error


argpna:
### Component

SQL collection scripts

### Performance Monitor Version

3.2.0

### SQL Server Version

Microsoft SQL Server 2025 (RTM-CU7) (KB5096981) - 17.0.4065.4 (X64) Jul 8 2026 23:26:08 Copyright (C) 2025 Microsoft Corporation Enterprise Developer Edition (64-bit) on Linux (Ubuntu 24.04.4 LTS) 

### Windows Version

N/A - reproduced via Docker - the bug is independent of the client/host OS

### Describe the Bug

The `report.trace_flag_changes` view contains a `CASE` expression that compares `previous_status` and `status` (both `bit` columns from `config.trace_flags_history`) against nvarchar literals:

```sql
WHEN rc.previous_status = N'OFF' AND rc.status = N'ON'  THEN ...
WHEN rc.previous_status = N'ON'  AND rc.status = N'OFF' THEN ...
```

SQL Server resolves the type mismatch by converting the string literal to `bit`. The error always names `'OFF'` here likely because it's the first literal SQL Server attempts to fold

```
Conversion failed when converting the nvarchar value 'OFF' to data type bit.
```

The view filters to `change_number > 1` (rows with a non-NULL `previous_status`), so the first INSERT into the history table never triggers the error. The view breaks the moment a second row exists for any trace flag - i.e. after any toggle is recorded.

### Steps to Reproduce

```bash
docker pull mcr.microsoft.com/mssql/server:2022-latest

docker run -e ACCEPT_EULA=Y -e MSSQL_SA_PASSWORD='HopeThisIsGoodEnough123!' \
  --name sqltest -p 1433:1433 -d mcr.microsoft.com/mssql/server:2022-latest

# brief pause for sql service to start up
sleep 15
```

Create a minimal history table in `tempdb` (not a `#temp` table - each `docker exec sqlcmd -Q` call is its own connection/session, so a session-scoped temp table wouldn't survive between steps) and insert two rows (initial + toggle):

```bash
docker exec -it sqltest /opt/mssql-tools18/bin/sqlcmd \
  -S localhost -U sa -P 'HopeThisIsGoodEnough123!' -C -Q \
  "CREATE TABLE tempdb.dbo.history (
     collection_time datetime2,
     trace_flag      int,
     status          bit,
     is_global       bit,
     is_session      bit
[...]

PR #1677, opened 2026-07-26
CI: execute every report view against seeded rows (#1669)

### LINKED ISSUE #1669: CI: report.* views are validated by existence only — execute them against seeded rows so throw-on-execute bugs get caught
Spun out of #1660 (and #1635 before it) — both were the same class: a view that compiles and exists but fails only when rows flow through it (Msg 245 bit→nvarchar in #1635; the reader-side cast in #1660 was its downstream twin). `.github/sql/ci_validate_installation.sql:128` checks `report.*` views by `OBJECT_ID` only.

A `SELECT TOP (0)` sweep would NOT close this: conversion errors like Msg 245 fire per-row, and a fresh CI install has empty tables anyway. To be meaningful the validation needs **seed rows** in the config/collection tables that the report views read (one representative row per source table, covering NULLable branches like `previous_status`), then `SELECT *` from every `report.*` view and fail on any error.

Sketch:
1. Seed step after install: minimal INSERTs into the tables `report.*` views read (incl. a trace-flag toggle pair so LAG paths execute).
2. Cursor over `sys.views WHERE schema = 'report'`, `EXEC (N'SELECT * FROM report.' + QUOTENAME(name))` inside TRY/CATCH, collect failures, THROW at the end with the list.
3. Keep it in the existing sql-validation workflow (SQL 2022 service container).

erikdarlingdata:
Closes #1669 — the follow-through on the #1635/#1666 bug class: views that compile and exist but only fail (or corrupt) when rows flow through the projection. `ci_validate_installation.sql` checks `report.*` by `OBJECT_ID` alone.

## What's added
- **`.github/sql/ci_seed_report_sources.sql`** — one guarded representative row per collect/config table, generated from the installed schema (NOT NULL/non-identity/non-computed columns get type-appropriate values; nullable columns stay NULL so those branches execute; compressed-LOB columns get real `COMPRESS()` payloads; timestamps land 5 minutes back so "today"/"last hour" windows include them). The three change-history tables get hand-authored correlated PAIRS (trace flag OFF→ON, server knob changed, database setting changed) so the LAG/toggle views emit actual rows. Every insert is guarded on the table being empty — re-runnable anywhere, inert on real data.
- **`.github/sql/ci_execute_report_views.sql`** — enumerates `sys.views` in the report schema at run time and materializes each with `SELECT * INTO` — `COUNT(*)` would let the optimizer prune the projection and skip exactly the per-row conversions this exists to catch. Failures are collected and re-thrown as one error listing every broken view; `sqlcmd -b` turns that into a red job. Dynamic enumeration covers future views automatically and tolerates the collection-time-built `report.query_snapshots` pair being absent on a fresh install.
- **`.github/sql/ci_generate_seed_rows.sql`** — dev tool (not run by CI) to regenerate the seed section after schema changes.
- **`sql-validation.yml`** — two new steps after installation validation, across the full 2017/2019/2022/2025 matrix. The workflow paths-filter includes `.github/sql/**` and the workflow file, so **this PR's own checks run the new sweep end-to-end on all four versions**.

## Validation
- Local full install on a real SQL Server 2022 (every install script 01–51 minus agent/test): seed clean, sweep **41 views executed, 0 failed**.
- Detection power: a planted view with a per-row-only `Msg 245` (`CONVERT(int, N'ci')` over a seeded column — the #1635 shape, invisible over zero rows) flips the sweep to `1 failed` with the view named, and the final `THROW` propagates.
- Known limit, stated in the script header: silent-semantic bugs like #1666's (valid execution, wrong value) still need targeted checks — this catches the throw-on-execute class.


erikdarlingdata:
**SQL Server 2016 coverage** (the min-supported version the matrix can't include — SQL Server on Linux starts at 2017, so no 2016 container image exists):

Ran the identical sequence manually against the local SQL2016 box (13.0.6300), scratch database, fixture `PerformanceMonitor` DB untouched:
- full install surface (01–51 minus agent/test scripts): clean
- `ci_seed_report_sources.sql`: clean, twice
- `ci_execute_report_views.sql`: **41 views executed, 0 failed**

The matrix now carries a comment recording why 2016 is absent and that it's validated manually as part of release testing.

Issue #991, opened 2026-05-26
[FEATURE] Availability Group Health Monitoring


mark-bentham:
### Which component(s) does this affect?

- [x] Full Dashboard
- [x] Lite
- [x] SQL collection scripts
- [ ] Installer
- [ ] Documentation

### Problem Statement

I'd like to be able to monitor the health of availability group replicas, with alerting should the secondaries fall significantly behind the primary, so we know if we need to use the replica it is not significantly out of date and can fail-over safely if required

### Proposed Solution

additional collector querying availability group dms collecting stats like redo_queue_size, redo_rate, log_send_queue_size, log_send_rate, synchronization_state etc, plus visualisation in the dashboard and alerting if the secondaries fall significantly behind the primary

### Use Case

Useful for anyone using always on availability groups for HA, to allow them to monitor the health of the availability group and to identify issues, such as the secondaries falling behind during periods of heavy usage, so that I can identify and resolve them.

erikdarlingdata:
@mark-bentham great, I've assigned this to you for now. I don't have any interest in AG monitoring, so it's not something I'm up for building here.

erikdarlingdata:
Closing as not planned, consistent with what was already said in-thread.

@mark-bentham — Availability Group health monitoring is a real gap, and it is honestly the one area a competing tool would point at. It is not something being built here.

To be precise about what does and does not exist, so nobody reads partial credit into it: three places touch HADR DMVs, and none of them is AG health. `ServerPropertiesCollector` derives a single Primary/Secondary/Standalone string for inventory display, `QueryStoreCollector` left-joins `dm_hadr_database_replica_states` to attribute Query Store rows to a replica role, and a FinOps recommendation counts non-basic AGs. There is no collector for redo queue, log send queue, send rate, or synchronization state, and no AG alerting.

Reopen or open a PR if you want to take it on.

PR #1690, opened 2026-07-26
Availability Group topology: web page + get_ag_health MCP tool (#991)

erikdarlingdata:
Closes the reader half of #991, stacked on the AG collectors from #1688 (rebased onto `dev` after that merged).

Two surfaces, one shared reader. `DarlingAgReader` serves both `GET /api/ag` (the new **Availability Groups** page) and the new **`get_ag_health`** MCP tool, the same way `DarlingFleetReader` backs `/api/fleet` + `get_fleet_overview` — so an agent and a browser cannot disagree about an AG's state.

## What it shows

One card per **(reporting server, AG)**: header with the AG name, a worst-health badge and the current primary; replicas as chips (role, connected/operational state, synchronization health, availability + failover mode, endpoint on hover); then a database grid with synchronization state, log-send/redo queue sizes, send/redo rates, secondary lag, derived drain estimates, and the suspend reason when data movement has stopped.

An AG whose replicas are all monitored appears **once per monitored replica**, each naming its reporting server. That is deliberate, not duplication: the DMV populates `operational_state` and `recovery_health` for the LOCAL replica only, `connected_state` is meaningful only from the primary, and a quorum-lost instance answers from locally cached metadata. Merging them would invent a consensus the DMVs never reached; comparing them is how a split view gets spotted.

## Judgement calls worth a reviewer's attention

Every value list below was verified against MS Learn rather than assumed.

- **`SYNCHRONIZING` is banded against the replica's availability mode, not flat.** It is the correct steady state for an ASYNCHRONOUS_COMMIT secondary (async never reaches SYNCHRONIZED) and a warning on a SYNCHRONOUS_COMMIT one. A flat state-to-color map necessarily gets one of those two wrong.
- **Suspension outranks every state string.** `secondary_lag_seconds` reads `0` — not null — while data movement is suspended, so a stopped replica otherwise looks perfectly caught up. The lag cell says `0 s (suspended)` for the same reason.
- **`ONLINE_IN_PROGRESS` belongs to `recovery_health_desc`, not `operational_state_desc`.** An early draft banded it on the wrong column, which made that arm unreachable while the column it belongs to went unbanded. `recovery_health` and `role` (RESOLVING) are now banded and fold into the roll-up.
- **A column the DMV leaves null bands `Unknown`**, the lowest enum value, so a remote replica's nulls never inflate a group's severity.
- **Lag and queue depth are deliberately NOT banded.** Every other severity here restates a verdict the DMVs already made; banding a raw magnitude means choosing "how many seconds behind is a warning," which is monitoring policy and belongs with the configurable alert thresholds, not hardcoded in a read. The magnitudes are surfaced per database instead. **This is the one open decision** — wiring AG lag into the alert engine with a real threshold is a follow-on, and I did not want to invent the number silently. Consequence today: a badly-lagging async secondary whose replica health still reports HEALTHY does not redden the card badge.
- `est_send_drain_minutes` / `est_redo_completion_minutes` are **derived** queue/rate, not collected columns; null when a non-empty queue is moving at zero, rather than a misleading `0` or an infinity.

## Choices not settled by an existing idiom

- **Nav gating.** There was no precedent for a conditionally-visible page, so the nearest one is `refreshViewList`'s conditional affordance plus `ComposeStoreAvailability`'s converge-then-stop probe: the entry is `hidden` in markup and revealed once the store reports AG rows, after which the probe stops. Always On is opt-in, so most fleets would otherwise carry a permanent dead link. `#/ag` routes regardless, with an honest empty state — a deep link is never broken by the entry being hidden. This needed one CSS rule (`.nav a[hidden] { display: none }`) because `.nav a`'s author-origin `display: flex` outranks the UA stylesheet's `[hidden]`.
- **`/api/ag` takes no `server` param**, mirroring `/api/fleet`; the per-server view is reachable through `/api/read/get_ag_health?server=`, which owns the name-resolution error path.
- **Both reads join the enabled server registry**, like every other fleet read, so a disabled server's AG cards leave with it instead of lingering until retention expires the rows.

## Validation

- Full `Darling.Tests` suite green against a **fresh** Postgres where #1688's V34 migration created the tables: **3363 passed, 0 failed**, zero build warnings.
- Ungated tests pin the banding rules, the drain estimates, the grouping, and the SQL dialect. Live-Postgres tests pin the things text assertions cannot: that the latest-snapshot join returns **only** the newest instant's rows (two instants planted with different replica sets), that the server filter actually **excludes** a second server, that a disabled server drops out, and that an empty store answers the empty envelope rather than an error.
- The page's render path was executed against a DOM shim with a null-heavy payload (null `ag_name`, null replica columns, a suspended database) before it ever reached a browser.

Not visually reviewed in a real browser — that is Erik's pass.

Heads-up unrelated to this change: while running the live suite repeatedly against one shared database I saw an intermittent failure that landed on a **different** unrelated test each time (`StoreConfigProviderTests.ConfigVersion_BumpTriggers_...`, then `ViewerMemoryLivePostgresTests.MemoryGrantTrend_...`), never on the AG tests. `StoreConfigProviderTests` is the one live class without `[Collection("live-postgres")]`, so it runs in parallel with the 62 that have it. CI uses a throwaway cluster per run, which is likely why it has stayed green. Worth its own issue.


erikdarlingdata:
Follow-up commit `0b90f715`: `get_ag_health` had to be recorded in `Lite.Tests/CrossAppMcpToolInventoryPinTests.KnownLiteMissingMcpTools`. That pin scans both apps' `Mcp/*.cs` and requires Darling's tool set to be a superset of Lite's, so a new Darling-only tool fails the `build` job — and a Darling-only local build never runs Lite.Tests, so it would only have shown up in CI.

Worth noting it is entered as a genuine **"not ported yet"**, not a Darling-only-by-architecture item like the Custom Views / alert-tuning / onboarding entries: #1688 landed the AG collectors in **both** apps, so Lite already has `ag_replica_states` + `ag_database_replica_states` in its local DuckDB. A Lite twin is a DuckDB reader over the same banding rules, and the per-replica / per-database severity logic currently in `DarlingAgReader` would move to Common on that port. Ratchet only shrinks, so removing the entry is the definition of done for the Lite side.

Issue #1706, opened 2026-07-26
RELEASE BLOCKER: bundle must ship PostgreSQL 18 (zstd blob compression) — prior assurance of 18 was false; runtime-upgrade machinery for existing 17 stores


erikdarlingdata:
Surfaced by #1705: field stores run whatever pg runtime + TimescaleDB they were initialized with, forever. Deploys swap app binaries but deliberately preserve the pg runtime dir, so DARLING01 still runs its original extension while the bundle ships 2.28.1 — which is exactly how the `scheduled =>` signature gap arose. There is no mechanism to push a runtime update with the package.

## Design sketch (versioned, healed-on-start — the ConfMarker/migrations discipline applied to the runtime)
- **Runtime manifest**: service knows its bundled versions; at service-owned start compare `SHOW server_version` + `SELECT extversion FROM pg_extension WHERE extname='timescaledb'` against bundled.
- **TimescaleDB extension update** (the near-term win): bundled tsdb > installed → stop pg, swap the runtime dir (keep the old one as `runtime_prev` for rollback), start on the SAME data dir, then run `ALTER EXTENSION timescaledb UPDATE;` as the FIRST statement on a fresh connection (Timescale's alone-first rule), stepwise through the shipped upgrade files. Verify extversion == bundled; on any failure swap back + fire a self-alert. Same flow covers **PG minor** bumps (17.x data dirs are binary-compatible).
- **PG MAJOR upgrades** (17→18): separate, bigger phase — pg_upgrade orchestration needing BOTH runtimes present, an offline window at service start, disk headroom for --link mode, and matching tsdb versions on both sides. Design when a major actually looms; the manifest machinery above is the prerequisite either way.
- **Test story**: an upgraded-in-place fixture (docker: init a store on an older image/extension, then run the current service against it) — the gap #1705 proved darling-pg CI cannot see with fresh 2.28.1 stores.
- **Field discipline**: one runtime version change per release, changelogged; the arm-when-safe pause pattern from #1680 applies to any post-upgrade policy work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
Erik's direction + a measured correction:

- **Target: PostgreSQL 18** — the driver is the better compression on blob columns (zstd) for the plan-XML/query-text heavy tables. PG 19 is explicitly not a near-term target.
[...]

Issue #1705, opened 2026-07-26
RELEASE BLOCKER: #1684's scheduled => false breaks retention on upgraded-in-place TimescaleDB stores (0/7 policies on DARLING01)


erikdarlingdata:
Found live on DARLING01 after the dev@35dc49a6 deploy: retention went from 7/7 policies to **0/7 with 7 failures**. Every tier grows unbounded until fixed.

Commit 0ef2f10a (PR #1684, for #1680) calls `add_retention_policy(..., scheduled => false)`. That named parameter does not exist in the TimescaleDB version actually installed in upgraded-in-place stores (DARLING01 upgraded from the 2.9.x era): `42883: function add_retention_policy(... scheduled => boolean) does not exist`. Creation fails, so #1680's arm-when-safe step never runs either.

**Why CI missed it**: darling-pg provisions a FRESH 2.28.1 store where the signature exists. Only upgraded-in-place stores hit it — i.e., every existing field installation. Shipping this in the release breaks retention fleet-wide while looking green in CI.

**Fix (compat-safe two-step, keeps #1680's semantics on old AND new extensions):** create the policy WITHOUT the new parameter, then pause the job explicitly — `SELECT alter_job(job_id, scheduled => false)` has existed across the 2.x line. The arming step is unchanged. Add a code comment naming the version-compat reason so nobody "simplifies" it back. Do NOT fix via ALTER EXTENSION UPDATE on field boxes — that trades a code bug for a fleet-wide extension upgrade requirement.

DARLING01 self-heals on the first service start after the fix deploys (policy creation is idempotent on start).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
Fixed by #1711 (merged to dev). The regression was `add_retention_policy(..., scheduled => false)`: that argument has never existed on any TimescaleDB 2.x, so every policy creation failed with 42883 and the per-policy try/catch downgraded it to a warning — field stores read `0/7 retention policies in place`. Replaced with create-then-`alter_job(job_id, scheduled => false)` inside ONE transaction, and the string-matching pin that let an unparseable statement through was replaced with a gated-live test that EXECUTES the SQL and asserts the applied count.

Verified in the field, unattended, on DARLING01 tonight: `7 retention policies` present via `timescaledb_information.jobs`, with PostgreSQL 18.4 / TimescaleDB 2.28.1.

PR #1718, opened 2026-07-26
Darling store runtime upgrade orchestration (#1706 leg 2) - in-place 17 to 18, validated end to end

erikdarlingdata:
Leg 2 of #1706: the service-orchestrated in-place upgrade of existing Darling stores. Leg 1 (the PG 18 + TimescaleDB-for-18 bundle) is separate.

**Read the blocker section before merging. The flow has never been observed to complete end to end.**

## What this adds

`Darling/PerformanceMonitor.Darling.Service/DarlingStoreUpgrade.cs`, wired into the existing `DarlingManagedPostgres` bootstrap.

The root problem: `EnsureRuntimeAsync` returns early the moment `pg-runtime\pgsql\bin\pg_ctl.exe` exists. A deploy that ships a newer `pg-runtime.zip` therefore changed nothing on any host that had already extracted one, which is exactly how a store ends up running its original extension forever (#1705).

The detector is now the zip's SHA256, recorded in `pg-runtime.stamp` at extraction. A hash that differs from the stamp means the package carries a new runtime, and that single signal covers both a TimescaleDB-only bump and a PostgreSQL major jump.

Flow, in order:

1. **Runtime advance.** Rescue the current `pg-runtime\pgsql` to `pg-runtime-prev\pgsql` (whole tree, not just `bin` - the old postmaster resolves `$libdir` relative to itself), then extract the new zip. Rescue-before-extract is load-bearing: pg_upgrade needs both runtimes present simultaneously, and getting the order wrong leaves a store that can never be upgraded. Refuses to swap under a live postmaster, since stopping a server this service did not start is not its call.
2. **Refuse a downgrade.** A data directory NEWER than the bundled runtime throws rather than starting - PostgreSQL cannot open a newer cluster with older binaries, and there is no supported downgrade.
3. **Headroom.** Measured, then copy mode (default, old cluster survives intact), hard-link mode (only when copy will not fit, only when a probe confirms the volume really makes hard links, and always with a LogCritical because it trades the rollback away), or abort-safe.
4. **TimescaleDB bridge, on the OLD cluster.** Mandatory, not hygiene: pg_upgrade recreates the extension pinned at whatever version is installed, and every TimescaleDB function resolves to a version-suffixed library that the new runtime ships exactly one of. Runs in every database that has the extension, template1 included, each on a fresh unpooled connection whose first statement is the ALTER (TimescaleDB rejects the update once its library is loaded into the session).
5. **initdb the new cluster to the old one's MEASURED identity.** Locale provider, collate, ctype, encoding and checksums are read from the live old cluster (`template0` plus `SHOW data_checksums`), never assumed - PostgreSQL 18 flipped initdb to checksums-on by default and pg_upgrade hard-refuses a mismatch.
6. **Managed conf BEFORE pg_upgrade.** `shared_preload_libraries = 'timescaledb'` has to be live on the new cluster or pg_upgrade's internal restore cannot load the extension it is restoring. Healing it afterwards would be too late, so `EnsureConfAppended` now takes a data directory.
7. **`--check` as a dry run, then the real pass.** Both via the no-redirect runner, because pg_upgrade starts postmasters that inherit redirected handles and hold them open; diagnostics come from the logs pg_upgrade writes under the new data directory and keeps on failure.
8. **Swap, verify, analyze.** Directory renames, then a server-version check, an extension-version check, a real read of `collect.collection_log`, and `vacuumdb --analyze-in-stages --missing-stats-only` (PG18 carries most statistics across; this fills the documented gaps).

Any failure at any step reverts the runtime, leaves the untouched old data directory exactly where it was, and records the failing zip's hash in `pg-runtime.blocked` so a known-bad package is not retried on every service restart. The store is never left down. The pre-upgrade data directory is kept as a rollback copy for two service starts, then deleted with a log line naming the space reclaimed.

Same-major runs still get the TimescaleDB extension update on their own - that is the #1705 case without any pg_upgrade at all.

## The blocker

`pg_upgrade --check`, EDB PostgreSQL 17.10 to 18.4, Windows 11, reproducibly stops after:

```
Checking cluster versions                                     ok
Checking database connection settings                         ok
```

`--verbose` shows the last thing it does is the `get_db_infos()` query, so it hangs as it begins opening its per-database connections. It never exits; I let it sit 13 minutes.

Ruled out by direct measurement, not reasoning:

| Suspect | How it was eliminated |
| --- | --- |
| TimescaleDB | A control cluster with no extension and no `shared_preload_libraries` entry hangs identically |
| The service's process runner | Reproduces from a plain foreground shell and from `Start-Process`, not only under the runner |
| Output-pipe blocking | Reproduces with stdout redirected to a file |
| Console-handle inheritance | Reproduces when launched with its own real console window |
| Unclean source cluster | `pg_controldata` reports `shut down` immediately before each run |
| Authentication | The old server logs no failed attempt, and `pg_stat_activity` shows pg_upgrade never opens a connection at all |
| The Windows restricted-token re-exec | `PG_RESTRICT_EXEC=1` skips the privilege-dropping re-exec PostgreSQL's Windows binaries perform for an Administrator. Confirmed it took effect (one `pg_upgrade.exe` process instead of the usual parent/child pair); the hang is identical, at the same line |

**No hypothesis survives.** The restricted-token re-exec was the last one standing and it is now eliminated, so the next step is a different machine and a CI runner rather than more variations on this box. I would rather hand over a clean elimination list than a guess.

### One thing is NOT settled, and it is the important one

The agent building leg 1 measured the **opposite** result on this same machine - `pg_upgrade --check` 17.10 to 18.4 completing in about 3 seconds while elevated - and attributes any hang to the caller's pipes. I could not reproduce that, and I specifically re-tested the things that claim rests on: clean slate with `Get-NetTCPConnection -LocalPort 50432` verified showing zero listeners beforehand, fresh 17.10 and 18.4 clusters with `--data-checksums` passed explicitly on both sides, output redirected to files via the exact `Start-Process ... -RedirectStandardOutput <file>` shape they document, same elevated identity (`IsInRole(Administrator)` = True). It hung at the same line and was killed at 150s. The old postmaster starts cleanly on 50432 every time (TimescaleDB even logs "the database is restoring or upgrading", so it sees binary-upgrade mode); pg_upgrade then never connects to it and never starts the new cluster.

I have asked them for their literal command line and, more importantly, what was actually IN their old cluster. If it was a bare `initdb` with no user database and no `shared_preload_libraries` line, that is a materially different test from mine and would reconcile both measurements - the blocker would be real but narrower than "pg_upgrade is broken on Windows". **Until that exchange lands, treat neither result as settled**, including mine. If they turn out to be right, this PR's framing is wrong and it can be unblocked the same day.

The `-o`/`-O` quiesce flags (`timescaledb.max_background_workers=0`) passed to both clusters are defensible independently - timescale/timescaledb#1593 documents the scheduler background worker deadlocking against restore-type workloads on `template0`, and nothing it would do during an upgrade window is wanted anyway - but they are a precaution, not the fix here, since the control had no scheduler to quiesce.

## Testing

- Full `Darling.Tests`: **3270 passed, 0 failed, 157 skipped** (the usual gated ones). No live `Installer.Tests` categories were run.
- 18 new ungated tests pin the pure decisions: version and control-file parsing, the transfer-mode arithmetic, and the initdb/pg_upgrade argument shapes - including that a checksum-less 17 store gets an explicit `--no-data-checksums` on 18 and nothing on 17, which is the case PostgreSQL 18's changed default would otherwise break silently.
- `Darling/tools/new-upgraded-store-fixture.ps1` builds the previous-major runtime (pinned URLs plus SHA256 values hashed from fresh downloads on 2026-07-26) and the current bundle, and prints the env vars for the gated test.
- The gated `UpgradeInPlace_...` test builds a REAL store on the old runtime - hypertable, TOAST-sized plan XML, continuous aggregate, compressed chunk - then runs the production bootstrap over a deployment whose shipped zip is the new runtime, and compares row counts, an ordered row checksum, continuous-aggregate rows and compressed-chunk counts before and after. **It currently hangs at the pg_upgrade step**, which is how the blocker was found. That fixture is the #1705 CI gap: darling-pg only ever creates fresh stores.

What I did verify directly: the PG 18.4 + TimescaleDB 2.28.1 bundle builds from dev's `fetch-pg-runtime.ps1` with both hashes matching their pins, and the orchestration's own steps all execute correctly against a real 17.10 store - runtime rescued, new zip extracted, new cluster initdb'd with derived settings, managed conf written, pgpass staged, pg_upgrade invoked with the right arguments - up to and including the call that hangs.

## Sequencing, which matters more than usual here

Leg 2 is **inert on its own** - its trigger is "the shipped zip's hash differs from the recorded stamp", and dev ships no new zip. It goes live the moment leg 1 ships a PG 18 bundle, and at that point every field store enters upgrade mode on its next service start, hangs in `pg_upgrade --check`, and is OFFLINE until that budget expires before reverting to 17. I cut the dry-run budget to 15 minutes for exactly this reason, and a failed attempt records the zip hash so the same package is never retried - so the worst case is a bounded one-time 15-minute outage per box rather than hours. That is still not something to ship deliberately. **Leg 1 and leg 2 must not both land until pg_upgrade is understood.** Auto-merge is deliberately not armed.

One premise correction now that #1713 has landed on dev: `zstd` is not a TOAST compression method in PostgreSQL 18, so there is no v7 zstd conf marker for this PR to heal, and the "better blob compression" driver behind the 18 move does not exist as stated. `default_toast_compression = lz4` has been in the v1 block all along. Nothing here depends on that either way, but it removes a step the original design sketch called for.

Closes nothing. #1706 stays open until pg_upgrade completes.

## Deliberately deferred, recorded rather than silent

**A `PG_VERSION` refuse-guard inside `RevertRuntime` itself is NOT in this PR.** The review raised it as a second, independent guard: `RevertRuntime` would read `PG_VERSION` at the configured path and refuse when the data directory is already on the new major, so the protection holds for any future caller rather than only for the callers that consult the commit-point flag. Verified absent at `f39fd960` -- `RevertRuntime(string runtimeRoot, string zipHash)` takes no data directory and reads no `PG_VERSION`.

It is deferred on the reviewer's explicit ruling that it is not a blocker, because at this sha the guard is redundant in practice: `RevertRuntime` has exactly two callers, the pre-commit handler (unreachable when `swapped`, since the filtered `catch ... when (swapped)` clause precedes the unfiltered one) and `RevertRuntimeForCancel` (gated by `if (!swapped)`), so it cannot be reached after the swap commits. The value of the in-function guard is defence-in-depth against a future third caller that forgets the flag -- real, but not a reason to hold a release.


erikdarlingdata:
CI note: the `Darling PostgreSQL tests` job is red here, and it is the runner's Postgres service, not this branch.

Two runs, two different sets of failing tests, both infrastructure-level rather than assertions:

- Run 1: `ViewerWaitStatsLivePostgresTests` (x2) and `ViewerDailyHealthLivePostgresTests` - `SocketException: An existing connection was forcibly closed by the remote host`.
- Run 2 (re-run of the failed job): `DarlingComposeTests` and `DarlingEndpointToggleCliTests` - `58P01: could not access file "plpgsql": No such file or directory`, thrown out of `PgMigrations.MigrateLockedAsync`. plpgsql is a core extension that ships with every PostgreSQL install, so that is a broken service container, not a code path.

Different tests each time with different transport/filesystem errors is a flapping service. The same job also failed on a `dev` push earlier today (run 30220251272), and none of the failing tests touch anything this branch changes - they are all BYO-store tests against the service container, while this branch changes managed-mode bootstrap, the store-upgrade orchestration, and the self-alert evaluator.

Locally the full suite is green: **3276 passed, 0 failed, 159 skipped**.

Not chasing it from here, since it is not this branch's failure and this PR is marked do-not-merge anyway - but flagging it because it will bite whoever tries to land the next PR.

erikdarlingdata:
## Status update: the hang is fixed, the PR title was stale

Superseding the blocker section in the description. Three things changed since it was written.

### 1. The `--check` hang is FIXED, and it was ours

Two product bugs, both found in pg_upgrade's own diagnostics rather than by reasoning:

- **Port collision.** pg_upgrade defaults BOTH throwaway postmasters to the fixed, well-known **50432**. A leftover postmaster — or a second upgrade anywhere on the host — collides there and pg_upgrade silently talks to a stranger's cluster. Caught red-handed as `FATAL: role "darling" does not exist` returned by a postmaster this service never started. Now `--old-port 55432 --new-port 55433`. This is a real field robustness fix: any stray postmaster on a customer box would have caused the same silent cross-talk.
- **IPv4/IPv6 loopback.** pg_upgrade has no Unix sockets on Windows so it dials its clusters **by name**, and Windows resolves `localhost` to `::1` first — while our v1 conf block pins `listen_addresses = '127.0.0.1'`, IPv4 only. Verbatim from its log: `connection to server at "localhost" (::1), port 50432 failed`. Now `-c listen_addresses=localhost` on both clusters for the upgrade window only; the store's own conf is never edited. This is the same trap `BuildConnectionString` already documents and dodges by using the literal address — the lesson had never been applied to a tool that dials on our behalf.

### 2. A genuine bug in the revert path, which would have bitten in the field

The revert did `TryDeleteDirectory(pgsql)` then `Move(prev, pgsql)`. A delete can partially succeed — a postmaster that has not finished exiting still holds its binaries — and the swallowed failure left `pg-runtime\pgsql` with **no `pg_ctl.exe`**: an unbootable install produced by the very path that exists to make failure safe. Both failure paths now move-aside instead of delete-first. Found in one run because the gated test now uses a capturing logger instead of `NullLogger`.

### 3. What still fails

The orchestration runs correctly end to end against a real populated store — runtime rescue, major detection, headroom, identity probe, TimescaleDB bridge, `--check` — and pg_upgrade then gets through `pg_dump` and deep into `pg_restore` (whole TimescaleDB catalog, ACLs, all four dumps, reaching globals) before exiting **-1 with no error in any of its own logs**. Reproducible at ~1m25s.

Ruled out for that: `--jobs` parallelism (serial fails identically — the correlation looked clean and did not survive testing; the code comment was corrected rather than left citing disproven evidence), TimescaleDB, elevation, the Windows restricted-token re-exec, pipes, console handles, and authentication.

Worth noting the box this ran on had two agents driving pg_upgrade concurrently plus 17+ postgres processes, and `initdb` itself intermittently failed with `pclose() failed` from process contention. A quiet machine or a CI runner is the next step.

**Still do-not-merge, and auto-merge stays unarmed.** Full suite green at **3283 passed, 0 failed, 0 warnings**.

erikdarlingdata:
## PASSES end to end on a quiet box — this supersedes every earlier blocker note

The gated `UpgradeInPlace_OldMajorStoreWithRealData_UpgradesAndKeepsEverything_Gated` fixture now **passes**. Evidence, read live from the upgraded store:

| Check | Result |
|---|---|
| Server version | **18.4** (from 17.10) |
| TimescaleDB | 2.28.1 |
| `collect.collection_log` rows | **20,000** — identical before and after |
| `collect.query_plans` rows | **500** — identical |
| Plan XML bytes | **41,359,100** (~41 MB of TOAST) — identical |
| Ordered row checksum | **identical** — not just counts; a truncated or reordered restore would fail this |
| Continuous aggregate rows | **1,336** — still answering through the upgraded cluster |
| Compressed chunks | **4** — survived as compressed chunks |
| Rollback copy | `store\pg-old-17` present, aged over 2 starts |
| Conf markers | v1 and v6 present exactly once each — written before pg_upgrade, not duplicated by the heal |

Full suite: **3283 passed, 0 failed, 0 warnings**.

## The four bugs the validation flushed out — three of them ours

1. **Port collision.** pg_upgrade defaults BOTH throwaway postmasters to the fixed well-known **50432**. Any leftover postmaster, or a second upgrade anywhere on the host, collides and pg_upgrade silently talks to a stranger's cluster — caught as `FATAL: role "darling" does not exist` from a postmaster this service never started. Now `--old-port 55432 --new-port 55433`. A field robustness fix, not just a dev-box one.
2. **IPv4-only loopback.** pg_upgrade has no Unix sockets on Windows so it dials its clusters **by name**, and Windows resolves `localhost` to `::1` first, while our v1 conf pins `listen_addresses = '127.0.0.1'`. Verbatim: `connection to server at "localhost" (::1), port 50432 failed`. Now `-c listen_addresses=localhost` for the upgrade window only; the store's conf is never edited.
3. **A revert that could brick the runtime.** It deleted the live runtime before moving the rescued one back; a partial delete (a postmaster still holding its binaries) left `pg-runtime\pgsql` with no `pg_ctl.exe` — an unbootable install produced by the safety path itself. Both failure paths now move-aside, never delete-first.
4. **A first-connection race.** `pg_ctl -w` returns when the postmaster accepts connections, which on Windows is not the same as the next backend spawning and surviving — each re-reserves the shared-memory region at the postmaster's base address (the error-487 surface behind our 1 GB `shared_buffers` cap). Losing it authenticates fine then dies on the **first query**, so retrying only the Open never helped; the whole open-and-first-query unit now retries with a fresh connection, while a `PostgresException` still fails fast.

Plus an off-by-one the passing run exposed: the retention sweep ran *after* the upgrade and aged the brand-new rollback copy on the very start that created it.

## Corrections to my own earlier claims on this PR

- The "reproducible pg_upgrade hang" was **not** a PostgreSQL or Windows defect. It was bug 1 above plus cross-agent contention on this box.
- I earlier claimed `listen_addresses` was the proven root cause from a clean-looking bisect. That bisect was contaminated by another agent's postmaster on 50432; under a port guard the same config passed once and failed once. The fix is still correct and evidence-backed, but the inference I sent was not, and I retracted it.
- `--jobs` was suspected and disproven (serial fails identically); the code comment says so rather than citing evidence that did not hold.

Ready for the shepherd under the standing ruling.

Issue #1696, opened 2026-07-26
AG alerts: post-release follow-ups (Lite parity, cross-node dedup, disconnect re-fire)


erikdarlingdata:
The #1692 AG alert family shipped with three deliberate, stated limits. Address after the current release:

1. **Lite AG alert parity** — Lite collects both AG grains (#1688) but has no AG alerts. `JudgeAgSync` is a pure static designed to lift into `PerformanceMonitor.Common` (the `ConnectionAlertPolicy`/#1659 discipline: shared class, identical test matrices in Lite.Tests + Darling.Tests). Wire Lite's alert path to the same four alerts with the same metric names.
2. **Cross-node de-duplication** — a 3-node AG with all nodes monitored reports one role change once per monitored server. The replica-grain collector doesn't carry `is_local`; add it (V-next migration + collector column) and de-dup fleet-side (prefer the primary's perspective, or is_local=1 rows) in the Darling evaluator.
3. **AG Replica Disconnected re-fire** — currently a pure edge; a standing disconnect never re-announces. Extend with the #1659 opt-in re-fire pattern (same `ConnectionAlertPolicy`-style clock, stamp-on-delivery, default off).

Also from the same round, still open elsewhere: #1687 (compose HardRowCap honest truncation). Topology-page follow-ons (failover history timeline off the collected role edges) can ride with item 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
Complete. All three legs merged to dev:
- **Lite parity** — #1726: Lite raises the same four AG conditions under the same metric names, off a shared `PerformanceMonitor.Common.AgAlertPolicy` (Darling refactored onto it with no behavior change).
- **Cross-node de-duplication** — #1734: `ag_replica_states.is_local` (read from the DMV, not inferred) plus vantage ranking so one server judges each AG; the AG edge state drops the server id because an AG is one object however many nodes watch it.
- **Disconnect re-fire** — #1734: `ag_disconnect_refire_minutes` (default 0 = off, clamped 0-1440), stamped on delivery only so a suppressed alert cannot consume the window, cleared on reconnect.

Both columns ride store migration V37. The rendering half followed in #1735, which surfaces `is_local` on every AG surface so the vantage the dedup depends on is visible to an operator.

Issue #1681, opened 2026-07-26
Self-alert FIRINGS are never logged — only resolutions are, so the log shows 'Recovered' with no preceding event


erikdarlingdata:
Field finding from prod-sql-use2-monitor-01. Raised about the compression-job stuck alert specifically; **verifying it showed the gap is broader — it affects every self-alert, not just that one.**

## Verified in code

`DarlingSelfAlertEvaluator.RecordResolutionAsync` (`:963`) logs at Information:

```csharp
_logger?.LogInformation("[{Server}] {Title}: {Message}", resolution.ServerName, resolution.Title, resolution.Message);
```

The **firing** path has no equivalent. It calls `_deliverer.DeliverAsync(new AlertOutcome(...))` (`:954-958`) with no logger call at all. The only `_logger` uses in the compression path are `LogError` for an exception (`:671`).

So the service log shows **"Compression Job Recovered"** with no preceding "Compression Job Stuck" — which is worse than logging neither, because it reads as a spontaneous recovery from nothing. Anyone tailing or grepping the log sees half a story.

This applies to every self-alert: disk pressure, capture-down, agent-not-running, collection-health. All fire silently and only announce themselves when they clear.

## Field evidence for why it matters

9 stuck events over 6 days on that store (2026-07-21 through 07-26), always jobs 1020/1026 — the compression policies for `procedure_stats` and `query_stats`, the two highest-write-volume hypertables. No other compression job (of 37) has ever appeared. Auto-heal has worked every time, so nothing was lost — but the only record is `config.config_alert_log`, and you have to already suspect the problem to query it.

## Fix

Log the firing where resolutions are logged, at a severity matching the alert (Warning for these), so the pair is symmetric. One change in the shared deliver path covers all self-alerts rather than special-casing compression.

## Related: no TimescaleDB job audit trail exists on that store

`timescaledb.enable_job_execution_logging` is off (the default, never explicitly enabled), so `timescaledb_information.job_errors` and `job_history` return **zero rows for every job** — including jobs with dozens of confirmed successful runs. There is no per-run start/finish/success/error record for any background job, so a hung run leaves no captured error, no worker PID, nothing.

Recommend enabling it in the managed-Postgres conf append (`DarlingManagedPostgres.BuildConfAppend`) so the next occurrence produces evidence. Cheap, and there is no way to diagnose the hang category without it.

## Hypothesis for the hang itself, explicitly unconfirmed

[...]

PR #1754, opened 2026-07-27
Restore the zero-warning build after #1750 (#1681)

erikdarlingdata:
Micro-PR. My #1750 merge left `CS8604` on dev; this restores the zero-warning build.

## The reported warning

`AlertOutcome.ShortMessage` is `string?` with a default of `null` — several engine families pass nothing — and `AlertFiringLog.Fired` declared it non-nullable, so every engine fire site was a possible-null argument.

Widened to `string?`, per pg-upgrade-builder's diagnosis. The method only interpolates the value, so a null is harmless — but rather than emit a line ending in a dangling `": "`, the separator is now omitted when there's no summary. An alert with no short message must still log its trigger; being unable to summarize it isn't a reason to lose the record that it fired.

## Two more, both mine, found by sweeping the full solution

- **`CS8602` in `Lite/MainWindow.AlertEngine.cs`** — and this one is **a real latent bug, not just a warning**. The #1696 AG evaluation dereferenced the nullable `_dataService` unguarded. That method runs **unawaited** off a UI-timer tick, so it can land before the data service is wired during startup, and an NRE on a fire-and-forget task is an unobserved exception nobody ever sees. Now captured once and null-checked, the same shape as the tray capture in `ShowAlertResolutionToastAsync`.
- **`xUnit2031` in `Lite.Tests`** — `Assert.Single(x.Where(p))` → `Assert.Single(x, p)`.

Both are from this evening's work, so they're mine to fix rather than someone else's warning to absorb.

## Why I shipped the regression

Worth recording, because it's the actual root cause: my build verification grepped for `error|Build succeeded`, which filters warnings out entirely. Every green build I reported this evening was blind to them. Checking the `N Warning(s)` line is the fix — and it's what caught the two extra ones here rather than shipping a second follow-up.

## Verification

Full solution: **0 Warning(s), 0 Error(s)**. Checked individually as requested — Notifications, Alerting, Lite, Darling Service, Darling Viewer all clean.

Suites green: **Darling 3365**, **Lite 1579**, **Dashboard 768**.


erikdarlingdata:
Superseded by #1753, which covers both warnings this PR was assigned for (the CS8604 in AlertFiringLog and the #1726 CS8602 in Lite's AG evaluation) — our messages crossed and I had already pushed. Closing so two armed PRs never cover the same fix.

One item here is NOT in #1753 and is still live on dev: `xUnit2031` in `Lite.Tests/AgAlertEvaluatorTests.cs:225` (`Assert.Single(x.Where(p))` → `Assert.Single(x, p)`), which my full-solution `-t:Rebuild` sweep surfaced. It is mine, from #1696. Shipping it as its own micro-PR covering only that line, so there is no overlap with #1753.

PR #1755, opened 2026-07-27
Fix the last zero-warning holdout: xUnit2031 in the AG evaluator tests

erikdarlingdata:
The remainder after #1753. One line, no overlap.

#1753 fixed the `CS8604` in `AlertFiringLog` and the `CS8602` in Lite's AG evaluation — both of which my now-closed #1754 also covered, so I closed that as superseded rather than leave two PRs carrying the same fix.

**This one warning was not in #1753 and was still live on dev**, so the zero-warning build was not yet restored:

```
Lite.Tests/AgAlertEvaluatorTests.cs(225,9): warning xUnit2031
```

`Assert.Single(x.Where(p))` → `Assert.Single(x, p)`. Mine, from #1696.

The analyzer is right that the overload is better, and not only stylistically: the filtering form reports *"the collection contained 0 items"* on failure, having already discarded everything that didn't match, while the overload reports what **was** there. That's the difference between a useful failure message and a puzzle.

Scoped to this single line deliberately.

## Verification

Full solution on `-t:Rebuild`: **0 Warning(s), 0 Error(s)**. Lite suite: **1580 passed**.

(The `-t:Rebuild` matters here — an incremental build reports zero warnings for projects it didn't recompile, which is exactly how this one stayed hidden behind the other two.)

### DIFF
--- a/Lite.Tests/AgAlertEvaluatorTests.cs
+++ b/Lite.Tests/AgAlertEvaluatorTests.cs
@@ -222,7 +222,7 @@ public class AgAlertEvaluatorTests
 
         /* This server catches up. Its snapshot says nothing about the other server, whose alert must stand. */
         var recovered = e.EvaluateDatabases(ServerId, new[] { Database(lagSeconds: 0) }, 300, 0, Cooldown);
-        Assert.Single(recovered.Where(a => a.MetricName == "AG Sync Recovered"));
+        Assert.Single(recovered, a => a.MetricName == "AG Sync Recovered");
 
         /* The other server is still inside its cooldown, so it stays quiet rather than re-firing fresh. */
         Assert.Empty(e.EvaluateDatabases(otherServerId, new[] { Database(lagSeconds: 600) }, 300, 0, Cooldown));


erikdarlingdata:
Closing as superseded by #1756, which contains this exact line plus one more thing — and that other thing is yours, which is the part worth reading.

**Your #1754 was not a pure duplicate of #1753, and closing it lost a real improvement.** #1753 shipped my weaker guard:

```csharp
if (_dataService == null) { return; }
// ...then dereferences the FIELD twice, across two awaits
```

Yours captured first:

```csharp
var data = _dataService;
if (data is null) { return; }
```

That is better, and specifically better here: `EvaluateAvailabilityGroupAlertsAsync` runs **unawaited** off a UI-timer tick and awaits twice, so the field can be reassigned between the check and either use. Your own framing of the consequence was sharper than mine too — an NRE on a fire-and-forget task is an unobserved exception nobody ever sees, so the symptom is AG alerts silently not evaluating on some launches with nothing in the log. It also matches the `var tray = _trayService;` idiom the same file already uses, which mine did not.

So when you closed #1754 believing #1753 covered it, the `CS8604` half was indeed covered — but that capture refinement was not, and it went with it. #1756 carries it back, credited to you in both the commit and the CHANGELOG.

**And this one line was yours as well** — I only found `xUnit2031` because your #1754 contained it. I had declared the zero-warning build restored in #1753 having rebuilt only the two projects I touched; `Lite.Tests` was not among them, so my claim was narrower than it sounded.

Nothing here is being discarded: both of your items are in #1756, which is armed. Closing this only to avoid two PRs editing the same line into a conflict.

erikdarlingdata:
Reopening — my close was correct on the facts I had and wrong on the facts I did not.

I closed this as superseded because my #1756 contained this exact line plus the capture-first guard, so #1756 was a strict superset. Team-lead has since ruled the better disposition: **your authorship of the xUnit2031 fix should be direct rather than credit-by-note**, which I agree with — a commit is a stronger record than a sentence in someone else's CHANGELOG entry.

So: this PR owns `xUnit2031`, and I am stripping #1756 down to the capture-first guard refinement only, so there is no overlap in either direction.

Your point about incremental builds is the one I am taking away from this, and it is now a house rule rather than a lesson: **an incremental build reports 0 warnings for projects it did not recompile**, so "0 warnings" from a normal build can be hiding one. Verification greps the `N Warning(s)` line rather than `Build succeeded`, and the final sweep uses `-t:Rebuild`. That is exactly how this warning stayed invisible behind the other two, and why my "zero-warning build restored" claim in #1753 was narrower than I stated it.

Issue #1912, opened 2026-07-31
Query Store rows collected before #1907 keep their split slices: read-side dedup is now deterministic but still under-counts, and the materialized CAGGs cannot be tie-broken at all


erikdarlingdata:
Spun out of #1907, which fixes the defect going forward. This is the part of #1907 that the collector-side
fix cannot reach: rows **already in the store**, and CAGG buckets already materialized from them.

## What is left wrong

#1907's root cause is that `sys.query_store_runtime_stats` returns the flushed and the still-in-memory
slice of one `runtime_stats_interval_id` as two ADDITIVE rows, and the collector stored both. The fix
combines them at collection, so every row collected from now on is one row per interval. It cannot
rewrite what is already stored:

- **Raw rows** collected before the fix still hold two (or more) partial slices that share the entire
  read-side dedup key *and* `collection_time`. The dedup keeps exactly one of them, so those windows
  report a fraction of the interval's true `execution_count` and count-weighted totals.
- **The CAGGs** (`query_store_stats_interval_hourly` and everything hierarchical above it, including the
  corrected/daygrain dailies) materialized `last(execution_count, collection_time)` over those tied rows.
  TimescaleDB's `last()` has no tie-break at all, so the survivor was arbitrary.

## What #1907 already did about it

Made the read side DETERMINISTIC rather than flapping. All 19 read-side dedup sites in both apps now
order by `collection_time DESC, execution_count DESC`, so a legacy tie resolves to the FLUSHED slice —
the one holding the bulk of the interval's work — instead of a different slice on each run. In the live
repro that is 100 of a true 125 rather than a coin flip between 100 and 25.

That is closest-available, not correct. It is deliberately not presented as a fix.

## Why the rest was not done in #1907

Summing the slices on the read side would mean a second aggregation stage (`GROUP BY` the partition +
`collection_time`, weighted means for every `avg_*`, MIN/MAX for the extremes) at 19 sites across two SQL
dialects — reproducing the collector's aggregation by hand in the place the #1907 analysis specifically
concluded it does not belong. And it still would not repair the CAGGs: a continuous aggregate cannot
contain a window function (probed and rejected in #1853), and the tied rows are gone once materialized.

## Scope, and why it mostly resolves itself

- **Raw tier**: Darling keeps raw for 4 days, so raw legacy rows age out on their own. Lite's retention is
  operator-set but bounded the same way.
- **Hourly CAGG**: 21 days.
[...]

Issue #1841, opened 2026-07-30
Query Store aggregates double-count re-collected cumulative intervals - tier 2: interval identity + CAGG rebuild (tier 1 read-side dedup shipped in #1845)


erikdarlingdata:
Spun out of #1833's triage, now **confirmed against a live store** rather than theoretical.

`query_store_stats` rows are CUMULATIVE per-interval snapshots (the documented Query Store trap: dedup before SUM), but the collector stores no interval identity (`runtime_stats_interval_id` / interval start are absent from the schema — verified against the live column list) and the read queries aggregate raw rows. `LocalDataService.QueryStore.cs:36-48` (time-slice buckets) SUMs `avg_* * execution_count` across every collected row in the window; the other reads in that file need the same audit.

Live evidence from a real store (read-only), same `(server_id, database_name, query_id, plan_id)` within ONE hour bucket:

| rows in bucket | min exec_count | max exec_count |
|---|---|---|
| 496 | 1 | 1 |
| 482 | 1 | 1 |
| 297 | 1 | 1 |
| 296 | 1 | 2 |
| 32 | 1 | 12 |
| 32 | 1 | 23 |

The 496-row case is one interval with `execution_count = 1` collected 496 times — its work is counted 496x in the bucket's `total_cpu_ms` / `total_duration_ms`. The 1→23 cases additionally re-count the cumulative growth on every re-collection. Net effect: the Query Store time-slice charts overstate hot queries' totals by one to two orders of magnitude.

Fix shape (two tiers):

1. **Read-side, shippable now**: dedup before aggregating — latest row per `(query_id, plan_id, first_execution_time)` in the window (`first_execution_time` is collected and serves as the interval-identity proxy), via a window function, in every `v_query_store_stats` aggregate read. Lite and the Darling viewer/web both (the Darling web views' dedup-before-SUM convention is the pattern).
2. **Write-side, properly**: collect the interval identity (schema addition + storage version bump in both stores) so dedup keys on the real thing, and/or store per-interval deltas the way query_stats does.

Tier 1 corrects the numbers for all existing data; tier 2 makes the model honest going forward.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
[...]

PR #1919, opened 2026-07-31
Combine Query Store's flushed and in-memory interval slices at collection (#1907)

### LINKED ISSUE #1907: Query Store: the in-memory and flushed slices of one runtime_stats_id collide in the read-side dedup, understating execution counts non-deterministically
Found during the #1887 end-to-end smoke (live Azure SQL Database, GP_S_Gen5_1, shipped Darling service
from `origin/dev` @ `4adf3236`). **Not caused by #1872** — see "Why this is not the replica flip" below.
The #1887 run's own checks all passed; this is a separate, pre-existing defect that only a *sustained*
end-to-end run surfaces, which is exactly why the hand-composed SQL in #1848/#1872 never saw it.

## Symptom

`query_store_stats` execution counts are **understated, non-deterministically**, on both the viewer grid
read (`ViewerDataService.QueryStoreTopSql`) and the `query_store_stats_interval_hourly` CAGG that the
corrected/daily rollups are built on.

Live example — the CAGG's `last(execution_count, collection_time)` versus the truth, same store, same run:

```
          module_name          | query_id | plan_id | cagg_execution_count | true_total
-------------------------------+----------+---------+----------------------+-----------
 dbo.pm_get_orders_by_customer |       12 |       2 |                  136 |        144
 dbo.pm_order_summary          |       14 |       3 |                    8 |         94   <-- 91% low
 dbo.pm_recent_orders          |       15 |       4 |                   61 |         69
 Adhoc                         |       38 |       6 |                    2 |          6
```

Two of the five rows report a number that is not merely stale but wildly wrong (8 instead of 94). Which
rows are wrong changes from run to run — it is a tie-break, not a consistent bias.

## Root cause

`sys.query_store_runtime_stats` returns the **flushed slice and the still-in-memory slice of the same
`runtime_stats_id` as two separate rows**. Live from the Azure database (`flush_interval_seconds = 900`):

```
runtime_stats_id|plan_id|count_executions|first_execution_time       |last_execution_time
12              |2      |136             |2026-07-31 04:20:35.767    |2026-07-31 04:32:18.570   <- flushed, static
12              |2      |17              |2026-07-31 04:20:35.767    |2026-07-31 04:37:59.093   <- in memory, growing
14              |3      |86              |2026-07-31 04:20:35.837    |2026-07-31 04:32:18.580
14              |3      |17              |2026-07-31 04:20:35.837    |2026-07-31 04:37:59.100
15              |4      |61              |2026-07-31 04:20:35.853    |2026-07-31 04:32:18.603
15              |4      |17              |2026-07-31 04:20:35.853    |2026-07-31 04:37:59.117
```

```
total_rows|distinct_ids
12        |8
```

The collector's `BuildPayloadBody` selects straight from that view with no aggregation, so **both slices
are stored**. They then share every column of the read-side dedup key —
`(database_name, query_id, plan_id, runtime_stats_interval_id, first_execution_time, execution_type_desc, replica_role)`
— *and* the same `collection_time`, proven in the store:

```
 query_id | plan_id | runtime_stats_interval_id |  first_execution_time   | execution_type_desc | replica_role | rows_sharing_the_entire_dedup_key | execution_counts
----------+---------+---------------------------+-------------------------+---------------------+--------------+-----------------------------------+------------------
       12 |       2 |                         1 | 2026-07-31 04:20:35.767 | Regular             | Primary      |                                 2 | {8,136}
       14 |       3 |                         1 | 2026-07-31 04:20:35.837 | Regular             | Primary      |                                 2 | {8,86}
       15 |       4 |                         1 | 2026-07-31 04:20:35.853 | Regular             | Primary      |                                 2 | {8,61}
       38 |       6 |                         1 | 2026-07-31 04:29:04.077 | Regular             | Primary      |                                 2 | {2,4}
```

So both the viewer's `ROW_NUMBER() OVER (PARTITION BY ... ORDER BY collection_time DESC)` and the CAGG's
`last(execution_count, collection_time)` are ordering by a value that is **identical for both rows**. The
survivor is whichever the engine happens to emit first. When the in-memory sliver wins, the grid shows it.

The dedup itself is correct and load-bearing (#1841) — it is there to collapse *re-collections of the same
interval across cycles*. The bug is that it now also silently collapses *two additive slices of one
interval within a single cycle*, which is a different thing entirely.

## The slices are additive — confirmed three independent ways

1. **Arithmetic against the known workload.** The seed ran each procedure a known number of times
   (100/50/25 in the burst, then one of each per 20s round; ~43 rounds by the last cycle → ~143/93/68).
   The **sums** land on 144/94/69. `max()` (136/86/61) does not.

2. **The product's own `procedure_stats` collector**, reading the entirely separate
   `sys.dm_exec_procedure_stats`, in the **same store on the same collection cycle** (`04:35:03`, against
   the `query_store` cycle at `04:35:03.870265` whose slices are tabulated above):

```
      collection_time       |        object_name        | execution_count
----------------------------+---------------------------+-----------------
 2026-07-31 04:35:03.753426 | pm_get_orders_by_customer |             144
 2026-07-31 04:35:03.753426 | pm_order_summary          |              94
 2026-07-31 04:35:03.753426 | pm_recent_orders          |              69
```

   Exactly the slice sums (`{8,136}` → 144, `{8,86}` → 94, `{8,61}` → 69), and not `max()` (136/86/61).
   The alignment holds at the next observation too: at `04:37:59` the DMV slices were `136+17`, `86+17`,
   `61+17` and `procedure_stats` at `04:38:05` recorded `153 / 103 / 78`. This is in-product corroboration
   that the correct total is `SUM`, not `last`/`max`.

3. **The flushed row never absorbs the in-memory one.** Across observations at 04:35 and 04:38 the `136`
   row stayed at 136 with a static `last_execution_time`, while its twin grew 8 → 11 → 17 with an advancing
   one. A superseding row would have replaced it; an additive slice behaves exactly like this.

## Why this is not the replica flip (#1872)

- `sys.query_store_replicas` holds **one row per `replica_group_id`** (4 groups, 4 names), so the
  `LEFT JOIN ... ON qsr.replica_group_id = qsrs.replica_group_id` the flip added **cannot fan out**:

```
replica_group_id|name_rows
1|1
2|1
3|1
4|1
```

- Both colliding rows carry the **same** `replica_role` (`Primary`), so the collision is identical to what
  it would be with the pre-flip `NULL` placeholder. `replica_role` is in the dedup partition either way.
- The duplication is visible in the raw DMV **before any join** (12 rows / 8 distinct ids, above).

## Scope

- Mechanism is Query Store's flush behavior, not an Azure peculiarity — but **live evidence here is Azure
  SQL Database only**. Whether box SQL Server surfaces the same same-cycle duplicate pair is untested and
  should be checked before assuming either way.
- Affects `query_store_stats_interval_hourly` and everything derived from it
  (`query_store_stats_corrected_hourly`, the daily tiers) plus the viewer's Query Store grid. Related to but
  distinct from #1869, which is about an hour-straddle residual in the daily tier, not a same-cycle tie.
- Lite reads the same shape (`GetQueryStoreTopQueriesAsync`) against DuckDB and looks to have the same tie —
  **not verified live**, flagged for the parity check rather than asserted.

## Fix direction (not applied — this is a design call)

Aggregating the slices belongs on the **collect** side, where the identity is unambiguous: group the
payload by the natural runtime-stats key and `SUM` the counters (and take `MAX(last_execution_time)`),
so one interval yields one row per cycle and the existing cross-cycle dedup keeps working unchanged.
Doing it on the read side instead would mean teaching every consumer the difference between "same interval,
later cycle" (keep latest) and "same interval, same cycle" (add) — the CAGG's `last()` cannot express that.

I did not implement it: it changes the collector's emitted row shape and the semantics the #1841 dedup was
built around, which is a decision worth making deliberately rather than inside a validation run.

Found by: #1887 live Azure end-to-end smoke. Azure resources torn down.

erikdarlingdata:
Fixes #1907.

## The bug

`sys.query_store_runtime_stats` returns the **flushed** slice and the **still-in-memory** slice of one
`runtime_stats_interval_id` as two separate rows, and they are **ADDITIVE members of one interval**, not
competing snapshots of it. The collector selected straight from the view, so both were stored — and they
then shared every column of the read-side dedup key (#1841/#1845/#1853) **and** `collection_time`. So the
viewer's `ROW_NUMBER() ... ORDER BY collection_time DESC` and the rollups' `last(execution_count,
collection_time)` were both ordering by a value identical for both rows, and the survivor was whichever
the engine happened to emit first. Live: **8 executions reported where 94 was true**, with a different
subset wrong on each run.

The dedup itself is correct and stays — it exists to collapse *re-collections of one interval across
cycles*. It just cannot also be asked to *add two slices within one cycle*, and no read-side rule can
express both. A TimescaleDB continuous aggregate cannot express it at all.

**It is not an Azure peculiarity.** #1907's evidence was Azure SQL Database only and flagged box SQL
Server as untested. It reproduces there — SQL Server 2022 (16.0.4255.1), `DATA_FLUSH_INTERVAL_SECONDS =
60`, 100 executions then a forced flush then 25 more:

```
rsi | plan_id | module               | count_executions
  1 |       3 | dbo.pm_order_summary |              100   <- flushed, static
  1 |       3 | dbo.pm_order_summary |               25   <- in memory, growing
  1 |       4 | dbo.pm_recent_orders |               60
  1 |       4 | dbo.pm_recent_orders |               10
```

`sys.dm_exec_procedure_stats`, a wholly separate source read at the same instant, said **125 and 70**.
SUM matches; the larger slice alone does not. Both slices carry an identical `first_execution_time`, so
the tier-1 proxy key had exactly the same exposure as the real interval id — which is why pre-tier-2 rows
are affected too.

With Query Store's default 900s flush against its default 3600s interval, one interval can hold several
flushed slices, so the split is not bounded at two.

## The fix

`BuildPayloadBody` — the ONE body both the on-prem and Azure execution shapes run (#1844's single-body
guarantee, drift guard still green) — groups on exactly the natural key of the view and combines:

| Column family | Combined by |
|---|---|
| `count_executions` | `SUM` |
| `avg_*` (11 of them) | count-**weighted** mean, `SUM(avg * count) / NULLIF(SUM(count), 0)` |
| `min_*` / `max_*` | `MIN` / `MAX` (includes `min_dop`/`max_dop`, which have no avg) |
| `first_execution_time` / `last_execution_time` | `MIN` / `MAX` — the interval's own span |

The weighted mean is the whole game: Query Store stores an average and a count but **never a total**, so
`avg * count` is what recovers a slice's total. A plain `AVG()` of the slice averages would weight a
25-execution sliver the same as a 100-execution flush. Verified live — slices of (1778.42 over 100) and
(2245.60 over 25) combine to **1871.856**, which is `(1778.42*100 + 2245.60*25) / 125`, and is not the
2012.01 an unweighted average gives.

Every `avg_*` column goes through one `WeightedAverage` helper rather than being written out by hand,
because the wrong form is not a compile error and does not look wrong. A test discovers the `avg_` columns
**from the emitted SQL** and requires the weighted shape on each, so a newly added one is covered the
moment it appears.

### The cutoff had to move from WHERE to HAVING

This is load-bearing, not tidiness. The flushed slice is **static**, so once the growing in-memory slice
pushes the watermark past its `last_execution_time` the flushed slice stops qualifying — and a sum over
the survivors is the sliver alone, the original defect with an aggregate bolted on top. `HAVING
MAX(last_execution_time) > @cutoff_time` asks the question at interval grain: did this interval see new
activity, and if so give me all of it. It is strictly more permissive than the predicate it replaces, so
nothing that used to be collected stops being.

### It is faster than what it replaces

Measured on a real 212,000-row Query Store (SQL 2025), full 55-column payload, warm, three runs:

| Shape | Elapsed | Rows |
|---|---|---|
| Pre-fix | 453 / 485 / 516 ms | 510 |
| **Post-fix** | **375 / 422 / 438 ms** | **262** |
| Post-fix without the interval pre-filter | 1203 / 1203 / 1235 ms | 263 |

Half the rows means half the `nvarchar(max)` query text and plan XML to materialize and ship, which more
than pays for the aggregate. The `IN (...)` pre-filter is a **prune, not a semantic** — its interval list
is by construction a superset of what the `HAVING` keeps, so it can never subtract a row — and the third
row is why it is there.

### Nothing about the stored row moved

Same 55 columns in the same order, same positional writers, `GoldenCollectorSchema` untouched, no
migration and no storage-version bump. Only the number of rows per interval (now at most one per interval
per collection). The `TOP` backstop now caps **intervals** rather than slices, and sits outside the
aggregate for cause: a cap falling mid-interval would emit a partial sum, which is worse than omitting the
interval.

`replica_group_id` joins the grouping key under the same 2022+/Azure gate the attribution column carries,
for the same bind-safety reason — naming a column that does not exist in a `GROUP BY` fails the whole
SELECT, and on Azure's per-database path that means every database. It has to be in the key where it does
exist: two replicas' rows for one interval are different work, and grouping without it would sum a
secondary's executions into the primary's — the exact bug replica attribution was added to prevent.

## Deprecated Dashboard: same defect, same fix

`install/09_collect_query_store.sql` reads the same view the same way and had the same bug. Same
aggregation, same `WHERE` to `HAVING` move, same gated `replica_group_id`. It is a proc body change, so an
upgrade re-applies it with no schema step and no `upgrades/` entry. Verified live against SQL 2022 by
executing the proc's own generated SQL in all three version shapes — it returns 125 and 70 where the raw
view holds the split slices.

## Legacy rows: made deterministic, and the residual is filed

Rows already collected cannot be rewritten. All **19** read-side dedup sites across both apps now order by
`collection_time DESC, execution_count DESC`, so a pre-fix tie resolves to the **flushed** slice — the one
holding the bulk of the interval's work — deterministically instead of flapping. On new rows the clause
can never fire, because there is one row per partition per collection.

That is **closest-available, not correct**. The correct value is the sum, and the materialized CAGGs
cannot be tie-broken at all (`last()` has no tie-break, and the tied rows are gone once materialized).
Filed as **#1912**, including the part that does not age out: the indefinitely-kept daily tier keeps
understated counts for the pre-fix period.

## Testing

**Watched red per mutation**, each reverted after:

| Mutation | Test that went red |
|---|---|
| weighted mean to `AVG(qsrs.avg_x)` | `Payload_EveryAverageColumn_IsTheCountWeightedMean` |
| `replica_group_id` ungated in the GROUP BY | `BuildPerItemQuery_ReplicaGroupIdEntersTheGroupingKey_OnlyWhereItBinds` |
| `HAVING` back to a per-slice `WHERE` | 3 tests, including both paths' cutoff pins |
| tie-break dropped from one Lite read | `TiedSlicesOfOneInterval_...` (**returned the 25 sliver**) + the source guard |
| tie-break dropped from one Darling read | the Darling source guard, naming the file |
| post-fix interval reseeded as split slices | the live-PG rollup test (**125 became 100**) |

The behavioural DuckDB test seeds **both insertion orders** as two independent queries, so it cannot pass
by accident on an engine that happens to favour the earlier row — without the tie-break one arm returns
the sliver.

**Live SQL Server 2022** — the emitted collector SQL (not a hand-composed approximation) executed in all
three version shapes; all 55 columns bind on the 2016 floor, on 2017, and on 2022. Against the repro
database it returns **125 and 70** where the raw view holds `{100, 25}` and `{60, 10}`, matching
`dm_exec_procedure_stats` exactly.

**Live PostgreSQL 18 + TimescaleDB 2.28.1** — a new `QueryStoreCorrectedRollupLiveTests` case asserts both
halves in one store on one refresh: the corrected rollups report the hand-computed **155** for an hour of
two combined intervals plus the exact execution-weighted mean off the same rows, while the same store fed
the pre-fix split slices returns a single slice and **can never reach the 125 they add up to**. Gating
verified honestly — it SKIPS without `DARLING_TEST_PG` and PASSES with it.

**Suites**: Lite **1854 passed / 0 failed**; Darling **3997 passed / 0 failed** (10 skipped = gated tests
needing a live SQL Server or the bundled runtime). Full-solution `-t:Rebuild`: **0 warnings, 0 errors**.
Installer.Tests deliberately not run.

## Deferred, filed before this PR opened

- **#1912** — the pre-#1907 rows and the CAGG buckets materialized from them. Three options written up,
  with the reason none of them repairs the indefinitely-kept daily tier.


erikdarlingdata:
GUARD WIDENED — pushed `a66e785f`.

`Lite.Tests/QueryStoreSliceTieBreakSourceTests` now mirrors Darling's counterpart: an enumerated `DedupSites` list covering all **3 files / 7 sites** (`LocalDataService.QueryStore.cs` 5, `Analysis/DrillDownCollector.Queries.cs` 1, `Analysis/DuckDbFactCollector.QueryPerf.cs` 1). Enumerated rather than globbed, so MOVING a read has to be re-declared instead of silently shrinking coverage; a missing tie-break fails naming the file. A third test reads Darling's guard and asserts its declared total of 12, so neither app can quietly fall behind the other.

The overclaiming comment is fixed in **both** apps — each now names the actual test and its real count instead of "a source-containment test pins all of them".

**Watched red**, and it demonstrates the gap was real: dropping the tie-break from `DuckDbFactCollector.QueryPerf.cs` leaves the OLD guard (`EveryQueryStoreAggregateInTheFile_CarriesADedupCte`) **green**, while the new guard fails with `Lite\Analysis\DuckDbFactCollector.QueryPerf.cs @ char 9829`. Restored, verified byte-identical to the committed file.

Lite 1903 passed / 0 failed; full-solution build 0 warnings / 0 errors.

Issue #1915, opened 2026-07-31
Retention: the CHUNK-GEOMETRY half of the coverage invariant is unguarded — ordering is now walked, but 'a consumer never retains less than its source' rests on a measured TimescaleDB default


erikdarlingdata:
Spun out of #1905, which derived the ORDERING half of this invariant from the policy list and documented the other half as not assertable there.

## The invariant has two halves

A healthy store must never read as coverage-SHORT, because since #1877 a positively measured shortfall re-holds a policy that is already armed — on a healthy store that is a purge stopped for no reason, and because holding lets the source grow deeper it never self-releases.

`min(bucket)` of the consumer has to stay at or before `min(time)` of the source. Two independent things make that true:

1. **Ordering** — every consumer's retention horizon strictly exceeds its source's. #1905 made this a walk over `TimescaleSupport.RetentionPolicies`, so it now covers policies nobody has written yet.
2. **Chunk geometry** — `drop_chunks` only removes chunks whose whole range is past the cutoff, so actual retention always runs LONGER than nominal, by up to one chunk interval. If a consumer's chunks were FINER than its source's, the consumer could round down to nearly its nominal horizon while the source rounded up well past its own, and the source could end up holding data older than anything the consumer kept — with both horizons correctly ordered.

Half 2 has no test.

## Why it holds today, and why that is not the same as being guarded

Measured on PostgreSQL 18.4 / TimescaleDB 2.28.1 during #1877: materialization hypertables come out at a **10-day** chunk interval, against the raw tables' explicit **1 day**. So every consumer's chunks are coarser than or equal to its source's and the rounding always runs the safe way, with days of margin at the tightest pair (raw 4d/1d-chunks against the interval layer's 7d/10d-chunks).

Two things could move it without any test noticing:

- **A TimescaleDB default change.** The 10 days is a default nobody in this codebase sets. Note it did NOT match the widely-repeated "10x the bucket width" rule — the 1-hour and 1-day bucket CAGGs both came out at 10 days — so the actual rule is not something to reason about from memory, only to measure.
- **Anyone calling `set_chunk_time_interval` on a CAGG,** which is a reasonable thing to want for compression or query-pruning reasons and has no obvious connection to retention safety.

Either one turns a healthy store into one whose purge stops.

## What a test looks like

[...]

Issue #1877, opened 2026-07-31
Retention gate is arm-only, so a policy whose coverage list GROWS is never re-held — a store upgrading into a new consumer keeps purging its source


erikdarlingdata:
Found while building #1869, which is the first change to add a consumer to a retention gate that stores already have ARMED.

## The gap

`EnsureRetentionPoliciesAsync` is arm-only. It creates a policy paused, evaluates `IsSafeToArmRetentionAsync`, and arms if the verdict is safe. When the verdict is UNSAFE it logs `HELD PAUSED` and moves on — but `add_retention_policy(if_not_exists => true)` returned `-1` for the already-existing policy, so nothing paused it, and **an already-armed policy stays armed**.

That is exactly right for the case the gate was built for (#1680/#1759: a policy this store already armed must not be re-paused on every restart, or a restart would undo an operator's backfill). It is wrong for a case that did not exist until now: a policy whose COVERAGE LIST GREW.

## Concretely, on a store already running #1870

`query_store_stats_interval_hourly`'s purge is armed and dropping at 7 days. #1869 adds `query_store_stats_interval_daily` as its third consumer, born empty. On the next start the gate correctly judges the coverage insufficient and logs:

> Retention policy for query_store_stats_interval_hourly created but HELD PAUSED - ... does not yet cover everything it holds

...and the job keeps running anyway.

## What it costs, and what it does not

**No data is destroyed and no read is wrong.** The interval layer keeps only 7 days by design, its consumers' refresh policies work a 3-day window well inside that, and every window older than the new daily's floor routes to `query_store_stats_corrected_daily`, which holds it (at #1869's documented ~2x). Steady state is completely unaffected.

**What it costs is REACHABLE DEPTH.** On such a store the new day-grain daily can only ever be backfilled as deep as the interval layer still holds — 7 days — rather than as deep as the store's history goes, because the deeper interval-grain rows were already purged and raw is at 4 days. An operator who runs `--backfill-rollups` promptly after upgrading loses nothing; one who waits a month permanently caps the new tier's depth at 7 days.

## Why it was NOT fixed in #1869

[...]

Issue #1564, opened 2026-07-18
Gated live-pg suite: two purge tests count globally in the shared store — order-dependent flakes (TimescaleSupport chunk-drop, DarlingRetention row-delete)


erikdarlingdata:
Nightly build-4 dispatch (dev @ 62ca1f50) failed the darling-pg job on two purge assertions that passed IDENTICAL code in #1560's branch dispatch an hour earlier:

- `TimescaleSupportTests.EndToEnd_DetectConvertAndDropChunksPurge_AgainstDevPostgres` — "expected the two old chunks (wait_stats + collection_log) to drop, got 1" (:239)
- `DarlingRetentionTests.EndToEnd_PurgeDeletesExpiredRowsAndKeepsFreshOnes_AgainstDevPostgres` — "expected the purge to delete at least the three expired test rows, got 2" (:325)

Both classes are members of the serialized `[Collection("live-postgres")]` (55 classes), so this is not concurrency — it is **order-varying data residue**: another member's planted rows land inside the time windows these tests count GLOBALLY (chunk counts across the whole hypertable; delete counts across the whole table), and xunit's class order within a collection varies run to run.

**Fix direction:**
- Chunk-drop assertions: plant into a dedicated, collision-proof time window (e.g. rows at now-3y, drop older-than-2y) that no other test writes into — chunk counting cannot be scoped by server_id, so the window must be reserved by convention (document it in the test).
- Row-delete assertions: count the test's OWN rows by its planted server_id before/after, never the purge's global return value.
- Sweep the other purge/retention assertions in the collection for the same global-count pattern.

Non-blocking: build-4 artifacts published fine (the build job is independent); the standard suites are green on dev. This is test-isolation debt in the gated suite — same family as the ScratchPostgres fix in #1551.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
[...]

PR #1928, opened 2026-07-31
Retention geometry: judge on the chunks on disk, not just the configured interval (#1925)

### LINKED ISSUE #1925: Retention geometry guard reads the CONFIGURED chunk interval, not the widths of chunks already on disk — a narrowing change would leave older, wider chunks it cannot see
Spun out of #1915, which guards the chunk-geometry half of the retention coverage invariant and deliberately stops at configuration.

## What the guard checks

For every (source, consumer) pair with a horizon on both sides, #1915 reads `timescaledb_information.dimensions.time_interval` — the interval the hypertable is CONFIGURED with — and requires either identical grids or that the consumer's horizon reach the source's horizon plus one of the source's chunks.

That is the right reading for the two scenarios it was filed about: a TimescaleDB default moving, and someone calling `set_chunk_time_interval` on an aggregate. Both change the configuration, and the guard sees them.

## What it cannot see

`set_chunk_time_interval` — and a change to `TimescaleSupport.ChunkIntervalDays`, which is the same thing by another route — affects only chunks created AFTER it. Existing chunks keep the width they were created with. So a store that has lived through a change carries a mix, and the guard judges only the newest width.

Direction matters, and only one of the two is dangerous:

- **Widening** (say raw 1 day to 7). New chunks round further past the horizon than old ones. The guard reads the wider, worse number and judges against it, so it is correct-and-conservative — verified while building #1915, where moving `ChunkIntervalDays` to 7 turned the guard red exactly as it should.
- **Narrowing** (7 back to 1). The guard reads the NEW, smaller interval and computes a smaller required margin, while chunks written under the old width are still on disk needing the larger one. That is a false green, and it lasts until the last wide chunk ages out.

## Why it is not urgent

Nothing narrows today. Darling has always created hypertables at `ChunkIntervalDays = 1` and calls `set_chunk_time_interval` nowhere; #1915's own test is the only caller in the repo and it does so against a scratch database to prove the guard has teeth. So the mixed-width state is currently unreachable — this is about the release that first changes that constant downward, which would be green in CI.

It is also worth being explicit that this whole guard is a CI-time check on OUR schema, not a runtime check on a customer's store. An operator who re-chunks an aggregate on a live store gets no signal from it. That is a deliberate scope line rather than an oversight — but it does mean the on-disk state is never inspected anywhere, which is what makes this residual the only place it could be.

## What a fix looks like

Read the actual per-chunk ranges (`timescaledb_information.chunks`, `range_start`/`range_end`) for the relations under judgement and take the WIDEST surviving chunk as the source's effective rounding, rather than the configured interval. That needs a store with chunks spanning enough time for the widths to differ, so it is a heavier fixture than #1915's — which builds the aggregates and reads catalog metadata without seeding a single row.

erikdarlingdata:
Closes #1925.

## What you asked me to do first

You said to measure rather than take the prescription, so here is what the store actually does, and the design follows from it rather than from the issue text.

**`set_chunk_time_interval` leaves existing chunks alone.** A hypertable created at 1 day, widened to 7, then given more data reports `7 days` as its configured interval while its six original 1-day chunks sit underneath it, unchanged. Both numbers are real and neither describes the store on its own.

**Widening is safe to read from configuration; narrowing is not.** After a widening the configured interval is the wider, worse number, so a configuration-only guard is conservative by accident. After a NARROWING the configuration returns to a small value while wide chunks remain — and that is the false green, because the rounding those wide chunks cause is invisible to it.

**Compression does not distort any of this.** `timescaledb_information.chunks` reports one row per chunk with its range intact whether or not the chunk is compressed — verified by compressing half a hypertable and re-reading. Darling compresses every hypertable, so this was load-bearing and worth retiring by measurement rather than assumption.

**Chunk starts are epoch-aligned to their own width.** This is the entire basis of the identical-grid regime #1915 shipped, and it was assumed there. It is now asserted by the test, against chunks at two different widths.

## The shape, and why it is this one

Your suggestion — read actual ranges and let the worst chunk carry the judgement per regime — is what the measurement supports, with one refinement that matters:

**Effective width = max(configured, widest on-disk chunk).** Not the widest on-disk width alone. Old chunks age toward the retention cutoff and eventually sit on it, and every FUTURE chunk is cut at the configured width, so both are candidates for the chunk that straddles the cutoff and the worst of them bounds the rounding. Dropping the configured half would go blind the other way — a store whose chunks are all narrow today but whose configuration has been widened is about to start cutting wide ones.

A useful consequence: a relation with no chunks yet has nothing on disk to consider, so it reports exactly its configured interval and is judged exactly as #1915 judged it. **This strictly extends that guard rather than replacing it**, and the fresh-store path — which is what CI exercises for the real schema — is unchanged.

**A relation carrying more than one width is refused the identical-grid argument outright.** Something sitting on two grids cannot support a superset argument whatever its widest chunk happens to equal.

## Watched-red

Both mutations land on the **judgement**, not on scaffolding — the first ordering of the new leg put the descriptive asserts first and the red landed on one of those, so the property is now asserted before the diagnosis that explains it.

| mutation | result |
|---|---|
| `Effective` reads the configured interval alone (i.e. the #1915 behaviour) | **RED** on the judgement — the mixed store passes, which is the false green this closes, demonstrated rather than argued |
| `Effective` takes the NARROWEST on-disk width | **RED** on the judgement |

**The second mutation initially PASSED, and that was a real hole in my fixture.** The scenario only ever cut 7-day chunks, so `Min` and `Max` were the same number and it could not tell "widest" from "narrowest" — a guard that took the narrowest, which is exactly backwards, would have looked correct. The fixture now seeds at the narrow width too, so the store is genuinely mixed (1-day AND 7-day chunks under a 1-day configuration), and the mutation goes red. This is the whole argument for watched-red as a practice: the test was green and wrong.

## One thing I did not oversell

The single-grid condition on `DistinctWidths` has **no mutation that turns it red**, and the code says so in as many words. In today's schema a mixed-width relation is always a raw table whose consumers are aggregates on a different interval, so the arithmetic refuses those pairs before the condition is consulted. It is kept because dropping it would be unsound the moment those widths coincided, but it is a deliberate conservative refusal rather than a proven-necessary one, and I would rather that be written down than implied by a test that does not exist.

## Tests

Darling **4023 passed / 0 failed / 10 skipped** against a live PostgreSQL 18.4 / TimescaleDB 2.28.1 rig. Lite **1895 passed / 0 failed**. `-t:Rebuild` clean on both, 0 warnings. No Installer.Tests. Cluster stopped after the run.

Port discipline per your note: enumerated listening ports before picking (5619/5787/5877/5985 were taken, took 6077), and verified the cluster answering on it was mine by reading `data_directory` off the server and matching it to my own datadir — `pg_ctl status` plus that, rather than a `createdb` exit code, which is precisely the check that would have caught the 6011 incident.


erikdarlingdata:
## CI turned this red on a test I did not set out to touch, and the fix is in this PR

First CI run failed on `QueryStoreCorrectedRollupLiveTests.RetentionSweep_TreatsAFailedProbeAsUnknown_NeverAsACoverageRegression` — one of the tests I added in #1877 — with:

```
Npgsql.PostgresException : 55P03: could not refresh continuous aggregate "query_store_stats_interval_hourly" due to a concurrent refresh
```

**It is a real race, not a flake to re-run past, and this change plausibly made it likelier.** `EnsureAggregatesWithoutRefreshPoliciesAsync` strips every refresh policy — but it can only do so AFTER `EnsureContinuousAggregatesAsync` has created them, and TimescaleDB runs a new policy's first check IMMEDIATELY rather than on its next interval (#1564/#1567). That leaves a window the length of two statements in which the background scheduler is already materializing the view the test is about to refresh by hand. The two collide as 55P03.

The window has always been there; what changed is pressure. This PR adds a live class that also builds the full aggregate set, and background workers are a CLUSTER-wide resource, so more concurrent aggregate creation on one cluster makes the collision likelier. Calling that someone else's flake and re-running would have been wrong on both counts.

**Fixed by retrying on 55P03, bounded.** A refresh is idempotent and the losing side is merely early, so retrying is the correct response rather than a workaround — the alternative is asserting that the first attempt wins, which the engine never promised. Bounded at 12 attempts 250ms apart, so a genuine stall still fails the test rather than hanging it.

**Both call sites, not just the one that failed.** `RefreshAllAsync` in the same class had the identical raw `CALL refresh_continuous_aggregate` and runs immediately after the aggregates are built, which is exactly when the scheduler is busiest — it was the same bug waiting for a different day. Both now go through one retry helper. I checked the other four files that mention `refresh_continuous_aggregate`: three only reference it in comments or string assertions, and `PayloadDimensionLiveTests` goes through the product's SQL builder with its own concurrency handling, so this class was the only live raw-call site.

Full Darling suite green against the live rig after the fix (4025 passed / 0 failed / 10 skipped).

erikdarlingdata:
## NIT FIXED

Both inline comments had the same root, so they take one fix.

`SetChunkIntervalAsync` resolved a continuous aggregate's materialization hypertable with server-side `format('%I.%I', ...)`, but built the raw-table fallback as a C# interpolated `collect.{relation}` — and then interpolated the result into the `set_chunk_time_interval` SQL text. Inert today, since every name it is called with is one of the product's own lowercase catalog constants, but it is two quoting rules where one will do and the next editor inherits the weaker one.

Fixed in both places rather than only the one that was flagged:

- **The fallback is now a `COALESCE` inside the same query**, so both arms quote through `format('%I.%I')`. There is one quoting rule in this method instead of two that agree only by luck of the catalog being lowercase.
- **The resulting identifier is BOUND as a `$1::regclass` parameter** rather than interpolated into the statement, so it never reaches the SQL text at all. That closes the interpolation the nit pointed at rather than just making the string it carried safer.

Both arms stay covered by the existing test without adding anything: the teeth leg re-chunks an aggregate (the catalog path) and the mixed-width leg re-chunks the raw `query_store_stats` table (the fallback path), so a regression in either shows up as a red test rather than as unexercised code.

Full Darling suite green against the live rig after the change — 4025 passed / 0 failed / 10 skipped, `-t:Rebuild` clean with 0 warnings. Pushed as `c0b0b3b4`; `origin/dev` was already an ancestor, so no merge was needed and the branch is ready for the next rollup. No other changes.

Issue #1849, opened 2026-07-30
Query Store hourly/daily CAGGs still bake in the pre-dedup inflation (#1841 tier 3)


erikdarlingdata:
Spun out of #1841 tier 2 (PR pending) after building and **live-probing** the corrected-rollup shapes against PostgreSQL 18.4 + TimescaleDB 2.28.1. Tier 2 collected the real interval identity (`runtime_stats_interval_id` + `interval_start_time_utc`, Lite v49 / Darling V41) and every RAW-tier read now dedups on it. The CAGGs are the one surface tier 2 did **not** repair, and this issue carries the design so it does not have to be re-derived.

## What is still wrong

`collect.query_store_stats_hourly` materializes `sum(execution_count)` and both weighted sums from **un-deduped** raw rows, and `query_store_stats_daily` inherits them hierarchically. Query Store rows are cumulative per-interval snapshots that the collector re-fetches every cycle, so one interval's work is baked in once per collection — measured on a live store at up to **496x** for a single interval inside one hour bucket.

No read-side edit can undo it: the duplicates are gone once materialized, and the CAGG output carries no interval identity to key on. Consequence today, unchanged by tier 2: a Custom Views panel whose window stays inside the raw tier (<= 3 days) reads corrected numbers; a window past it reads inflated ones, with a **visible step at the routing boundary**.

## Live probe results (TimescaleDB 2.28.1, PG 18.4)

Five CAGG shapes, created against a hypertable with `query_store_stats`' exact column set:

| Probe | Shape | Result |
|---|---|---|
| 1 | `time_bucket('1 hour', interval_start_time_utc)` | **REJECTED** — `ERROR: time bucket function must reference the primary hypertable dimension column` |
| 2 | `last(execution_count, collection_time)` grouped by full interval identity, bucketed on `collection_time` | **ACCEPTED** |
| 3 | `row_number() OVER (...)` inside the CAGG | **REJECTED** — `ERROR: invalid continuous aggregate query` / `HINT: Enable experimental window function support by setting timescaledb.enable_cagg_window_functions` |
| 4 | Hierarchical CAGG on probe 2, re-dedup with `last(x, bucket)` over the interval identity | **ACCEPTED** |
| 5 | Hierarchical CAGG on probe 2, collapsing interval grain to the composer dims with `sum(...)` | **ACCEPTED** |

[...]

Issue #1839, opened 2026-07-30
[FEATURE] Blocking alert: threshold on total blocked wait time (level-triggered), not just event count


sdinu13:
### Which component(s) does this affect?

- [ ] Lite
- [x] Darling
- [ ] Documentation

### Problem Statement

Hi Erik,

Billions of thanks for your work!

Written with ai but checked by brilliant brains:

We've been going through AlertEngine.cs / RollingCountAlertGate.cs to understand how the Blocking and Deadlocks alerts behave, and we'd like to propose a change specifically for Blocking Detected — we think the current mechanism is well suited for Deadlocks, but not for blocking.

Current behavior (both Blocking and Deadlocks)

Both checks share the same edge-trigger gate (RollingCountAlertGate) over a 1-hour rolling window: an alert fires only when the current event count climbs above a watermark (i.e., a genuinely new event), and the watermark resets to 0 once the count drops back below the threshold. This makes a lot of sense for deadlocks — a deadlock is a discrete, always-significant event, and without the watermark a single deadlock lingering in the 1-hour window would re-fire an alert every cooldown cycle for the whole hour.

Why it doesn't fit Blocking

Blocking is a different kind of signal. In production environments, short blocks (10-15 seconds) happen routinely and are rarely worth an alert if they resolve on their own — but the count-based threshold doesn't distinguish a handful of trivial, short-lived blocks from one (or several) genuinely long, damaging block. Right now the check thresholds on how many blocked-process reports/rows are in the window, not on how long anything was actually blocked — so we either get noise from routine short blocks, or we set the count threshold high enough to be quiet and then miss the case that actually matters: one session blocked for several minutes.

When monitoring a large number of instances, this distinction matters a lot — we want to be paged when something is actually bad, not every time transient contention shows up.

Thanks!
Romanian dbas.

### Proposed Solution

Proposed change

For the Blocking check, replace (or add as an alternative mode) a threshold on the sum of wait time across the currently blocked sessions, in seconds, instead of (or in addition to) the event count. For example: alert when the total blocking wait time across all currently blocked sessions exceeds 600 seconds — regardless of whether that's one session blocked for 600s or several summing to 600s.

[...]

Issue #1832, opened 2026-07-30
[BUG] Webhook and some other settings, alerts and history are lost after 3.2.0.->3.3.0 update


ZedZipDev:
### Component

Lite

### Performance Monitor Version

3.3.0

### SQL Server Version

Azure SQL db

### Windows Version

Windows 11

### Describe the Bug

I have installed PMLite 3.2.0 and it worked, then detected the version 3.3.0 appeared. I downloaded the Setup and tun it, it wrote the v.3.2.0 already installed and suggested some options, I have selected "Update".
It was installed ok, I go to Settings and see webhook is lost. I.e. no my webhook settings and all by default.

### Steps to Reproduce

Install 3.2.0
Then run 3.3.0 Setup.
The webhook settings is lost.

### Expected Behavior

The webhook should be kept.

### Actual Behavior

The webhook is lost after update to the v3.3.0.

### Error Messages / Log Output

```text

```

ZedZipDev:
Also, the "Last 24 hours" -> "Last 4 Hours"  after the update.
and no Alerts History. Queries, Deadlocks after the update. 
Tabula rasa...

erikdarlingdata:
Confirmed, and this is the most serious bug report in a while - thank you for it. Root cause is found. Short version: the 3.3.0 Setup.exe you downloaded installs into the same folder Lite keeps its data in, and re-running Setup over an existing install clears that folder first. Your settings file and the local history database were deleted by the installer. The "Last 4 Hours" detail in your second comment confirmed it: that value is the hardcoded default that only appears when the settings file is missing entirely.

Two things before anything else, in this order:

**1. Your webhook URLs are probably still recoverable - but do not press Save in Lite's Settings window until you have copied them out.** The URLs are not stored in the settings file; they live in Windows Credential Manager, which the installer never touches. But with the settings file gone, Lite does not load them, shows you empty boxes, and pressing Save writes those empty boxes back - which deletes the stored credentials. So first: Control Panel > Credential Manager > Windows Credentials, look for entries named `PerformanceMonitorLite_TeamsWebhook`, `_SlackWebhook`, `_GenericWebhook`, and copy the URLs somewhere safe. Then you can re-enter them in Lite normally.

**2. There is a chance your entire pre-update history survived under a renamed folder.** The installer renames the old folder aside before deleting it, and if any file was still locked the delete fails silently. Check with:

```
[...]

PR #1931, opened 2026-07-31
Repair the Query Store split slices stored before #1907, and disclose what cannot be reached (#1912)

### LINKED ISSUE #1912: Query Store rows collected before #1907 keep their split slices: read-side dedup is now deterministic but still under-counts, and the materialized CAGGs cannot be tie-broken at all

erikdarlingdata:
Fixes #1912.

#1907 fixed Query Store's split slices at collection and made the read deterministic. It could not touch
what was already stored. This repairs that — fully in Lite, and as far as raw reaches in Darling — and
discloses, plainly, the part that is permanent.

## The defect, in the stored rows

`sys.query_store_runtime_stats` returns the flushed and the still-in-memory slice of one
`runtime_stats_interval_id` as two **additive** rows. Builds before #1907 stored both, so every read that
keeps one row per interval reports a fraction of that interval's work, and every rollup materialized from
those rows carries that fraction. On the live evidence: **8 executions where 94 was true**.

The pre-fix signature is exact — two or more rows sharing the **entire** dedup key *and* `collection_time`.
No row collected since #1907 can match it, because the collector now emits at most one row per interval per
cycle. That single fact is what makes the repair idempotent, safe to re-run, and incapable of touching
correctly-collected data.

## Darling: `--collapse-legacy-slices`

An operator verb, mirroring `--backfill-rollups` (#1849): same Windows/managed-credential guard, same
`--dry-run`, same compiler-required disclosure sink (#1797). It surveys, collapses, then re-materializes the
rollups the collapsed rows fed.

Column handling is **derived** from `QueryStoreCollector.PayloadColumns` and each column's declared type
rather than restated — `execution_count` sums, every `avg_` takes the count-weighted mean, `min_`/`max_` take
the extreme. That derivation paid for itself on the first live run: PostgreSQL has no `min(boolean)`, and
because the classifier reads the declared type, `bool_or` was a one-line fix rather than a hunt.

### The refresh clamp is a safety property

Measured on PostgreSQL 18.4 + TimescaleDB 2.28.1, three fresh CAGGs in identical state, one variable each:

| Refresh | Range | Pre-raw buckets surviving |
|---|---|---|
| `force => TRUE` | dropped-range only | **0 of 144 — destroyed** |
| `force => TRUE` | spanning dropped + live | 144 — preserved |
| plain, no force | dropped-range only | **0 — destroyed** |

A refresh whose range lies entirely within dropped raw chunks **destroys** the materialization there, with
`force` and without. My first probe said history survived; that was an artifact of the range spanning into
live data, which is why it took a single-variable design to see. Aiming at a nominal "pre-fix period" would
therefore have blanked the 21-day hourly and the indefinitely-kept daily below raw's floor — the one thing
#1759/#1793 forbid. The verb derives its window from the rows it actually collapsed, which are inside raw's
extent by definition, so safety falls out of the data flow instead of a bolted-on check. **Pinned as a live
test**, so a future TimescaleDB that makes it safe fails loudly rather than quietly widening the blast radius.

The shipped `--backfill-rollups` was checked for the same exposure and is **safe by construction** — it
converges down to `min(collection_time)` of raw and never refreshes below it. No bug to file.

Also measured: DML against **compressed** chunks works on 2.28.1 and leaves them compressed (DELETE 600 rows
6.0 ms, INSERT 6.4 ms, UPDATE 0.6 ms), so the collapse reaches the older half of the raw window without a
decompress cycle.

## Lite: automatic, on the first launch after upgrading

Per the lead's call, and the parity argument is that consistency means the same **semantics** for the same
defect — same collapse, same weighted math, same disclosure — not the same invocation surface. Each app
follows its own precedent: Darling's is #1849's verb, Lite's is the automatic startup store migration that
v39 and #1832's data-root move established. A button would leave the users least equipped to find it holding
wrong numbers permanently.

**Lite gets a FULL repair where Darling gets a bounded one**, because Lite's long tier IS the parquet
archive — the same rows, rewritable — rather than a materialized rollup that cannot be rebuilt from raw that
no longer exists. Measured against a real archive: ~46 MB across three monthly files, **6.6–8.5% of rows in
split groups** with exactly 2 rows per group every time, the largest file (28 MB / 533,109 rows) rewriting in
under a second. A real group verified: (4 execs @ 161093µs) + (10 @ 203358µs) → 14 @ 191282µs, the
count-weighted mean exactly.

### Migration discipline (#1832 / v39 pattern)

Each file is rewritten beside the original and promoted only after **two conservation invariants** — row
count landing on exactly `rows - rowsRemoved`, and `SUM(execution_count)` **unchanged**, which is the
invariant the collapse's own arithmetic guarantees and the one that would catch a bad aggregate that still
produced the right row count. No backup copies: verify-before-promote is the safety.

Failures are isolated **per file**, in the rewrite and in the survey. Both of those came from tests catching
real gaps rather than foresight: the mid-rewrite failure test showed the first bad file aborted the whole run
— and since files process in name order, one bad old file could permanently block every newer one — and a
second run showed the same hole in the survey. The marker is written only on a fully successful pass, so a
partial repair retries next launch.

**The marker is its own table, not a schema-version bump**, for two load-bearing reasons: `RunMigrationsAsync`
drops and recreates tables per version step, so a data repair has no business there; and withholding the
schema version to force a retry would re-run those drops.

### The DuckDB finding that would have shipped silently

**An in-place parquet rewrite poisons DuckDB's external file cache.** It keys on **path** at **instance**
scope, so replacing the bytes at a path already read makes every later read fail with `No magic bytes found
at end of file` — on new connections too — until the process restarts. That would have left Lite unable to
read the archive it had just repaired.

The instance scope is the trap: my first isolation used separate in-memory instances, which made it look
per-connection and harmless. `PRAGMA clear_cache` does not exist and `enable_object_cache` is a different
cache; toggling `enable_external_file_cache` off and back on evicts the entries and leaves caching enabled.
Verified standalone.

Two smaller ones, also caught by running it: DuckDB widens `SUM(BIGINT)` to HUGEINT and returns a
`BigInteger` that `Convert.ToInt64` refuses (the #1839 trap), cast in SQL; and `ArchiveService`'s 4GB
`memory_limit` raise is now `internal` and reused rather than copied, because that floor is a
known-broken-below-2GB constraint and a second drifting literal is exactly how #942 broke compaction.

## What is permanent, and said plainly

Darling's repair reaches only what raw still holds — a few days. **Query Store numbers older than that stay
understated for the life of the store**, because the daily tiers are kept indefinitely and cannot be rebuilt
from raw retention has dropped. No ordering of these operations reaches further back.

That is disclosed in three places: the #1845-style annotation in `ComposeCompiler`'s doc beside the #1849
boundary it shares, the CHANGELOG, and a CHANGELOG operator action that **leads with the timing** — the verb
gets *less* useful every day you wait, so run it right after upgrading. DARLING01 named explicitly.

## Testing

**Watched red per mutation**, each reverted:

| Mutation | Went red |
|---|---|
| `KeyColumnsFor` assumes the newest schema | the unit assertion **and** the archive test, the latter with the field symptom exactly: `Binder Error: Referenced column runtime_stats_interval_id not found in FROM clause` |
| archive file that fails mid-rewrite | showed the run aborted instead of isolating — **fixed the code**, not the test |
| survey of an uncombinable file | same hole one layer up — **fixed the code** |

The era-schema fixture is a parquet written **without** `runtime_stats_interval_id` or `replica_role`,
matching the real June 2026 archive file, so the era-appropriate key has to engage.

**Live PostgreSQL 18.4 + TimescaleDB 2.28.1** — rollup shows a slice before the repair; the survey finds
exactly the split cycles; the collapse leaves a correctly-collected interval untouched; a re-run is a no-op;
after re-materializing the rollup reports the interval's true **125** with the weighted mean **1871** (not
the 2011 an unweighted average gives). Plus the clamp-safety pin and the verb's dry-run → repair →
nothing-left cycle.

**Live DuckDB** — hot-store collapse, old-era archive rewrite, failure isolation, and the startup
marker/retry contract.

Archive discipline: every archive in the tests is a scratch file the test wrote into a temp directory it
deletes. The measurements that justified this work were taken against Erik's real archive by **copying** it;
the originals still carry their original mtimes.

**Suites**: Darling **4033 passed / 0 failed** against live PG (10 skipped = gated); Lite **1927 passed / 0
failed**. Full-solution `-t:Rebuild`: **0 warnings, 0 errors**. Installer.Tests deliberately not run.

## Deferred

None. The permanent residue is a disclosure rather than a deferral — no further work reaches it.


erikdarlingdata:
PR UP — #1931, all checks green, `d7fb1849`, MERGEABLE / CLEAN.

Both halves shipped on the shape you decided: Darling's `--collapse-legacy-slices` operator verb, Lite's automatic one-time startup repair with the #1832/v39 discipline (verify-before-promote on **two conservation invariants**, per-file isolation, marker withheld on partial so it retries, no backups).

**Three engine findings, each of which would have shipped silently:**
1. A refresh entirely within dropped raw chunks **destroys** the materialization — force or not. Pinned as a live test; the clamp derives from the collapsed rows so safety falls out of the data flow. `--backfill-rollups` checked and safe by construction.
2. An in-place parquet rewrite **poisons DuckDB's external file cache** (path-keyed, **instance** scope), which would have left Lite unable to read the archive it just repaired. Toggling `enable_external_file_cache` evicts it.
3. `SUM(BIGINT)` → HUGEINT → `BigInteger` (the #1839 trap), cast in SQL.

**Two of the watched-reds fixed the CODE rather than the test** — the mid-rewrite failure showed one bad file aborted the whole run, and the survey had the same hole a layer up.

Suites: Darling 4033/0 against live PG 18.4 + TSDB 2.28.1, Lite 1927/0, `-t:Rebuild` 0 warnings.

The permanent residue is disclosed in three places (ComposeCompiler doc beside the #1849 boundary, CHANGELOG, and an operator action that leads with the timing). DARLING01 named. No deferrals — nothing further reaches it.

erikdarlingdata:
PUSHED — `bb1f48c0`, all checks green, MERGEABLE / CLEAN.

**Confirming: option (a) landed. There is no Lite UI and none was ever started.** The "remaining is the Lite operator UI" phrase was from a message sent before the (a) decision arrived; Lite's half has been the automatic startup migration since, wired into `MainWindow` and covered by the marker/retry tests.

**Cache eviction moved inside the swap helper**, as directed. `PromoteRewrittenFileAsync` now owns delete + move + evict + rebuild-views, because replacing the bytes at a path and making the store able to READ that path are one operation, not two. Any future in-place rewrite that goes through it is correct without its author knowing about DuckDB's instance-scoped path cache; one that hand-rolls `File.Move` breaks the archive it just fixed. The comment also records why the monthly archive cycle is deliberately **not** given the same treatment — it only writes new names and deletes old ones, never replaces bytes behind a read path, so it was never exposed and adding it there would imply a hazard it does not have.

**Watched red, and it separates the two halves cleanly:** removing the eviction while *keeping* the view rebuild fails both archive tests with the exact production symptom, `No magic bytes found at end of file`. The view rebuild alone does not save it — the eviction is the load-bearing part.

Also corrected a stale claim the restructure surfaced: the class doc still said the Lite repair is "operator-invoked, never automatic", which was true of the pre-(a) shape and is now the opposite of what ships.

Darling 4033/0 against live PG 18.4 + TSDB 2.28.1, Lite 1927/0, build 0 warnings.

erikdarlingdata:
BOTH FIXED — `dd201c94`, all checks green, MERGEABLE / CLEAN.

**1. Lock correctness.** The mutating phases now take `AcquireWriteLock`, mirroring `ArchiveService`'s shipped idiom. The race was real rather than theoretical: the read lock excludes the collector, not the UI, and `RepairOnStartupAsync` is fired un-awaited *precisely* so the UI stays interactive — so a UI reader can be mid-`read_parquet` on the exact file being swapped.

Write-lock spans are **tight and per file**, not one lock around the run. Reading the original and writing the temp sibling touches nothing a reader can see, so only the swap instants need exclusivity — a large archive repair must not freeze the UI for its full duration when the swaps are the fast part.

Guarded by a source pin in the shape of the other #1907/#1912 pins. **Watched red:** reverting the hot collapse to the read lock drops the write-lock count 2 → 1 and fails it.

**2. Lock duration (#1564/#1567 family).** Fixed at the **call site**, not the doc: the verb now loops per-day slices over the survey span. A default 4-day raw tier is a handful; a widened retention gets the protection `CollapseSliceAsync`'s doc actually promises. Each slice is its own transaction, so a mid-run failure keeps the committed ones and the idempotent re-run continues — the output says so. The nothing-left assertion after the loop is unchanged.

Darling 4033/0 against live PG 18.4 + TSDB 2.28.1, Lite 1928/0, `-t:Rebuild` **0 warnings** (two xUnit2013 analyzer warnings my new pin introduced were fixed — the repo's zero-warning rule caught them).

PostgreSQL on 5619 is down, verified: connection refused, zero listeners.

Issue #1892, opened 2026-07-31
Query Store time-slice: window filters on collection_time but buckets on interval_start_time_utc, so a bucket can fall outside the requested range


erikdarlingdata:
Raised in review of #1853 and confirmed still present on `dev`. Applies identically to **both** apps.

## The mismatch

The Query Store time-slice queries filter the window on `collection_time` but bucket on `interval_start_time_utc`:

- `Lite/Services/LocalDataService.QueryStore.cs` - CTE filters `collection_time >= $2 AND collection_time <= $3`, outer query groups on `date_trunc('hour', COALESCE(interval_start_time_utc, collection_time))`
- `Darling/PerformanceMonitor.Darling.Viewer/ViewerDataService.QueryStore.cs` and `ViewerDataService.QueryTrends.cs` - identical construction

Those are no longer the same instant. `interval_start_time_utc` is the Query Store interval's START; the row that survives the window filter is selected by its CLOSING `collection_time`, which for a 60-minute interval lands up to roughly one interval-length later.

## Effect

An interval that started at `windowStart - 30min` and closed with its final collection at `windowStart + 2min` passes the `collection_time` filter, but its bucket is `date_trunc('hour', windowStart - 30min)` - dated BEFORE the nominal window start. A "last N hours" chart can therefore show one extra bar to the left of the requested range, bounded by one Query Store interval length (default 60 minutes).

Before #1841/#1853 this could not happen: bucket == `collection_time` == the filtered column, so every bucket was guaranteed inside `[start, end]`.

The totals inside that bucket are accurate - this is a placement/range-edge issue, not a data-correctness one. The bar is real work that really happened at that time; it is just outside the window the user asked for.

## Not covered by the existing test

The live-Postgres test added in #1853 seeds an interval start (`h0`) that is never earlier than the queried start (`h0.AddMinutes(-1)`), so it does not exercise this boundary.

## Options (needs a decision, which is why this is filed rather than fixed)

1. **Document and accept.** One extra bar of real data at the left edge is arguably better than dropping it. Cheapest, and the numbers stay honest.
[...]

Issue #683, opened 2026-03-25
Highlight selected grid row on performance trend charts


erikdarlingdata:
## Summary

When a user selects a row in a query/procedure grid, the performance trend chart above doesn't react. The user has to double-click, open a modal history window, and mentally correlate timing.

## Proposed Behavior

When a row is selected in a query grid (Top Queries, Top Procedures, Query Store), overlay that specific query's trend line on the existing performance trend chart — or highlight its data points. This instantly answers "is this query always expensive or did it just get bad recently?"

This is lightweight — no popup or navigation required. Just a visual reaction: click row → chart responds.

## Example Flow

1. User is on the Query Performance tab, looking at the duration trend chart and the Top Queries grid
2. User clicks a row for a query with high avg duration
3. The duration trend chart overlays that query's individual duration trend as a highlighted line
4. User can immediately see whether the query recently regressed or has always been slow

## Design Notes

- The overlay should be visually distinct (brighter color, thicker line) from the aggregate trend
- Deselecting the row (or selecting a different row) should clear/replace the overlay
- Data for the individual query trend is already available via `GetQueryTrendAsync` by query_hash
- Applies to both Dashboard and Lite

Complexity: Low — visual overlay on existing charts, data fetch method already exists.

erikdarlingdata:
**Implementation order**: Lite first, then port to Dashboard once the pattern is validated. This avoids the complexity of iterating on both apps simultaneously.

PR #1938, opened 2026-07-31
Place the Query Store slicer overlay where the work ran (#1921, option 1)

### LINKED ISSUE #1921: Query Store item-timeline overlay plots at collection_time while the bars it annotates are keyed on the interval start
Found while fixing #1892 and verified against the code; filed rather than folded in because the fix is a product decision about what the overlay is FOR, not a mechanical edit.

## The disagreement

`QueryStoreItemTimelineSql` draws its points at `collection_time`:

- `Darling/PerformanceMonitor.Darling.Viewer/ViewerDataService.ItemTimeline.cs` — projects `collection_time` as the x-value, windows on `collection_time >= $5 / <= $6`

That series is drawn **over** the Query Store slicer bars, which since #1841 are keyed on `COALESCE(interval_start_time_utc, collection_time)` — the hour the work RAN. The two axes are no longer the same instant, so a point sits up to one Query Store interval (default 60 minutes, max 1440) to the RIGHT of the bar describing the very same work.

The file's own comment states the invariant this breaks:

> This series is drawn OVER the Query Store slicer bars, which ARE deduped (QueryStoreSlicerSql) — leaving the overlay raw would make the overlay disagree with the bars it annotates.

The dedup half of that invariant holds. The PLACEMENT half silently stopped holding when #1841 moved the bars and left the overlay where it was.

#1892 does not create this and does not worsen the x-axis, but it does add a second axis of disagreement: after #1892 the bars are windowed on the effective timestamp while the overlay is still windowed on `collection_time`, so at a range edge the two surfaces can disagree about whether a given interval is in the window at all.

## Why this is a decision, not a mechanical edit

The overlay answers a genuinely different question from the bars. A bar says "this much work ran in this hour". An overlay point arguably says "at this moment, the collector observed these cumulative totals" — which is a real thing to want, and `collection_time` is the honest x-value for it.

So there are two defensible fixes and they are not equivalent:

1. **Move the overlay to the interval start**, matching the bars. The overlay becomes "work, placed when it ran", consistent with everything else post-#1841, and the file's stated invariant holds again. Cost: the overlay stops being able to show WHEN something was observed, and several points from one collection cycle for different intervals now spread across the x-axis instead of stacking.
2. **Keep `collection_time` and document the overlay as a collection-event series**, updating the comment that currently claims agreement. Cost: the two series on one chart keep using different clocks, which is exactly the kind of thing that produces a support question.

Option 1 is probably right — it is the direction #1841 took everything else — but it changes what a shipped chart means, so it wants a deliberate call rather than being folded into an edge-case fix.

## Scope when it is done

Both apps, and the window filter moves with the x-value (the #1892 shape: filter on the same expression you place on, keep a bare `collection_time` floor for chunk exclusion and no ceiling).

- `Darling/PerformanceMonitor.Darling.Viewer/ViewerDataService.ItemTimeline.cs`
- Lite's equivalent item-timeline read
- The live pin `ViewerQueriesTests.QueryStoreAggregates_CountARecollectedIntervalOnce_AtItsLatestValues_AgainstDevPostgres` asserts the timeline point lands at `bucketStart+15m` (a collection time), so it moves with the decision.

erikdarlingdata:
Fixes #1921. Erik's call: **option 1** — the overlay becomes work-placed-when-it-ran.

## The disagreement

Selecting a Query Store row overlays that item's curve on the slicer. #1841 moved the **bars** to the hour an
interval ran and left the **overlay** on the hour it was collected, so a point sat up to one Query Store
interval — 60 minutes by default, 1440 at the maximum — to the RIGHT of the bar describing the very same work.

The file's own comment stated the invariant that broke:

> leaving the overlay raw would make the overlay disagree with the bars it annotates

Its dedup half kept holding. Its **placement** half silently stopped.

## The fix

Both apps place the overlay on `COALESCE(interval_start_time_utc, collection_time)` — the same expression the
bars use — and pre-tier-2 rows keep collection-time placement, the same two-generation handling every other
post-#1841 read uses.

**The window moved with the x-value**, which is the part that would otherwise still disagree. Filtering on
collection time while plotting on the interval start disagrees at *both* edges — an interval that started
before the window but closed inside it draws a point to the left of the range, and the window's own last
interval, whose closing fetch lands after the range ends, vanishes — and worse, it disagrees with the bars
about whether an interval is in the window at all. Same `collection_time` pruning bounds the slicer already
ships (#1892/#1923); **the issue's own "no ceiling" scope note is stale**, written before that round added one.

`ORDER BY` moved too — ordering on collection time while plotting on the interval start hands the chart a
non-monotonic series, which a line control draws as a zig-zag rather than a curve.

## Two things the brief didn't anticipate

**1. The nominated watched-red pin does not move — and shouldn't.**
`QueryStoreAggregates_CountARecollectedIntervalOnce_...` seeds its rows *without* an interval identity
(`intervalId`/`intervalStartUtc` default to NULL), so `COALESCE` correctly falls back and the point stays at
`bucketStart+15m`. It passed before and after the change.

Rather than force it, its assertion is unchanged and now pins the legacy **fallback**, and the behavioral
proof went into `QueryStoreIntervalIdentity_PlacesWorkWhenItRan_...`, whose tier-2 rows genuinely disagree —
the interval starts in `h0` and every collection lands in `h1`. Watched red there: reverting the x-value fails
it with **`Expected 06:00, Actual 07:10`**, the 70-minute offset, while the legacy pin keeps passing. Both
generations are now covered by the test whose data can actually demonstrate each.

**2. Lite has no dedicated overlay read.**
It fed the overlay from the history **grid's** read — a deliberately raw per-collection projection (#1845)
windowed on collection time. That is right for a list of collection events and wrong for a series drawn over
the bars, and moving the shared read would have silently changed the history grid's window too, a surface
outside this issue.

So Lite gained a dedicated overlay read, matching the structure Darling has had since #683. It picks up the
**dedup the overlay never had**: an interval collected repeatedly used to draw as a rising staircase of
restatements and now draws as one point at its final values — which is what the bar beside it always showed.
That is the same invariant ("the overlay agrees with the bars"), not extra scope.

## Testing

**Watched red** on the behavioral pin, as above. Both source guards caught the new read and were updated
deliberately — 5 → 6 dedup sites, 7 → 8 rank filters, tie-break total 7 → 8. The #1912 tie-break guard fired
from the direction of a read being **added** rather than edited, which is exactly what it was written for.

New live-DuckDB pin `Overlay_PlacesPointsWhereTheWorkRan_AgreeingWithTheBarsItIsDrawnOver` asserts the
invariant as an assertion rather than a comment: same x and same value as the bar computed from the same rows.

Sibling timelines are untouched and correctly so — `query_stats`/`procedure_stats` are per-collection deltas
with no interval concept, so they stay on `collection_time`.

**Suites**: Darling **4033 passed / 0 failed** against live PostgreSQL 18.4 + TimescaleDB 2.28.1 (10 skipped =
gated); Lite **1937 passed / 0 failed**. Full-solution `-t:Rebuild`: **0 warnings, 0 errors**.
Installer.Tests deliberately not run.

## Accepted costs, recorded so they aren't re-reported as bugs

The overlay can no longer show WHEN a value was observed — a real thing to want, and what option 2 would have
kept — and several intervals from one collection cycle now spread across the x-axis instead of stacking.
Spreading is the truthful picture: those intervals really did run at different times.

## Deferred

None.


erikdarlingdata:
PR UP — #1938, all checks green, `46b3d9e6`, MERGEABLE / CLEAN (checked right after opening too).

**Two things the brief didn't anticipate, both verified before acting rather than worked around:**

**1. The nominated watched-red pin does not move — and shouldn't.** `QueryStoreAggregates_CountARecollectedIntervalOnce_...` seeds its rows *without* an interval identity, so `COALESCE` correctly falls back and the point stays at `bucketStart+15m`. It passed before and after. Rather than force it, its assertion is unchanged and now pins the legacy **fallback**; the behavioral proof went into `QueryStoreIntervalIdentity_PlacesWorkWhenItRan_...`, whose tier-2 rows genuinely disagree (interval starts h0, collections land h1). Watched red there: reverting the x-value fails with **`Expected 06:00, Actual 07:10`** — the 70-minute offset — while the legacy pin keeps passing. Both generations now covered by the test whose data can demonstrate each.

**2. Lite has no dedicated overlay read.** It fed the overlay from the history *grid's* read — a deliberately raw per-collection projection (#1845) windowed on collection time, right for a list of collection events and wrong for a series drawn over the bars. Moving that shared read would have silently changed the history grid too, a surface outside this issue. Lite gained a dedicated read instead, matching Darling's structure since #683 — and it picks up the **dedup the overlay never had**, so a repeatedly-collected interval now draws as one point at its final values instead of a rising staircase. Same invariant, not extra scope.

Also followed your correction over the issue text: the `+ interval '30 days'` pruning ceiling from #1923 is in, and `ORDER BY` moved with the x-value (a non-monotonic series draws as a zig-zag).

Both source guards caught the new read and were updated deliberately (5→6 dedup sites, 7→8 rank filters, tie-break total 7→8) — the #1912 guard fired from the direction of a read being **added** rather than edited, which is what it was written for.

Darling 4033/0 against live PG 18.4 + TSDB 2.28.1, Lite 1937/0, `-t:Rebuild` 0 warnings. PostgreSQL on 5619 is down, verified.

PR #1943, opened 2026-07-31
Adding PagerDuty as a first-class webhook notification channel

ianwalkeruk:
## What does this PR do?

Adds PagerDuty as a first-class webhook channel across Lite and Darling editions, implementing native Events API v2 integration. This provides proper incident correlation via dedup_key, severity mapping (CRITICAL→critical, ALERT→error, WARNING→warning), and support for both US and EU data centers.

- [ ] Lite
- [ ] Darling
- [ ] Lite Tests
- [ ] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## How was this tested?

Describe the testing you've done. Include:
- SQL Server version(s) tested against
- Steps to verify the change works

## Checklist

- [ ] I have read the [contributing guide](https://github.com/erikdarlingdata/PerformanceMonitor/blob/main/CONTRIBUTING.md)
- [ ] I have tested my changes against at least one SQL Server version
- [ ] I have not introduced any hardcoded credentials or server names


erikdarlingdata:
Merged - and thank you, this was a genuinely well-studied contribution: the secret tiers in both apps, the V42 migration claim, and the viewer projections were all exactly right, which is rare for a first PR here. Three maintainer fixes rode along (details in the commit messages): the credential-load block re-read siblings inside the settings guard and re-broke the #1832 hoist contract (two shipped tests caught it), the Viewer's Settings window never wired the PagerDuty boxes into the notification row round-trip, and two small doc corrections. Proxy support for the channel follows immediately in #1945, and the whole thing ships to a field site in tonight's build. PagerDuty is in the next release's headline notes with your credit.

ianwalkeruk:
## Follow-up on the PagerDuty channel

Two follow-up commits are now on the fork's `dev` after this PR merged. They are **not** part of this merged PR (they cannot be — it is already merged) and would need a new PR to reach upstream.

### 1. Connection edges now correlate to one incident — `c5b2b50f`

`"Server Unreachable"` and `"Server Restored"` derived distinct `dedup_key` values (`<serverId>:Server Unreachable` vs `<serverId>:Server Restored`) because the fallback key embedded the metric name, so a single outage opened two PagerDuty incidents. Both edges now share `<serverId>:ServerConnection`. The normalization lives in `DerivePagerDutyDedupKey`, which the generic `{{dedup_key}}` token and the triage link also consume, so all channels agree on the key.

### 2. Opt-in auto-resolve for connection restores — `34179322`

The tool deliberately does not auto-resolve incidents with third parties, so the recovery no longer resolves by default:

- **Unchecked (default):** `"Server Restored"` is delivered as an info-severity `trigger` on the shared connection `dedup_key`; the incident stays open.
- **Checked — "Auto-resolve the incident when the server recovers":** the recovery sends `event_action: "resolve"` on the same key, closing the incident its `"Server Unreachable"` trigger opened.
- Scoped to the connection edge, so other RESOLVED notices (`"AG Replica Reconnected"`) keep triggering either way.

Surfaces:

- **Lite:** `pagerduty_auto_resolve` in `settings.json`, authored in the Settings window.
- **Darling:** `config_notification.pagerduty_auto_resolve` (V127 rung; non-secret, so the read-only `viewer` role keeps its column grant), authored in the Viewer Settings window, with the matching `ViewerDataService` probe sentinel and schema-version rung.
- **Deprecated Dashboard:** reports `false` (no new settings surface).

### Validation

- Lite.Tests, Darling.Tests and the deprecated Dashboard build with 0 warnings.
- Darling: 9862 tests, 0 failed, 389 skipped (env-gated live tests).
- Lite: 4014 tests, 0 failed except 3 `CrossAppGuardCiGateTests` that fail only because the pinned SDK 10.0.302 is not installed locally (they shell out to `dotnet msbuild`).
- Caveat: the Darling V127 migration and probe are compile- and test-verified only — no live PostgreSQL/TimescaleDB here, so the `ALTER TABLE` has not been executed against a store.

erikdarlingdata:
Thank you for the careful follow-up note — and for getting the mechanics right unprompted: you're correct that these need a new PR, and both changes are welcome. The connection-edge normalization in particular fixes a real wart (one outage opening two incidents is exactly the kind of thing this channel exists to prevent), and the opt-in default for auto-resolve is the right instinct — this tool doesn't close incidents with third parties unless someone explicitly asks it to.

Three things to fold into the rebase before you open it, all consequences of how much the trunk has moved since July:

1. **The V127 rung is taken — several times over.** The migration ladder is at **V135** as of today (it moved four rungs in the last 24 hours alone). Renumber to the next free rung at the moment you open the PR, and expect to renumber again if something lands ahead of you — the house rule is one un-landed rung at a time, and the end of the ladder belongs to the newest PR. The rung's shape (nullable ADD COLUMN, non-secret so the `viewer` grant keeps working) is exactly right as described.

2. **Dedup keys grew a consumer since your fork.** As of today the notification path routes findings by corroboration and stamps its routing decision into the alert-history ledger alongside the dedup key (#3712, merged this afternoon). Your `DerivePagerDutyDedupKey` normalization should be fine — the generic `{{dedup_key}}` token agreeing across channels is precisely the property we want — but rebase onto current `dev` and re-run the notification tests rather than trusting the July baseline; the files you're touching have moved.

3. **State the store caveat on the PR face.** Your note that the V-rung is compile/test-verified but has never executed against a live TimescaleDB store is exactly the honest disclosure we want — put it in the PR body, and CI's PostgreSQL job will close that gap (it runs the full ladder against a real store).

Open it whenever suits you — base `dev` as before. Looking forward to it; the first PR set a high bar and this follow-up reads like more of the same.

Issue #1964, opened 2026-07-31
Lite.Tests: two xUnit2013 analyzer warnings on dev (ProcedureStatsCollectorDefinitionTests)


erikdarlingdata:
`dotnet build Lite.Tests/Lite.Tests.csproj -c Debug -t:Rebuild` on current `dev` emits two warnings:

```
Lite.Tests/ProcedureStatsCollectorDefinitionTests.cs(151,9): warning xUnit2013: Do not use Assert.Equal() to check for collection size. Use Assert.Single instead.
Lite.Tests/ProcedureStatsCollectorDefinitionTests.cs(152,9): warning xUnit2013: Do not use Assert.Equal() to check for collection size. Use Assert.Single instead.
```

Introduced by #1959's test additions (commits 537ec280 / 7ff4d359). The fix is mechanical - `Assert.Equal(1, x.Count)` becomes `Assert.Single(x)` on both lines.

Not fixed in #1953's branch on purpose: that branch touches no Lite files, and the file had just been merged by a parallel session. Filed rather than folded in so the zero-warning build is restored deliberately.

Not urgent: `Lite.Tests.csproj` does not set TreatWarningsAsErrors, so CI on `dev` is green - this is warning hygiene, not a broken build.

erikdarlingdata:
Already fixed on `dev` independently — commit `a9e2f2ae`-era work turned both assertions into `Assert.Single(...)`, which came in with the merge I just took into #1968's branch. Verified: `dotnet build Lite.Tests -t:Rebuild` on the merged tree is clean. Closing as done rather than leaving a filed-then-fixed issue open.

Issue #1959, opened 2026-07-31
query_stats and procedure_stats render plan XML BELOW the TOP: 2,434 TVF executions for 200 kept rows - validated 6x fix from field investigation


erikdarlingdata:
Field investigation from the prod-sql-use2 fleet (52 RDS replicas), source-analyzed at 0af464e with live A/B measurements and captured actual plans. Fleet query_stats sweeps: avg 5.1s / p95 9.3s / max 34.4s over 24h, including 30-SECOND COMMAND-TIMEOUT MISSES on the heaviest boxes - lost collection cycles, not just slowness.

## Mechanism (from the captured plan, 21,316 cache entries)

The dm_exec_text_query_plan OUTER APPLY is spliced into the FROM, and the optimizer does NOT defer it past the TOP even though query_plan_xml is only referenced in the SELECT list: the TVF sat at node 24, deep below the Top, executed 2,434 times to keep 200 rows - 92% of the plan-XML renders discarded, 5,293 ms of a 6,499 ms plan (81%). The optimizer estimated 3 rows reaching the Sort (46x under) and never considered deferral. The usual suspects (NOT LIKE self-filter 361 ms, sql_text 17 ms, DMV scan 80 ms) were noise by comparison.

## The fix, validated on the field fleet

Rank on the cheap DMV columns first in a derived table (inner TOP with headroom, ORDER BY total_elapsed_time), then run the text/plan applies and the NOT LIKE self-filter against survivors only - deferral made STRUCTURAL instead of hoped-for. Measured interleaved on the alpha box: median 7,736 ms -> 1,286 ms (6.0x), TVF executions 2,434 -> 200, row-identity parity 99.2% with the residual fully explained by cache churn between paired runs. The fleet's 30s timeout misses concentrate on exactly the big-cache boxes where discarded renders scale worst - expected to disappear.

## Scope and caveats (the reporter's, all sound)

- The big win is Darling-specific (CapturePlanXml=true); Lite's shape pays only the text-apply + NOT LIKE (~380 ms here) - same rewrite applies, smaller win, parity is the rule so both ship together.
- Inner TOP needs headroom over the outer 200 because the self-filter moves post-ranking (240 clean on their alpha; 300 costs nothing - use 300).
- Evicted-text rows excluded identically in both shapes - no parity change.
[...]

Issue #1954, opened 2026-07-31
Viewer diagnostics: say WHERE darling.json and server.crt were read from, what was parsed, and which connection layer failed


erikdarlingdata:
Field feedback: debugging Viewer configuration problems is very hard - you cannot tell where it read darling.json or server.crt from, or what it got from them. Verified against dev: the Viewer's config load logs NEITHER path today, and a failed connection surfaces as a generic failure with no layer attribution.

## The fix shape

1. **At startup, log (and show in the connection-failure UI) the FULL RESOLVED PATHS**: which darling.json was read (DARLING_CONFIG override vs the Roaming default), which certificate file the connection string named, and whether each existed.
2. **A non-secret parse summary**: host, port, username, database, sslmode, cert path, the managed flag - everything except credentials - so 'it read the wrong file' and 'it read the right file but the value is wrong' stop being indistinguishable.
3. **A layered connection self-test** (button in the connect-failure dialog or a --test viewer flag): DNS resolution, TCP reach on the port, TLS handshake incl. cert validation, auth, and the schema-version gate - reporting exactly which layer failed. This is also the honest answer to 'firewall problems are hard to debug': a TCP-layer failure with DNS succeeding points at the firewall by elimination, from the viewer's own seat.

The existing service-side --test-connection proves the shape works; this is its viewer-side counterpart.

erikdarlingdata:
Core shipped in #1966 (bullets 1 and 2). **Leaving this open for bullet 3.**

**Done:**

- **Where the config came from.** Startup logs which of the four resolution rules won (explicit command-line path, `DARLING_CONFIG`, beside the viewer, the service root one level up), the absolute path it produced, and whether that file exists — emitted *before* the load is attempted, so a missing or unparseable file still says where the viewer looked.
- **The non-secret parse summary.** Host, port, username, database, SSL mode, search path, whether the string was read verbatim or DERIVED from `postgres.managed`, and the certificate: the value as written, the absolute path Npgsql will actually open, and whether it exists. A bare `Root Certificate=server.crt` resolves against the process *working directory*, so a shortcut launch and a shell launch look in different places — now stated rather than inferred.
- **Both in the connection-failure window**, under a Copy details button, as well as the log.
- Redaction is an allowlist rather than a habit, pinned by a test that feeds a live password through and asserts it appears nowhere in the output.

[...]

Issue #1581, opened 2026-07-19
Darling managed store: 30h climbing memory + Windows 487 backend-spawn failures on a field box


erikdarlingdata:
**Field box** (managed co-located PostgreSQL, Windows). Over ~30 hours: OS free 11.9 -> 9.6 GB; postgres working-set SUM 8.15 -> 18.1 GB; checkpointer WS ~1066 MB; Darling service WS 255 -> 467 MB; Windows "could not reserve shared memory region" (error 487) count 10 -> 67 and still firing. Has not crashed, but the trajectory resembles the pre-round-4 OOM pattern (#1557) unfolding over a day+ instead of hours.

### Reading the numbers (confirmed from code)
- Checkpointer WS (~1066 MB) ~= shared_buffers, so **shared_buffers ~= 1 GB and the #1559 cap IS active on this box** (min(25% RAM, 1 GB), re-applied every start via `EnsureConfAppended`).
- On Windows shared_buffers is charged to every backend's working set, so WS SUM ~= shared_buffers x N_backends. 8.15 -> 18.1 GB => **backend count ~8 -> ~18: the climb is ~10 additional persistent backends over 30h.** OS free (-2.3 GB) is the true consumption; the WS "sum" is mostly the shared segment counted N times.
- **Not a missing/uncapped pool.** All three managed store pools route through `DarlingManagedPostgres.BuildRoleConnectionString` (`MaxPoolSize=24`, DarlingManagedPostgres.cs:466) -- owner (:442), mcp (:503), viewer (:522); WPF viewer = 10 (ViewerSettings.cs:198). A capped pool plateaus; it can't produce an unbounded 30h climb.

### Leading suspect (needs box data to confirm)
**TimescaleDB compression + autovacuum background workers, amplified by the 487 retry loop.** The managed conf allows ~29 timescale background workers + ~40 max_worker_processes (`DarlingManagedPostgres.cs:294-307`, sized HypertableCount+2); ~27 compression policies (`PgMigrations.cs:905`, compress_after = 1 day) begin firing as chunks cross ~24-48h old -- which matches the onset window -- and autovacuum ramps as hypertables grow. On Windows each worker spawn is a 487 risk; a failed spawn is rescheduled by the TS job scheduler -> churn that both climbs the 487 count and keeps re-spawning. (Retention itself is clean -- inline drop_chunks/DELETE via the shared pool with `await using`, no add_retention_policy jobs.)

[...]

PR #1968, opened 2026-07-31
Export the viewer's whole config folder instead of documenting it (#1953)

### LINKED ISSUE #1953: Viewer onboarding: the service should hand you a working viewer config, not a documentation exercise
Field feedback from a real setup: getting the remote Viewer working was harder than any other part - copy-pasting JSON from the documentation, merging it with --print-viewer-connection's terminal output, and discovering by trial that the VIEWER's darling.json wants "managed": false even though the SERVER runs "managed": true. Every one of those steps is work the service could do itself, since it already has every value the viewer needs.

## The fix shape

1. **A --export-viewer-config verb** (or an extension of --print-viewer-connection) that writes the COMPLETE viewer-side darling.json - connection string with the resolved host, Root Certificate path wired, "managed": false already set - plus the server.crt beside it, as files ready to drop on the viewer machine. No hand-merging. The managed-flag confusion dissolves structurally: nobody writes that file by hand anymore.
2. **Consider the full step further**: the service ships the entire Viewer's code already - evaluate exporting a pre-configured viewer ZIP (binaries + config + cert) so the viewer machine unzips and runs. Weigh the binary-duplication cost honestly; the config-plus-cert export alone removes most of the pain if the ZIP is too heavy.
3. **--print-viewer-connection warns BEFORE it prints, not after.** The reporter saw the secret hit their scrollback and THEN the redirect advice. Order the STDERR guidance ahead of the STDOUT payload, and verify the interleaving on a real console rather than assuming stream order.
4. The exported file carries its own comments documenting each field, including the valid 'Root Certificate=' values - the export becomes the reference the docs currently fail to be.

Alternative also worth weighing: a first-run GUI prompt in the Viewer asking for host/port/credential/cert instead of requiring a file at all. The export and the prompt are complementary, not competing.

erikdarlingdata:
Closes #1953 (items 1, 3 and 4; item 2 assessed and declined below).

## What changed

**`--export-viewer-config`** — run on the service host, it writes the viewer machine's entire configuration folder:

```
viewer-config\darling.json    resolved host/port, the role's decrypted credential, verify-full TLS,
                              Root Certificate wired, "managed": false already set - and every field
                              documented in // comments IN the file
viewer-config\server.crt      the store's certificate, the file the connection pins
viewer-config\README.txt      the same reference in plain text, including the valid
                              "Root Certificate=" values and the one-line install instruction
```

Copy the folder over, point `DARLING_CONFIG` at it, start the Viewer. Nothing to edit, nothing to merge, and nobody writes that file by hand anymore — which is what dissolves the `managed` confusion structurally rather than with a doc paragraph.

Default destination is `viewer-config` beside the service's `darling.json`; a positional argument names another (`--export-viewer-config D:\handoff`).

**`--print-viewer-connection` warns before it prints.** All STDERR guidance now precedes the STDOUT payload — including the missing-certificate NOTE, which used to trail the secret. Verified with both streams redirected to a single handle, which is the only way to observe the real order (PowerShell's `2>&1 |` merge reorders and cannot be trusted for this).

**Shared resolution.** One `ResolveViewerHandoffAsync` now feeds both verbs, so they cannot disagree about host, role, credential or certificate. The print verb's behavior is otherwise unchanged.

## Decisions worth reviewing

- **The exported JSON carries comments.** `System.Text.Json` cannot *write* comments, so the file is hand-composed — the viewer parses with `JsonCommentHandling.Skip` (the same tolerance `darling.sample.json` relies on), which is what makes issue item 4 possible in the file itself and not only in the README. A test feeds the exported text to the real `ViewerSettings` parser to keep that honest.
- **`SSL Mode=VerifyFull` is kept, not downgraded to `Require`.** `Root Certificate` is only consulted under `VerifyCA`/`VerifyFull`; under `Require` the connection stays encrypted but stops verifying, silently voiding the pin the export exists to wire up. Both the JSON comments and the README say so explicitly, since "downgrade it until it connects" is the obvious wrong move when TLS fails.
- **A missing certificate is fatal to the export** (unlike the print verb, which still has a useful string). A verify-full folder without its PEM is a handoff that fails on the operator's desk — exactly the half-finished setup being eliminated.
- **The positional argument is the DESTINATION**, inverting the sibling verbs' "positional = config path" convention, because that is how the verb reads. An explicit config path is `--config <path>`, and a destination that is really a config file is refused by name — before any credential is decrypted.

## Secret discipline

The exported `darling.json` holds a live credential (that is what the viewer authenticates with). The verb prints a loud STDERR warning naming the file *before* writing it, and hardens the file to SYSTEM + Administrators + the running account + INTERACTIVE, dropping any inherited `BUILTIN\Users` read. The password value is never echoed to either stream, and the README deliberately takes no password parameter. Verified on the live run:

```
NT AUTHORITY\INTERACTIVE : Read, Synchronize
NT AUTHORITY\SYSTEM : FullControl
BUILTIN\Administrators : FullControl
NADAFACCIA\edarl : FullControl
```

## Registration

`IsExportViewerConfigVerb` is in the `IsKnownVerb` allow-list, `UsageText`, and the `Program` dispatch. #1950's reflection reachability pin was watched red first — with the allow-list entry removed it fails by name: `IsKnownVerb does not reach IsExportViewerConfigVerb's verb '--export-viewer-config' - the Program.cs dispatch for it is UNREACHABLE`.

## Item 2 (pre-configured viewer ZIP): declined

The issue asked for it to be weighed honestly rather than assumed. The Viewer is a WPF app whose publish output is tens of MB against a ~6 KB config folder, the service would have to carry or reach the viewer's build output at runtime, and the operator still has to solve unblocking a downloaded archive and keeping the binaries in step with the service version on every upgrade — a versioning surface the config folder does not have. The config-plus-cert export removes the pain the field report actually described; the ZIP adds a distribution channel to maintain. Not nearly free, so out.

## Verification

- `Darling.Tests`: **3857 passed / 0 failed / 221 gated-live skipped**
- `Lite.Tests`: **1962 passed / 0 failed**
- `-t:Rebuild` on both test projects: 0 warnings, 0 errors. (Two pre-existing xUnit2013 warnings in `Lite.Tests/ProcedureStatsCollectorDefinitionTests.cs` come from #1959, a file this branch does not touch — filed as #1964.)
- No `Installer.Tests` run.

Every behavior was watched red before green: the allow-list entry removed (reachability pin fails by name), `"managed": false` flipped to `true` (the real viewer parser then goes hunting for a local bundled PostgreSQL — the reporter's exact bug), the STDERR/STDOUT reorder reverted to the shipped shape (both ordering tests fail on the interleaved log), and the missing-certificate refusal removed (the export test fails).

Live proof against a scratch managed layout with the built binary — the error paths first, since those are what an operator hits before anything works:

```
===== RUN 1: no credential material yet =====
The 'viewer' role credential (...\store\pg-viewer-credential.dpapi) does not exist yet. Start the
PerformanceMonitor Darling service once so its first run provisions the least-privilege roles and
their credentials, then re-run this command.
exit=1

===== RUN 2: credential present, TLS certificate still missing =====
Cannot export: the server TLS certificate (...\store\server.crt) does not exist yet, and the exported
connection uses SSL Mode=VerifyFull - without the cert the viewer could not connect. The service
generates it on its first managed start with postgres.network exposed: set postgres.network
(listen + allowFrom), restart the service, then re-run this command.
exit=1

===== RUN 5: config path typed where the destination goes =====
'...\1953-live\darling.json' is a file, not a directory. This verb's argument is the DESTINATION
folder for the exported viewer files; to point it at a different darling.json, use --config <path>
(e.g. --export-viewer-config D:\handoff --config C:\Darling\darling.json).
exit=1
```

Successful export, both streams to one handle (STDERR entirely ahead of the payload):

```
WARNING: ...\handoff\darling.json will contain a LIVE database password (the 'viewer' role) in
cleartext, because that is what the viewer connects with. Copy the folder over a channel you trust,
and keep the file ACL'd to the operator account on the viewer machine.

Copy the whole folder to the viewer machine, then point DARLING_CONFIG at its darling.json
(or drop the files beside the Viewer executable). README.txt documents every field.
Re-run this command after the store's certificate is regenerated (a changed bind IP rotates it).

...\handoff
...\handoff\darling.json
...\handoff\server.crt
...\handoff\README.txt
exit=0
```

And the reordered print verb, same single-handle capture:

```
WARNING: the connection string below contains a LIVE database password (the 'viewer' role), written
to STDOUT. Redirect it to an ACL'd file or pipe it to the clipboard; do not leave it in shell
scrollback, CI logs, or a screenshare.
  Example (file):      ... --print-viewer-connection > viewer-connection.txt
  Example (clipboard): ... --print-viewer-connection | clip
  Example (no paste):  ... --export-viewer-config   (writes the whole viewer folder for you)
Save the certificate block below as 'server.crt' on the viewer machine ...

# Paste into the viewer machine's darling.json -> postgres.connectionString (with postgres.managed = false):
Host=192.168.1.205;Port=5641;Username=viewer;Password=...;Database=darling;Search Path=collect,config,public;SSL Mode=VerifyFull;Root Certificate=server.crt
```

Docs: `Darling/README.md`'s remote-viewer handoff now leads with the export (it previously walked operators through the exact hand-merge this removes), states the `managed` flag's meaning outright, and corrects the bare-`server.crt` claim — it resolves against the **working directory**, not "beside the viewer", which a desktop shortcut can change. The `--configure-network` wizard's next-steps now point at the export too.


erikdarlingdata:
## Security round: the export could destroy files it had no business touching

An adversarial review of this branch found a data-loss defect in the new verb serious enough to write up separately. **Reproduced against the built binary before fixing:**

```
service darling.json BEFORE: 390 bytes
> --export-viewer-config <the install directory>
exit=0
service darling.json AFTER: 2813 bytes
first line AFTER: // PerformanceMonitor Darling - VIEWER configuration.
```

Exporting into the service's **own** config directory replaced the service's `darling.json` with the viewer's — every monitored server, every DPAPI `encryptedPassword`, the MCP bearer token and the web access token, none of which exist anywhere else. Exit 0, no warning, unrecoverable. And the install directory is the obvious place an operator would put a handoff folder, so this was one keystroke from a legitimate command.

### Fixed

Destination guards now run **before** any config load or credential decrypt (they are pure path questions, and the failure they prevent cannot be undone):

- the destination's `darling.json` resolving to the **service's own config** — refused, naming both paths
- a `darling.json` in the destination that **this verb did not write** — refused. That covers an operator's own file and one pre-created by a local user who would keep **ownership** of it through the harden (a Windows owner retains `WRITE_DAC` and can re-grant themselves read afterwards, whatever DACL we apply)
- a destination that is a **junction or symlink** — refused. Creating one needs no privilege on Windows, which makes it the way an unprivileged local user redirects a cleartext credential into a folder they control
- re-exporting over this verb's **own** output stays silent — it is the documented step after a credential or certificate rotation, and a guard that broke it would be a worse bug than the one it fixed

Also from the review:

- **The file is deleted before being rewritten**, so this process owns it instead of inheriting a pre-planted owner.
- **The harden is verified, not assumed.** Five sibling call sites pair `HardenFile` with `IsReadableByOrdinaryUsers`; this one did not, while its own doc comment claimed it mirrored them. A secret left readable is now a `WARNING` with delete-it advice and a **non-zero exit**, so automation cannot read "exported" as "protected". `NOTE:` was also the wrong prefix — it is what the same verb uses for a benign admin-role tip.
- **A write that dies part-way** takes the partial secret with it and says so; previously "the export failed" left an unmentioned file holding the start of a live password.
- **Argument parsing is strict**, in the spirit of #1581's classifier. An unrecognized flag used to become the destination, and a bare `--config` with no value would have written a live password to a folder literally named `--config` under the working directory — `C:\Windows\System32` for the elevated prompt the docs prescribe.
- **A UNC destination is warned about**: the local ACL guarantee does not carry to a remote share.
- `.gitignore` now covers `viewer-config/` and `darling.json`, so an export run inside a working tree cannot be staged by `git add -A`.

### Verified after the fix, same binary

```
service darling.json BEFORE: 331 bytes
--- A: export into the service config directory ---
Refusing to export into ...\1953-live: that would overwrite the SERVICE's own darling.json
(...\1953-live\darling.json) with the viewer's, destroying its monitored servers, encrypted
passwords and tokens. Name a different destination - the default (no argument) is a
viewer-config subfolder beside it.
exit=1  service darling.json AFTER: 331 bytes

--- B: unknown flag ---        Unknown option for --export-viewer-config: --force            exit=1
--- C: --config, no value ---  --config needs a path: ...                                    exit=1
--- D: two destinations ---    --export-viewer-config takes ONE destination directory; ...   exit=1
--- E: foreign darling.json in the destination ---
Refusing to overwrite ...\foreign\darling.json: it was not written by --export-viewer-config.
Export to an empty directory, or move that file aside first.
exit=1
foreign file still says: { "postgres": { "managed": true }, "servers": [ { "name": "MINE" } ] }

--- F: happy path unaffected ---   first export exit=0   re-export exit=0
                                   exported files: darling.json, README.txt, server.crt
--- G: exported secret ACL ---
  NT AUTHORITY\INTERACTIVE : Read, Synchronize
  NT AUTHORITY\SYSTEM : FullControl
  BUILTIN\Administrators : FullControl
  NADAFACCIA\edarl : FullControl
```

Both new guards were watched red first: with the service-config check disabled the test fails and the config is destroyed; with the identity check weakened to any-JSON, the foreign-file test fails on the clobbered content. (One mutation attempt was a no-op worth mentioning — changing `ViewerConfigMarker` moves the writer and the checker together, so it proves nothing. The mutation that bites is weakening the *check* alone.)

`Darling.Tests` **3861 passed / 0 failed / 221 gated-live skipped**, 0 warnings on `-t:Rebuild`.

### Filed, not fixed here

- **#1970** — a relative `Root Certificate` resolves against the **Viewer's working directory**, not the folder holding `darling.json`. The export emits a relative name deliberately (the service host cannot know the viewer machine's path) and documents the absolute-path alternative, but the real fix is viewer-side anchoring in `ViewerSettings`, which changes behavior for every BYO connection string and deserves its own change. The issue also flags that Npgsql's behavior with a *missing* root-certificate file should be pinned by a test rather than assumed.
- **#1964** — two pre-existing xUnit2013 warnings in `Lite.Tests`, from a file this branch does not touch.

### Reviewed and deliberately not changed

Hardening the exported **directory** as well as the file. The file is the secret and it is now ACL'd and verified; the pre-plant and junction vectors that a directory ACL would have addressed are closed above by the ownership and reparse-point guards. A directory ACL would leak only filenames if absent, and applying one to an operator-named folder risks the copy-it-to-the-viewer workflow this verb exists to make easy.

erikdarlingdata:
## Correctness round: the headline instruction did not work

A second review (correctness, not security) caught that the export's **primary advertised workflow was wrong**.

`Root Certificate=server.crt` is a bare name, and Npgsql resolves it against the **Viewer process's working directory** — which for a normally-launched Viewer is its own install folder, not wherever `DARLING_CONFIG` points. So:

- "copy the three files next to the Viewer executable" — works unedited
- "put the folder anywhere and point `DARLING_CONFIG` at it" — works for the JSON, **fails on the certificate**

Every surface led with the second one and promised "nothing to edit". The caveat was documented three layers down, which is exactly where a field report would not find it. Both exported files, the verb's closing STDERR advice, `Darling/README.md` and the CHANGELOG now lead with the placement that works and state the one edit the other route requires. #1970 is the viewer-side anchoring that would remove the condition entirely; until it lands the export must not promise what it cannot deliver. Pinned by a test that asserts the working instruction is offered *before* the conditional one, in both files.

### Also fixed

- **The argument parser moved out of `Program.cs`** into the pure `TryParseExportViewerConfigArgs`, following the file's own stated idiom ("Pure so it pins directly"), with a six-case theory. It was the largest untested inline parser in that file, guarding the case where a wrong answer writes a cleartext password to a path nobody chose.
- **The junction refusal gained a test** — the only one of the three destination guards without one, and the one whose failure mode is silent (a wrong path or a short-circuit would just export through the junction).
- **The rerun test is now load-bearing.** It rotates the *credential* as well as the certificate and asserts the new password lands in the re-exported JSON. Asserting the marker alone would have passed against an implementation that saw its own marker and skipped the rewrite — the marker is already in the file it left alone.
- **Dropped an assertion that could not fail**: `Assert.DoesNotContain(password, stderr)` on the foreign-file guard, which returns *before* the credential is ever decrypted. It would have passed against a version that leaks the credential on every other path. The equivalent assertion on the missing-certificate path, which has decrypted by then, stays.
- Three tests carried DPAPI setup their code path never reaches; `TaggedWriter` now flushes a trailing partial line instead of dropping it; a one-column misalignment in the JSON's certificate table; an interpolated string with no holes; an over-broad "every sibling harden site" claim (`DarlingStoreUpgrade` is not one); a trailing-separator hole in the `.json` destination guard; and the operator-facing docs omitting INTERACTIVE from the ACL they describe.

`Darling.Tests` **3872 passed / 0 failed / 221 gated-live skipped**; `Lite.Tests` **1962 passed / 0 failed**; `-t:Rebuild` clean on both. Rendered output of both exported files re-checked against the rebuilt binary.

erikdarlingdata:
## FIXES IN — all three findings addressed

### 1. Dangling `--config` as the last token — already fixed on HEAD, now tested

Verified before changing anything: the current parser rejects it. The finding was reported against the **pre-extraction** `Program.cs` loop, whose `&& i + 1 < rest.Length` let a trailing `--config` fall through to the positional branch; that condition went away in `0420f446` and the parser moved to the pure `TryParseExportViewerConfigArgs` in `c6b429e7`. What was genuinely missing is the **test**, including the clobber shape you named. Added three cases: `[--config]`, `[D:\handoff, --config]` (a valid destination followed by a dangling flag), and `[--config, C:\c.json, --config]`.

Watched red by restoring that exact `&& i + 1 < rest.Length` shape — all three cases fail; the last two are the clobber, where the good destination is lost or `--config` becomes the destination.

### 2 + 3. All three files now go through one path

Confirmed as asked: **`darling.json`, `server.crt` and `README.txt` all pass through the same two helpers**, no triplication.

- `TryCheckExportTarget(path, marker)` — refuses a reparse point by name, refuses an existing file that is not recognizably a previous export's. Side-effect-free and runs before the config load, so a run that fails later has not thrown away the previous export.
- `WriteExportFileAsync(path, content)` — deletes, then creates with `FileMode.CreateNew`. The delete drops a pre-existing owner (a Windows owner keeps `WRITE_DAC` and could undo any later ACL); `CreateNew` closes the check-to-write gap by failing if anything reappeared at the path instead of writing through it.

Identity per file: `darling.json` and `README.txt` carry their own header lines; `server.crt` is a copy of the store's certificate and carries no marker of ours, so the bar there is that the file is a PEM certificate block at all.

**One deliberate asymmetry, flagged rather than silently kept:** hardening stays `darling.json`-only, because it is the only one of the three holding a secret — the certificate is public by construction and the README has nothing in it. That is a difference in what the files *are*, not the inconsistency you flagged, which was about how they were *written*. Say the word if you want the ACL on all three too.

**A correction worth recording:** my first version of the link check used `File.ResolveLinkTarget` with a comment claiming `File.Exists` follows a link and reports false for a dangling one. That is a Unix assumption — on Windows a dangling symlink reports as **existing**, which my own test caught as a failed precondition. The check now reads the path's attributes, and the comment says what actually happens: without it, the identity read would follow the link, fail, and refuse with a puzzling "could not read" instead of naming the link.

Watched red by reverting the check list to `darling.json` only: the foreign-`server.crt`, foreign-`README.txt` and planted-symlink tests all fail.

### Live, against the rebuilt binary

```
--- 1: dangling --config as the LAST token, after a valid destination ---
--config needs a path: --export-viewer-config [directory] --config <path to darling.json>
exit=1  (destination created? False)

--- 2: a foreign server.crt in the destination ---
Refusing to overwrite ...\foreign\server.crt: it was not written by --export-viewer-config.
Export to an empty directory, or move that file aside first.
exit=1
server.crt still says: this is not a certificate
files in destination: server.crt

--- 3: a foreign README.txt in the destination ---
Refusing to overwrite ...\foreign\README.txt: it was not written by --export-viewer-config. ...
exit=1
README.txt still says: my own notes

--- 4: a SYMLINK planted at the server.crt path, target does not exist ---
Refusing to write ...\linked\server.crt: it is a symbolic link (to ...\elsewhere.crt), so the
real destination is somewhere else. Export to a directory that does not contain one.
exit=1
was anything written through the link? False

--- 5: the happy path and a re-export still work ---
first export exit=0
re-export exit=0
files: darling.json, README.txt, server.crt
```

### Suites, on the merged tree

`dev` had moved (#1954 among them) — merged, CHANGELOG resolved keep-both-sides with both `[Unreleased] Added` entries and both bottom link-refs intact.

- `Darling.Tests` **3898 passed / 0 failed / 222 gated-live skipped**
- `Lite.Tests` **1975 passed / 0 failed**
- `-t:Rebuild` clean on both, 0 warnings

erikdarlingdata:
Dispositioning the two findings from the final review round (both posted on the rerun head) before merge. The failure-path sweep gap is real and minor: the catch removes the secret-bearing darling.json but can leave a mixed new/stale cert+README folder unnamed in the error. The ancestor-junction finding is narrower than written: the catastrophic case (export resolving into the service's own config directory) is already blocked by TryCheckExportTarget, which runs per file through the OS's REAL path resolution - File.GetAttributes and File.ReadAllText follow the whole junction chain, the service's darling.json carries no export marker, and the verb refuses before anything is deleted. The residue is files landing in a directory the junction-planter already controls, with the secret still ACL'd and confirmed wherever it lands. Both are ~15-minute mechanical fixes, tracked with full analysis in #1973 rather than another CI cycle tonight (merge is reboot-blocking). The darling-pg red on the first cycle of this head was the pre-existing midnight seeding window, #1972 - the no-change rerun passed.

PR #1974, opened 2026-08-01
Sweep every exported file on a failed export, and record the guards' ancestor blind spot (#1973)

### LINKED ISSUE #1973: --export-viewer-config: sweep all three files on a failed export; document the ancestor-junction limit of the directory guards
Two low-grade review findings from #1968's final review round, deferred at merge time only because the merge was reboot-blocking; both are ~15-minute mechanical fixes in `Darling/PerformanceMonitor.Darling.Service/DarlingCliCommands.cs`.

## 1. A failed export sweeps only the secret file, leaving a half-finished folder

The catch around the three `WriteExportFileAsync` calls deletes `darling.json` on any failure (correct - it is the secret-bearing file, and the message says so). But if the failing write was `server.crt` or `README.txt`, the files written BEFORE the failure stay behind - and because each write deletes the PREVIOUS export's copy first, the folder can end up as a mix of new and stale files with no config, unnamed in the error. No secret exposure; it just cuts against the verb's premise of refusing to hand over a partial setup.

**Fix:** on failure, best-effort delete all three (the cert and README are always safe to remove), and name in the error exactly what was removed and what, if anything, was left.

## 2. Directory guards do not see a reparse point on an ANCESTOR of the destination

The self-destruct guard (lexical `Path.GetFullPath` compare) and the junction guard (`DirectoryInfo.Attributes` on the leaf) both miss a junction earlier in the chain - e.g. destination `C:\Data\subfolder` where `C:\Data` is a junction. `GetFullPath` is lexical and leaf attributes do not reflect intermediate links.

**Analysis that downgrades this (verified against the code, record it in whatever fix ships):** the catastrophic case the reviewer named - the export resolving back into the service's config directory and destroying the service's own `darling.json` - is ALREADY blocked one layer down. `TryCheckExportTarget` runs per FILE through the OS's real path resolution (`File.GetAttributes` / `File.ReadAllText` follow the whole junction chain), and the service's `darling.json` does not carry the export marker, so the verb refuses with "not written by --export-viewer-config" before anything is deleted. The residual exposure is only "files land in an unexpected directory the junction-planter controls" - and someone who can plant junctions in the ancestor chain of the operator's chosen destination already controls that destination's fate. The exported secret is still ACL'd wherever it lands, and the harden result is confirmed, not assumed.

**Fix:** at minimum a comment on the two guards recording the limitation and naming the marker guard as the backstop (the reviewer's own suggested floor); optionally resolve the chain (`Directory.ResolveLinkTarget` walk or `Path.GetFullPath` on a `FileInfo.LinkTarget`-resolved base) and refuse when the RESOLVED destination differs from the lexical one - but do not let that replace the marker guard, which is the layer that actually holds.

Both were dispositioned on the PR before merge; review context: https://github.com/erikdarlingdata/PerformanceMonitor/pull/1968

erikdarlingdata:
Closes #1973. Both items are the two review findings dispositioned on #1968 but deferred at merge time; all changes are in `Darling/PerformanceMonitor.Darling.Service/DarlingCliCommands.cs`.

## Item 1 - a failed export sweeps all three files and names what survived

The catch around the three `WriteExportFileAsync` calls deleted only `darling.json`. Since each write deletes the previous export's copy before recreating it, a failure on `server.crt` or `README.txt` left a mix of new and stale files with no config beside them, and the error named none of it - which reads to an operator as "nothing was written" when something was.

The catch now sweeps all three through a new `TryRemovePartialExport` helper. Every file still on disk is named, either as removed or as left behind with the OS reason. The `darling.json` line keeps its live-password framing verbatim; the cert and README are phrased as cleanup, because neither carries a secret and neither is usable without the config. Each removal is independent, so a delete that itself fails masks neither the original error nor the other two removals.

One detail worth calling out: the existence probe is `File.GetAttributes` in a try, not `File.Exists` - the idiom `TryCheckExportTarget` already uses ten lines away. `File.Exists` answers **false** for a dangling symlink, which would leave a planted link unmentioned by the one sweep whose entire job is to account for what is on disk.

### The error shape, verbatim from a live run of the failure path

Captured from the rebuilt binary (`ExportViewerConfigAsync` with the cert write forced to fail), stderr as written:

```
WARNING: C:\...\handoff\darling.json will contain a LIVE database password (the 'viewer' role) in cleartext, because that is what the viewer connects with. Copy the folder over a channel you trust, and keep the file ACL'd to the operator account on the viewer machine.
Could not write the viewer config to C:\...\handoff: The process cannot access the file 'C:\...\handoff\server.crt' because it is being used by another process.
Removed C:\...\handoff\darling.json (it held a live password - the whole of one if the write that failed was a later file).
WARNING: C:\...\handoff\server.crt was left behind (The process cannot access the file 'C:\...\handoff\server.crt' because it is being used by another process.) - it is a leftover of the failed export, not a secret. Delete it yourself.
```

(Paths elided to `C:\...` for width; the run used a temp directory. The em dashes in the "held a live password" and "not a secret" clauses are real em dashes in the source, matching this file's existing operator-facing voice. `README.txt` is absent from the output because the run never reached its write - there is nothing on disk to name, which is the correct silence.)

## Item 2 - the directory guards' ancestor blind spot, recorded

Comment only, on both the lexical self-destruct compare and the leaf reparse-point check. `Path.GetFullPath` normalizes text without resolving links, and `DirectoryInfo.Attributes` describes only the last path component, so a junction on an **ancestor** of the destination slips past both.

Deliberately not chased with a chain-resolving walk, and the comment records why: the destructive case the reviewer named - the export resolving back into the service's config directory and destroying the service's own `darling.json` - is already refused one layer down. `TryCheckExportTarget` runs per file through the OS's real path resolution (`File.GetAttributes` and `File.ReadAllText` follow the whole junction chain to the actual file), and the service's own `darling.json` carries no export marker, so such a run stops with "not written by --export-viewer-config" before anything is deleted. What survives the gap is only "the files land in a directory the junction-planter chose" - and planting a junction in the ancestor chain of the operator's destination already means controlling that destination's fate.

## Test

`ExportViewerConfigAsync_FailsPartWayThrough_SweepsAllThreeAndNamesWhatSurvived` in `Darling/Darling.Tests/DarlingExportViewerConfigTests.cs`.

The failure is forced where it actually bites: after `darling.json` has been written and hardened, on the `server.crt` write. A pre-existing (marker-carrying) cert is held open with `FileShare.Read` - readable, so the pre-flight identity check passes and the run reaches the writes at all, but not deletable, so `WriteExportFileAsync`'s delete-then-`CreateNew` fails with the same `IOException` a live antivirus or an open editor produces. No stub, no seam injection; this is the production write path.

Asserts exit 1, that `darling.json` is gone and named as the secret it was, that the surviving `server.crt` is named as left behind (and still holds its stale content), that `README.txt` was never reached, and - the claim the whole sweep rests on - that **every** file remaining in the folder appears by name in stderr, enumerated from `Directory.GetFiles`. Also that the password reaches neither stream on the failure path.

### Watched red

With the source change reverted and the test project rebuilt (not `--no-build` against a stale DLL), the new test fails on exactly the old behavior:

```
[FAIL] ExportViewerConfigAsync_FailsPartWayThrough_SweepsAllThreeAndNamesWhatSurvived
Assert.Contains() Failure: Sub-string not found
DarlingExportViewerConfigTests.cs:line 845
```

Line 845 is `Assert.Contains($"{exportedCert} was left behind", stderr, ...)` - the old code removed `darling.json` and left the cert behind unnamed. Source change restored, retested green.

## Gates

- Build: `Build succeeded. 0 Warning(s) 0 Error(s)` (checked the warnings line, not just the exit code).
- Test project rebuilt before every run.
- `DarlingExportViewerConfigTests`: 39 passed, 1 skipped, 0 failed. The skip is the pre-existing symlink test, which needs Developer Mode or elevation.
- Full ungated `Darling.Tests`: **3898 passed, 0 failed, 223 skipped** (the live-Postgres gated classes).
- CHANGELOG entry under `[Unreleased]` > `Fixed` with the bottom link-ref.

Windows-only conventions followed: `Assert.SkipUnless(OperatingSystem.IsWindows(), "DPAPI requires Windows.")`, temp subdirectory with `finally` cleanup, `StringWriter` capture, exit-code asserts.


erikdarlingdata:
## CI: all four checks green, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`

```
build                    SUCCESS   3m3s
check-branches           SUCCESS   2s
review                   SUCCESS   3m52s
Darling PostgreSQL tests SUCCESS   3m16s
```

Checked `mergeable` immediately after opening (MERGEABLE, so CI actually ran) and again after settle - still MERGEABLE / CLEAN, so nothing on `dev` has moved under the CHANGELOG entry yet. Three sibling PRs are landing tonight; if one merges first this will need a keep-both-sides resolve on `CHANGELOG.md` and nothing else, since the two code files are touched by no other PR in flight.

### On the review bot's one non-blocking note

It flags a TOCTOU gap between the `File.GetAttributes` probe and the `File.Delete` in `TryRemovePartialExport`, reasoning that a file vanishing in between would land in the generic `catch (Exception cleanupFailure)` and produce a misleading "left behind" warning.

The mechanism is not right: `File.Delete` is documented as a **no-op** for a path that does not exist and does not throw `FileNotFoundException`, so the race cannot reach that catch. The actual residual is the opposite and smaller - we would print `Removed <path>` for a file some other process removed a moment earlier, which is a true statement about the end state and a false one about who did it. On a single-process CLI cleanup path that is not worth a lock or a re-probe, and adding either would make the block less clear than the thing it guards against. Leaving as written.

PR #1976, opened 2026-08-01
Anchor live compression-test chunk seeds to midday (#1972)

### LINKED ISSUE #1972: Live compression tests have two ~200-second midnight blind windows from now()-relative chunk seeding
Three live compression tests seed chunks with `now()`-relative timestamps, which makes chunk PLACEMENT depend on the time of day the suite runs. There are two ~200-second windows around midnight UTC where the tests fail deterministically - not flaky in the reused-database sense, just blind windows the clock walks through every day.

## Evidence (caught live tonight)

- dev run 30673966983 (head 3695dcf7): **green**, darling-pg started 23:50 UTC 2026-07-31.
- PR #1968 run 30674344636 (head a1cc3d27, byte-identical `TimescaleSupportTests.cs`): **failed** at 00:01:38 UTC 2026-08-01.

`EndToEnd_CompressionRun_TakesEligibleChunksAndLeavesYoungOnes_AgainstDevPostgres` - `Assert.Equal()` expected 1, actual 2, at `TimescaleSupportTests.cs:740` (the UNcompressed chunk count). A rerun outside the window passed with no code change.

## Mechanism

The tick tables are day-aligned 1-day chunks. The young-chunk seed is:

```
INSERT ... SELECT now()::timestamp - (g || ' seconds')::interval ... generate_series(1, 200)
```

which writes rows spanning `[now-200s, now-1s]`. Run between **00:00:00 and 00:03:20 UTC**, that span straddles the day boundary, so the "one young chunk" is actually TWO chunks (yesterday's tail + today's). Total chunks = 3, uncompressed = 2, and line 740 fails exactly as observed.

The old-chunk seeds have the symmetric problem in the other direction:

```
INSERT ... SELECT now()::timestamp - INTERVAL 'N days' + (g || ' seconds')::interval ...
```

spans `[now-Nd+1s, now-Nd+200s]`. Run between **23:56:40 and 23:59:59 UTC**, the `+200s` end crosses midnight N days back, splitting the old data into two chunks - both closed, both past the eligibility delay, so the policy compresses BOTH and the compressed-count assert (line 739) fails instead.

## Affected sites (all in `Darling/Darling.Tests/TimescaleSupportTests.cs`)

| Line | Test | Window |
|---|---|---|
| 735 | `EndToEnd_CompressionRun_TakesEligibleChunksAndLeavesYoungOnes` | 00:00:00-00:03:20 (hit tonight) |
| 733 | same test, old-chunk seed | 23:56:40-23:59:59 |
| 798 | `EndToEnd_CompressionRun_OneUncompressibleChunkDoesNotBlockTheRest` (parameterized `daysBack`) | 23:56:40-23:59:59 |
| 986 | `EndToEnd_CompressionActivity_ReportsBacklogAndClearsAfterARun` | 23:56:40-23:59:59 |

Nightly scheduled darling-pg runs and evening PR pushes can both land in these windows; tonight's did.

## Fix direction

Anchor the seeds to midday instead of `now()`:

```
date_trunc('day', now()) - INTERVAL 'N days' + INTERVAL '12 hours' + (g || ' seconds')::interval
```

A +/-200s span around midday can never cross a day boundary, chunk placement becomes deterministic for any run time, and no assertion changes are needed. The young chunk keeps its "still open, still young" property (its chunk is today's) - worth one careful look at the `range_end > now() - CompressAfterDays` assert at line 749 while touching it, but midday-today satisfies it with hours to spare.

Mechanical fix, four lines plus comments; the only care point is keeping the young/old semantics each test pins.

erikdarlingdata:
Fixes #1972.

## The mechanism

The three live compression tests seed their chunks with `now()`-relative timestamps spanning 200 seconds, into chunks that are day-aligned (`ChunkIntervalDays = 1`). That makes chunk PLACEMENT a function of what time of day the suite runs, and there are two windows per day where the 200-second span crosses a day boundary and one intended chunk becomes two:

- **00:00:00-00:03:20** - the young seed `[now-200s, now-1s]` straddles midnight, so the "one young chunk" is yesterday's tail plus today's. Uncompressed count asserts 1, gets 2. This is the failure caught live: `dev` run 30673966983 green at 23:50 UTC, PR run 30674344636 carrying a byte-identical `TimescaleSupportTests.cs` failed at 00:01:38 UTC at line 740, and a rerun outside the window passed with no code change.
- **23:56:40-23:59:59** - an old seed `[now-Nd+1s, now-Nd+200s]` crosses the boundary N days back into two chunks that are BOTH closed and BOTH past the eligibility delay, so the policy compresses both. Compressed count asserts 1, gets 2.

Deterministic inside the window, green outside it - not the reused-database flake that moves between unrelated classes.

## The four sites

All in `Darling/Darling.Tests/TimescaleSupportTests.cs`:

| Test | Seed | Window |
|---|---|---|
| `EndToEnd_CompressionRun_TakesEligibleChunksAndLeavesYoungOnes` | old, 5 days back | 23:56:40-23:59:59 |
| same test | young | 00:00:00-00:03:20 |
| `EndToEnd_CompressionRun_OneUncompressibleChunkDoesNotBlockTheRest` | old, `daysBack` 7/5/3 | 23:56:40-23:59:59 |
| `EndToEnd_CompressionActivity_ReportsBacklogAndClearsAfterARun` | old, 5 days back | 23:56:40-23:59:59 |

Each gets a comment stating why midday, citing #1972.

## Why midday works

A 200-second span centred 12 hours from either boundary cannot reach one. The anchor depends only on which DAY it is, not the time of day, so for any two runs on the same day the seeded values are identical - the placement is time-independent by construction, not merely unlikely to break.

One deviation from the issue's suggested expression, and it is deliberate: `date_trunc('day', now()::timestamp)` rather than `date_trunc('day', now())`. The column is `timestamp` (naive) and the code being replaced was `now()::timestamp`, so truncating the naive value keeps every step in the domain the column actually is. The timestamptz form would take an implicit cast back on insert whose result depends on session `TimeZone`, and its `- INTERVAL 'N days'` step is wall-clock arithmetic that shifts across a DST transition. Same midday anchor, no zone or DST reasoning needed.

## The line-749 semantics check

The young assert - `NOT is_compressed AND range_end > now() - INTERVAL 'CompressAfterDays days'`, with `CompressAfterDays = 1` - still holds, and it is not close. Midday today sits in today's chunk, whose `range_end` is tomorrow's midnight; that is greater than `now() - 1 day` at every hour of the day. The chunk is also still the OPEN one (it contains `now()`, so it is where live inserts would land), which is the property the test's doc comment pins. Measured on the rig rather than argued: `young_line749 = 1` for the midday scenario.

The same probe found something worth recording - in the BUGGY young case this predicate still returns 1, because yesterday's tail chunk has `range_end` = today's midnight and falls outside it. So only the line-740 count fails, never line 750. That matches the reported failure exactly, which is a fidelity check on the repro.

## Verification 1: the mechanism, proven on a live rig

A throwaway repro (kept in the scratchpad, not committed) mirrors `CreateTickTableAsync` / `AddCompressionPolicyParkedAsync` / `RunPolicyAsync` / `ChunkCountAsync` exactly, but substitutes literal anchors for `now()` so both windows can be ENTERED on demand instead of waited for. Run against a live PostgreSQL 18.4 + TimescaleDB 2.28.1 cluster at 20:50 local, nowhere near either window:

```
     scenario      | compressed | uncompressed | young_line749
-------------------+------------+--------------+---------------
 repro_buggy_old   |          2 |            1 |             1
 repro_buggy_young |          1 |            2 |             1
 repro_fixed       |          1 |            1 |             1
```

`repro_buggy_young` is the reported failure reproduced: expected 1, actual 2 on the uncompressed count. `repro_buggy_old` is the other direction: expected 1, actual 2 on the compressed count. `repro_fixed` passes both, plus the young assert.

Then the fix's time-independence, swept over every minute of the day:

```
 buggy_young | buggy_old | fixed_young | fixed_old
-------------+-----------+-------------+-----------
           3 |         3 |           0 |         0
```

Three splitting minutes per day for each `now()`-relative form (`00:01`, `00:02`, `00:03` for the young seed; `23:57`, `23:58`, `23:59` for the old), and **0 of 1440** for both midday forms.

## Verification 2: the gated-live suite

Local rig per the recipe - own port (5573) and data dir verified mine via `pg_ctl status` before any database was created, superuser `darling`, `shared_preload_libraries = 'timescaledb'`, and the worker sizing re-derived from `build.yml` rather than pasted (`timescaledb.max_background_workers = 41`, `max_worker_processes = 52`). Connection string matched CI's shape exactly. Probe SQL ran in a SECOND database, never the suite's.

- `TimescaleSupportTests`: **29 passed, 0 failed, 0 skipped** - zero skips confirms the gated-live tests actually ran.
- Full `Darling.Tests`: **4109 passed, 0 failed, 11 skipped** in 38s. The 11 skips gate on a live SQL Server and on `DARLING_TEST_PGRUNTIME`, neither of which I set.
- `dotnet build` on `Darling.Tests`: **0 Warning(s), 0 Error(s)**; the test project was rebuilt before running, so `--no-build` ran the fresh DLL.

## Scope check

The other three files that assert chunk counts are NOT exposed and are deliberately left alone: `RetentionChunkGeometryLiveTests` seeds from fixed literal dates (`"2026-01-01"`), and `DarlingStoreUpgradeTests` seeds continuously across ~14 and ~21 days and compares a before/after snapshot of the same data across a `pg_upgrade`, so a boundary shift moves both sides identically.

Test-suite only - no shipped code changed.


erikdarlingdata:
**CI green on the merged tree.** `statusCheckRollup` on head `4f294b45`, all four checks that ran:

| Check | Conclusion | Duration |
|---|---|---|
| build | SUCCESS | 3m9s |
| check-branches | SUCCESS | 4s |
| review | SUCCESS | 2m1s |
| Darling PostgreSQL tests | SUCCESS | 3m29s |

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

Sibling #1974 (issue #1973) landed on `dev` while the first run was finishing and put this PR into `CONFLICTING` on the CHANGELOG link-ref block, which would have silently stopped CI from running at all. Resolved keep-both-sides - both `[#1972]` and `[#1973]` link-refs survive, both Fixed entries survive, blank lines kept - then re-run from scratch, which is the green above. The prose entries auto-merged; only the link-ref line collided.

**One honest note on what CI does and does not prove here.** This run executed around 01:25 UTC, outside both blind windows, so a green run is not by itself evidence the windows are closed - a green run at this hour is exactly what the BUGGY code produced too. The evidence that they are closed is the rig work in the PR body: both windows entered on demand with literal anchors reproduce the failures against the current code, the midday form passes in the same conditions, and the 1440-minute sweep shows the midday anchor splitting on zero minutes of the day versus three for each `now()`-relative form. The fix is time-independent by construction - the seed value depends only on which day it is, so two runs on the same day seed byte-identical rows.

Issue #1956, opened 2026-07-31
Portable Viewer failed to connect to a store the Setup-installed Viewer (and an IDE) reached - investigate the packaging difference


erikdarlingdata:
Field report, not yet reproduced locally: a state existed where a generic IDE/database tool could connect to the Darling Postgres store and the PORTABLE Viewer could not - while the Setup-installed Viewer just worked, same machine, same store.

The two artifacts genuinely differ: the nightly builds the Viewer twice - a framework-dependent publish for the ZIP and a SELF-CONTAINED build for the Velopack Setup - so runtime-level differences are structurally possible. Leading suspects to check first:

- Native/TLS dependency differences between the FDD portable and the self-contained build (certificate validation path, ICU, OpenSSL-vs-SChannel behavior under Npgsql's sslmode handling).
- Working-directory or path-resolution differences affecting how a relative Root Certificate path in the connection string resolves (a portable run from a Downloads folder resolves relative paths differently than an installed app - which would also explain the IDE connecting fine with an absolute path).
- Config discovery: whether both builds resolve the same darling.json (DARLING_CONFIG vs Roaming default) - #1954's path logging will make this class diagnosable in the field; until then this needs a side-by-side repro.

Repro shape: same machine, same darling.json, both artifacts from one nightly, capture #1954-style diagnostics from each. If the relative-path suspect confirms, the fix likely belongs in path resolution (resolve Root Certificate relative to the config file's directory, not the process working directory) - which would be a real bug fix, not a packaging note.

erikdarlingdata:
[...]

PR #1977, opened 2026-08-01
Anchor a relative Root Certificate to darling.json's directory (#1970)

### LINKED ISSUE #1970: Viewer: anchor a relative `Root Certificate` to darling.json's directory, not the process working directory
Split out of #1953's security review rather than folded into PR #1968, because it changes **Viewer** behavior for every bring-your-own connection string, not just exported ones - that is a decision worth its own change.

## The gap

`ViewerSettings.Parse` returns a BYO `postgres.connectionString` **verbatim** (`Darling/PerformanceMonitor.Darling.Viewer/ViewerSettings.cs:134-141`), so Npgsql resolves a relative `Root Certificate=server.crt` against the **Viewer process's working directory** - which is not necessarily the folder holding `darling.json` and the certificate beside it. A desktop shortcut's "Start in", or launching from Explorer's last-used directory, silently changes the anchor.

That makes the working directory the trust anchor for a verify-full pin. If the Viewer is ever launched with a working directory an attacker can write (Downloads, a shared folder), a planted `server.crt` becomes the pinned root; combined with LAN position that is interception of the store session and the credential inside it.

`--export-viewer-config` (#1953) deliberately emits the relative name - the service host cannot know the viewer machine's absolute path - and documents the caveat and the absolute-path alternative in three places. So this is a sharp edge that is currently *documented*, not *removed*.

## The fix

In `ViewerSettings.TryLoad`, when the parsed BYO connection string's `Root Certificate` is a relative path, rewrite it to `Path.Combine(Path.GetDirectoryName(resolvedConfigPath)!, value)`. The Viewer already knows the config path it loaded, so "the certificate beside darling.json" becomes true regardless of working directory, the exported folder keeps working when copied anywhere, and an absolute path set by hand is left alone.

## Also worth pinning while in there

What Npgsql 10 does when the configured `RootCertificate` file does **not exist**. The shipped assembly's metadata suggests an existence guard with a fall-through to the machine trust store, which would mean VerifyFull validating against public CAs instead of the pinned PEM. A self-signed store certificate would not chain to a public root, so the practical outcome is probably a connection failure - but "probably" is not the standard for a pinning control, and a test should settle it.

erikdarlingdata:
Fixes #1970.

## The defect

`ViewerSettings.Parse` returned a bring-your-own `postgres.connectionString` verbatim, and Npgsql hands a `Root Certificate` value to the file system as given - so a relative one resolved against the viewer **process's working directory**, which is not necessarily the folder holding `darling.json` and the certificate exported beside it. A desktop shortcut's "Start in", or launching from Explorer's last-used folder, silently changed which file a `SSL Mode=VerifyFull` connection pinned.

That makes the working directory the trust anchor for the pin. Launched with a working directory an attacker can write (Downloads, a shared folder), a planted `server.crt` becomes the pinned root; combined with LAN position that is interception of the store session and the credential inside it.

## One resolver, two consumers

There are now two places that must agree about which certificate file Npgsql will open: the connection string itself, and the #1966/#1954 startup diagnostics, which print "the absolute path Npgsql will actually open". A diagnostic that computes its own answer is worse than none once the two drift, so this adds a single implementation of the rule and routes both through it.

**`Darling/PerformanceMonitor.Darling.Viewer/ViewerCertificateAnchor.cs`** (new) - pure and total, nothing here throws, because both callers sit on the viewer's startup path where an exception is a window that never opens.

- `Resolve(rootCertificate, configDirectory)` -> the absolute path, or null when there is no value or it cannot be expressed as a path. Absolute passes through; a bare name or relative subpath anchors to `configDirectory`; a null anchor falls back to the working directory, which is what Npgsql itself would do (the honest report rather than an invented path).
- `Anchor(connectionString, configDirectory)` -> a thin wrapper: the same string except a relative `Root Certificate` rewritten to `Resolve`'s answer.

**Rewrite sites - every place the resolved path is produced:**

| Site | What it does |
| --- | --- |
| `ViewerSettings` private constructor | Applies `Anchor`. Deliberately the constructor and not a parse branch: it is the one choke point every construction passes through, so no branch can produce settings whose certificate still hangs off the working directory. |
| `ViewerSettings.TryLoad` | Now resolves via `ResolveConfigLocation` (not `ResolveConfigPath`) so the file it reads and the folder it anchors to are the same resolution. |
| `ViewerSettings.Parse(json, configDirectory)` | New overload; the existing string-only `Parse` keeps its unanchored behavior for callers with no file behind the JSON. |
| `ViewerConfigLocation.Directory` | New - the anchor, defined **once**, so `TryLoad` and the diagnostics do not each derive it. |
| `ViewerConfigDiagnostics.DescribeRootCertificate` | Calls `ViewerCertificateAnchor.Resolve` instead of doing its own `Path.GetFullPath`; the `workingDirectory` parameter is renamed `configDirectory` throughout because the semantics changed. |
| `MainWindow.OnLoaded` (3 call sites) | Every diagnostics call now carries `configLocation.Directory`. The post-load calls summarize `settings.ConfiguredConnectionString` (new property - the string as the file carried it) so the block still shows the value the operator **wrote** alongside the absolute path it resolves to. |

**Rules honored:** absolute passes through untouched; the keyword is never ADDED to a string that lacked it (a connection that was not pinning is not made to pin); an unparseable string is returned verbatim for Npgsql to reject in its own words. **No `SSL Mode` gate** - `Root Certificate` is only consulted under `VerifyCA`/`VerifyFull`, so anchoring it under `Require` changes nothing Npgsql reads, while gating would leave a stale working-directory anchor behind for an operator who later tightens the mode, which is the one moment the pin has to be right.

Verified against the shipped Npgsql 10.0.3 by reflection: `Root Certificate` and `rootcertificate` are both accepted in any casing (`sslrootcert` is **not** an alias), and `NpgsqlConnectionStringBuilder.ConnectionString` re-emits only the keys that were set - no defaults injected - so the rewritten string differs from the operator's in the certificate value alone.

## Export-docs ripple

`--export-viewer-config` emits a bare `Root Certificate=server.crt` (the service host cannot know the viewer machine's absolute path) and documented the working-directory caveat in four places, telling operators to hand-edit it to a full path when the folder lived anywhere but beside the Viewer executable. **That caveat is now obsolete** - a bare name beside `darling.json` just works, so `DARLING_CONFIG` pointed at an out-of-the-way folder needs no edit at all.

Updated: `BuildViewerConfigJson`'s comment block, `BuildViewerConfigReadme`, the verb's STDERR epilogue, the `--print-viewer-connection` placeholder comment, and `Darling/README.md` (the troubleshooting entry, the export section, the certificate-placement bullet, and the viewer startup-summary description). All texts now carry one canonical unsplit phrase, "the folder holding darling.json", so a line wrap cannot break the pin.

`DarlingExportViewerConfigTests` pinned the old text, so those expectations flipped: `ExportedDocs_LeadWithTheInstructionThatWorksUnedited_AndQualifyTheOtherOne` became `ExportedDocs_LeadWithTheSimplestPlacement_AndNoLongerCarryTheWorkingDirectoryCaveat`, asserting the caveat's **absence** (both spellings, plus "change Root Certificate") and the anchor rule's presence. A documented workaround that is no longer true is worse than none: it gets you to hardcode a path that breaks the next time the folder moves.

## Tests

New in `ViewerConfigDiagnosticsTests` (following the suite's idiom, including the glob over the `MainWindow` partial family):

- Resolver behavior: absolute passthrough, bare name, relative subpath, no-anchor fallback, null/whitespace.
- `Anchor`: no-`Root Certificate` passthrough (and never adds the keyword), absolute passthrough, unparseable passthrough, and the rewrite preserving `VerifyFull` plus host/port/user/database/search-path/password.
- `TryLoad_AnchorsARelativeCertificateToTheDirectoryOfTheFileItActuallyRead` - the end-to-end behavior change.
- `TheDiagnosticsCertificateLine_IsThePathTheConnectionStringActuallyCarries` - functional anti-drift pin across all three value shapes.
- `BothConsumersOfTheCertificatePath_RouteThroughTheOneResolver` - structural: the diagnostics must call `Resolve` and must contain no `Path.GetFullPath` of its own; `ViewerSettings` must call `Anchor`.
- `EveryDiagnosticsCallInTheShellCarriesTheConfigDirectoryAsItsAnchor` - glob-scans `MainWindow*.cs` and requires an anchor argument for every diagnostics call, since an omitted one silently falls back to the working directory.

Existing `DescribeConnection_AbsoluteRootCertificate_...` was strengthened to pass a config directory **different** from the certificate's, so an anchor wrongly applied to an absolute path would move it somewhere the assertion can see.

### Watched-red

Test project rebuilt before each run (stale-DLL trap).

**Mutation A** - `Anchor` returns the connection string verbatim (the exact pre-fix behavior): **3 red**.

```
Failed TryLoad_AnchorsARelativeCertificateToTheDirectoryOfTheFileItActuallyRead
  Assert.Equal() Failure: Strings differ
  Expected: "C:\\Users\\edarl\\AppData\\Local\\Temp\\darling-viewer-t"...
  Actual:   "server.crt"
Failed Anchor_RelativeRootCertificate_RewritesOnlyTheCertificate
Failed TheDiagnosticsCertificateLine_IsThePathTheConnectionStringActuallyCarries
```

**Mutation B** - `Resolve` ignores the anchor and uses `Path.GetFullPath(rootCertificate)`, the old working-directory rule: **4 red**, including `DescribeConnection_RelativeRootCertificate_ResolvesAgainstTheConfigDirectory_AndReportsExistence` and `Resolve_BareNameAndRelativeSubpath_BothAnchorToTheConfigDirectory`.

Worth stating: the agreement pin stays **green** under Mutation B, because both sides move together. That is the honest limit of a functional agreement pin, and why the behavior tests sit beside it.

### Results

Full `Darling.Tests` ungated: **3908 passed, 0 failed, 223 skipped** (the gated live-Postgres suites). Viewer, Service and test projects all build **0 Warning(s), 0 Error(s)**.

## Note on #1956

Not touched or closed here (the team lead owns it), but this is the likely mechanism behind it: a portable Viewer run from a different working directory is exactly the case where the bare `Root Certificate` stopped resolving.

Issue #1970's secondary question - what Npgsql 10 does when the configured `RootCertificate` file does not exist (an existence guard falling through to the machine trust store, versus a hard failure) - is **not** settled here. It needs a live TLS store to answer honestly rather than by reading metadata, so it stays open under #1970.


erikdarlingdata:
Taking the review's one finding rather than dispositioning it away: Anchor's builder construction was guarded but the RootCertificate setter and ConnectionString re-serialization were not, while the class doc claims nothing here throws - and that claim is load-bearing, since both callers sit on the viewer's startup path where an exception is a window that never opens. The guard now covers them with the same contract as the parse fallback: on any failure the operator's string reaches Npgsql unchanged. Realistically non-throwing for a GetFullPath-produced value, but the invariant is now enforced by code rather than promised by comment. Built clean, ViewerConfigDiagnosticsTests 26/26 on a rebuilt test DLL. This restarts CI; merging off the fresh rollup.

Issue #1563, opened 2026-07-18
Darling: custom views/notebooks (Datadog-style composition) on the web surface


erikdarlingdata:
Services like Datadog win on composition UX — easy custom views/notebooks — while their SQL Server collection is shallow. Darling has the opposite shape: deep collectors, fixed views. Close the gap on top of the Darling Web surface (prerequisite issue):

- A **view definition** is a stored JSON document in the config schema: `panels: [{ read, params, viz, span }]`, where `read` names one of the catalog reads the MCP data tools already expose (get_wait_trend, get_top_queries_by_cpu, get_memory_trend, ...). No new query engine — reuse the exact reads.
- The web UI renders/edits the panel grid; definitions are per-store (shared by every seat), versioned rows, exportable JSON.
- The viewer (WPF) can later render the same definitions if it earns its keep; web-first.
- Longer-term: with the MCP surface adjacent, "ask AI to draft a view for this incident" is a small step once definitions exist (the model emits the same JSON).

Depends on: Darling Web phase 1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
Shipped to dev in #1576. Custom views/notebooks on the Darling Web surface: compose named, shareable, versioned dashboards from the ~69 read-only reads (table/line/stat/bandlist), stored as JSON and rendered through the same renderPanel seam the built-in pages use. V31 config.custom_views with a narrow single-table viewer grant; server-side ValidateDefinition authority; catalog + CRUD API. Editing is available to any authenticated seat (networked = normal mode) behind the token->cookie + CIDR gate, CSRF-hardened (always-on host-header allowlist, SameSite cookie, application/json). Independently design-reviewed (1 blocking + 5 major all fixed) and security-reviewed (2 majors hardened). The composer: read/param/viz pickers with a live preview beside the config, auto-derived fields, export/import. Verified live on DARLING01 incl. networked create/edit/delete from a remote seat.

Issue #1963, opened 2026-07-31
Decision: move deadlocks collection from the 2-minute to the 6-minute tier - the query is a measured no-op, cadence is the only lever


erikdarlingdata:
Field decomposition on the alpha box (n=3 each): DATALENGTH-only probe 119 ms, + TRY_CAST/materialize 218 ms, full collector query with shred 167 ms - all one noise band, measured while the ring buffer held 284 BYTES. The entire ~273 ms fleet-average cost is fixed dm_xe_session_targets serialization overhead, invariant to buffer content. A change-detection pre-probe pays the same fixed cost as the real read, so there is NO query-side win available - the collector query itself should be left alone.

The only real lever is cadence: the ring buffer retains events between polls, so moving deadlocks from the 2-minute to the 6-minute tier keeps the same data (the watermark catches up) at 3x fewer runs - roughly 2.3 SQL-hours/day back fleet-wide - trading up to 6 minutes of deadlock-detection latency. That is a policy call about alert latency, not a defect, which is why this is filed as a decision rather than a fix. The one loss scenario: a deadlock storm large enough to cycle the ring buffer between 6-minute polls drops events that a 2-minute poll would have caught - and the buffer's own capacity bounds that exposure either way.

Erik decides: take the 3x cadence saving, keep the 2-minute latency, or split the difference. Per the defaults-over-config rule this should be a DEFAULT change, not a knob.

erikdarlingdata:
Ruled by Erik 2026-08-01 and shipped in #1978: deadlocks collection defaults to the 5-minute tier in both apps. Five rather than the six the issue sketched, deliberately - the product's tier vocabulary is 1/5/60/1440, and 5 is where the sibling event-buffer readers (system_health_events, default_trace_events) already sit; it strictly dominates six on latency at the same order of savings, with no one-off tier invented. Note the shipped default was actually EVERY MINUTE (the issue's 2-minute figure was the field fleet's own scheduling), so default installs save 5x the fixed ~273ms serialization cost per server. The schedule-editor presets map deadlocks the way they map those siblings (Aggressive 2 / Balanced 5 / Low-Impact 15), a new test pins the ruled value so a quiet edit cannot regress the decision, and the CHANGELOG states who picks it up when: Darling stores without an override row converge automatically on upgrade; Lite installs keep their stored user-editable schedule and move on preset apply, edit, or reset-to-defaults. The field fleet keeps its own override until they change it - worth one line in the next client note.

PR #1979, opened 2026-08-01
Docs: promote remote-Viewer setup to its own section, audit every grant snippet against provisioning (#1955)

### LINKED ISSUE #1955: Docs: remote Viewer setup gets its own top-level section, grant snippets audited against provision-roles.sql, Root Certificate values documented
Field feedback on the docs, verified against the current README:

1. **The remote Viewer setup is buried.** The README has one top-level 'Darling Edition (service + viewer)' section and the viewer-only path lives so deep that GitHub's table of contents does not surface it. Someone whose ONLY goal is pointing a Viewer at an existing service has to read service-setup material to find their three steps. Promote 'Connect a remote Viewer' to its own ToC-visible section, written for the person who has nothing installed on their machine yet, with the service-side prerequisites stated as one line and a pointer.
2. **The reporter says the permission-grant code block is out of date.** Audit EVERY grant snippet in the docs against what the code actually provisions (Darling/tools/provision-roles.sql and DarlingManagedRoles are the truth for the PG side; the SQL Server monitoring-login block near README line 443 gets the same check against what the collectors actually require). Where a snippet exists in both places, state which one is authoritative and make the other a pointer - duplicated grant lists are how this went stale.
3. **Document the valid 'Root Certificate=' values** - what the field accepts, where the service writes server.crt, and what a bring-your-own-PostgreSQL setup should put there instead. (#1953's exported config self-documents this at the file level; the docs still need the reference form.)

Item 2 is a correctness fix and should not wait on the restructure if they end up different sizes.

erikdarlingdata:
Closes #1955.

Docs-only (three markdown files). No `.cs` touched, so no pin updates were needed — see "Pinned-text check" below for why.

## Scope: three of the issue's complaints were already shipped tonight

#1955 was filed from field feedback before tonight's fixes landed, and three of its four sources are now code:

- **#1966 (viewer startup diagnostics)** — the viewer already logs which resolution rule won, the absolute path, and a non-secret parse summary, and renders the same block in the failure window with **Copy details**. The issue's implied "the viewer says nothing about what it read" is gone.
- **#1968 (`--export-viewer-config`)** — the verb writes the whole viewer handoff folder, and the exported `darling.json` + `README.txt` are now the field reference. This is what turned the docs restructure from "reorder some prose" into "lead with the verb."
- **#1970 (certificate anchoring)** — a relative or bare `Root Certificate` anchors to the folder holding the `darling.json` the viewer read. **The hand-edit-to-an-absolute-path caveat is dead**, and I swept the living docs for survivors of it (none in prose; the remaining mentions offer absolute as one valid *option*, which is still true). CHANGELOG history entries were deliberately left alone.

What remained, and what this PR does: the restructure (item 1) and the grant audit (item 2), plus the `Root Certificate=` reference form (item 3) written against the post-#1970 behavior.

## 1. Restructure

`Darling/README.md` gains a top-level, ToC-visible **`## Connect a Remote Viewer`**, placed between Operations and Troubleshooting, written for a machine with nothing installed on it. It **leads with the export verb** — the three steps are literally: run `--export-viewer-config` on the service host, copy the folder next to the Viewer exe, start the Viewer. The service-side prerequisite (the store must be LAN-exposed) is one line with a pointer, per the issue.

Manual configuration is demoted to **`### Manual configuration (fallback)`** inside that section: `--print-viewer-connection`, the minimal `darling.json`, the `managed: false` explanation, and the new `Root Certificate=` reference table (bare name / relative subpath / absolute path / omitted, each with what it resolves to), where the service writes `server.crt` and what rotates it, and what a BYO PostgreSQL points the field at instead (its own CA or self-signed server certificate — the file `psql` takes as `sslrootcert`).

Two places that described the manual path as the *only* path were fixed rather than left:

- `### Store endpoint (viewer over the LAN)` kept ~45 lines of handoff material that is now duplicated by the new section. It keeps the **service-side** config (network block, role, TLS, firewall) and points at the new section for the viewer side.
- `### The Viewer` said "a viewer seat on another machine needs only a minimal `darling.json` containing the `postgres.connectionString`" — the manual path stated as the setup. Now points at the new section.

All internal anchors were checked programmatically after the move: **0 broken** across both READMEs (including the cross-file `../README.md#lite--darling-on-premises` added below).

## 2. Grant audit — every snippet, with before/after

Audited against `Darling/tools/provision-roles.sql`, `DarlingManagedRoles.BuildProvisioningSql`, and `grep -rn "GRANT" Darling/PerformanceMonitor.Darling.Storage Darling/PerformanceMonitor.Darling.Service`.

### 2a. `ALTER SETTINGS` was missing from BOTH SQL Server blocks (the real correctness bug)

The Darling guide's per-grant table listed `ALTER SETTINGS` as the grant behind the blocked-process-threshold bootstrap — but **neither** grant block granted it. `DarlingXeSessions.EnsureBlockedProcessOnPremAsync` really does run `sp_configure` + `RECONFIGURE`, which requires it. So a login built by copy-pasting either block could not do the bootstrap, and blocked process reports stayed empty for a reason the docs described but never granted. Two copies, both stale in the same way — exactly the failure mode the issue names.

**Before** (root `README.md`, and a near-identical copy in `Darling/README.md`): `VIEW SERVER STATE`, `CONNECT ANY DATABASE`, `VIEW ANY DEFINITION`, `ALTER ANY EVENT SESSION`, `ALTER TRACE`, msdb job tables + `agent_datetime`.

**After** (root `README.md`, now the stated authoritative copy):

```sql
/* Optional: lets the app bootstrap 'blocked process threshold (s)' to 5 when it is still 0.
   sp_configure + RECONFIGURE require ALTER SETTINGS; without it the bootstrap is logged and
   skipped, blocked process reports stay empty, and you set the threshold yourself (or via an
   RDS Parameter Group). Blocking stays visible either way through the DMV blocking snapshot. */
GRANT ALTER SETTINGS TO [YourLogin];
```

Verified live on SQL Server 2022 (16.0.4255.1) with two scratch logins — one built the old way, one with the added grant — using the product's own gate expression, then both dropped:

```
login_name      has_alter_settings
--------------  ------------------
Rig1955Probe                     1     <- WITH the new GRANT ALTER SETTINGS
Rig1955NoSet                     0     <- old block (VIEW SERVER STATE only)

leftover Rig1955% principals: 0
```

### 2b. Duplicated grant lists collapsed to one authority

Per the issue ("state which one is authoritative and make the other a pointer"): the root README's block is authoritative — Lite and Darling need the identical target-server grants, and the root README already positioned itself that way ("Darling uses the same target-server grants"). `Darling/README.md` now points there and keeps only what is genuinely Darling-specific: the per-grant "what breaks without it" table (not duplicated anywhere), and the integrated-auth note, which I made sharper rather than dropping — the grants belong to the account **the service runs as**, not the one that ran `--test-connection`.

### 2c. The PostgreSQL role table understated `viewer`

**Before:** `viewer` = "SELECT on both schemas + INSERT/UPDATE/DELETE on `config.custom_views` only".

That is wrong in two ways the field would hit: `viewer`'s SELECT on `config` is **column-carved fail-closed** on `config_monitored_servers` / `config_command` / `config_notification`, and the role carries `ALTER ROLE viewer SET statement_timeout = '15s'` (provision-roles.sql step 1a), which was documented nowhere in the README.

**After:** the table states the carve and the timeout, and the `admin` row states that it *can* read those secret columns (the Settings window needs them) and has no timeout.

### 2d. Stale schema version

`Postgres store ready (schema v34, ...)` in the console-run walkthrough. The current migration count is **v44** (measured, below). Updated, with a note that the number tracks the migration count.

### Checked and found correct (no change)

- "`provision-roles.sql` creates two login roles — `admin` and `viewer`" — proven live: `pg_roles` after provisioning holds exactly `darling`, `admin`, `viewer`. No `mcp`, matching the deliberate BYO asymmetry pinned by `ProvisionRolesAclDriftTests.ProvisionRolesSql_ProvisionsAdminAndViewerOnly_NeverTheMcpRole`.
- The `mcp` row's "viewer's exact read surface" — code applies `BuildViewerColumnAclSql(config, mcp)`, the same builder as viewer's carve. (Managed-only, so not provable on a BYO rig; verified by reading the generator.)
- `admin` = "SELECT both schemas + INSERT/UPDATE/DELETE on `config` only" — proven live below.

## Live grant evidence

Local rig per the recorded recipe: repo `pg-runtime.zip` → **PostgreSQL 18.4**, throwaway cluster on a unique datadir and port 5578, port ownership confirmed as mine via `pg_ctl status` + `netstat` (PID 16992) before any database creation, database created with `psql -c` rather than `createdb`, superuser `darling`, TimescaleDB preloaded with the product's own worker sizing (41/52).

Store brought up **the way the product does** — a harness calling the real `PgMigrations.MigrateAsync` then `TrySetDatabaseSearchPathAsync`, the same two calls the service makes at startup:

```
MigrateAsync -> schema v44
TrySetDatabaseSearchPathAsync -> done
collect relations: 76
config relations: 13
```

Then the **documented** provisioning command, unmodified except the two password placeholders and the port:

```
psql -h 127.0.0.1 -p 5578 -U darling -d darling -v ON_ERROR_STOP=1 -f provision-roles.sql
```

Clean end to end under `ON_ERROR_STOP=1`. Roles created: `admin`, `viewer` (+ the `darling` superuser). Then probed as each role.

**`viewer` — reads what the viewer reads:**

```
 current_user |       search_path       | statement_timeout
--------------+-------------------------+-------------------
 viewer       | collect, config, public | 15s

 v_collection_log_rows | servers_rows | wait_stats_rows | alert_log_rows | mute_rule_rows | bare_name_resolves
                     0 |            0 |               0 |              0 |              0 |                  0
```

(All six SELECTs succeeded — a collector view, a collect table, two config tables, and a bare unqualified name resolving through the search path. Zero rows because nothing has collected into this rig; the point is the grant, not the data.)

**`viewer` — cannot read the secret columns (fail-closed carve):**

```
ERROR:  permission denied for table config_notification   <- smtp_encrypted_password
ERROR:  permission denied for table config_notification   <- smtp_username
ERROR:  permission denied for table config_notification   <- teams_url
ERROR:  permission denied for table config_notification   <- generic_headers
ERROR:  permission denied for table config_notification   <- pagerduty_routing_key
ERROR:  permission denied for table config_monitored_servers  <- encrypted_password
ERROR:  permission denied for table config_command            <- args_json
```

...while the **non-secret** columns of those same tables read fine (`server_id, name, host` and `id, smtp_host, smtp_port` both returned). `pagerduty_routing_key` is the fail-closed proof: it is not in the GRANT list, so it is denied without anyone having revoked it.

**`viewer` — cannot write, cannot DDL:**

```
ERROR:  permission denied for table servers            <- INSERT collect.servers
ERROR:  permission denied for table servers            <- UPDATE collect.servers
ERROR:  permission denied for table config_mute_rules  <- INSERT
ERROR:  permission denied for table config_mute_rules  <- UPDATE
ERROR:  permission denied for table wait_stats         <- DELETE
ERROR:  permission denied for schema collect           <- CREATE TABLE
ERROR:  must be owner of view v_collection_log         <- DROP VIEW
```

**`viewer` — the one write it IS granted (`config.custom_views`, #1563):**

```
 id |  name
----+---------
  1 | rig1955
INSERT 0 1
UPDATE 1
DELETE 1
```

**`admin` — reads the secret columns, writes config, still cannot touch collect or do DDL:**

```
 current_user | search_path             | statement_timeout
 admin        | collect, config, public | 0

 notification_secret_readable | monitored_secret_readable
                            0 |                         0     <- SELECT succeeded (0 rows on an empty rig)

 INSERT 0 1 / UPDATE 1 / UPDATE 0 / DELETE 1   <- config_mute_rules CRUD + config_alert_log dismissal

ERROR:  permission denied for table servers      <- INSERT collect.servers
ERROR:  permission denied for table wait_stats   <- DELETE
ERROR:  permission denied for schema collect     <- CREATE TABLE
ERROR:  must be owner of table wait_stats        <- DROP TABLE
```

One honest note on method: three probes in my first pass failed on wrong column names / a type error rather than on permission, which proves nothing. I re-ran those against the real column list; every denial quoted above is a genuine `permission denied`, not a schema error wearing its clothes.

## Pinned-text check

Per the trap: `grep -rn "README" Darling/Darling.Tests --include=*.cs -l` → `DarlingExportViewerConfigTests`, `ProvisionRolesAclDriftTests`, `DeadlockGraphParserTests`, `ViewerDeadlockSeverityStatsTests`.

- `DarlingExportViewerConfigTests` pins the **exported** `README.txt` / `darling.json` phrases, which are built by `DarlingCliCommands.BuildViewerConfigReadme` — generated strings, not this repo's markdown. Untouched by this PR.
- `ProvisionRolesAclDriftTests` parses `Darling/tools/provision-roles.sql`. Not modified here (the audit found the *docs* wrong, not the script). Its comment referencing "the BYO section of `Darling/README.md`" is prose; my edits keep that section's claim (BYO = two roles, managed adds `mcp`) true.
- The other two are deadlock-fixture READMEs, unrelated.
- No test reads `Darling/README.md` or the root `README.md` from disk (verified by grep, and no `README` content-copy items in `Darling.Tests.csproj`).

So there is no pin to update, and no watched-red cycle to run — nothing asserts on the text I changed. No `.cs` changed means no build/warning gate applies either.

## CHANGELOG

Two entries under `[Unreleased]` → `### Changed` (the restructure, and the grant audit), with the `[#1955]` bottom link-ref.


erikdarlingdata:
## Final status — green on head `6f3414bb`

```
mergeable: MERGEABLE   (was CONFLICTING after dev moved; merged origin/dev, CHANGELOG resolved keep-both-sides)

build:                    COMPLETED SUCCESS
check-branches:           COMPLETED SUCCESS
review:                   COMPLETED SUCCESS
Darling PostgreSQL tests: COMPLETED SUCCESS
```

Check-runs verified against the current head SHA (`6f3414bb73169ba60fd1dc3ebc0278132128b1db`), not a stale one.

**One thing worth flagging about the CI signal.** The first cycle on the docs-only head reported all four checks green in under 30 seconds, which is the path-filter short-circuit rather than a real build. The merge of `origin/dev` brought #1963's collector-schedule `.cs` changes onto this branch, so this cycle ran the full build and the Darling PostgreSQL job for real. That is the cycle the green above comes from.

Merge conflict resolution: `CHANGELOG.md` only (both sides had prepended to `### Changed`). Kept both — this PR's two #1955 entries and dev's #1963 entry — with both bottom link-refs intact. Verified neither parent contained a bare `=======` line, so the marker cleanup removed exactly the three conflict lines and nothing legitimate; line count 2432 = base + 9 (mine) + 2 (dev), as expected.

Review feedback from the first cycle was addressed rather than deferred: the reviewer caught that splitting the permissions intro left `The full least-privilege grant set (...).` as a verbless fragment, fixed in `0c3c2961`.

Not merging — leaving that to the team lead.

PR #1987, opened 2026-08-01
Surface automatic plan correction in both apps (#1952)

### LINKED ISSUE #1952: Detect automatic plan correction and surface the queries the engine is actively working on
When FORCE_LAST_GOOD_PLAN is enabled, the engine is quietly doing plan-regression detection and correction on the operator's behalf - detecting regressed queries, forcing last-good plans, verifying the result, and backing out when the forced plan does not help. None of that activity is visible in the product today, and it is exactly the kind of thing an operator should see: the engine just told you which queries regressed.

## What to collect

Two layers:

1. **Enablement inventory**: sys.database_automatic_tuning_options per database - which databases have FORCE_LAST_GOOD_PLAN on (desired vs actual state, since Query Store conditions can make them differ).
2. **The active work**: sys.dm_db_tuning_recommendations per enabled database - the engine's live recommendation set. The details JSON carries the query id, the regressed and recommended plans, and the estimated gain; the state JSON carries where the engine is in the detect/force/verify/revert cycle. Surfacing the ACTIVE and VERIFYING ones is the headline: those are queries the engine is working on right now.

## Platform and version gates

- 2017+ on box (version-gated like the other post-2016 columns; min supported here is 2016).
- Azure SQL DB supports automatic plan correction and can have it on by default; Managed Instance supports it. Verify per-platform behavior against MS Learn at build time - especially whether the DMV binds on each edition through the same paths the #1886 probes exercised - per the sourced-advice rule.
- Requires Query Store enabled per database, which the QS collector already establishes per target; reuse that knowledge rather than re-probing.

## Surfacing

Both apps, catalog-driven collector per the standing pattern (catalog + dispatch + StorageVersion bump + viewer probe sentinel). The natural home is beside the Query Store surfaces, since every recommendation names a Query Store query id - and in-app plan navigation (#1184 family) should make the regressed/forced plans one click away. NOT in scope: acting on recommendations from the product; this is visibility into what the engine itself is doing.

erikdarlingdata:
Closes #1952.

When `FORCE_LAST_GOOD_PLAN` is on, the engine is quietly doing plan-regression detection and correction on the operator's behalf. None of it was visible in the product, even though the engine is effectively handing you a list of which queries regressed and by how much.

One catalog-driven collector, `plan_correction`, captures both layers per database in a single round trip. Both objects are database-scoped, so it takes the enumeration path (`[db].sys.sp_executesql`) like `database_scoped_config` and `query_store`, with the same Azure SQL DB per-database-connection branch.

## Documentation verification

Every claim below was checked against the page before it was written into SQL.

- [sys.database_automatic_tuning_options](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-database-automatic-tuning-options-transact-sql) - column list, state/reason semantics, 2017+.
- [sys.dm_db_tuning_recommendations](https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-objects/sys-dm-db-tuning-recommendations-transact-sql) - column list, `state` and `details` JSON shape, the shredding examples, and the remark that recommendations live **only until the instance restarts**.
- [Automatic tuning (SQL Server)](https://learn.microsoft.com/en-us/sql/relational-databases/automatic-tuning/automatic-tuning) - the detect/force/verify/revert lifecycle, the 10-second estimated-gain forcing threshold, and that a restart mid-verification unforces the plan.
- [ALTER DATABASE SET options](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-database-transact-sql-set-options) - the `AUTOMATIC_TUNING (FORCE_LAST_GOOD_PLAN = ...)` syntax, 2017+.
- [Automatic tuning overview (Azure SQL)](https://learn.microsoft.com/en-us/azure/azure-sql/database/automatic-tuning-overview) and [Enable automatic tuning](https://learn.microsoft.com/en-us/azure/azure-sql/database/automatic-tuning-enable) - Azure has its own tuning service with CREATE_INDEX/DROP_INDEX, which box does not have and Managed Instance does not support; MI is FORCE_LAST_GOOD_PLAN only.
- [sys.query_store_plan](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-query-store-plan-transact-sql) / [sys.query_store_query](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-query-store-query-transact-sql) / [sys.query_store_query_text](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-query-store-query-text-transact-sql) - the join path from the recommendation's queryId to query text.
- [Editions and supported features](https://learn.microsoft.com/en-us/sql/sql-server/editions-and-components-of-sql-server-2022) - automatic tuning is **Enterprise only**. On Standard the enablement row reports NOT_SUPPORTED rather than going quiet.

Version gate: 2017+ (`SqlMajorVersion >= 14`), Azure SQL DB and MI ungated, `SqlMajorVersion == 0` = assume newest per the house convention. The enumeration query re-evaluates `PRODUCTVERSION >= 14` **on the server** for exactly that unknown-version case, in the WHERE clause rather than an IF so the batch always returns the enumeration contract's first result set.

## Three places the docs are wrong, and what the server actually did

**1. AbortedCount vs ErrorCount.** The column table documents `regressedPlanAbortedCount`; every example query on the same page shreds `regressedPlanErrorCount`. They cannot both be right, and picking wrong binds a silent NULL rather than raising anything. Live capture on SQL2022 settles it - the engine emits **ErrorCount**. The shipped SQL still reads both spellings and COALESCEs, because that cannot be wrong.

**2. OPENJSON is not safe here.** Microsoft's documented shredding query uses `OPENJSON`, which requires database compatibility level 130 or higher. A compat-100 database on a 2017+ instance is perfectly legal. Measured on SQL2022 against a compat-100 database:

- `JSON_VALUE(...)` returned `42`.
- `OPENJSON(...) WITH (queryId bigint ...)` failed outright: `Msg 102, Level 15, State 1 - Incorrect syntax near '$.queryId'`.

The documented query would take out collection for that database, so this ships JSON_VALUE with explicit paths.

**3. reason_desc values do not match the documented codes.** Docs say reason code 2 = `DISABLED`; a default SQL2022 database returns code 2 with `reason_desc = AUTO_CONFIGURED`. The collector stores the engine's own `_desc` string rather than mapping codes itself. Also undocumented and confirmed live: `reason` / `reason_desc` are **NULL** when desired state equals actual state.

One more deliberate divergence: Microsoft's Example 3 INNER JOINs `sys.query_store_plan` twice, which silently drops any recommendation whose plans have aged out of Query Store. That recommendation is still live and still names a real regression, so the joins here are LEFT - it lands with a null `query_text` instead of vanishing.

## Live verification

**The recommendation repro succeeded.** What follows is a real `sys.dm_db_tuning_recommendations` row, not a synthesized one.

Getting there took a detour worth recording: the classic parameter-sniffing regression **cannot be provoked on SQL Server 2022 at compat 160**, because Parameter Sensitive Plan optimization splits the query into plan variants that each get their own `query_id` (observed: QueryVariantID 1 and 3 as two separate query ids). Automatic plan correction compares plans *within* one query_id, so a PSP-dispatched query never shows the regression. Setting `PARAMETER_SENSITIVE_PLAN_OPTIMIZATION = OFF` restored the old behaviour, the engine detected the regression, and it auto-forced the plan - the estimated gain cleared the documented 10-second threshold.

Raw payload the engine emitted:

```
state   = {"currentValue":"Verifying","reason":"LastGoodPlanForced"}
details = {"planForceDetails":{"queryId":1,"queryHash":"0x805988133769D0AB","regressedPlanId":2,
           "regressedPlanExecutionCount":15,"regressedPlanErrorCount":0,
           "regressedPlanCpuTimeAverage":1.054473333333333e+005,"recommendedPlanId":1,
           "recommendedPlanExecutionCount":600,"recommendedPlanErrorCount":0,
           "recommendedPlanCpuTimeAverage":1.629666666666667e+001, ...},
           "implementationDetails":{"method":"TSql",
           "script":"exec sp_query_store_force_plan @query_id = 1, @plan_id = 1"}}
```

End to end through the real `DarlingCollectorRunner`, from live SQL2022 into live PostgreSQL 18.4 + TimescaleDB 2.28.1 - 12 databases enumerated, 12 rows landed:

```
ENABLEMENT | Crap               | desired=DEFAULT | actual=OFF    | reason=AUTO_CONFIGURED
ENABLEMENT | PlanCorrection1952 | desired=ON      | actual=ON     | reason=(null)
ENABLEMENT | StackOverflow2013  | desired=ON      | actual=ON     | reason=(null)
ENABLEMENT | tempdb             | desired=(null)  | actual=(null) | reason=(null)
RECOMMENDATION | db=PlanCorrection1952 | name=PR_1 | state=Verifying/LastGoodPlanForced
  score=100 query_id=1 regressed_plan=2 last_good_plan=1 forcing=AUTO
  regressed_cpu_ms=105.447 last_good_cpu_ms=0.0163 gain_s=64.84 error_count=0
  text=(@skew_key int)SELECT c = COUNT_BIG(*), m = MAX(s.filler) ...
  script=exec sp_query_store_force_plan @query_id = 1, @plan_id = 1
```

`tempdb` is the case the schema shape exists for: it has **no** row in `sys.database_automatic_tuning_options` at all. The enablement half hangs off a one-row anchor via OUTER APPLY, so the database still lands a row with nulls instead of dropping out of the collection.

Also verified:

- The QUERY_STORE_OFF diagnostic: with desired ON and Query Store off, the engine reports `actual=OFF, reason=QUERY_STORE_OFF`. That is why every database is enumerated rather than only the Query-Store-enabled ones - filtering would drop exactly the rows worth looking at.
- **SQL2019 (15.0.4410)**, **SQL2022 (16.0.4255)** and **SQL2025 (17.0.4045)**: the enumeration and the per-item read run clean on all three. The per-item text used was extracted from the shipped source and quote-doubled exactly as `BuildPerItemQuery` does it, so the escaping was exercised against the server rather than only asserted in a unit test.
- The DMV binds and returns that database's rows through the three-part `sp_executesql` call - the enumeration path's core assumption, which the docs never actually state.

Not verified: Azure SQL DB and Managed Instance were not exercised. The Azure branch mirrors the shape `query_store` and `deadlocks` already use.

## Darling V46 on fresh and upgraded stores

Proven against a real PostgreSQL 18.4 + TimescaleDB 2.28.1 cluster, both paths, because they build the table by different code: on a **fresh** store V1's catalog-generated schema creates it and V46 is a no-op, so nothing else anywhere ever executes `V46Sql`. `DarlingPlanCorrectionLiveMigrationTests` winds that one table back to its pre-V46 state and lets the migration rebuild it, then asserts both paths produce column-for-column identical tables.

That matters because both writers are **positional**: a literal that disagreed with the generator by one column or one type would put every value in the wrong column on exactly the stores that upgraded, while fresh stores stayed correct. Red-watched by injecting `score bigint` into `V46Sql` alone - the test failed naming `score`, expected `integer`, actual `bigint`.

**Numbering:** V45 is deliberately absent - a sibling lane holds it and this lane was assigned V46. The applier takes every script above `MAX(version)` and never assumes contiguity, so a gap is inert; a collision would not be. A test pins uniqueness and ascent rather than contiguity, and the upgrade path asserts exactly one migration is applied.

## Surfacing

- **Plan Corrections** grid beside the Query Store surfaces, born #1949-compliant: `Collected`, `Query Text`, then the metrics. Pinned in both apps' column-order tables, and twinned so the two column sequences cannot drift.
- **Automatic Tuning** grid under Configuration, beside Scoped Configuration, for the enablement half - one row per database from the latest capture.

## Tests

- `Lite.Tests`: **2015 passed, 0 failed.**
- `Darling.Tests` with `DARLING_TEST_PG` pointed at the local rig: **4152 passed, 0 failed**, 11 skipped (those need env vars this run did not set).
- Builds - `PerformanceMonitor.Collectors`, `PerformanceMonitor.Common`, `Lite`, `Darling.Service`, `Darling.Viewer`, both test projects: **0 Warnings, 0 Errors** each.

New tests were watched red before green:

- The `sp_executesql` quote-doubling assertion: removing the escaping fails it. The first version of this assertion was **vacuous** - `''X''` trivially contains `'X'` - so it is anchored on `", '"` instead, which the escaped and unescaped forms cannot both match.
- The V46 drift guard, as described above.

`PlanCorrectionCollectorDefinitionTests` covers the version gate, both query shapes, the Azure branch, and the 38-ordinal read using values taken from the real captured recommendation, so an off-by-one lands as a value on the wrong property. It also pins the JSON_VALUE, dual-spelling and LEFT JOIN decisions so they cannot silently rot back to the documented forms.

## One pre-existing failure, verified not mine

`DarlingCollectorRunnerTests.EndToEnd_QueryStats_PlanCaptureFlag_TogglesStoredPlanXml` fails against this SQL2022 + Postgres pair with "flag-on capture should store at least one non-empty plan". It is deterministic, not flaky - it failed on three consecutive runs. I built a clean `origin/dev` worktree and ran it against the same two servers: **it fails there identically**, so it is environmental (plan cache state on SQL2022), not a regression from this branch. It is skipped in CI, which does not set `DARLING_TEST_SQL`.

## Pins moved

Catalog 38 to 39, Lite DuckDB schema v49 to v50, Darling `StorageVersion` 44 to 46, both grid column-order pin tables (+1 grid each; the non-vacuity floors were raised by exactly the 27 columns added, so the guard keeps its existing slack rather than quietly loosening), `GoldenCollectorSchema`, the enumerating-collector inventory in both apps plus `ServerHealthBands.ExpectsUserDatabases`, all four preset tables (Lite `s_presets` and Darling `CollectorSchedulePresets`, plus the two cross-app pins), and CI's throwaway-cluster worker sizing - 41 to 42 background workers, 52 to 53 worker processes - which derives from the hypertable count.

Not in scope, per the issue: acting on recommendations from the product. This is visibility into what the engine itself is doing.


erikdarlingdata:
## Final status: green, `MERGEABLE` / `CLEAN`

`statusCheckRollup`:

| Check | Conclusion |
| --- | --- |
| build | SUCCESS |
| check-branches | SUCCESS |
| review | SUCCESS |
| Darling PostgreSQL tests | SUCCESS |

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `state: OPEN`. Not merged, and #1952 is left open.

Worth noting that **Darling PostgreSQL tests** passing means CI re-ran the V46 fresh-vs-upgraded store proof on its own throwaway cluster, independently of my local rig - including the worker-sizing bump (42 background workers / 53 worker processes) that this branch had to make, since that cluster is sized off the hypertable count.

Local runs for the record: `Lite.Tests` 2015 passed / 0 failed; `Darling.Tests` against a local PostgreSQL 18.4 + TimescaleDB 2.28.1 rig 4152 passed / 0 failed / 11 skipped; every project builds at 0 warnings.

The one failure I hit locally (`EndToEnd_QueryStats_PlanCaptureFlag_TogglesStoredPlanXml`) is documented in the PR body: it reproduces identically on a clean `origin/dev` worktree against the same servers, so it is environmental and not from this branch. It does not run in CI.

erikdarlingdata:
Took the review's one nit: the estimated-gain formula now carries its MS Learn provenance and the deliberate both-plans execution-count sum, with the live repro number as the cross-check. SQL-string-only change inside PayloadBodyText; collector project builds clean. Merging off the fresh rollup.

Issue #1743, opened 2026-07-27
Darling: anomaly-engine statistical depth (MAD/EWMA/changepoint; real confidence)


erikdarlingdata:
Split from #1606. Engine is z-score/ratio with confidence hardcoded 1.0. Add robust statistics + honest confidence. Already ahead of DBM's threshold-only model; this deepens it.

erikdarlingdata:
Calibration evidence, measured 2026-07-31 on DARLING01's live store against a real HammerDB run (30 min TPC-C on SQL2025, 40,362 NOPM / 95,665 TPM, window 11:48-12:20 UTC). Read-only analysis; query preserved in session records.

**The current engine is fully blind to a 26x workload surge, and the mechanism is exactly the robust-statistics argument.**

Batch Requests/sec on SQL2025, from the engine's own perfmon_baseline population (29,408 samples):

| | value |
|---|---|
| idle reality (median) | 64/sec |
| baseline MEAN | 1,107/sec |
| baseline STDDEV | 11,994 |
| baseline MAD | 11 |
| load window average | 1,682/sec (peak 2,315) |

The mean is 17x the median because weeks of prior HammerDB bursts live in the baseline - this store's real history IS the one-bad-afternoon pathology, permanently. Consequences during today's genuine 26x-over-idle load:

- **Ratio detector (the shipped semantics for rate metrics): 1,682 / 1,107 = 1.5x, under the 4.0 threshold. No alert. The engine ran and could not see it.**
- z against mean/stddev: **0.0** - the inflated deviation swallows a 26x surge without registering a tenth of a sigma.
- z against median/MAD: **99.2** - unmissable.

What it would take to fire TOMORROW (after today's load joins the baseline): stddev model needs ~25,083/sec (~390x idle); ratio model needs ~4,431/sec; MAD model fires from ~97/sec, which the existing absolute floors (BatchRequestFloor = 500/sec) correctly clamp so quiet-metric noise stays out - the floors and robust statistics compose, they do not compete.

Methodology notes, honestly: the population was pooled across hour-of-day/day-of-week buckets rather than per-bucket (the skew is global - burst runs happened at many hours - so bucketing does not rescue the mean), and the live-fired alerts during the window were the threshold family (Blocking x10, Deadlocks x50 - real TPC-C contention, correctly caught) plus a correct CPU non-fire (the box held ~10%, under the 50% floor).

[...]

PR #2001, opened 2026-08-03
Dedup get_analysis_findings to one entry per diagnostic chain (~28x smaller reads)

### LINKED ISSUE #2000: get_analysis_findings re-ships every story once per engine cycle - ~28x payload duplication on a 24h read
## Problem

`get_analysis_findings` returns one entry per persisted finding ROW, but the analysis engine persists the same stories every cycle. A multi-hour read therefore returns the same diagnostic chain dozens of times, each occurrence re-carrying the full composed advice prose (headline/investigation/remediation), the rendered `remediation_command`, `story_path`, and a `co_fired` array. Agent consumers burn context on a wall of repeats; this surfaced during the #1743 dogfood readout, where a single server's `hours_back=10` read weighed ~257KB.

Measured on the production monitor store (52 servers, 24h window, 2026-08-03):

| measure | value |
|---|---|
| finding rows | 21,623 |
| distinct (server, story_path_hash, incident_id) groups | 774 |
| **duplication factor** | **27.9x** |
| worst single server (upsilon) | 547 rows / 14 groups = 39.1x |
| avg frozen story_text per row | 859 chars (before advice prose + remediation rendering) |

Severity is NOT constant within a group — 275/774 groups moved >0.1 severity across the window (max spread 0.976) — so a naive "keep one" would erase real signal.

## Fix

Group by `(story_path_hash, incident_id)` and emit ONE entry per group:

- representative fields from the **latest** occurrence (value-bearing advice is frozen at analysis time, so latest = current truth): severity, confidence, category, facts, story, advice, `remediation_command`, `co_fired`
- occurrence stats replacing the lost timeline: `occurrences`, `first_seen`, `last_seen`, `peak_severity` (and latest severity is the top-level one), `time_range` spanning the group
- response header keeps `finding_count` (deduped) and adds `total_occurrences` (raw rows) so the compression is visible

The store keeps every occurrence — viewer timelines are untouched; this is a read-shaping change only. Applies to both twins (Darling `DarlingMcpTools.GetAnalysisFindings`, Lite `McpAnalysisTools.GetAnalysisFindings`) and rides through the web API tool bridge for free. Dedup helper belongs in the shared Analysis project so the twins cannot drift.

Expected effect: a 24h read shrinks ~28x on typical servers (547 entries → 14 on the worst measured), with severity evolution preserved via the occurrence stats.

erikdarlingdata:
## What does this PR do?

Fixes #2000. `get_analysis_findings` returned one entry per persisted finding ROW, but the engine re-persists the same stories every analysis cycle — so a multi-hour read returned each diagnostic chain dozens of times, every occurrence re-carrying the full advice prose and copy-paste remediation command. Measured on the 52-server production monitor (24h window): **21,623 rows collapse to 774 (story_path_hash, incident_id) groups — 27.9x mean duplication, 39.1x on the worst server**, ~257KB for a single server's 10-hour read during the #1743 dogfood readout.

Both MCP twins now collapse through a shared `FindingOccurrences` helper in PerformanceMonitor.Analysis (so they cannot drift; the Darling web API bridge rides the same method):

- **One entry per (story_path_hash, incident_id)** — the story hash is the chain's identity (what muting keys on); the incident id separates distinct episodes of the same chain.
- **The LATEST occurrence is the representative** — value-bearing advice is frozen into StoryText at analysis time, so latest = current truth.
- **Occurrence stats replace the collapsed timeline**: `occurrences`, `first_seen`, `last_seen`, `peak_severity`. Severity genuinely moves within a chain (275 of the 774 measured groups spread >0.1, max 0.976), so a naive keep-one would erase real signal.
- **`finding_count` counts chains; new `total_occurrences` reports raw rows** — the compression is visible, not silent.
- **The read stops being quietly wrong at depth**: the store read's default `limit 100` covered only the ~5 most recent engine cycles of a 24h request and silently dropped the rest — occurrence stats computed under it would lie about `first_seen`. The tools now pass a window-covering cap (10,000), sized for the deepest legal read (7 days × ~23 rows/hour measured = ~3,900).

The store keeps every occurrence — this shapes the READ only. Viewer timelines, persistence, and `analyze_server` (single-run, no cross-run duplication) are untouched. Tool descriptions and both MCP instruction tables state the new semantics.

- [x] Lite
- [x] Darling
- [x] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## How was this tested?

- **New `FindingOccurrencesTests` (Lite.Tests, pure in-memory)**: grouping by (hash, incident) with same-hash/different-incident staying separate; latest-occurrence selection with severity-then-finding-id tie-breaks for equal-time rows; occurrence stats (count, first/last seen, peak severity); spanning time range with null ends skipped and all-null yielding null; still-firing-first ordering; empty input.
- **Darling gated e2e extended (`EndToEnd_FindingsEnvelope_MuteFlow_AgainstDevPostgres`)**: plants a second occurrence of the same chain (same hash + incident, 30 minutes later, LOWER severity 1.8) and asserts through the real Postgres store and MCP envelope: 3 rows → 2 chains, `total_occurrences` 3, the collapsed chain is first (later last_seen), represents the latest occurrence (severity 1.8, leaf value 81), carries `peak_severity` 2.5 from the first occurrence, `first_seen` < `last_seen`; the single-occurrence chain keeps trivial stats. Envelope field pins extended with the four new fields; the null/non-null `remediation_command` assertions ride the latest occurrence unchanged.
- Existing `McpAnalysisFindingsCommandTests` (distinct hashes → unchanged counts) pins the Lite envelope still carries `remediation_command` per entry.
- Local builds of PerformanceMonitor.Analysis and the Darling service are clean (0 warnings); the Windows-only Lite/test projects compile in CI.
- Duplication/severity-spread numbers measured against the production monitor's live store (52 servers) via psql — queries in the issue.

Dogfood plan after merge: kick the nightly, upgrade the monitor box, and re-read the same server that measured 257KB/10h to confirm the shrink and the occurrence stats against real fleet data.

## Checklist


erikdarlingdata:
Good catch on the Lite lane — addressed in the follow-up commit. One correction to the premise: CI *does* set `DARLING_TEST_PG` (the **Darling PostgreSQL tests** job, build.yml), so the Darling e2e's collapse assertions ran and passed in this PR's runs. But the core point stands: the tool bodies are twin copies — only `FindingOccurrences.Collapse` is shared — so Lite's own JSON wiring for the new fields ran in no test. The new fact exercises it through the real DuckDB store and serializer: 3 rows → 2 chains, the later-but-lower occurrence as representative (severity 0.5) with peak_severity 1.0 preserved, still-firing chain ordered first, spanning time_range, trivial stats on the single-occurrence chain.

erikdarlingdata:
Both round-2 notes addressed: the `co_fired` pre-pass now composes titles only for runs that produced a group representative (steady state: the single most recent run) instead of all window-covering-cap rows, and a read that fills the 10,000-row cap now discloses it in a `truncation_note` — present-but-null on healthy reads, pinned in both twins' tests — rather than letting `occurrences`/`first_seen` under-report silently.

Issue #2005, opened 2026-08-03
Headless --test flag for the Viewer: run the #1954 connection self-test without the UI


erikdarlingdata:
Deferred scope from #1954 (the issue offered "a button in the connect-failure dialog **or** a `--test` viewer flag"; the button shipped). A headless flag would run the same layered ladder (`StoreConnectionSelfTest.RunAsync` in Darling.Storage) against the resolved config and print/log the report without starting WPF — useful for scripted diagnostics over RDP/SSM sessions where clicking the overlay button isn't possible, and it would let the monitoring-box dogfood loop exercise the ladder remotely.

Why deferred: a WPF app has no console by default — honest stdout needs `AttachConsole(-1)`/alloc fallback p/invoke and exit-code plumbing, which deserves its own focused change rather than riding the button PR blind (the flag path can't be tested on the development Mac; the button path's core is fully covered by unit + live tests).

Shape sketch:
- `PerformanceMonitor.Darling.Viewer.exe --test [--config <path>]` → resolve config exactly like startup (all four rules), run the ladder, print `FormatReport`, exit 0 when no layer failed / 1 otherwise.
- Reuse the #1966 provenance block in the output so the headless report carries the same "which file, which values" context as the overlay.

erikdarlingdata:
Shipped in #2045 (merged to dev): `--test [--config <path>]` resolves config exactly like startup, prints the #1966 provenance block and the six-layer ladder report to console + log, and exits 0/1 — no window, no single-instance dance. AttachConsole(-1)/AllocConsole handles the GUI-subsystem console; scripts sequence on the exit code via Start-Process -Wait. First live run rides the next box upgrade's SSM session.

erikdarlingdata:
Live validation complete, second attempt. The first real SSM run found the classic WPF sync-over-async deadlock (fixed in #2053 with `Task.Run` hopping the ladder off the dispatcher's context — exactly the risk the original deferral called out about not being testable on the dev Mac). Tonight's rerun on the monitor box over SSM:

```
exit code: 0
darling.json source: the service root, one level above the viewer
postgres.managed: true (connection derived by the viewer)
[PASS] config parsed; endpoint 127.0.0.1:5641, ssl mode Prefer
[skip] dns    host is an IP literal; no DNS lookup needed
[PASS] tcp    connected to 127.0.0.1:5641 (13 ms)
[...]

PR #2006, opened 2026-08-03
Viewer connect-failure overlay gains a layered connection self-test (#1954 bullet 3)

### LINKED ISSUE #1954: Viewer diagnostics: say WHERE darling.json and server.crt were read from, what was parsed, and which connection layer failed

erikdarlingdata:
## What does this PR do?

Fixes #1954 — the remaining bullet 3 (bullets 1–2, path provenance + non-secret parse summary, shipped in #1966). The connect-failure overlay gains **Run self-test**: a layered probe against the SAME effective connection string the failed connect used — config parse → DNS → TCP → TLS → auth → schema gate — with each layer reporting pass/fail/skip, prose detail, and elapsed time, and a verdict naming the FIRST failed layer. "Firewall problem" becomes diagnosable by elimination (DNS passed, TCP did not) from the Viewer's own seat, as the issue asked.

Design notes:

- **The core is shared and never throws** (`StoreConnectionSelfTest` in Darling.Storage, which the Viewer already references and where `StorageVersion`/`PgSchemaGenerator` live). Every failure becomes a layer result; later layers report NotReached.
- **TLS vs auth attribution is by exception shape** (they share one Npgsql `Open`): PostgreSQL auth SQLSTATEs (28P01 / 28000 / 3D000) prove the handshake already succeeded; an `AuthenticationException` in the chain proves auth was never reached. The table is pure and pinned.
- **Two honesty rules, both earned in live testing**: `SSL Mode=Prefer/Allow` opening successfully proves *no* handshake (falls back to plaintext) → skipped-with-caveat, not a hollow pass. And the least-privilege viewer role can't read the owner-only `darling_schema_version` — the fallback probes `to_regclass` (catalog-visible to every role, both historical schema homes), **not** `information_schema`, which hides unprivileged tables and misread a healthy store as not-a-Darling-store in a live viewer-role run before I fixed it.
- Report re-runs **replace** the previous report in the overlay (spliced on the core's public `ReportHeader`); everything also goes to the log, flushed, matching the #1966 failure treatment.
- **Deferred per the standing rule**: the headless `--test` flag the issue offered as an alternative → filed as #2005 (WPF console-attach plumbing deserves its own change and can't be tested on this dev machine).

## How was this tested?

- **Live matrix against a real TimescaleDB store** (all eight cases attribute to the right layer): happy path (schema v48 PASS), wrong password (auth FAIL 28P01 with the username lifted from the server message), wrong port (tcp FAIL, later layers NotReached), unknown host (dns FAIL), unknown database (auth FAIL 3D000), authenticated-but-not-a-store (schema FAIL with guidance), **least-privilege viewer role against a healthy store** (schema SKIP with the owner-only explanation — the case that caught the `information_schema` bug), and a garbage connection string (config FAIL, everything NotReached).
- **Pure pins** (run in CI): the ClassifyOpenFailure table (including nested/wrapped exceptions), EvaluateSchemaGate's OLDER/NEWER wording (which half to upgrade), FormatReport verdict honesty (skips are not overclaimed as "all layers passed"), and the no-network RunAsync shapes.
- **Gated live ladder** (`[Collection("live-postgres")]`, runs in the Darling PostgreSQL tests job): happy path against the just-migrated store, the wrong-port firewall signature, and the 3D000 auth arm — chosen over wrong-password because CI's Postgres trusts local connections, so a wrong password would still authenticate there; the 28P01 shape is pinned purely instead.
- Darling.Storage builds locally with 0 warnings; the Viewer (WPF) compiles in CI.

- [ ] Lite
- [x] Darling
- [ ] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## Checklist


erikdarlingdata:
Took the cancellation nit even though it was non-blocking — it's a contract fix better made before the API ships. Every catch site now rethrows on the caller's token (house filter pattern); internal per-layer timeouts still report as that layer's failure. Pinned with an already-cancelled token rather than a timer: validating the timer version locally showed it racing — my network rejects TEST-NET-1 with "Network is unreachable" in microseconds instead of hanging, so a 250ms timer would never fire before the TCP verdict on hosts like mine.

Issue #2008, opened 2026-08-03
Server search/filter, coloured tag pills, and tag-tree grouping across all three surfaces


jakemorgangit:
Hi Erik — a feature proposal for finding and organising servers as the monitored list grows. Flagging up front what already exists so this is framed as *extend + bring to parity*, not reinvent.

## What's there today

- **Viewer** already has hierarchical server tags (`config.server_tags` with `parent_id` / `sort_order`, many-to-many via `server_tag_map`), and renders them as a collapsible **tag tree** in the sidebar (servers nested under tag headers). Nice foundation.
- Tags have **no colour** (no colour column), are **not** shown as pills, and don't appear on the Overview cards.
- **Lite** and the **web viewer** have **no tag concept at all** — no tree, no grouping.
- **Search / filter by server name (or tag) does not exist on any surface** — the Viewer's `FleetView` even reserves a seam for it, but it's unbuilt. Lite's Overview and the web fleet page have a sort control only.

## A bug I hit while auditing this: tag editing silently fails for the viewer role

The "Assign Tags" submenu (and create/rename/delete) appears to do nothing on a managed store. The wiring is fine — `AssignTag_Click` calls `AssignServerTagAsync` and reloads — but every tag write routes through `ExecuteWriteAsync`, and **the `viewer` role has no write grant on `config.server_tags` / `config.server_tag_map`** (neither `provision-roles.sql` nor `DarlingManagedRoles` grants them — `custom_views` got its viewer write grant, the tag tables were missed). So on a store where the Viewer connects as the least-privilege `viewer` role, the INSERT hits `42501`, `ExecuteWriteAsync` degrades it to `ViewerReadOnlyException`, and the click's catch logs `"Assign/unassign failed"` and shows nothing. Net effect: tags are effectively read-only from the Viewer on any provisioned store.

Happy to send this as a tiny standalone fix (add the `viewer` INSERT/UPDATE/DELETE grants on the two tag tables, mirroring the `custom_views` grant) ahead of the feature below, if you'd prefer to decouple it.

## The gap

On a large fleet you can't quickly *find* a server, and — even where tags exist (the Viewer) — they're editable only in theory, invisible except as sidebar grouping, and absent from Lite and the web entirely. DEV/PROD/TEST/etc. live only in people's heads or server names.

## Proposal (suggested as two stages)

[...]

PR #2011, opened 2026-08-03
Gate the Viewer's tag editors on the read-only seat (#2008 audit finding)

erikdarlingdata:
## What does this PR do?

Fixes the bug half of #2008 — with the opposite remedy from the one proposed there. The audit's diagnosis of the symptom is verified and correct: on a provisioned store, a least-privilege `viewer`-role seat's tag edits (Assign Tags, create/rename/delete, the Manage Tags bulk editor) all route through `ExecuteWriteAsync`, hit `42501` (no grants on `config.server_tags` / `server_tag_map`), and surface only as a status-bar line under working-looking menus.

But granting the viewer role those writes would **broaden the role model**: the viewer role's only write is `config.custom_views`, deliberately, and tags are shared fleet configuration. The actual defect is that the tag UI shipped without the `IsReadOnly` gate every other write surface carries (`RecommendationsGenerateButton.IsEnabled = !_dataService.IsReadOnly` et al. — the tag files have zero references to the probe).

The fix gates all three entry points:

- **Tag-header context menu**: all five entries (New/New Child/Rename/Delete/Manage) disable on a read-only seat — every one mutates the tree or opens the write-only editor.
- **Server-row "Assign Tags" submenu**: disabled-with-reason instead of built-then-failing.
- **`ManageTags_Click`**: backstop for any entry point the menus miss — status-bar reason instead of a window of dead buttons.

Disabled editors carry a tooltip saying *why* ("this seat is connected with the read-only viewer role — tag editing needs the admin connection"), with `ToolTipService.SetShowOnDisabled` because WPF suppresses tooltips on disabled elements — a silent grey-out would be #2008's silence back again. Admin-seat and self-managed single-connection stores keep editing tags exactly as before.

## How was this tested?

- Verified the audit's claims against dev before touching anything: no viewer grants on either tag table in `provision-roles.sql` or `DarlingManagedRoles` (only `custom_views` has the exception); every tag write routes through `ExecuteWriteAsync`; the tag files carry zero `IsReadOnly` references while the sibling write surfaces all gate on it.
- New source pins (the established worker-pin idiom): every tag entry point must consult `CanEditTags`, the gate must read `IsReadOnly`, and the disabled reason must ship with `SetShowOnDisabled`.
- WPF compile is CI's (Viewer is Windows-only); the gate logic is the same one-line pattern the other write surfaces use.
- The read-only probe itself (`DetectReadOnlyAsync`) is pre-existing and already pinned elsewhere.

- [ ] Lite
- [x] Darling
- [ ] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## Checklist


erikdarlingdata:
Took the window-level note: ManageTagsWindow now carries its own gate — read-only seats get inert controls with the reason appended to the title, generic over the whole content tree so it can't drift as buttons are added. Pinned next to the entry-point pins.

Issue #2012, opened 2026-08-03
Hash-grouped query readers blend INSERT...EXEC callers and label them with one arbitrary text


erikdarlingdata:
## The mechanism (reproduced, not theoretical)

`INSERT INTO #t EXEC <proc>` statements produce the **same `query_hash` across different callee procedures** — the hash normalizes the callee away. Reproduced on SQL Server 2022:

| query_hash | statement | host object |
|---|---|---|
| 0x9C01BDF7F0F05FC4 | `INSERT INTO #items EXEC dbo.inner_v3` | wrapper_v3 |
| 0x9C01BDF7F0F05FC4 | `INSERT INTO #items EXEC dbo.inner_v4` | wrapper_v4 |

This bit during live production triage on 2026-08-01: hash-keyed text served a V3-core's text for stats that belonged to a V4-core call on two of three per-tenant attribution datapoints — the wrong-callee label survived until disambiguated by hand via `OBJECT_DEFINITION`.

## Where it lands in the product

Per-row storage is **correct**: the collector captures each row's own statement text (offset SUBSTRING) and #1767's payload dims key by content digest, which explicitly rejected `query_hash` for this class of reason. The corruption is read-side, in every reader that groups by `query_hash` and picks one representative text:

- `DarlingDataReader.TopQueriesSql` (`get_top_queries_by_cpu`): groups `(database_name, query_hash)`, takes `MAX(sql_handle)` + a latest-text LATERAL — colliding callers' **stats are summed together** and the line is **labeled with one arbitrary caller's text**.
- `ViewerDataService.QueryStats.cs` (four hash-grouped sites) and `ViewerDataService.FinOps.Workload.cs` — same shape.
- Lite's twins of the above.
- The plan-analysis MCP tools resolve per-row and are unaffected; `query_store_stats` composers group by (module_name, query_hash), which already carries the host-object dimension and is much less exposed.

## Staged fix

**Stage 1 (no schema, this issue):** make the merged groups honest. The #1767 digest column is on every row, so the hash-grouped readers can carry `COUNT(DISTINCT query_text_digest) AS distinct_texts` almost for free. When a group has more than one distinct text, the MCP payload says so (`distinct_texts` + a caveat that the shown text is one representative — literal variants are the common benign cause, INSERT...EXEC callers the misattributing one) and the Viewer/Lite grids get the count as a column so a "1 of 4 texts" group stops masquerading as a single statement.

[...]

PR #2013, opened 2026-08-03
Top-queries reads count distinct texts per hash group (#2012 stage 1)

erikdarlingdata:
## What does this PR do?

Stage 1 of #2012 (see the issue for the live SQL Server 2022 repro and the production-triage provenance). `query_hash` normalizes `INSERT...EXEC` callees away, so both apps' hash-grouped top-queries reads could sum different callers' stats into one line labeled by one arbitrary caller's text. Per-row storage was never wrong — this is read-side honesty:

- **Darling** `TopQueriesSql` counts `COUNT(DISTINCT query_text_digest)` per group (the #1767 content digest is already on every row — effectively free); **Lite** counts distinct inline text.
- **`get_top_queries_by_cpu` (both twins)** emits `distinct_texts`, plus a `text_note` only when >1: the shown text is one representative — attribute per-caller work via `sql_handle`/`OBJECT_DEFINITION` before naming a caller. `0` = only legacy rows predating the digest column (honest unknown; ages out with 4-day raw retention).
- Tool descriptions updated in both twins.

Stage 2 (tracked on #2012) captures `dm_exec_sql_text.objectid` at collection so proc-hosted callees can be **grouped apart** rather than annotated — that's a collector column + migration, so it gets its own PR.

## How was this tested?

- The collision itself was reproduced live on SQL Server 2022 before any code was written (two wrapper procs, two different callees, one `query_hash` — table in #2012).
- **New live Postgres test** (`[Collection("live-postgres")]`): plants the collision shape (one hash, two digests/texts), a single-text hash, and a legacy NULL-digest hash; pins `distinct_texts` = 2/1/0 through both the reader record and the tool's JSON envelope, and that `text_note` fires only on the blend.
- **Lite DuckDB twin fact** added to the existing top-queries reader test class (blend = 2, stable = 1).
- Darling service builds locally, 0 warnings; Lite compiles in CI.
- Ordinal-shift risk in Lite's `r.*` reader mapping handled and covered by the same reader tests.

Dogfood plan after merge: nightly → box upgrade → `get_top_queries_by_cpu` against the fleet servers where the real INSERT...EXEC collisions live (the triage that motivated this), confirming `distinct_texts > 1` surfaces on the known hashes.

- [x] Lite
- [x] Darling
- [x] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## Checklist


erikdarlingdata:
Round-1 CI failure was a test-helper namespace split (Darling.Tests vs PerformanceMonitor.Darling.Tests — usings added). Both nits taken: Lite's description now states its own zero-case ("0 = no stored text for the group" — Lite has no digest column to predate), and Lite's envelope wiring got its own ungated test with the ServerManager-resolver fixture pattern, since the tool body is a twin copy and its Darling counterpart's envelope pin is DARLING_TEST_PG-gated.

PR #2014, opened 2026-08-03
Server search box across all three surfaces (stage 1 of #2008)

jakemorgangit:
Stage 1 of #2008 (server search), as you greenlit — self-contained, no schema, and it slots into the `FleetView.Rebuild` seam that was reserved for exactly this. A live search box on the server list across all three surfaces, filtering by server name (and by tag name in the Viewer, where tags exist).

### One shared rule, three surfaces
The match logic is a single pure helper, `ServerOverviewFilter.Matches` in `PerformanceMonitor.Ui` (the same home as `ServerOverviewSort`) — case-insensitive substring over any supplied field. Both desktop apps call it; the web viewer mirrors the same rule in JavaScript. Pinned identically in both test suites, the way the sort helper is.

### Per surface
- **Viewer** — search box on the sidebar, wired into `FleetView.SetSearch` → the reserved `Rebuild` seam, so grouping, favourites, and selection all follow with no consumer changes. Matches server name **and** any tag assigned to the server. Tag groups with no match anywhere in their subtree drop out; a match inside a collapsed group is revealed (every group renders expanded while a search is active) and the persisted collapse state is restored untouched when the box is cleared; the sidebar count reads "N of total" while filtered.
- **Lite** — search box beside the existing Overview sort control. Filter and sort compose through a single `ApplyOverviewView`, and alerts still run over the **whole** fleet, never the filtered view (a search narrowing the screen must not silence alerts for hidden servers).
- **Web** — a live name filter on the fleet page beside the sort control, with a "No servers match …" empty state. The term persists across the 60s re-render exactly like the sort choice does.

### Not in this PR (per your staging)
The tag-editing grant is untouched — #2011 already resolved that the right way (read-only gate on the admin seat). Tag colour + pills (2a) and Lite/web tag parity (2b) will follow as separate PRs.

### Tests + verification
- New: `ServerOverviewFilterTests` in both suites; `FleetView` search cases (name/tag match, subtree suppression, nested-ancestor kept, collapse reveal + restore).
- Build clean (0 warnings / 0 errors); **Darling 4008** pass, **Lite 2070** pass.
- The web filter (the one surface with no automated harness) was driven live in a browser against a stub `/api/fleet`: `prod` narrows to the two prod servers, `zzz` shows the empty state, clearing restores all three.
- No schema change and no new store reads — filtering is an in-memory pass over the fleet already loaded.

Based on `dev`; CHANGELOG entry added under Added with the `#2008` bottom link ref.


erikdarlingdata (approved):
Reviewed the full diff — this is exactly the Stage 1 that was greenlit, and the details hold up under scrutiny:

- **Shared rule**: `ServerOverviewFilter` is pure, `OrdinalIgnoreCase` (no culture trap), whitespace-normalized to a match-everything sentinel, selector-parameterized like `ServerOverviewSort`, and pinned identically in both suites.
- **Web**: the empty-state term renders through `el()`'s `text` prop (textContent — the helper bans `html` outright), so no injection path; term persistence mirrors the sort control's pattern.
- **Lite**: the alert loop deliberately iterates the UNFILTERED fleet — the exact trap a display filter invites — and filter+sort compose through one `ApplyOverviewView`.
- **Viewer**: the reserved `Rebuild` seam, memoized + cycle-guarded subtree matching (same idiom as the existing tag DFS), whole-branch drop for matchless subtrees, and the `IsExpanded` overlay that reveals matches without ever touching the persisted `_collapsed` set. All six semantic cases pinned in `FleetViewTests`.

CI green across the board. Merging — 2a (colour + pills) and 2b (Lite/web parity) welcome as follow-ups per the staging.

PR #2015, opened 2026-08-03
Tag colour + Overview pills in the Viewer (stage 2a of #2008)

jakemorgangit:
Stage 2a of #2008, per the staging you laid out — a small, self-contained follow-up to the search PR (#2014). Tags gain a colour, and a server's tags show as coloured pills on the Viewer's Overview cards.

### Store / schema
- **V50 migration**: one nullable `colour` column on `config.server_tags` (created back in V32), added additively with no backfill — existing tags stay NULL and render as a neutral pill until coloured. `StorageVersion` → 50, and the viewer's connect-time schema probe gains the V50 rung (newest arm + `server_tags.colour` sentinel + a cap test, mirroring the V49 rung).
- **Shared `TagColorPalette`** (`PerformanceMonitor.Ui`, alongside `ServerOverviewSort`/`Filter`): a fixed set of theme-safe swatches plus a deterministic `ForTagId` rotation, so a new tag gets a stable, reproducible colour — re-creating the same tags gives the same colours. Pure and pinned in both suites.
- `DarlingTag` carries `Colour`. `CreateServerTagAsync` auto-assigns the id-derived palette colour at creation; `SetServerTagColorAsync` writes a user override. Both go through `ExecuteWriteAsync`, so a read-only seat degrades to `ViewerReadOnlyException`.

### UI
- **Manage Tags** gains a **Colour…** button and a swatch picker, and shows each tag's swatch in the tree. The whole editor stays behind #2011's read-only gate — its constructor disables the content tree on a read-only seat, so the new control is inert automatically (no new per-control gate).
- **Overview cards** render a server's tags as coloured pills — neutral when a tag has no colour, hidden when a server has none. The fill is the stored colour and the label a constant light, so a pill looks the same in the light and dark themes; pills re-stamp when tags reload.

### Not in this PR
Lite/web tag parity (2b) — DuckDB tag tables, a manage-tags UI, tree grouping + pills, and `/api/fleet` gaining tag data — is the big lift and follows as its own issue/PR, as agreed.

### Tests
`TagColorPalette` (both suites), the V50 migration + colour-SQL pins, the probe V50 rung + cap, and the version-pin bumps. Build clean (0 warnings / 0 errors); Darling 4021, Lite 2080 pass. CHANGELOG entry added under Added with the bottom link ref.


erikdarlingdata (approved):
Reviewed the full diff — disciplined stage 2a, and it pre-answers the questions a colour column invites:

- **Store**: V50 additive nullable (`ADD COLUMN IF NOT EXISTS`), no backfill, correct numbering against dev, probe gains the newest-first V50 rung with the cap rationale documented (the tag SELECT names `colour`, so a healthy V50 store must not trip the gate at 49).
- **Palette**: pure, WPF-free `#RRGGBB` strings, deterministic id rotation (stable colours across re-creation), pinned in both suites — and `IsValidStoredColour` means a hand-edited store value degrades to a neutral pill instead of reaching `ColorConverter` as garbage; `TagColorBrushes.Fill` actually routes through it, with frozen-brush caching per the converter idiom.
- **Writes** ride the existing insufficient-privilege-aware paths, and the Colour picker inherits #2011's whole-content-tree read-only gate with no new per-control gating to drift.
- Pills restamp on tag edits without a full reload; constant light label mirrors the severity-dot contrast approach in both themes.

CI green, merging. 2b (Lite/web parity) welcome as its own issue/PR when you get to it — and thanks for the fast turnaround on the whole #2008 arc.

Issue #1568, opened 2026-07-18
Queries pane: attribute Top Queries rows to a named module (procedure/function) or ad hoc — Lite + Darling viewer


erikdarlingdata:
## Ask

In the Queries pane, **Top Queries by Duration** should carry a column identifying whether each statement is **ad hoc** or belongs to a **named module** (procedure, function, trigger) — with the module name when attributable. The full Dashboard already does this; Lite and the Darling viewer do not.

## How the Dashboard does it (the exemplar)

- `install/10_collect_procedure_stats.sql` collects procedures/triggers/functions per database with `object_name = OBJECT_NAME(x.object_id, x.database_id)` + `schema_name` + `database_name` + `sql_handle` (procedures :285, triggers :357-358, functions :532).
- It then ties statements to modules by handle (:788-802): `UPDATE collect.query_stats SET object_type = ISNULL(ps.object_type, 'STATEMENT'), schema_name = ISNULL(ps.schema_name, N'N/A'), object_name = ISNULL(ps.object_name, N'N/A') ... LEFT JOIN collect.procedure_stats AS ps ON ps.sql_handle = qs.sql_handle` — statements whose `sql_handle` matches a cached module inherit its identity; everything else is a plain `STATEMENT` (ad hoc).
- The `database_id` half of Erik's note matters because `OBJECT_NAME(object_id)` without the module's own `database_id` resolves against the wrong catalog — the Dashboard always passes both.

## Current state in Lite/Darling (verified)

- Shared `PerformanceMonitor.Collectors/QueryStatsCollector.cs` reads `dm_exec_query_stats` and already `OUTER APPLY sys.dm_exec_sql_text(qs.sql_handle)` (:151, :175) — but does **not** select `st.objectid`/`st.dbid`, and the row schema has no object columns at all.
- Shared `PerformanceMonitor.Collectors/ProcedureStatsCollector.cs` **already collects everything needed for the stitch**: `database_name` (:75), `object_name = OBJECT_NAME(s.object_id, s.database_id)` (:77, :120-121, :174-176), and `sql_handle = CONVERT(varchar(130), s.sql_handle, 1)` (:98, :153, :197) — the **same normalized handle format** `QueryStatsCollector` stores (:120-121). The join key already lines up in both stores.
[...]

PR #2017, opened 2026-08-03
ProcStats comparison grid gains the representative-statement text column (#1981)

### LINKED ISSUE #1981: ProcStatsComparisonGrid carries no query text column while the other two comparison grids do
Found by the #1949 census, out of that issue's scope. The Query Stats and Query Store comparison grids both carry query text; the Procedure Stats comparison grid does not (procedure NAME is arguably the identity there, but the inconsistency is real and symmetric across both apps). Decide whether to add text (the procedure's definition/statement) or record the asymmetry as deliberate.

erikdarlingdata:
## What does this PR do?

Fixes #1981 (the #1949 census asymmetry, per Erik's call to add text rather than record the asymmetry as deliberate). The Procedure Stats comparison grid in **both apps** now carries a text column like its Query Stats / Query Store siblings.

Design honesty note: `procedure_stats` stores no text of its own, so the column shows a **representative statement** — the latest `query_stats` text sharing the module's normalized `sql_handle` (the exact join #1568's module attribution proved lines up across both stores). The header says "Statement (representative)", not "Definition": it's one statement the plan cache held, for orientation and copy-paste, not for diffing procedure versions. Both period CTEs carry `MAX(sql_handle)` and the LATERAL coalesces them, so a GONE row (baseline-only) still resolves its text. Darling reads through `v_query_stats` (#1767 dimension); Lite through its own view.

## How was this tested?

- Shared `PerformanceMonitor.Ui` builds locally (0 warnings); the WPF apps and both suites compile in CI, where the existing comparison-reader tests exercise the changed SQL shape end-to-end against live stores.
- The sql_handle join semantics are the already-pinned #1568 path (both stores persist the identical normalized handle text) — no new join assumptions introduced.
- Visual verification of the new column is Erik's Viewer pass, batched with today's pills/search artifacts.

**Note**: #1984 part 2 (the PVS trend chart) was originally scoped for this same batch — it's a genuine ScottPlot feature and follows as its own PR immediately after, rather than padding this one.

- [x] Lite
- [x] Darling
- [ ] Lite Tests
- [ ] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## Checklist


erikdarlingdata:
Both taken: the mirrored `idx_query_stats_server_handle_time` joins the PgTableTuning set (10 statements, count pin updated) — the LATERAL's probe shape now has the same support its procedure_stats twin always had, and the build stays cheap because raw retention bounds query_stats to ~4 days of chunks. DuckDB gets no twin on purpose: Lite-scale columnar scans serve the probe fine without index maintenance. Spacing nit fixed too.

PR #2018, opened 2026-08-03
PVS trend chart on the FinOps Version Store tab, both apps (#1984 part 2)

erikdarlingdata:
## What does this PR do?

Completes #1984 (the alert half shipped as the PVS pressure alert; this is the deferred trend chart). Above the PVS grid in **both apps**: the last 7 days of hourly `pvs_stats` for the **top-5 databases by current PVS size** — one line per database, legend labels carrying each database's latest %-of-database (per-point ratio from the same denominator the grid uses, so the surfaces can't disagree). Erik's approved series shape: off-row MB as the lines, %-of-DB on the labels — both numbers on one chart instead of two stacked plots. Hides entirely when there's no PVS history (no dead axis on ADR-less servers); the top-5 cap keeps a 90-day series readable while the grid still lists every database.

- **Readers**: twinned `GetPvsTrendAsync` (Lite original + Darling port — same top-N pin at the newest collection, same columns/ordering), reading `v_pvs_stats`.
- **Charts**: the QueryTrends ScottPlot idiom exactly — `TimeSeries` over OADate, `ChartPalette.CyclingColor` per series (stable across redraws), `StyleScatter`, `ChartStyle.ApplyThemeToChart`, datetime bottom axis. Loaded alongside the existing grid refresh on the PVS sub-tab.

## How was this tested?

- **Both WPF apps compiled locally** (`-p:EnableWindowsTargeting=true` on macOS): 0 warnings, 0 errors — this also caught a `DateTime[]` vs OADate `TimeSeries` overload mistake before CI ever saw it, and means Viewer/Lite XAML changes are no longer CI-blind from this dev machine.
- The trend SQL is the latest-read's shape plus a window filter — same view, same twin discipline; CI's suites compile both apps and run the reader-adjacent tests.
- Visual verification rides Erik's next Viewer session (with the search box, pills, tag gate, self-test button, and #2017's comparison column all in the same artifact).

- [x] Lite
- [x] Darling
- [ ] Lite Tests
- [ ] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## Checklist


erikdarlingdata:
All three taken: Lite's X axis now goes through ServerTimeHelper.ToServerTime like every other Lite chart (Darling's already used ViewerTimeHelper.ForDisplay), and the PvsTrendPoint records moved below the PvsStatsRow classes in both apps — the doc-hygiene guard was right that my insertion split summary from class. Both WPF apps rebuilt locally clean.

Issue #2022, opened 2026-08-03
Query Store phase 2: newest-first backfill worker for the history the live path never takes


erikdarlingdata:
Phase 2 of #1960, split out at phase-1 build time per the plan posted there. Phase 1 (oldest-first byte-budgeted shipping, watermark-exact resume) makes the LIVE path hole-free, but two bounded windows still discard history by design:

- **First contact** takes only the trailing 60 minutes of a ~30-day Query Store catalog (`BuildCutoffParameters`'s no-watermark fallback).
- **Post-outage catch-up** is clamped to 24h (`WatermarkPolicy.ClampCatchup`, the #1556 incident fix), logged as a bounded hole.

The phase-2 worker backfills both with one mechanism:

- Walks **newest-first** from the live watermark's floor (the most recent history is the most valuable — the onboarding case wants yesterday before it wants week-4).
- Runs on its **own budgeted tick**, so backfill can never delay or starve the main collection cadence; the same per-database byte budget bounds each backfill slice.
- **Staged at the raw-retention horizon first** (4 days): backfilling deeper than raw retention interacts with rollup/CAGG refresh-window routing (whether the hourly/daily tiers would even materialize rows landed that far in the past), so the deeper stage is a separate decision after the horizon-bounded version proves out.
- Write path is the same COPY + dimension-upsert batch the live path uses; backfilled rows carry their own interval identity, and readers already dedupe re-shipped interval rows (the #1907 HAVING re-ships growing intervals every cycle).

Design constraint carried over from #1960: the backfill worker must leave the LIVE watermark alone — it fills strictly BEHIND the live path's floor, so the two can never race for the same boundary.

erikdarlingdata:
Shipped in PR #2056 (merged to dev). The worker fills both windows with one mechanism, as specified:

- **First-contact tail**: derived boundary (stored MIN(last_execution_time) per server+db — the mirror of the live MAX() watermark), newest-first byte-budgeted slices strictly below it; phase 1's tie-group completion makes the strict bounds hole-free in both directions. Pre-existing stores mark themselves done without shipping a row, so fleet deploy is a steady-state no-op.
- **24h-clamp holes**: recorded by the runner at clamp time under the worker's own collector_state name ((raw, clamped), merged wider on repeat outages), serviced newest-first, shrunk as slices land. query_store still declares no StateKeys.
[...]

Issue #1556, opened 2026-07-17
Darling: unbounded memory growth in the collector row accumulation — query_store OOM cascades to Postgres commit-limit failures and process crash


erikdarlingdata:
Field report (24-server fleet, 16GB host, nightly 3.1.0-nightly.20260717 post-#1553): service memory grew 0 → ~13GB in under 9 minutes, OutOfMemoryException in query_store collection (caught, logged, growth continued), then system-wide commit-limit exhaustion — bundled Postgres threw `could not reattach to shared memory (error code 1455)` / `database system is in recovery mode`, RDP and SSM on the box degraded, and the service process died twice (exit 1067, no WER dump). Reproduced on two consecutive restarts.

## Root cause (grounded)

1. **Unbounded accumulation across databases**: `DarlingCollectorRunner.RunAsync` creates ONE `List<TRow>` per collector run and every database's rows accumulate into it; storage happens only after the full database loop (`DarlingCollectorRunner.cs:262-278`). A server's entire per-database haul is held in memory at once.
2. **The per-database catch retains the accumulation**: the `Failed to collect {Collector} from [{Database}]` handler (`:272-277` — the exact line the field log captured) swallows the OOM and CONTINUES the loop with the already-accumulated list fully retained.
3. **The query_store read is unbounded and plan-heavy**: no TOP/row cap ("no size guard" per its own comment), and `CONVERT(nvarchar(max), qsp.query_plan)` is selected on EVERY runtime-stats interval row — the same plan XML duplicated per interval (`QueryStoreCollector.cs:347-353, :430-445`). The store-derived watermark makes each restart after an outage re-read everything since the last successful collect.
4. **#1553's 4-wide sweep multiplies the peak** ×4 (four servers' hauls concurrently). Exacerbation, not cause — the reporter's hypothesis was exactly right.

## Fix directions

- Flush-and-clear rows per database (or per N rows) instead of accumulating the whole run — bounds peak to one batch.
- Bound the query_store read (row cap per database per cycle, logged when hit) and stop duplicating plan XML per interval row (capture it once per plan per cycle — newest interval row only).
- OOM-specific handling in the item loop: clear the accumulation and ABORT the collector run rather than continuing.
- Cap watermark catch-up staleness so post-outage restarts do not re-read unbounded history.
- Viewer: surface a store-unreachable indicator instead of an empty screen (reporter ask 5).
[...]

PR #2023, opened 2026-08-03
Ship Query Store oldest-first on a byte budget with watermark-exact resume (#1960 phase 1)

### LINKED ISSUE #1960: Query Store ships GB-scale plan XML on catch-up and re-ships hot plans every cycle - add a per-database plan_id watermark
The third finding from the prod-sql-use2 field investigation, filed as a measured-decision marker per the reporter's own recommendation: NOT worth acting on blind.

The Query Store collector is already defended against the below-TOP TVF defect its siblings had (its expensive payloads are catalog COLUMNS via 1:1 joins, plan text dedupes to one copy per plan_id via the ROW_NUMBER CASE that composes with the TOP by design, and the @cutoff_time watermark bounds each cycle). The one theoretical exposure: the ROW_NUMBER window forces the joins plus the rn=1 plan-text conversion to evaluate for ALL qualifying rows BEFORE the TOP - so a pathological burst converts XML for every distinct plan in the window, including rows the TOP discards.

Burst-only, and unvalidatable on the reporting fleet (replicas serve zero QS rows). Tripwire: when a QS-BEARING store shows query_store sweep times out of family with its row counts, measure this exposure there first - the reporter's ab_test.py methodology (interleaved pairs, warmup discarded, row-identity parity) is the template. Do not restructure the QS read speculatively; its current shape's properties (dedup-compose-with-TOP) are load-bearing and pinned.

erikdarlingdata:
## Summary

Phase 1 of #1960, built to the reconstructed plan posted on the issue. The #1556 per-database bounds (TOP 50000 + the client text-byte budget) stopped one database from ballooning the process, but they were hole-makers: rows were read newest-first, the watermark derives from the newest row STORED, so everything a bounded cycle didn't reach fell permanently behind the strict `> @cutoff_time` — logged as "oldest rows dropped."

## What changed

- **Oldest-first shipping** (`ORDER BY last_execution_time ASC`): the derived watermark now sits exactly at the shipped boundary when a cycle is bounded, so the next cycle resumes there. A bound costs catch-up latency, never data.
- **Two tie-group guards** make the boundary exact — required because raw rows are COPY-appended (no upsert), so a split tie group would strand its unshipped half behind the strict comparison forever: `TOP ... WITH TIES` server-side, and the client budget finishes the boundary tie group before stopping (tied rows are adjacent under ASC; overshoot is one group).
- **Budget 256 MB → 64 MB**: with cuts hole-free, the smaller budget is safe — the July field catalog (~1.33 GB fresh) converges in ~21 bounded cycles, and the #1553 4-wide sweep's transient peak drops ~1 GB → ~256 MB.
- **Observability** (plan point 5): both hosts' bounded-cycle WARNINGs now report megabytes shipped and the boundary reached, via two new `CollectorContext` signals (`PerItemTextBytesShipped`, `PerItemShippedBoundary`) written alongside the existing truncation flag on every budgeted read.
- **Plan-XML dedupe** keeps its newest-interval-carries-the-plan rule (partition ORDER BY stays DESC); during a multi-cycle catch-up a plan's XML can trail its interval rows by a cycle, then lands when the window reaches its newest row. Steady-state unaffected.

## The design check the plan required

The plan flagged one build-time check: whether the write path's upsert keying makes partial-interval re-ship idempotent. Answer: raw `query_store_stats` rows are **binary-COPY appended, no upsert** — but re-shipped interval rows are already the steady-state (#1907's interval-grain HAVING re-ships a growing interval every cycle), so readers already dedupe by latest-per-key. What must NOT happen is a stranded tie, hence the exact boundary via tie-group completion rather than reliance on idempotency.

## Unchanged, deliberately

The 60-minute first-contact window and the 24h catch-up clamp. Backfilling the ~30 days both leave behind is phase 2 — the newest-first backfill worker on its own budgeted tick, filed as #2022 per the plan.

## Testing

- SQL pins updated: `WITH TIES` + ASC in both capture modes and both path shapes; the rn=1 plan-dedupe pin unchanged.
- New pure pins on the production 64 MB constant (no test seam): budget cut completes the boundary tie group and stops at the first row past it, boundary/bytes signals match the kept rows exactly, every kept row ≤ boundary (the never-a-hole invariant); a null-boundary trip stops at the next row.
- Signal-reset test extended to the two new context signals.
- Both apps + Lite.Tests build clean locally (`EnableWindowsTargeting`); the WPF-targeting test assemblies run on CI's Windows runners.

Closes #1960 (phase 1; phase 2 tracked in #2022).


erikdarlingdata:
Review responses:

**CHANGELOG bug** — real, fixed in 9186cc04: the insertion edit matched only the front half of #2012's one-line bullet and fused its remainder onto the new entry. Both bullets stand separately again.

**Tie-group overshoot** — considered, leaving as-is, with the reasoning on the record: a tie group needs N DISTINCT plans whose per-(plan, interval, execution_type) MAX(last_execution_time) lands on the same datetime2(7) tick — 100ns resolution, one row per plan per interval after the #1907 aggregation. A burst of distinct ad-hoc queries in the same flush window doesn't produce this: each plan's last execution is its own instant, and the flush stamps don't overwrite last_execution_time. An engineered worst case overshoots for one cycle, transiently, and the bounded-cycle WARNING reports the true bytes shipped (the new signal is the actual count, not the budget). A defensive hard cap on the tie group would reopen the exact stranded-rows hole this PR closes, which is the worse trade.

**Fully-qualified Select/List in the new tests** — deliberately copied from the neighboring tests: the pre-existing self-marker and ReadAsync tests in this file use `System.Linq.Enumerable.Select(rows, ...)` and `new System.Collections.Generic.List<QueryStoreCollector.Row>()` verbatim, so the new ones match their immediate surroundings.

**Deprecated Dashboard proc** — deliberate scope call, same as the plan-text dedupe before it: `deprecated/` doesn't get behavior changes.

Issue #2020, opened 2026-08-03
Stage 2b of the tag/search proposal: Lite + web tag parity (tree, pills, /api/fleet tags)


erikdarlingdata:
The remaining stage of #2008, split out so the shipped umbrella can close. Stages 1 (search box on all three surfaces, #2014) and 2a (tag colour + Overview pills in the Viewer, #2015) shipped 2026-08-03.

Scope, per the accepted proposal (@jakemorgangit's, with Erik's design answers on the thread):

- **Lite**: a tag model mirroring Darling's hierarchical many-to-many shape (`server_tags` tree + `server_tag_map`, plus the V50 colour column equivalent) in DuckDB — parity-of-shape so the twins stay honest and future import/export stays possible; a Manage Tags UI behind the same admin-seat conventions; sidebar tree grouping (opt-in drill-down) and coloured pills on the Overview; the search box already matches tag names via the shared `ServerOverviewFilter` once tags exist.
- **Web**: `/api/fleet` gains tag data (read-only — assignment stays in the desktop apps, per the role model settled in #2011); pills on the fleet cards; tree/group rendering; the existing search input matches tag names.

House notes: `TagColorPalette` and `ServerOverviewFilter` are already shared and pinned in both suites; the Viewer's `FleetView` projection semantics (subtree drop, collapse-reveal) are the behavioral reference, pinned in `FleetViewTests`.

jakemorgangit:
Picking this up. Planning to ship it in two PRs, Lite first:

- **2b-i — Lite** (the lift): the DuckDB tag model mirroring Darling's shape (`server_tags` tree + `server_tag_map` + the colour column), a Manage Tags UI ported from the Viewer's, sidebar tree grouping + Overview pills, and wiring the existing search to match tag names via the shared `ServerOverviewFilter`.
- **2b-ii — Web** (read-only): `/api/fleet` gains tag data, then pills + tag-matching search + tree/group rendering on the fleet page — web tree included per your scope.

FleetView's projection semantics (subtree drop, collapse-reveal) are the behavioural reference I'll mirror in Lite. Starting on 2b-i now; will open it against dev when it's green.

erikdarlingdata:
All three stages shipped, all by @jakemorgangit:

- **2b-i** — Lite tag model + Manage Tags editor, colour, pills, search-by-tag (PR #2026)
- **2b-i-b** — Lite sidebar tag tree with inline assign + tag CRUD (PR #2040)
[...]

PR #2026, opened 2026-08-04
Lite server tags: Manage Tags editor, colour, pills, search-by-tag (#2020 stage 2b-i)

jakemorgangit:
Stage 2b-i of #2020 (Lite tag parity), split into the additive half you'd expect first: the tag store, the Manage Tags editor, colour, Overview pills, and search-by-tag. The **sidebar tree-grouping** drill-down is deliberately its own follow-up PR (2b-i-b) — it's a `ListView`→`ListBox` refactor that reroutes every sidebar read-back site, a different risk profile from these additive bits, so it shouldn't ride with them. Web parity is 2b-ii.

### Store — the DuckDB twin of the Viewer's tables
- `server_tags` (hierarchical: `parent_id` null = root, + the `#RRGGBB` `colour` column) and `server_tag_map` (many-to-many, keyed on the same `GetServerId` hash the Overview cards use). Same shape/semantics as Darling's `config.server_tags` (V32) + the V50 colour column.
- Where DuckDB differs from Postgres, the invariants move into code: **ids assigned in C#** (no IDENTITY), **cascade-on-delete** via a recursive walk (no self-FK), **unique-name-per-parent** enforced in the store (no partial index). The two tables are created idempotently in `Schema.GetAllTableStatements()` on next launch — like every other Lite config table (the `config_database_state_expected` precedent), so an upgraded store gains them with no migration step.
- `LocalDataService.ServerTags`: full CRUD + read + assign/unassign/clear. A new tag's colour is auto-assigned from the shared `TagColorPalette` (rotated by id, stable across re-creation). **Lite has no read-only seat** (single connection, the user's own credentials), so there's no #2011-style gate to add.

### UI — ported from the merged Viewer (2a)
- **Manage Tags** window: the tag tree with colour swatches, New Tag / New Child / Rename / **Colour…** / Delete, and a searchable per-tag server-assignment checklist. Plus the `TagColorDialog` swatch picker and `TagNameDialog`. A new **Manage Tags** footer button opens it and refreshes the Overview on close.
- **Overview pills**: each server's tags render as coloured chips (neutral when uncoloured, hidden when a server has none) — the same `ServerTagPill` + `TagColorBrushes` shape as the Viewer.
- **Search by tag**: the existing Overview search now matches tag names as well as server names, via the shared `ServerOverviewFilter` (multi-field, already unit-tested).

### Tests
`LiteServerTagsStoreTests` — real-DuckDB round-trips for the store's DuckDB-specific mechanics: the subtree cascade, duplicate-name rejection, `ON CONFLICT` idempotent bulk assign, palette-colour-by-id, and colour clear (the exact SQL a mock wouldn't catch). Schema table-count pin `50 → 52`. Build clean (0 warnings / 0 errors); Lite 2087, Darling 4021 pass.


erikdarlingdata (approved):
Reviewed against source with the store checked line-by-line against the Darling twin:

- **Duplicate-name semantics match exactly**: `COALESCE(parent_id, 0) = COALESCE($1, 0) AND lower(name) = lower($2)` is the same predicate as V32's `ux_server_tags_parent_name` expression index — case-insensitive, per-parent, roots collapsed to 0. The invariant genuinely moved into code rather than being approximated.
- **Cascade order is right** (map rows first, then tags), the recursive-CTE walk is correct, and the inline IN list is safe (DB-sourced ints only).
- **Editor enforces what the store can't**: MaxDepth=4 + cycle-safety live in ManageTagsWindow, mirroring the viewer's split.
- **Schema is cleanly additive** — rides the existing CREATE-IF-NOT-EXISTS startup enumeration, so upgraded stores just grow the tables; the schema-test pins were updated in the same PR.
- **Palette/brush parity**: shared `TagColorPalette` for values + validation, frozen-brush cache like the viewer's twin.
- Store tests cover ids/colour assignment, duplicate rejection, round-trips, idempotent assign, cascade, and the clear-for-server read.

One functional gap, merging anyway and fixing in an immediate follow-up since maintainer-edit pushes aren't available to my token: **`ClearServerTagsForServerAsync` has no call site** — its doc says 'called when a server is removed,' but the remove path (MainWindow's `ServerContextMenu_Remove_Click`, which already does the #1696-style deep cleanup) never calls it, so a removed-then-re-added server resurrects its old tags (the id hashes deterministically — the exact scenario the method exists for). The wire-up lands right after this merge.

Also noting for a separate issue: ManageServersWindow's delete path skips ALL the deep cleanup (health, AG state, and now tags) — pre-existing, not this PR's doing.

Really nice scope discipline splitting the sidebar refactor out. Thank you!

PR #2035, opened 2026-08-04
Lite: clear a removed server's tag assignments on remove (#2020 follow-up)

jakemorgangit:
The immediate follow-up you asked for on #2026: wire up `ClearServerTagsForServerAsync` so removing a Lite server drops its tag assignments.

## The gap

As you flagged, the method existed but had no call site. Lite's server id is a deterministic hash of host+database+intent, so remove-then-re-add returns the **same id** and any `server_tag_map` rows keyed on it silently reattach â€” the exact scenario the method was written for.

## The fix

`ServerContextMenu_Remove_Click` now clears the server's tag assignments alongside the two sibling cleanups already on that path, which drop deterministic-id state on remove for the identical reason:

- the `#1696` AG edge-state `Forget` (or a re-add pages a phantom failover against the old role), and
- the health-cache clear.

The tag clear sits right next to them with the same rationale in the comment. It's best-effort and logged on failure, so a store hiccup can't block the removal itself.

## Scope / testing

- The store method was already covered by the tag round-trip tests (`LiteServerTagsStoreTests`); the gap was purely the missing call, which lives in a WPF event handler â€” so this is a wire-up, not new store behaviour.
- Build clean; full Lite suite **2089/2089** green.
- CHANGELOG `Fixed` entry added.

Kept tightly scoped to the flagged gap. Noted in the changelog (and here) that `ManageServersWindow`'s delete path skips **all** this deep cleanup â€” health, AG state, and now tags â€” which is pre-existing and broader; happy to take that as its own issue if you'd like.


jakemorgangit:
Crossed in the mail with #2032 — I had this in flight before I saw you'd already wired it up. Same fix, right down to the placement beside the AG forget and the deterministic-id rationale, so nothing to add here. Closing in favour of yours. Thanks!

Issue #2031, opened 2026-08-04
Surface "server silenced" state: a muted-bell indicator + make Silence/Unsilence mutually exclusive


jakemorgangit:
A server's alerts can be silenced (a whole-server mute) while it keeps being monitored — but nothing shows that a server *is* silenced, so it's easy to forget one is muted and miss real alerts. Flagging what already exists so this is *surface-an-existing-state*, not new machinery.

### What's there today
- **Silence works on both desktop apps.** Viewer: a store-side whole-server `MuteRule` (`IsWholeServerSilence` detects it) the service honours. Lite: `AlertStateService.IsServerSilenced(serverId)`, persisted. Both already *know* the silenced state.
- **No indicator anywhere.** The status dot is health-only; the silenced state is never rendered on the Viewer sidebar, the Lite list, the Overview cards, or the web fleet.
- **Menu:** the Viewer's context menu shows **both** *Silence This Server* and *Unsilence* unconditionally. Lite already disables the inapplicable one.
- **Web** has no silenced concept — `/api/fleet` emits none.

### Proposal
1. **A subtle, greyed-out muted-bell icon (not an emoji), immediately to the right of the status dot**, when a server is silenced — on all three surfaces. Viewer + Lite read their existing silenced state; the web needs `/api/fleet` to expose a read-only `is_silenced` (computed from the mute rules) and render it on the card.
2. **Make Silence / Unsilence mutually exclusive in the Viewer** — show only the applicable action, matching Lite which already gates it.

### Open questions
1. **Hide vs. disable** the inapplicable menu item — Lite disables today; which would you prefer, applied consistently to both apps?
2. **Web** — expose `is_silenced` on `/api/fleet` for display only (no silence action on the least-privilege web seat) — OK?

erikdarlingdata:
Verified all four claims in source (the state machinery, the missing indicator, the Viewer's ungated menu pair, the web's absent concept) — accurate as filed, and framing it as surface-an-existing-state is exactly right. Answers to the open questions:

[...]

PR #2040, opened 2026-08-04
Lite sidebar: group the fleet into its tag tree, with inline assign + tag CRUD (#2020 stage 2b-i-b)

jakemorgangit:
The sidebar half of #2020's Lite tag work (stage **2b-i-b**), riding on top of the merged 2b-i-a. Stage 2b-i-a gave Lite the tag store, Manage Tags editor, colour, Overview pills, and search-by-tag; the sidebar was still a flat server list. It now renders the same **Favorites → tags → Untagged** tree the Darling Viewer's sidebar does.

## What's in it

- **Ported `FleetView` / `FleetRow` / `FleetRowTemplateSelector` into Lite**, retyped over a thin `FleetServer(ServerId = storage-hash, IsFavorite, DisplayName, Connection)` adapter — so the projection keys on the *same deterministic storage id the tag map uses*, and the tree and the Overview pills can never disagree about identity. WPF-free, so the projection is unit-tested without a window (`LiteFleetViewTests`, ~20 invariants mirroring the viewer's `FleetViewTests`).
- **Sidebar `ListView` → `ListBox`** rendering a mixed header+server projection via a template selector; the single row template splits into a header row (chevron + title + `(n)` badge + depth indent) and the existing server card.
- **Full-parity interaction**: inline **Assign Tags** checkable submenu on server rows, and tag CRUD (**New / New Child / Rename / Delete**) on group headers — disabled on the Favorites/Untagged pseudo-groups. Reuses the existing `LocalDataService` tag CRUD; no new store code. No read-only seat to gate (Lite is single-connection), unlike the viewer's #2011 gate.
- **Expand/collapse persists across restarts** via a new `collapsed_fleet_groups` list in `settings.json` (keyed by group, so a tag rename keeps its state) — the Lite twin of the viewer's `ViewerPreferences.CollapsedFleetGroups`.
- Opt-in exactly like the viewer: with no tags defined the sidebar stays the flat list it always was.
- Small related fix: the Overview search box's placeholder/tooltip now say **server name / tag**, since 2b-i already made that box match tag names.

## Scope discipline

Lite-only — the Viewer is untouched. The port was smaller and lower-risk than the viewer's original refactor because **only three sites coupled to the sidebar** (the bind, the double-click, and `GetServerFromContextMenu`); every "what servers exist?" consumer already reads `ServerManager.GetAllServers()` rather than the bound list, so the projection changes only what the sidebar *shows*. No Silence/Unsilence on the sidebar — that's #2031, which rides after this per your note on #2037.

## Testing

- Full **solution** build clean.
- **2125** Lite tests green (28 new `LiteFleetViewTests` among them: opt-in flat default, Favorites/nesting/Untagged ordering, collapse hiding descendants, multi-tag duplication, selection skipping headers, cycle / dangling-parent safety, and the collapse-key persistence round-trip).

This is the last of the 2b-i work; web parity (`/api/fleet` + `get_fleet_overview`) is 2b-ii.


erikdarlingdata (approved):
Reviewed with the load-bearing invariants checked against source:

- **Identity is exactly right**: `ToFleetServer` keys on `GetDeterministicHashCode(GetServerNameForStorage(s))` — the same composition the tag map, the Overview pill stamping, and the remove-path cleanup all use, so the tree and the pills cannot disagree about which server is which.
- **The read-back rewiring is complete**: a full sweep of `ServerListView` usages and sidebar `DataContext`/`SelectedItem` casts finds every site retyped to the projection (double-click guards header rows, the context-menu helper unwraps `FleetServerRow`); the remaining `ServerConnection` casts in the codebase belong to FinOps/Recommendations' own selector combos, untouched by design.
- **The 2b-i remove-path wire-up survives the refactor** (`ClearServerTagsForServerAsync` still called with the same id composition).
- **Selection re-entry is handled properly**: the suppression guard wraps every programmatic rebind/toggle with finally-restore, and header clicks recover the previously selected server rather than stranding selection on a group row.
- Collapse state persisting to settings mirrors the viewer's `CollapsedFleetGroups` — nice parity detail.
- The WPF-free projection with ~20 pinned invariants running in CI is exactly how this control should be tested.

Splitting this from 2b-i-a was the right call and it paid off — clean, reviewable, green. Merging. This also unblocks the two halves that were deliberately parked on it: the Lite muted-bell (#2031) and the rest of the #2020 tail.

PR #2050, opened 2026-08-04
Lite: surface the silenced state with a muted-bell on the sidebar and Overview (#2031, Lite half)

jakemorgangit:
Completes #2031 for Lite — the muted-bell silenced-state indicator, mirroring the Viewer + web halves you shipped in #2037. A whole-server alert silence already worked in Lite (tab badge → Silence), but nothing showed that a server WAS silenced, so a muted server looked identical to a healthy-quiet one.

## What's in it

- **`ServerConnection.IsSilenced` + `SetSilenced(bool)`** — notify-on-flip, the Lite twin of your `DarlingServer.SetSilenced`.
- **`RefreshSilencedIndicators()`** — the Lite twin of `UpdateServerSilencedAsync`: an in-memory read of the persisted silence set (no store hit), keyed on the `ServerConnection` GUID that `AlertStateService` stores. Rides the existing 30-second status poll and every list refresh, and the Silence/Unsilence handlers nudge it so the bell flips instantly.
- **The bell glyph** — your exact vector from #2037 (bell `Path` + diagonal `Line` slash in `ForegroundMutedBrush`, never an emoji), right of the status dot.
- **No sidebar menu change** — Lite's silence lives on the tab badge, which already greys out the inapplicable Silence/Unsilence item (#2031's disable-not-hide, already present in Lite).

## Two things worth your eyes (both deliberate choices, easy to change)

1. **The bell also renders on Lite's Overview cards, not just the sidebar.** #2037 put it only on the Viewer's sidebar (not its Overview cards), so this is a small divergence — but the web already shows it on its fleet cards, and within Lite it felt inconsistent to show silenced state in the sidebar but not the NOC grid. Flagging in case you'd like the Viewer's Overview cards to gain it too for full symmetry. Happy to drop it from the Lite Overview if you'd rather keep strict parity with the Viewer.

2. **Discovered gap (not addressed here, per scope): Lite can only silence a server via the transient tab alert badge**, which is only visible while the server has an active blocking/deadlock alert. So there's no way to proactively silence a *healthy* server in Lite — unlike the Viewer, which has Silence/Unsilence on the server-row context menu. That makes the bell only reachable via the alert-then-silence flow. Kept this PR bell-only as scoped, but wanted to raise it: worth adding a sidebar-row Silence/Unsilence to Lite (matching the Viewer) as a follow-up? Happy to take it if you agree.

## Testing

- Full **solution** build clean.
- **2125** Lite tests green.
- Verified the bell live on both surfaces (sidebar tree + Overview cards) by exercising the real silence path.

This is the Lite half; the Viewer + web halves shipped in #2037.


erikdarlingdata (approved):
Reviewed against source — this is a faithful Lite twin of the Viewer half, and a couple of choices here are genuinely better than parity required:

- `RefreshSilencedIndicators` is an in-memory lookup riding the existing 30-second poll and every list refresh, with instant nudges from the Silence/Unsilence handlers — no store read at all, which is cheaper than the Viewer's mute-rules read (Lite's silence state lives locally, so this is the right shape, not a shortcut).
- The Overview rebind only fires on a REAL flip, so a quiet poll never churns the card list or resets scroll — nice touch.
- Stamping before the sidebar binds means the bell is correct on first render, not after the first tick.
- `IsSilenced` is runtime-only with `[JsonIgnore]`, so servers.json stays clean; the notify-on-flip setter matches `DarlingServer.SetSilenced`.
- The XAML paths the bell through the new FleetRow projection (`Server.Connection.IsSilenced`) with the same vector bell-in-Viewbox as the Viewer — no emoji, per the issue.

One observed-but-not-blocking note: the Overview sync matches summaries by `ServerName` while silence keys on the connection GUID, so two entries for the same host (e.g. a read-intent sibling) would flip together on the Overview. That matches how Overview summaries are already keyed, so it's the existing convention rather than a new gap.

Merging — this completes #2031 across all three surfaces. Thanks for turning it around same-day!

PR #2052, opened 2026-08-04
Web fleet: read-only tag pills, tag tree, and tag-name search (#2020 stage 2b-ii)

jakemorgangit:
Completes the #2020 tag rollout: read-only tag parity on the Darling web fleet (stage **2b-ii**) — all three things the issue's web scope asks for: **pills, a group-by-tag tree, and tag-name search**. The Viewer got tags in 2a, Lite in 2b-i; the web dashboard is the last surface.

## What's in it

**Reader (`DarlingFleetReader`)** — one change serves both `/api/fleet` and `get_fleet_overview`:
- `FleetTagsSql` — a bounded `server_tag_map` → `server_tags` join keyed on the integer `server_id` (1-to-many, so a `ReadCpuAsync`-style dictionary read, not the `is_silenced` `EXISTS` scalar). Per-server `tags` (id/name/colour) on each `FleetServerCard` → the pills.
- `FleetTagForestSql` — the **full tag forest** (id/name/parent_id/sort_order/colour) as a top-level `tags` on the fleet result, so the tree can nest an organisational parent tag that has **no directly-assigned servers** — the desktop `FleetView` is handed the whole tag list for exactly this reason.
- **No new DB grant** — the read-only role already `SELECT`s the config-schema tag tables.

**Colour handling** — raw nullable `#RRGGBB`, neutral fallback; deliberately **not** via `TagColorPalette` (Windows-only WPF `PerformanceMonitor.Ui`; the service builds for Linux). Matches the desktop's neutral-for-uncoloured rule exactly.

**Web (`fleet.js` + `app.css`)**:
- Coloured **pills** below each card's status line.
- A **Group by tag** toggle (shown only when tags exist) rearranges the cards into a **collapsible tag tree** — the web twin of `FleetView`'s projection: forest DFS (child tags before a tag's own servers) then Untagged, collapse hides the whole subtree, multi-tag servers appear under each, cycle/dangling-parent safe. Toggle + collapsed groups persist client-side.
- The existing **search** now matches tag names too, and composes with grouping. No Favorites group (the web has no per-user favourites).

**Display-only**, matching the web seat's read-only philosophy: tags are created/assigned in the Viewer or Lite; the web only shows them.

## Testing
- `FleetTagsSql` + `FleetTagForestSql` SQL-shape pins, both added to the PG-dialect theory; the DTO field pins now cover `tags` (per-card) and the top-level `tags` forest (`parent_id`/`sort_order`).
- Full **solution** build clean; **Darling.Tests 4075 passed**; `fleet.js` `node --check` clean.

This closes out [#2020] across all three surfaces.

---
*(This PR grew from the earlier pills+search cut per the scope note above — it now delivers the full 2b-ii, tree included.)*


jakemorgangit:
Scope correction on this one, owning it up front: I built it as pills + tag-search and deferred the tree/group rendering — but that's part of #2020's web scope (pills + tree/group + tag search), and I'd committed to including it on the issue thread, so quietly narrowing it wasn't mine to decide. I'm completing the tree/group rendering now and will push it onto this branch so the PR delivers the full 2b-ii rather than a partial. Could you hold the merge until it lands (shortly)? I'll update the description when it does. Apologies for the churn.

jakemorgangit:
Pushed the tree/group rendering — this now delivers the full 2b-ii (pills + collapsible tag tree + tag search), so it's complete against #2020's web scope rather than a partial. Description updated. Reader gained a `FleetTagForestSql` (the whole tag forest as a top-level `tags`) so orphan parent tags nest correctly like FleetView; the JS group view mirrors FleetView's DFS + collapse-reveal. Build + Darling.Tests (4075) green.

erikdarlingdata (approved):
Reviewed against the FleetView behavioral reference and the web seat's read-only philosophy — this is a clean completion of #2020.

What I checked:
- **Escaping/XSS**: tag names only ever reach the DOM through `el(..., { text: ... })` (textContent), and the stored colour is format-checked (`^#[0-9a-fA-F]{6}$`) before it touches the style attribute — the one injection vector a stored string could reach is closed.
- **Tree semantics**: walked the `hideBelow` depth-gate through nested/sibling/collapsed-mid-tree cases — collapse correctly hides the whole subtree and re-reveals siblings, matching FleetView's collapse-reveal. `buildTagGroups` is cycle- and dangling-parent-safe (orphans surface as roots rather than vanishing), multi-tag duplication and Untagged-last both match the desktop projection.
- **Server side**: both SQL consts are dialect-clean, bare-name/search_path-consistent with FleetServersSql, read-only, and pinned; the separate forest read (so an organisational parent with no direct servers still nests) is the right call and the doc says why. `BuildRollup`'s optional param keeps existing callers source-compatible; snake_case pins updated.
- **Claim check**: verified 2b-i actually shipped (#2026, #2040 merged) before trusting the "completes the issue" wording — it did, so it does.
- Nice touches worth calling out: localStorage guarded for locked-down browsers, keyboard/ARIA on the group headers, and the neutral-pill fallback matching TagColorBrushes without dragging the WPF assembly into the Linux build.

CI is fully green including the live PG job and the Linux build. Merging.

Issue #2068, opened 2026-08-05
Store self-metrics: persist per-hypertable size, compression ratio, and dimension-table series for capacity forecasting


erikdarlingdata:
Capacity forecasting for the monitor store currently requires ad-hoc archaeology over the TimescaleDB chunk catalog (doable — 1-day chunks make per-day ingest reconstructible — but ephemeral: raw chunks age out in 4 days, so the reconstructable window is always tiny). With **primaries joining the fleet** — which carry Query Store data and plan capture, the store's dominant payloads — the operator needs a real series to extrapolate from.

**Measured motivation (2026-08-05, the production 52-replica store):**
- Total 147 GB, of which `query_plan_dim` alone is **101 GB (69%)** — the single biggest forecasting term, and invisible to any current surface.
- Hypertable compression is excellent (5.1–16.2% of original, 6–36×) but only measurable by hand.
- Compressed daily ingest jumped **~3× around Aug 1** (620 MB/day → ~2.3 GB/day) and nothing recorded it — it was only visible because the chunk catalog still held both eras. In four more days that evidence is gone.

**Proposal**: a small fleet-level self-metrics sweep (the disk-check/compression-check idiom — hourly is plenty) persisting one row per hypertable/dimension per run into a tiny hypertable of its own: total/compressed/uncompressed bytes, chunk counts, and the dimension tables' row counts + total sizes. Surfaces: a store-growth panel (web + viewer) and an MCP read, so "how fast are we growing and what's driving it" is a query, not an expedition. Bounded by its own retention (say 400 days at ~50 rows/hour ≈ nothing).

Also worth including: the derived **per-server ingest rate** (daily bytes ÷ enabled servers) so onboarding N primaries has a measured baseline to multiply — with the explicit caveat that primaries add query_store payloads replicas never ship, so the first primary onboarded becomes the calibration point for that term.

erikdarlingdata:
[...]

Issue #2060, opened 2026-08-05
Findings compute per-query drill-downs but no surface enumerates them — expose finding drill-downs through MCP/REST


erikdarlingdata:
**The gap, from live triage (2026-08-04/05, multi-48):** the PARAMETER_SENSITIVITY finding reported *"the worst-affected plan ran 345.9× more expensive … and 13 plans showed the pattern"* — but `get_analysis_findings` returns advice, co-fired titles, and the time range with **no drill_down field**, and `get_analysis_facts` is scalar observations by contract. The engine computes and persists exactly the list a responder needs (`finding.DrillDown`: parameter-sensitive plans, top spill queries, abnormal CPU plans, blocking chains…), and the alert e-mail proves it exists by flattening its first 3 rows — but no programmatic surface enumerates it.

**The cost in practice:** during the fleet ambient-chain triage (the #2054 evidence), attributing the QUERY_SPILLS leg meant re-deriving the driver queries through `get_top_queries_by_cpu` + hand-crossing against the finding's window — work the finding had already done and thrown away at the surface boundary. An agent or human asking the obvious follow-up — *"which 13 plans?"* — currently has no answer short of the remediation/advice pipeline's 3-row e-mail flatten or independent re-derivation.

**Proposal:** surface the persisted drill-down on the finding read-outs:
- `get_analysis_findings`: an opt-in `include_drilldown` parameter (default off — drill-downs can be bulky and the dedup summary is usually enough), returning the finding's drill-down sections as structured rows, capped per section.
- Same data on the REST twin, and the viewer's finding card already renders it — this is about parity for the programmatic surfaces.
- No engine work: the rows are already in the persisted finding JSON; this is a serialization gate, not a pipeline change.

Related observation while here: the advice/remediation split is working as designed (parameter sensitivity deliberately refuses plan-forcing T-SQL; regression findings get it) — the gap is only that the *evidence list* rides the same pipeline as the *advice* instead of being independently queryable.

erikdarlingdata:
**Premise correction after reading the persistence path**: the drill-down is NOT in the persisted finding — it is deliberately ephemeral (`AnalysisModels.cs`), which is why `get_analysis_findings`' code comment says it's "gone on read-back" while `analyze_server` (live) returns it. So this is a persistence decision, not a serialization gate.

[...]

Issue #2069, opened 2026-08-05
Gzip the plan-XML dimension content at the application level — measured 14.0x vs lz4's 8.9x on live content


erikdarlingdata:
**Decision** (Erik, 2026-08-05): realize compression gains today with app-level gzip. Native TOAST tops out at lz4 in PG 18 (`pglz, lz4` are the only accepted values — verified against the production binary; zstd TOAST is still an in-progress upstream patch), and gzip is not a TOAST option at all, so the gain comes from compressing in the service.

**Measured on live store content** (stratified 400 plans + 2,000 texts, per-value):
| | plan XML | query text |
|---|---|---|
| pglz | 6.6× | 1.7× |
| lz4 (current TOAST) | 8.9× | 1.5× |
| **gzip Optimal (app-level)** | **14.0×** | 2.4× |

Scope is **plan XML only**: `query_plan_dim` is 101 GB (69% of the 147 GB store) while `query_text_dim` is 21 MB — text is not worth reader churn. Projected: **101 GB → ~64 GB**, self-realizing within the ~9-day GC turnover with no backfill/rewrite. Throughput is a non-issue (51 MB in 5s single-threaded PowerShell; the service will do better against ~4 GB/hour raw intake).

**Design constraints, from the code:**
- The #1767 resolving view (`v_query_stats`) serves `query_plan_xml` as SQL `text` via COALESCE — PG cannot gunzip in SQL, so the dim gains a `query_plan_gz bytea` column (migration + the full schema-gate ladder ceremony) and the RESOLVING VIEW exposes presence, not content: plan-fetching readers (`GetQueryStatsPlanXmlAsync`, the MCP plan tools, the alert context builder) fetch the gz column and decompress in C#. The has_query_plan presence flag already tests digest/inline without resolving — extend it to the gz arm.
- Write path: `PayloadDimensionWriter`/batch gzips before upsert (System.IO.Compression, Optimal). Digest stays computed over the UNCOMPRESSED text so identity is stable across the format change.
- Transition: old rows keep text, new rows carry gz — readers COALESCE (decompress gz if present, else text), and the GC window ages the text era out in ~9 days. No migration of existing content.
- Lite is unaffected (no plan capture, no dims).
- Same treatment optional later for `analysis_findings.drill_down_json`/plan XML inside alert ContextJson — out of scope here.

[...]

PR #2070, opened 2026-08-05
#2068: hourly store self-metrics — per-hypertable size, compression, and dimension series for capacity forecasting

erikdarlingdata:
## What / why

Implements #2068: the service now measures its own store hourly and persists the series, so "how fast are we growing and what's driving it" is a query instead of an expedition over the TimescaleDB chunk catalog.

**The measured motivation (2026-08-05, the production 52-replica store):**
- Total 147 GB, of which `query_plan_dim` alone is **101 GB (69%)** — the single biggest forecasting term, and invisible to any current surface (the payload dims are deliberately plain tables no hypertable-shaped read ever counts).
- Hypertable compression is excellent (6–36x) but was only measurable by hand.
- Compressed daily ingest jumped **~3x around Aug 1** (620 MB/day → ~2.3 GB/day) and nothing recorded it — it was only discoverable by chunk archaeology, and raw chunks age out in 4 days, so that evidence window is always tiny. With primaries joining the fleet (Query Store + plan capture, the dominant payloads), the operator needs a real series to extrapolate from.

**The sweep** (`StoreSelfMetrics`, called from `DarlingWorker`'s main loop on the `_nextDiskCheckUtc` / `_nextCompressionCheckUtc` fleet-level precedent, hourly via `_nextStoreMetricsUtc`): one row per hypertable (total bytes from `hypertable_detailed_size`, pre/post-compression bytes from `chunk_compression_stats`, chunk count from `timescaledb_information.hypertables`), one row per payload dimension (`pg_total_relation_size` — the TOAST where the plan XML lives — plus exact row count), and one whole-store summary row (`pg_database_size` + enabled-server count, so the per-server ingest rate is derivable). All rows share one `metric_time` per run. The hypertable arm gates on the worker's cached TimescaleSupport detection — plain-PostgreSQL stores skip it silently and still record their dimension + store rows. Retention is the sweep's own bounded DELETE (400 days), no policy machinery.

**The table** (`collect.store_metrics`, migration V53): a PLAIN table, deliberately NOT a collector — it stays out of the collector catalog so the compression/retention machinery it measures can never recurse onto it (pinned by test). ~30 narrow rows/hour.

**The read surface**: a read-only `get_store_metrics` MCP tool (own tool class, the `DarlingMcpPvsTools` shape; registered in the MCP host and wired into `/api/read` beside the other read routes; instructions text bumped to ninety tools). Returns the latest snapshot per object (with the derived compression ratio) plus a settled last-sample-per-day series (`days_back` default 30, capped at the sweep's own 400-day retention), including the whole-store daily growth and the derived per-server rate. No `server_name` parameter — the store is the subject.

**The full V53 ladder** (the #2012/V51 and #2060/V52 drill):
- `StorageVersion.SchemaVersion` 52 → 53
- Viewer schema gate: 36th probe sentinel (`store_metrics` table existence), `GetBoolean(35)`, `hasStoreMetrics` parameter, newest-first arm returning 53 (doc comment in the V51/V52 style — a don't-under-report rung like V44's; the viewer never reads the table)
- Version pins: `DarlingObservabilityTests` (newest-rung identity), `PvsStatsStoreTests` (both 52→53 pins + comment), `ViewerDataServiceTests` (full-sentinel pin gains its 36th `true`, plus a new V53 cap-at-52 rung test), `DarlingPlanCorrectionLiveMigrationTests` (`applied` 7→8 + script list)
- CHANGELOG: one bullet under [Unreleased] ### Added with the measured numbers + `[#2068]` link ref (CRLF preserved)

## Tests

- `StoreSelfMetricsTests` (new): V53 migration identity (name, collect.-qualified idempotent CREATE, the `(metric_time)` index, plain-table pin — no `create_hypertable`), every sweep column present in the migration, the not-in-collector-catalog invariant, the viewer gate's V53 arm, and per-statement SQL pins (the three Timescale catalog surfaces; `pg_total_relation_size` + exact row count on both dims; `pg_database_size` + the `is_enabled` denominator; the 400-day DELETE) plus dialect pins (PG positional `$N`, no bare `now()` — every statement stamps the one caller-supplied `metric_time` — no T-SQL-isms).
- `DarlingMcpStoreMetricsToolsTests` (new): tool surface exactly `get_store_metrics` (static, `[McpServerToolType]`, `Task<string>`), param contract (`days_back` only, default 30, ceiling pinned equal to `StoreSelfMetrics.RetentionDays`), both read SQL shapes (`DISTINCT ON` newest-per-object; settled last-sample-per-day windowed on `$1`), dialect pins, instructions-text presence, and pure unit tests for `ComputeDailyGrowth` (deltas between settled days, denominator is the measured day's server count, missing denominator → null rate never zero/infinity, unrecorded totals skipped not invented, negative deltas preserved, empty/single-day inputs).
- Existing web-surface parity test (`ReadEndpoints_AreExactlyTheReadToolCatalog...`) covers the new tool automatically via the reflection catalog + the new dispatch entry.
- Full solution builds clean with the pinned SDK (the only two warnings are pre-existing in `Lite.Tests/LiteServerTagsStoreTests.cs`, untouched here). The `net10.0-windows` test host cannot run on this Mac — CI runs the suite.

## Scope notes

- The issue's proposed **store-growth panel (web + viewer)** is deferred — this PR delivers the collection + persistence + MCP/REST read the panel would sit on. The series accrues from deploy day either way, which is the time-sensitive part.
- The first sweep runs within one loop tick of startup (the `DateTime.MinValue` idiom), so the series starts immediately after upgrade; the daily-growth series needs two settled days before it emits its first delta.
- `enabled_server_count` counts the `servers` registry's `is_enabled` rows — the same predicate the fleet reader uses, so the per-server rate divides by exactly the servers the fleet surfaces count.


erikdarlingdata:
Review (as comment — same-identity PRs can't self-approve): V53 is the V47 collect-qualified plain-table idiom exactly, and correctly defines NO view — the v_query_stats last-definer tripwire stays pinned at V51, untouched. The full schema-gate ladder is present and the live PG job exercised it for real (V53 apply + the 7→8 rewind pin + gate pins). The per-server-rate semantics are better than asked: settled-day deltas, null-not-zero on missing denominators, negative deltas preserved because retention drops genuinely shrink the store. Plain-PG guard on the Timescale catalog reads confirmed. This is the instrument the #2069 gzip change gets measured with.

Issue #1141, opened 2026-06-17
[FEATURE] Add a "per‑event" notification option (vs. the batched per‑cycle summary) for deadlocks & blocking


gotqn:
### Which component(s) does this affect?

- [x] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

### Problem Statement

**Summary:**

Deadlock and blocking alerts currently aggregate all recent events in the window into a single notification (one tray/email/webhook card with N detail blocks, and the detail list appears capped). There's no way to receive one notification per event, which teams need for per‑incident tracking (e.g., one Azure DevOps ticket per deadlock). Downstream automation can't split the batched payload reliably because the per‑event details are flattened into repeated name/value facts rather than a structured array.

**Current behavior:**

- BuildDeadlockContextAsync / BuildBlockingContextAsync (Lite MainWindow.xaml.cs:1668,1739; Dashboard equivalents) build one AlertContext from GetRecentDeadlocksAsync/…Blocking(hoursBack: 1) and send a single alert (TrySendAlertEmailAsync).
- The card shows Current Value = N but lists only a few events (apparent Take(3) cap), so high‑volume incidents are truncated.

### Proposed Solution

**Proposed:**

1. A notification mode setting (global and/or per‑server): Summary (current default) | Per‑event (one notification per deadlock/blocking event).
2. In Per‑event mode, a configurable max‑per‑cycle cap (e.g. default 10) to prevent floods during a storm, with a final "+X more" summary if the cap is hit.
3. Each per‑event message should carry a stable per‑event identifier/fingerprint (the involved objects — ties into the "stable dedup fingerprint" request) so downstream dedup/ticketing is per incident.
4. Apply to deadlocks and blocking first; consider extending to long‑running query.

**Keep**: Summary as the default; Per‑event is opt‑in. Existing cooldown/gating still applies per event.

### Use Case

I want to be able to create an issue about particular problem and if the message is grouping and not showing all errors, I can miss something. 

**Why it matters:**

- One ticket/notification per incident is impossible: the webhook delivers one payload with multiple events as repeated Victim SQL/Processes facts — not parseable into distinct events without fragile heuristics.
- Truncation hides events beyond the cap.

erikdarlingdata:
The grouping + per-incident fingerprint half of this is implemented and merged to `dev` (so it's in tonight's nightly); the per-event delivery *mode* itself is the next step.

[...]

PR #2110, opened 2026-08-07
Multi-incident alerts render as self-contained units; every database-scoped alert carries a discrete Database fact

### LINKED ISSUE #2108: [BUG]  Multi-incident alerts (Deadlocks Detected) serialize as a flat, un-grouped FactSet — emit each incident as a self-contained unit
### Component

Darling

Darling 3.3.0

### SQL Server Version

n/a

### Windows Version

n/a

### Describe the Bug

**Environment:** Darling 3.3.0, webhook → Teams / Azure Logic App.

**What happens:** when one alert aggregates several incidents - a burst of deadlocks collapsing into multiple #1140 fingerprints — sections[0].facts[] comes through as one flat list: every Victim SQL/Processes pair first, then each Dedup Key followed by its Involved Objects, with Occurrences interleaved. Real example (3 events, 2 fingerprints):

> Victim SQL / Processes  (×3, no incident label)
> Dedup Key e7d9559c…  /  Involved Objects  smModel…TeamGk.dbo.T2, …T3
> Occurrences 2
> Dedup Key 908d2df…   /  Involved Objects  smModel…TeamGk.dbo.T1, …T2

**Problem:** a Victim SQL/Processes pair can't be associated with its Dedup Key/Involved Objects — they're in different parts of the list. In Teams it reads as one long undifferentiated block ("which deadlock is which?"), and downstream automation (Logic App → Azure DevOps) can't split into per-incident work items because the victim→fingerprint mapping is gone.

### Steps to Reproduce

Generate two different deadlocks. 

### Expected Behavior

**Request:** render each incident as a self-contained, labeled unit — ideally one sections[] entry per incident (each carrying its own Victim SQL, Processes, Dedup Key, Involved Objects), or failing that, keep one section but group facts per incident with an index ("Deadlock 1 of 2") and each incident's Dedup Key/Involved Objects adjacent to its own Victim SQL/Processes.

**Benefit:** readable multi-incident Teams cards, and downstream tooling can open one correctly-scoped ticket per incident.

### Actual Behavior

Image

```text

```

### Screenshots

_No response_

### Additional Context

_No response_

### LINKED ISSUE #2109: [BUG] Database exposed as a discrete alert fact on only 2 of the database-scoped alerts
### Component

Darling

3.3.0

### SQL Server Version

SQL Server 2022

### Windows Version

Windows Server 2022

### Describe the Bug

The alert/webhook payload exposes a discrete Database fact for only Blocking Detected and Long-Running Query (AlertContextBuilders.cs:84 and :245). Other database-scoped alerts carry the database name only inside a heading or free text, so a downstream consumer can't read it as a field:

- Deadlocks - database(s) appear only inside Involved Objects (smModel_20260727_TeamGk.dbo.T1), never as a Database fact.
- Version Store / PVS -database is in the heading ("{db} - PVS X%"), not a field.
- Database State and AG Database Suspended / AG Sync Fell Behind - database is in the message text only.
- This blocks routing/tagging by database and populating a Database field on the work items our Logic App creates.

### Steps to Reproduce

1. Trigger a blocking event and a deadlock on a monitored instance.
2. Inspect each alert's webhook payload sections[0].facts[].
3. Blocking has { "name": "Database", "value": "…" }; the deadlock has no Database fact — only Involved Objects.

### Expected Behavior

Every database-scoped alert emits the same discrete ("Database", <db>) fact that Blocking and Long-Running Query already add. For deadlocks (which can span databases), a comma-separated list of the distinct databases parsed from the involved objects is perfectly fine.

### Actual Behavior

Only Blocking and Long-Running Query add the Database field; Deadlocks, Version Store, Database State, and the AG database alerts expose the database only in the heading/message text.

```text

```

### Screenshots

_No response_

### Additional Context

_No response_

erikdarlingdata:
Closes #2108. Closes #2109.

## What

Both from gotqn, both about the same wire surface, shipped together because they meet in the deadlock payload.

**#2108** — a multi-fingerprint deadlock alert serialized as one flat `sections[0].facts[]`: every victim's fields first, every fingerprint's fields after, no way to associate them. The association was lost twice: the builder kept victims and fingerprint metadata in separate detail items (the #1141 renderer deliberately omitted forensic fields from summary incident items *because* the builder's separate items carried them), and the Teams serializer then poured all items into one fact list.

**#2109** — only Blocking and Long-Running Query exposed a discrete `Database` fact; Deadlocks, PVS, Database State, and the AG database alerts named the database only in headings or prose.

## How

- **Deadlock builder**: each fingerprinted deadlock is now ONE self-contained item — Database (#2109), Victim SQL, Processes, Dedup Key, Involved Objects, Occurrences — headed `Deadlock N of M`. Standalone "Deadlock Victim" items remain only for graphs with no parseable lock objects (no fingerprint identity — they'd otherwise vanish), and those get the Database fact too. One parse pass feeds both the object fingerprint and the new `DatabasesFromGraphXml` (distinct `currentdbname`, the same attribute the excluded-database filter already reads).
- **Teams serializer**: every fields-carrying detail item becomes its own `sections[]` entry titled by its heading, on every alert type. Advice prose and remediation-T-SQL hints stay folded into the lead section — commentary on the whole alert, not incidents. Slack/email surfaces are unchanged (their renderers already show per-item structure).
- **Database facts**: PVS items lead with `Database`; Database State fires with a structured context (`Database` / `Current State` / `Expected State`) instead of `Context: null`; both AG database alerts carry `Database` / `Availability Group` / `Replica` (+ `Suspend Reason`) via a new shared `AgAlertContexts` builder used by BOTH SKUs' evaluators — fact names are a wire contract, one builder means they can't drift. Darling's self-evaluator `FireAsync` gains a trailing optional context param so its dozens of positional call sites stay untouched; Lite's `AgAlert` record gains a trailing optional `Context`.

## Compatibility

All #2109 facts are additive. The #2108 section change is the one shape change consumers can feel: automation keyed on `sections[0].facts[]` containing *every* fact should iterate `sections[]` — flagged to the reporter on the issue and in the changelog, and it's the shape they explicitly requested.

## Tests

- Deadlock builder: self-contained single-incident pin (exact field list incl. Database + Dedup Key), recurrence collapse (`Occurrences 4`, one item), two-fingerprint indexing (`Deadlock 1 of 2` / `2 of 2`, each with its own Database/Victim SQL), unfingerprintable-graph fallback (victim item kept, Database fact present, `Incidents` null).
- Teams: structural JSON test — lead section + one section per fields item, advice folded into lead, each section carrying only its own facts.
- Extractor: `DatabasesFromGraphXml` distinct/sorted/case-insensitive + malformed-input pins.
- AG: database-scoped alerts carry the discrete facts through the shared builder; resolutions stay context-less.


selormwalker (commented):
Approved changes. The implementation adheres to repo guidelines.

Issue #2111, opened 2026-08-07
Query Store backfill contends with the live path on busy replicas — yield the slice when live collection is failing


erikdarlingdata:
Follow-up to #2102, from validating the fix on the prod monitor box. The hour-chunked catch-up recovered 7 of 10 wedged databases immediately, but four servers stayed in a failure churn — and the profile changed: QS catalog size no longer predicts failure (a 538MB database fails while a 2.6GB one succeeds), and on the affected servers **every** user database fails together. Whole servers, not databases.

What those servers have in common: the backfill worker is running a 64MB slice against them every tick, concurrently with the live QS sweep, on MAXDOP=1 read replicas during peak hours. The live query that normally finishes in seconds waits behind the backfill slice's scan of the same QS internal tables and dies at the 60s command timeout; on two servers the backfill slices themselves also time out, so their holes never drain and the churn self-sustains. Recovery-phase contention, self-inflicted.

Fix: **yield-to-live** — the backfill worker skips a server's slice whenever that server's live query_store collection has failed within the last 10 minutes. The class doc already states the contract this enforces: "backfill can be slow forever without delaying collection." Backfill hammering a replica whose live path is struggling is the opposite of that. The signal is server-grain (any live QS per-database failure stamps the server) because the contention is server-wide; the policy (window + decision) lives beside the shared backfill state so both SKUs' workers behave identically.

Held in reserve if yield alone doesn't clear the stragglers: adaptive slice-span shrink on failed hole slices (halve toward 15m; the hole records already persist state).

erikdarlingdata:
Merged to dev via #2112. Deploying to the prod monitor box for validation against the four churning servers — live QS WARNs should fall off within a few cycles as backfill yields, then the hole records drain. Adaptive slice-shrink stays in reserve if stragglers persist.

PR #2125, opened 2026-08-08
Adaptive catch-up shrink: a member that can't fit the window in the timeout halves it until it can

erikdarlingdata:
The #2111 reserve, promoted on box evidence: after yield-to-live shipped, most members recovered, but Redstone sat **3+ hours stuck through quiet overnight hours** — its 1h live window intermittently exceeds the 60s command timeout, and a fixed-width window gives such a member no path back (every retry asks the same impossible question).

## How

One shared pure policy — `QueryStoreBackfillState.AdaptiveSpan(fullSpan, consecutiveFailures)`: halve per failure, floor at 15 minutes (fits a 60s read on every store the fleet has shown; narrower than a flush interval would mostly return empty), exponent-capped, pinned by tests.

- **Live path** (Darling + Lite, enumerated + Azure arms — the #2115 lesson applied up front): per-(server, database) consecutive-failure counters ride beside the #2111 yield stamps; when >0, the watermark tightens to `now − AdaptiveSpan(MaxCatchup, failures)` and the skipped range is recorded as a **backfill hole** — the same records the clamp writes, so nothing is dropped, only deferred to the trickle. Success (the completion callbacks / Azure per-db success) resets to full width. A WARN names the database, count, and narrowed window.
- **Backfill workers** (both SKUs): consecutive failed slices per server narrow the chunk span the same way via the counted-slice wrapper; a completed slice resets. The `BoundSliceFloor` adaptive overload keeps the parameterless form byte-identical.
- Counters are in-memory on purpose (like the yield stamps): a restart costs one full-width attempt.

## Expected effect on the box

Redstone's next failure streak walks 60m → 30m → 15m; a 15m window fits comfortably (raw QS reads of 24h took 0.6s in diagnosis — the cost cliff is aggregation width), so it recovers within ~3 cycles instead of never, and the hole drains via backfill. Deploy-and-watch is the validation plan, same as #2112.

## Tests

`AdaptiveSpan_HalvesPerFailure_FloorsAtFifteenMinutes_AndResetsAtZero` + `BoundSliceFloor_AdaptiveForm_CapsToThePassedSpan`, alongside the existing yield/chunking pins.


erikdarlingdata:
Both review points addressed in 33c06d91:

1. **Null-watermark gap — real, fixed.** All four adaptive sites (Darling + Lite, enumerated + Azure) now tighten the first-run fallback window by the same halving schedule when a never-succeeded database accrues consecutive failures. No hole record on that path deliberately: there is no watermark to record a hole *from*, and pre-watermark history is the backfill tail's job by design.

2. **Slice-failure counter granularity — per-server is deliberate.** The slice timeout contention this counter tracks is server-wide (one command timeout budget, one server's Query Store under load), and the slice-failure log line carries no database name to key on anyway. Per-database granularity would just make N databases each re-learn the same server's limit. The Darling counter is now a `ConcurrentDictionary` to match the Lite twin, though — that asymmetry was real.

Issue #2105, opened 2026-08-07
[QUESTION] How to update the service-version (not urgent)


ghauan:
### Which component is your question about?

Darling

### Performance Monitor Version

3.4.0

### Is your question about how it works, or the results?

Setup / installation

### What's your question?

Not urgent, take this at a suitable time: 
I was wondering a bit how the update works now that it is a service. Is it just to unzip to where it runs today at C:\PerformanceMonitorDarling? 
If yes, do I then need to backup/copy my existing darling.json-file first? 
Or do I need to run the install-darling.ps1 again?

Sorry if these are dumb questions 🫤

erikdarlingdata:
Not dumb questions at all — the service deliberately never auto-updates, so this is a real procedure and worth spelling out.

**The short version:** stop the service, extract the new zip over the same folder, run the installer script again. Concretely, from an elevated PowerShell:

```powershell
Stop-Service 'PerformanceMonitor Darling'
# extract the new PerformanceMonitorDarling-<version>.zip over C:\PerformanceMonitorDarling
Set-Location C:\PerformanceMonitorDarling
.\install-darling.ps1
```

The stop comes first because the running service holds its binaries locked — extraction over a running install fails. `install-darling.ps1` detects the existing service and treats it as an upgrade: binPath updated in place, config/store/credentials preserved, ACL hardening and firewall rules re-checked, service started and confirmed Running.

**No, you don't need to back up `darling.json`.** The zip ships only `darling.sample.json` — extracting never touches your real `darling.json`, and the installer explicitly preserves it, the Postgres store, and everything encrypted with `--encrypt-password` (those are DPAPI machine-scope blobs). A copy never hurts, but it isn't part of the procedure. Store schema migrations run automatically on the first service start after the upgrade.

**One 3.4.0-specific note** since that's the version you're on: if you upgraded from an earlier version, run the one-time repair verb **promptly** — it fixes Query Store rows the old builds stored as split slices, and it can only reach rows still in the raw retention tier, so it gets less useful every day you wait:

```powershell
.\PerformanceMonitor.Darling.Service.exe --collapse-legacy-slices --dry-run
.\PerformanceMonitor.Darling.Service.exe --collapse-legacy-slices
```

Details in the 3.4.0 changelog section for [#1912].

[#1912]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/1912

erikdarlingdata:
[...]

PR #2127, opened 2026-08-08
Collapse slices lift TimescaleDB's decompression cap for their own transaction

erikdarlingdata:
#2105 round two (ghauan). The #2123 timeout fix got the repair four minutes into real work — far enough to hit the next wall: `53400: tuple decompression limit exceeded by operation`.

## Why

The repair's DELETE touches **compressed** chunks — a store old enough to need `--collapse-legacy-slices` has had its compression policy running the whole time, so this is the normal case, not an edge. TimescaleDB rails DML-triggered decompression at 100k tuples per transaction by default.

## Fix

`SET LOCAL timescaledb.max_tuples_decompressed_per_dml_transaction = 0` at the top of each slice transaction — scoped to die with the transaction, mirroring the rail-lift `DarlingRetention`'s fallback DELETE has always carried (same comment rationale: deliberate bulk decompression is exactly this operation's job). Placeholder-GUC-safe on stores without the extension.


erikdarlingdata:
Both review points taken, in 3ae492f0:

1. **`[#2105]` link-reference** — added to the definition list; both dangling uses now resolve.

2. **Compressed-chunk live test** — agreed this verb earned its own #1564-style pin after two field failures. `CollapsingRowsInsideACompressedChunk_SurvivesTheDecompressionRail_TheSecondFieldFailure` compresses the seeded chunk synchronously, PROVES the compressed shape, then arms the session rail at **one tuple** before calling the collapse — so the transaction-local `SET LOCAL ... = 0` is the only thing between the DELETE and the operator's exact `53400`. Dropping the lift fails CI with the field error instead of a silent coverage hole. Both halves verified against live TimescaleDB (pg17) before pushing: cap=1 + DELETE on a compressed two-row batch trips `tuple decompression limit exceeded (current limit: 1, tuples decompressed: 2)`; the `SET LOCAL` lift lets the same DELETE commit.

PR #2128, opened 2026-08-08
Job History tab speaks display names

### LINKED ISSUE #2126: [FEATURE] Use Server Display Name in Job History tab

- [ ] Lite
- [x] Darling
- [ ] Documentation

### Problem Statement

Both the filter dropdown and column header in the Job History tab in the viewer uses the real servername. 

### Proposed Solution

It would be preferable that it showed the Display Name the same as the other tabs do.

### Use Case

Normally I don't remember all the actual servernames, rather I use the aliases we have for them.

### Alternatives Considered

_No response_

### Additional Context

_No response_

erikdarlingdata:
Closes #2126.

## The gap

ghauan runs the fleet by aliases and the Job History tab was the one place that refused to speak them: both the Server filter dropdown and the Server column carried the raw collected `server_name`, while every other tab resolves the operator's display alias.

## The fix

**Darling** — both Job History readers now resolve names through the `servers` registry with `COALESCE(reg.display_name, <raw name>)` — the exact pattern the FinOps inventory reader already uses:

- **`GetJobHistoryAsync`** — the base CTE joins `servers` and emits the alias as `server_name`, so the Server column, the Server filter combo (which keys on `server_id`, so filtering is unchanged), the per-column filter popup, and the CSV export all show display names with zero tab-side changes.
- **`GetAgentStatusAsync`** — same join; the fleet Agent roll-up now also sorts by display name.

`display_name` is never empty on the write path (`DarlingConfig.DisplayName` falls back to the host), and the `COALESCE` covers pre-existing rows written before the column existed.

**Lite** (robot review catch — same gap, different mechanism): Lite's display-name concept lives at the config layer (`ServerConnection`), not in DuckDB — the stored `servers.display_name` column is unpopulated by design. So the shell passes a `server_id → alias` snapshot into the tab Overview-style (the same `GetDeterministicHashCode(GetServerNameForStorage(...))` identity the Overview refresh already uses), and rows swap in the alias on every refresh. A server no longer in config keeps its raw collected name — job history is a durable record and rows can outlive their registration.

## Tests

No SQL-shape pins exist for these readers (checked — only schema-probe sentinels reference `job_history`), and the Darling change is a projection swap with the join precedent already live in two other readers. The Lite change is display-only row post-processing behind an optional `Initialize` parameter, defaulting to off.


erikdarlingdata:
Review taken — the Lite parity gap is real and is now closed in 4468d501, your way rather than a copy-paste: Lite's display-name concept lives at the config layer (the stored `servers.display_name` column is unpopulated by design, as you found), so the shell passes a `server_id → alias` snapshot into the tab Overview-style — the same `GetDeterministicHashCode(GetServerNameForStorage(...))` identity the Overview refresh already uses — and rows swap in the alias on every refresh. A server no longer in config keeps its raw collected name, since job history is a durable record and rows can outlive their registration.

Also updated the PR title's scope: this now closes the gap in BOTH SKUs, not just Darling.

(The earlier `Darling PostgreSQL tests` failure was unrelated to this diff — `CompressionJobSelfHeal_DetectionQueryValid` tripped on `next_start = -infinity` compression jobs left in the shared CI fixture by sibling tests; the same suite passed on three other runs in the same window. Rerun kicked.)

PR #2132, opened 2026-08-08
Collapse verb narrows its slice adaptively instead of dying at the timeout

erikdarlingdata:
#2105 round three (ghauan, same store). With the decompression rail lifted by #2127, their run survived past the old four-minute wall and died ~15 minutes in with the same bare stream exception — which is exactly `SliceStatementTimeoutSeconds` (900s): a **day-wide** stage aggregation on a store carrying 60,654 split intervals doesn't fit the statement timeout, and the verb's slice width was a fixed `AddDays(1)`.

## The fix

The slice loop now runs the same shared adaptive schedule the Query Store backfill worker shipped this week (`QueryStoreBackfillState.AdaptiveSpan`, 24h base):

- a failed slice **halves the window and retries the same start**, announcing itself with a `[RETRY]` line that names the error — narrowing reads as progress, not a hang;
- a completed slice resets to full width, so healthy stores still repair in a handful of day-wide slices — the narrowing costs nothing until a slice actually fails;
- only a slice that fails at the ~22-minute floor gives up, to the existing idempotent re-run message (now naming the failing slice's start and width).

Raising the timeout instead would be the wrong lever: the slice transaction holds chunk locks the live service's compression jobs also want, and 15 minutes is already generous — the right response to "too big to fit" is a smaller bite, the exact design conclusion #2125 reached for the live path.

## Tests

The halving schedule itself is the already-pinned `AdaptiveSpan` (floor and cap pins shipped with #2125). The verb's happy path stays covered by `CollapseVerb_DryRunsWithoutChanging_ThenRepairs_ThenFindsNothingLeft`; the retry arm is driven by real statement timeouts that can't be provoked deterministically in a live test without multi-minute CI stalls, so its correctness rides the shared policy's pins plus the loop's structure.


erikdarlingdata:
Both catches were real — thank you, the first one would have made this fix a no-op in exactly the field case it targets. Fixed in ce45eed7:

1. **Dead-connection retry**: the retry path now cycles the connection before re-attempting (close is safe on a broken one, reopen draws a fresh physical connection; nothing session-scoped is lost — the slice's `SET LOCAL` and command timeouts are transaction/command scoped).

2. **Clamped-width honesty**: the retry decision and both operator messages now use the width the slice ACTUALLY covered (`sliceEnd - sliceStart`) — and the halving search skips nominal steps that wouldn't narrow a clamped final slice, so it never reruns an identical window just to walk the schedule down.

erikdarlingdata:
Round two both taken, in 0eb07318:

1. **Unguarded reopen** — a failed reopen now degrades to the same clean idempotent-rerun message as every other failure mode (naming both the slice failure and the reopen failure), never an unhandled crash. You're right that this one-shot verb has no caller safety net, unlike the backfill worker's per-tick catch.

2. **Below-floor tail slice** — a slice at/below the ~22.5m floor now gets ONE same-width retry on a fresh connection before giving up, tracked per-slice (`retriedAtWidth`, reset on success and on any narrowing step). The run's usual last slice is no longer exempt from the retry mechanism.

3. Took the extraction suggestion too: the narrowing decision is now the pure `NextNarrowingFailureCount`, pinned for the first-halving step, the clamped-tail step-skipping, and the at/below-floor null that hands over to the same-width retry — exactly the case that would otherwise only surface in the field.

Issue #2102, opened 2026-08-07
[BUG] Query Store catch-up death spiral: window width is unbounded below the 24h clamp, so one missed cycle wedges a big database permanently


erikdarlingdata:
## What happened

After the use1 region migration, Query Store collection on the migrated servers' **largest databases** began timing out at the 60s command timeout on every cycle (~120 WARNs/hr, flat for 6+ hours) and never recovered. Small databases on the same servers stayed current to within minutes.

Field evidence (prod monitor box, 2026-08-07):

- `sys.database_query_store_options` on the affected server: every database `READ_WRITE`, well under `max_storage_size_mb`, AUTO cleanup — **the source Query Store is healthy**.
- Raw QS reads are fast: `COUNT(*)` over 1.9M `query_store_runtime_stats` rows = 0.2s; a plain 24h join across runtime_stats/plan/query/text = 0.6s.
- Watermark staleness at diagnosis time was bimodal by database size: big tenants 0.5–6.5h behind, small ones minutes. **Zero `catch-up clamped` WARNs in the log** — the 24h clamp never fired.
- The backfill worker's slices on those same big databases also fail: `query_store backfill slice on 'multi32' failed: Execution Timeout Expired`, while other databases ship 64MB slices fine.

## Why

The live path's incremental window is `(watermark, now)`, and the watermark only advances when a cycle **succeeds**. The per-database query aggregates the slice-combining derived table, computes the plan-XML window function, and sorts — over the **whole window** — before `TOP (50000) WITH TIES` or the client byte budget can bound anything. A row cap and a byte budget are not a cost cap: query cost grows with window width.

So one missed 60s cycle (a flush burst, a CPU blip, a migration cutover) widens the next window, which costs more, which times out again. The window grows without bound below the clamp, and the 24h clamp sits far above the tipping point (~1–2h on the big tenants), so it never interrupts the spiral. The database wedges permanently while its neighbors stay healthy.

The backfill worker (#2022/#2058) has the same latent flaw one layer down: a hole or tail slice queries the **full remaining range** in one statement — the byte budget bounds what ships, not the aggregation window — so a wide hole on a big database can never drain either.

## Design intent

Catch-up was always supposed to trickle in ~hour-sized chunks through the backfill machinery, not one-shot an arbitrary window on the live path. The fix restores that:

[...]

PR #2134, opened 2026-08-08
Query Store payload stages its aggregate — the fixed cost big catalogs couldn't pay

### LINKED ISSUE #2133: [BUG] query_store collector: monolithic TVF join is a fixed ~30s+ cost on big catalogs — stage through #temp (sp_QuickieStore architecture)
Field evidence from the dogfood fleet (2026-08-08): after #2125's adaptive shrink deployed, most stragglers converged, but a hard core (AppDatabaseThirteen@multi22, AppDatabaseFourteen@multi-01, AppDatabaseFifteen@multi32, AppDatabaseSixteen@psi, AppDatabaseOne@multi-49, AppDatabaseTwelve@multi22, and rotating peers) kept failing at every window width — 60m, 30m, 15m identically — with zero rows shipped for hours. Width-independence was the tell: some fixed cost eats the 60s command timeout before the window predicate matters.

## Bisection (AppDatabaseThirteen@multi22, SQL 2022 16.0.4215.2, 3 GB QS catalog, 82k plans / 64k queries / 526 intervals, all probes 30s-capped, read-only)

| probe | result |
|---|---|
| current interval pre-filter (runtime_stats scan on last_execution_time) | 426 ms — **not** the wall |
| interval-catalog pre-filter (`runtime_stats_interval.end_time`) | 20 ms |
| raw rows in last-hour intervals | 10,292 rows, 252 ms |
| inner aggregate alone (GROUP BY plan_id, interval_id) | 7,346 groups, **81 ms** |
| bare `SELECT COUNT(*) sys.query_store_plan` | 82k rows, **319 ms** |
| bare `SELECT COUNT(*) sys.query_store_query` | 64k rows, **208 ms** |
| aggregate JOIN qsp JOIN qsq (the collector's shape, WITH the LOOP JOIN hint) | **timeout ≥30s** |
| same WITHOUT the hint | **timeout ≥30s** |
| aggregate joined to qsp ONLY | **timeout ≥30s** |
| **staged: aggregate INTO #x, then #x JOIN qsp JOIN qsq** | **stage 56 ms + join 409 ms = 524 ms total** |

Two relations that each materialize in well under half a second join catastrophically when the optimizer has only TVF fixed-guess cardinalities to work with — the inner side re-materializes per probe (7k probes × ~300 ms ≈ 35 minutes of work against a 60s timeout). The `OPTION(LOOP JOIN)` hint neither causes nor cures it.

## Fix

Restructure `QueryStoreCollector.BuildPayloadBody` to stage the runtime-stats aggregate into a `#temp` and run the plan/query/text joins FROM the temp — sp_QuickieStore's architecture, for exactly this reason. Constraints to preserve:

- ONE result set at the end of the batch (the reader/byte-budget machinery is single-result-set);
- TOP WITH TIES + derived-watermark semantics unchanged on the final SELECT;
- the `LOOP JOIN` hint must NOT survive onto the staged join (looping from the temp into the TVFs is the pathology itself; the 409 ms join is unhinted);
- both engine arms (on-prem/RDS sp_executesql nesting and the Azure per-database parameter path — #temp scope lives and dies with the batch in both), both SKUs, live + backfill variants;
- while here: the interval pre-filter can resolve ids from `sys.query_store_runtime_stats_interval` (20 ms) instead of scanning runtime_stats (426 ms) — not the wall, but strictly better and the same tie-group semantics via the existing HAVING.

Expected field effect: the wedged members drop from 'never completes inside 60s at any width' to sub-second, and the adaptive shrink becomes a rarely-needed safety net instead of a perpetually-pinned floor.

Evidence gathered via read-only probes from the monitor box; full transcript in the session that produced #2125/#2132.

erikdarlingdata:
Closes #2133 — and this is the actual root cause under the whole catch-up saga: #2102's death spiral, #2111's yield-to-live, and #2125's adaptive shrink were all mitigating a cost none of them touched.

## The bug

The payload joined its slice aggregate straight into the `query_store_plan`/`query`/`text` catalog TVFs. The optimizer only has fixed-guess cardinalities for TVFs, and the shape it picked re-materialized a TVF per probe — a **fixed** ≥30-second cost on an 82k-plan catalog that no catch-up window width could reduce. That's why the fleet's big databases (AppDatabaseThirteen, AppDatabaseFourteen, AppDatabaseFifteen, AppDatabaseSixteen, AppDatabaseOne…) pinned at the 15-minute shrink floor and never converged while smaller neighbors on the same servers stayed current.

## Bisected live on the wedged store (AppDatabaseThirteen, SQL 2022, 3 GB QS catalog)

| shape | result |
|---|---|
| aggregate alone | 7,346 groups, **81 ms** |
| `sys.query_store_plan` bare scan | 82k rows, **319 ms** |
| aggregate JOIN plan (current shape) | **>30s, hinted or not** |
| staged: aggregate INTO #temp, join FROM it | **524 ms** (56 stage + 409 join) |
| the full emitted 55-column batch, plan capture ON, one-hour backlog | **21.3 s / 8,111 rows / 55 cols** — where the old shape never finished inside the 60s command timeout |

## The fix

- `SELECT ... INTO #pm_qs_slice`, then the plan/query/text joins run FROM the temp — real row counts instead of TVF guesses, each TVF scanned exactly once. sp_QuickieStore stages for exactly this reason.
- The `LOOP JOIN` hint is **gone for good** and pinned absent: looping from the temp into the TVFs is the same per-probe re-materialization by another name. The 524 ms join is unhinted, chosen from true cardinalities.
- The interval pre-filter resolves ids from the tiny interval catalog (**20 ms**) instead of scanning `runtime_stats` itself (**426 ms**) — same superset bound, the exact HAVING unchanged.
- Batch mechanics preserved: `SELECT INTO` emits no result set so the batch still returns exactly ONE (the reader/byte-budget contract); `TOP … WITH TIES`, ship order, and derived-watermark semantics live on the final SELECT unchanged; the leading `DROP TABLE IF EXISTS` covers Azure's pooled direct connections while the on-prem `sp_executesql` scope self-cleans. One shared body, so both SKUs and both engine arms (live + backfill) get the fix by construction.

## Validation

All timings above are from read-only probes against the wedged field store, run today. Shape pins updated: staging present, catalog pre-filter, no-LOOP-JOIN (both live and backfill arms), HAVING indentation.

After merge: nightly → dogfood box → the eight wedged members should converge and the adaptive shrink demote to a rarely-needed safety net.


erikdarlingdata:
Both taken, in 92896147:

1. **Staging OPTION(RECOMPILE)** — completely right, and a sharp catch: split out on its own, the staging statement would be cached via `sp_executesql`'s parameterized text and sniffed across live vs backfill windows — the exact fixed-guess failure mode this PR removes, reintroduced one statement earlier. Both statements now carry `OPTION(RECOMPILE)`, with the reasoning documented in the template comment.

2. **Indentation** — the staged block (aggregate lines, the spliced `replicaGroupKey`/gated-family fragments, and the backfill window continuations) now dedents to the file's 4-space convention, and the pins assert the clean shape instead of encoding the mixed one. Done with per-fragment edits and the diff hunks read before pushing — the earlier attempt at this exact cleanup is the reverted 82dea99b, whose block-rewrite anchor matched the wrong marker 700 lines early.

Issue #2136, opened 2026-08-09
Query Store aggregate tax scales serially with fleet size — measure, alert, and evaluate before large onboardings


erikdarlingdata:
From the #2105/#2133 plan work, generalized: the aggregation shape over `query_store_stats` (wide, near-unique grouping keys → partials combine nothing → single-threaded finalize hash) isn't just the collapse verb's cost — it's what the store's own background jobs pay continuously, and it scales linearly with raw volume with no parallel rescue.

## Measured today (dogfood store, 52 servers, 1.2–1.6M raw QS rows/day)

The four most expensive background jobs in the store are all the query_store_stats family:

| job | last run |
|---|---|
| policy_compression query_store_stats | **157s** |
| query_store_stats_interval_hourly refresh | **96s** |
| query_store_stats_hourly refresh | 35s |
| query_store_stats_corrected_hourly refresh | 30s |

Everything else in the store is single-digit seconds. Plan analysis (full EXPLAIN in #2133's thread): ~50% serial finalize re-hash of near-unique ~160-byte keys, ~25% backward index scan over the uncompressed hot chunk, ~25% scan+partial hashing. work_mem is NOT the lever (measured: 256MB traded away parallelism and net-lost).

## Why file it

Onboarding ~50 primaries ≈ 2× raw volume ≈ interval_hourly at ~3+ min and compression at ~5 min per run. Nothing breaks day one, but: refresh jobs take chunk locks that contend with collector writes (#1564/#1567 family), the CPU competes with collection sweeps (60–75s slow-sweep skips already visible), and refresh lag compounds silently if any job exceeds its cadence.

## Asks

1. **Visibility first**: surface background-job durations (job_stats last/mean duration per proc+hypertable) as self-metrics with a threshold alert — 'refresh exceeded N% of its schedule interval' is the honest tripwire. (Check overlap with V53 self-metrics before building.)
2. **Scale test**: synthetic 2×–4× ingest on a scratch store to find the knee for the three heavy jobs.
3. **Structural options to evaluate** (only if 1–2 say so): group keys that let partials actually combine (bucket-first aggregation), refresh cadence/window tuning for interval_hourly, and the hot-chunk access path (backward index scan vs seq).

Not urgent at 52 servers; becomes urgent the week someone points a 100-server fleet at one store.

erikdarlingdata:
[...]

Issue #2119, opened 2026-08-07
[BUG] 3.4.0 Darling upgrade from a ≤V50 store fails migration: rung 51 replays the resolving view with the V54 gz column three rungs early


erikdarlingdata:
Field report on #2105 (ghauan): upgrading a 3.3.0-era Darling store to 3.4.0 dies at service start with

```
[CRIT] Cannot reach or migrate the Postgres store: 42703: column qpd.query_plan_gz does not exist
```

and the store shuts down cleanly (each rung runs in its own transaction, so nothing is corrupted — the ladder just can't pass).

## Why

Migration rung 51 (`query-stats-host-object`) is assembled at RUNTIME as `V51Sql + PgSchemaGenerator.GenerateQueryStatsResolvingView()`. The generator emits the view as today's code knows it — which since #2069 includes `qpd.query_plan_gz`, the column **rung 54** adds. A store at ≤V50 replays rung 51 with that current-generator SQL three rungs before the column exists: `42703`, ladder halts, service stops. Every 3.3.0 Darling store upgrading to 3.4.0 hits this.

It never showed on the dogfood box because that store walked 51→52→53→54 one nightly at a time — each rung ran while the generator matched that rung's era. The V54 design note anticipated exactly this class for V38 ("V38's generated body pre-adds the column for stores upgrading from below it") but missed that rung 51 re-emits the view too.

## Fix

Pre-add the column in rung 51 by prepending V54's (idempotent) ALTERs ahead of the generated view — a ≤V50 store gets the column at 51, the view validates, and rung 54's `ADD COLUMN IF NOT EXISTS` no-ops. Plus a pin test: any migration whose SQL references the compressed-content column must itself contain the pre-add, so the next generated-rung/late-column collision fails in CI instead of on an operator's store.

## Operator impact until the fix ships

The failed upgrade is safe to retry after the fix; no data is touched. Eager unblock: manually `ALTER TABLE collect.query_plan_dim ADD COLUMN IF NOT EXISTS query_plan_gz bytea;` on the store, then start the 3.4.0 service — the ladder passes.

erikdarlingdata:
[...]

PR #2137, opened 2026-08-09
Store self-metrics measure background jobs — duration vs cadence as a series

erikdarlingdata:
The visibility half of #2136 (alerting is the next half).

## What

The hourly #2068 self-metrics sweep writes one row per TimescaleDB background job — `object_kind = 'background_job'`, with **last run duration, schedule interval, total runs, total failures** (V56: four nullable bigint columns, appended per the V55 ordinal rule). TimescaleDB-gated like the hypertable arm; plain-PG stores skip it silently.

## Why

The store's heaviest recurring work is its own job machinery: on the production 52-server store the four most expensive background jobs are all the `query_store_stats` family (compression **157s**, interval_hourly refresh **96s**), their runtimes scale **serially** with raw volume (the finalize hash-aggregate runs in one process — plan analysis in #2136), and a job that outgrows its own cadence compounds refresh lag silently. Storing the interval beside the duration makes 'how close is each job to its ceiling' one division over a 400-day series — the number an onboarding wave moves first.

## Ladder dance (the #2119/#2122 checklist)

- V56 migration (`store-metrics-background-jobs`), idempotent ALTERs, replay-safe.
- `StorageVersion.SchemaVersion` → 56.
- Viewer probe: `store_metrics.last_run_duration_ms` sentinel + newest-first V56 arm (the viewer never reads these columns — under-reporting to the connect gate is the guarded failure, like V53's rung) + reader ordinal 38.
- Every literal version pin visited (StoreSelfMetricsTests, DarlingObservabilityTests, PvsStatsStoreTests) and the all-sentinels tripwire extended.

## Tests

Beyond the pins (migration↔sweep column agreement, probe arm exact-56/caps-at-55, SQL shape): a **new end-to-end live sweep test** — scratch store, real migrations, real compression policies so real jobs exist, one `SweepAsync` run, then every object_kind asserted present including job rows carrying intervals. That's the drift-catcher for 'sweep writes a column the migrations didn't create', which no static pin can be — and this suite previously had no live execution at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **Every previously-hardcoded alert threshold is now a real setting** ([#2107], the split-out from gotqn's #2101 - "it was fine to hardcode these for development but any serious monitoring allows configuring of alert thresholds") - six new knobs ride the store control plane (V55), the Viewer's Settings window, and `get_alert_settings`/`update_alert_settings`, clamped on read like their siblings: the monitor store volume's self-alert warning percent (was 10), the Collection Stopped staleness window (was 30 minutes) and consecutive-failure fast path (was 10), the low-disk CRITICAL severity tier's percent and GB floors (were 3% / 2 GB - these grade the target-volume alert in BOTH apps, and Lite reads its pair from `settings.json` as `alert_disk_critical_free_percent` / `alert_disk_critical_free_gb`), and the analysis notification cooldown (was a hardcoded 360 in Darling while Lite always honored a configured value - the parity gap closed). MCP shape: `low_disk.critical_free_percent` / `low_disk.critical_free_gb`, a new `self_alerts` group, and `analysis.notify_cooldown_minutes`.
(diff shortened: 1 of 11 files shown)


erikdarlingdata:
Taken, in 201ae75a — and it was the right catch: written-but-never-read would have left `get_store_metrics` returning job rows with a name and nulls, which defeats the visibility this PR exists for. The reader SQLs, row records, and the MCP/REST JSON projection now carry all four fields, plus a derived `duration_vs_cadence_percent` on the latest snapshot — the ceiling-proximity number in one field. The live sweep test walks the read path too now, asserting a `background_job` row comes back through `DarlingStoreMetricsReader` with its interval intact.

One deliberate note: the reader naming V56 columns carries no skew hazard — it runs service-side against the service's own just-migrated store. The viewer (which CAN lag) still never reads them, exactly as the probe-arm comment claims.

erikdarlingdata:
Taken in ea6e2582 — job rows now carry a `[job_id]` suffix in `object_name`. You're right that DISTINCT ON had no tiebreaker: two `add_job()` customs sharing a proc_name (or two hypertable-less jobs) would have silently dropped one job's series from both the latest view and the daily grouping. job_id is already joined, unique by construction, and stable for a job's lifetime, so per-job continuity holds — and since V56 hasn't shipped anywhere, there's no existing series shape to preserve.

Issue #2138, opened 2026-08-09
[FEATURE] Auto force plan bot — detect bad Query Store plans (CPU-weighted), force with self-review, evict-first option


erikdarlingdata:
From Erik's feature email ('Auto force plan bot', 2026-08-08/09). Faithful decomposition of the ask, then a proposed phasing.

## The ask, verbatim requirements

1. **Detection**: find opportunities to force Query Store plans when a bad plan is detected. Signal = CPU + duration combined, **CPU weighted stronger** (duration is confounded by factors outside the plan).
2. **Importance gate**: prerequisites before acting — significant executions AND resource expenditure judge whether a query matters enough.
3. **Self-review + regression ledger**: post-force review phases, and tracking for plan forces that were **not a net benefit**.
4. **Live-activity component**: watch server activity for queries currently running hot; target them with **DBCC FREEPROCCACHE** (targeted, per-plan) so a newly forced plan takes effect immediately.
5. **Evict-first strategy**: optionally clear the bad plan FIRST and observe whether the optimizer picks a better plan before resorting to forcing.
6. **Goal**: parameter-sensitive plans, and queries adversely affected by a stats update.

## What already exists to build on

- `query_store_stats` series: per-plan CPU/duration/executions over time, both SKUs — the detection substrate is already collected.
- `PlanCorrectionCollector`: `sys.dm_db_tuning_recommendations` + FORCE_LAST_GOOD_PLAN desired/actual state — the engine's own regression opinions, already shredded and stored.
- `remediation_action_json` precedent: findings that carry executable T-SQL.
- Active-query snapshots for the currently-hot component.
- House rule to honor: the bot's self-review may only unforce **its own** forces — operator-placed forces are never touched.

## Proposed phasing (the write-access question is the product decision)

- **Phase 0 — advise-only detection** (no target writes; shippable alone): plan-regression scoring from the stored QS series — per (query, plan) compare candidate vs incumbent on CPU-weighted deltas, gated by executions × resource spend; classify parameter-sensitivity (plan alternation) vs stats-update regression (correlate with stats timestamps); surface as findings with the exact `sp_query_store_force_plan` / `DBCC FREEPROCCACHE(plan_handle)` remediation attached.
- **Phase 1 — evict-first, operator-armed**: one-click/one-call execution of the targeted eviction with before/after observation windows, still human-initiated.
[...]

PR #2139, opened 2026-08-09
#2138 Phase 0: CPU-primary plan-regression scoring + absolute-spend noise floor

erikdarlingdata:
Phase 0 of #2138 (gaps 1 + 2 from the gap analysis on the issue): make the PLAN_REGRESSION verdict trustworthy enough to someday act on, while everything stays advise-only.

## What changed

**CPU is now the primary signal.** The old score was `GREATEST(cpu ratio, duration ratio)`, so a plan whose CPU never moved could fire on duration alone — but duration is confounded by blocking, IO waits, and machine contention that no plan choice caused, exactly the false positive a plan-forcing bot must never act on. Now:

- A CPU regression scores at its own ratio (>= 2x fires, unchanged).
- A duration-dominant regression fires only when EXTREME (>= 4x) **and** corroborated by at least mild CPU worsening (>= 1.25x), scored at **half** the duration ratio so it competes honestly with CPU-detected rows.

**A 10-CPU-second absolute-spend floor.** The latest plan must have burned >= 10 CPU-seconds across the 14-day comparison window. This is a noise gate, not an importance ranking — a 12x ratio on a query costing 12ms/run over 100 runs is sampling jitter, not a finding. Magnitude ranking stays with the factor and the scorer.

**Fact and drill-down agree by construction.** The regressed-queries drill-down kept its OWN copy of the old GREATEST in both SKUs — left alone, it could surface rows the fact never counted. Both drill-downs now run the identical CASE + floor.

## Where

All four SQL sites: `PgFactCollector.QueryPerf.cs` + `PgDrillDownCollector.Queries.cs` (Darling), `DuckDbFactCollector.QueryPerf.cs` + `DrillDownCollector.Queries.cs` (Lite). The `regressed_dimension` metadata stays honest for the new path: a duration-fired row reports dimension 2.

## Pins

Split-signal seeds (CPU and duration moved independently) in the DuckDB suite and mirrored as new arms of the live-Postgres twin, so the REAL Postgres SQL is exercised too:

- duration-only (CPU flat, duration 5x) does NOT fire — and the drill-down returns empty for the same seed;
- extreme duration (6x) with mild CPU corroboration (1.5x) fires at 3.0 (dur/2), not 6.0;
- the same 12x CPU ratio that passes everywhere else does NOT fire when the query burns only 1.2 CPU-seconds across the whole window;
- all classic 12x/3x replica-split assertions unchanged and still passing gates (existing seeds regress on both signals and spend 30–120 CPU-seconds).

Existing consumers are unaffected mechanically: scorer threshold formula, remediation generation, and force-plan replica-scope logic all consume the factor/rows downstream of this filter.

## What this is NOT

No behavior toward forcing plans. Phases 1–2 (evict-first, live-hot targeting, self-review windows, auto-unforce ledger, write-credential tier) remain design-gated on the issue.

Part of #2138 — leave the issue open.


erikdarlingdata:
Good catch on `regressed_dimension` — confirmed real: the magnitude comparison was only correct when the score WAS the max of the two ratios. Fixed in 5dd9e755 by deriving the dimension from which CASE branch fired (`cpuRatio >= 2 ? 1 : 2` — exhaustive, since a surviving row must have fired one branch), both SKUs, and dropped the now-dead `latestDur`/`bestDur` locals so nobody trusts them later. Pinned with the exact failure case you described (cpu 2.5x + dur 10x → factor 2.5, dimension cpu) plus a dimension assert on the duration-fired arm, in the DuckDB suite and the live-Postgres twin.

Issue #1882, opened 2026-07-31
Force-plan recommendation is replica-blind now that plan regression splits by replica_role


erikdarlingdata:
Spun out of #1850, which made the analysis-layer Query Store dedups key on `replica_role` and carry it through the drill-down row shape. That change is correct on its own terms, and it surfaces a question the remediation layer has never had to answer.

## What #1850 changed

On a SQL Server 2022+ AG with Query Store for secondary replicas enabled, one query can now legitimately appear in the `regressed_queries` drill-down TWICE — once as `PRIMARY`, once as `SECONDARY` — with its own best plan on each, because the two replicas run genuinely different workloads against the same shared Query Store.

## The mechanical half, already fixed in #1850

`FactRemediation.ExtractPlanRegressionTargets` reads a fixed set of keys and does not know about `replica_role`. Two rows for one query would have produced two `ForcePlanTarget`s naming the SAME `(database, query_id)` with DIFFERENT `best_plan_id`, which renders as two `sp_query_store_force_plan` calls for one query — mutually exclusive instructions where whichever the operator runs last silently wins. #1850 dedups on `(database, query_id)`, keeping the first, which the SQL orders `regression_factor DESC` so it is the worst. That stops the contradiction and nothing more.

## What is left, and why it is not mechanical

1. **Should a regression observed only on a read-only SECONDARY recommend forcing a plan at all?** Forcing is done on the primary. A recommendation driven by read-only reporting workload changes the plan the primary's write workload gets.
2. **When both replicas regressed, is "the worse regression factor" the right tiebreak?** Today's dedup keeps it because it is consistent with the existing cap-of-5-worst behavior, not because anyone decided a secondary's 12x should outrank a primary's 3x.
3. **Should `ForcePlanTarget` and the rendered recommendation NAME the replica?** It is a positional record and the renderer is pinned by a render-stability golden test, so adding the field is a deliberate change rather than an additive one. The drill-down and the MCP payload already carry `replica_role` (`AnalysisNotificationService.FlattenInto` enumerates every property, so it renders in email/webhook/toast automatically) — it is only the typed remediation path that drops it.

Anyone picking this up should confirm the current SQL Server 2022+ semantics for plan forcing under Query Store for secondary replicas against MS Learn first, rather than reasoning from the row shape — whether forcing is per-query or can be scoped per replica is the fact the whole decision turns on, and #1850 deliberately did not assume an answer.

## Impact bound

[...]

PR #2140, opened 2026-08-09
#2138 gap 3: parameter-sensitivity co-fire flag + force-plan caution

erikdarlingdata:
Gap 3 of #2138, per the design posted on the issue (advise-only; no change to what fires, only to what the recommendation SAYS and what the future bot will be allowed to touch).

## What changed

**Per-row `parameter_sensitivity_cofired` in the `regressed_queries` drill-down (both SKUs).** Computed inside the drill-down SQL with the PARAMETER_SENSITIVITY detector's own firing signature — >= 10x worker-time spread on one cached plan, same floors (min >= 10ms, max >= 250ms, execs >= 20, compiled before the window), same standard analysis window — joined to the regressed query by (database, query_hash). Using the detector's exact thresholds is the honesty guarantee: the flag can never claim evidence the detector would not report. Grant/spill divergence stay metadata on the PSP side; they don't fire the detector alone, so they don't fire this flag alone.

**Force-plan preview caution** for flagged targets (the #1882 disclosure pattern): names the plan-cache variance evidence, states the risk plainly — forcing pins ONE shape for every parameter value, and the population that preferred the other plan inherits the wrong one permanently, quietly, because a forced plan no longer recompiles away — then names the gentler levers first (statistics updates; PSP optimization / Query Store hints on 2022+). The copy-paste surface gets a compact two-line version, because that's the surface that gets executed. Comment-only on both: the runnable statement is byte-identical.

**Render stability**: unflagged targets — absent key (old persisted drill-downs, deprecated Dashboard) and explicit false alike — render byte-identically to before, pinned.

**`ForcePlanTarget.ParameterSensitivityCoFired`** appended with a default, same wire-compatibility discipline as `ReplicaRole` (#1882); legacy persisted-action JSON round-trips to false, pinned.

## Recon finding (recorded on the issue)

The `parameter_sensitivity_cofired` hook that already existed in `FactRemediation`'s CLEAR_PLAN path is written ONLY by the deprecated Dashboard's drill-down — Lite and Darling never produce `abnormal_cpu_plans` rows, so that flag is dormant in the live SKUs. This PR builds the live-SKU flag fresh, per-row and data-driven.

## Bot consequence (standing, documented in the record and renderer docs)

A flagged target is **never auto-forced** by the future Phase 1+ bot — it gets an investigate verdict. This flag is the first hard gate that machinery inherits.

## Pins

- Drill-down: flag TRUE at 20x cache spread, FALSE at 2x (same floors — pins the threshold, not mere cache presence). DuckDB suite + live-Postgres twin (real SQL both stores; live cleanup now clears query_stats too).
- Extractor carries the flag; preview renders the caution; paste surface renders the two-line version with the runnable statement untouched; legacy JSON defaults false with no caution.

Part of #2138 — leave the issue open (phases 1–2 remain design-gated).


erikdarlingdata:
Confirmed real, and the sharpest catch of the arc: the render path in both apps goes through the DESERIALIZED action, so the missing DTO mirror meant the caution never reached production surfaces at all — the in-memory tests passed while the actual wire dropped the flag. Fixed in 6a628eb8 following the ClearPlanFiguresDto pattern (ForcePlanTargetDto + ToDto + FromDto), pinned by a true-flag round-trip test through SerializeAction → DeserializeAction → RenderCopyPasteCommand. The legacy-JSON default-false pin already covered the other direction.

Issue #2143, opened 2026-08-09
Retention-purge e2e can deadlock against the background jobs it just created (nightly flake; consider product-side 40P01 retry)


erikdarlingdata:
## Nightly flake: retention-purge e2e deadlocks against background jobs on the shared dev store

`TimescaleSupportTests.EndToEnd_DetectConvertAndDropChunksPurge_AgainstDevPostgres` failed in nightly run 31334656715:

- `drop_chunks` on collection_log hit `40P01: deadlock detected` (deadlock partner is almost certainly a TimescaleDB background job — the test converts hypertables and applies compression policies moments before purging, so a policy job waking up on the same chunk is a live race);
- the DELETE fallback then ALSO deadlocked, so the 70-day row survived and the assertion failed.

**Production behavior is already graceful** — the purge logged the warning, counted 1 failed table, and moved on; a real store retries next cycle and self-heals. This is a TEST-hardness gap, not a product bug: the e2e asserts single-shot success against a store where its own just-created policy jobs can race it.

Fix options (pick one):
1. Retry the purge once on `40P01` inside the TEST (deadlocks are transient by definition; one retry makes the race window vanishingly small), or
2. have the test disable/pause the compression+retention jobs it created before running the purge assertion (deterministic, slightly more test code), or
3. teach the PRODUCT purge to retry once on 40P01 before falling back to DELETE — defensible on its own merits (a deadlocked drop_chunks wastes a whole cycle for a lock that clears in milliseconds), and the test inherits the fix.

Leaning 3 + 1: a single product-side retry is cheap and honest (log the retry), and the test keeps one retry as belt-and-suspenders.

Caught by the nightly on the #2141 merge commit; unrelated to that PR's content (nothing in it touches retention). The other failure in that run (Settings-box wiring) is fixed in #2142.

erikdarlingdata:
Fixed by #2145: drop_chunks retries exactly once on 40P01 (the partner's locks clear in milliseconds), a second deadlock gives up to the existing DELETE-fallback posture, non-deadlock failures stay single-shot, cancellation propagates. Four unit pins through the delegate seam; the purge e2e inherits the retry, closing the flake at its root. Robot review: no findings — it also confirmed each retry attempt opens a fresh connection, so no aborted-transaction reuse.

Issue #1888, opened 2026-07-31
CI's throwaway PostgreSQL runs max_worker_processes = 8, so TimescaleDB background jobs mostly cannot launch — the gated-live suite tests a configuration the product deliberately avoids


erikdarlingdata:
Found while root-causing the #1862 CI failure. Not fixed there, because raising the limit makes background policy jobs launch RELIABLY, which is a behaviour change for the whole gated-live suite and wants its own verification pass.

## The gap

The product sizes worker processes deliberately. `DarlingManagedPostgres.BuildConfAppend`'s v2 worker block derives them from the live hypertable count:

```
timescaledb.max_background_workers = HypertableCount + 2
max_worker_processes               = 3 + (HypertableCount + 2) + 8
```

and its comment records exactly why, from a live capture: PostgreSQL's default `max_worker_processes = 8` "cannot launch TimescaleDB's 26 per-hypertable compression policy jobs — the postmaster logs 'failed to launch job ... failed to start a background worker' storms and most policy runs fail (caught live: 21 failures vs 7 successes in `timescaledb_information.job_stats`)".

CI's throwaway cluster does not use `BuildConfAppend`. The `Initialize and start throwaway PostgreSQL` step in `build.yml` (and the mirror in `nightly.yml`) appends only three settings:

```
shared_preload_libraries = 'timescaledb'
port = 5541
listen_addresses = '127.0.0.1'
```

So the gated-live suite runs against `max_worker_processes = 8` (PostgreSQL's default) with `timescaledb.max_background_workers = 16` (TimescaleDB's default) — the exact under-provisioned shape the product exists to avoid.

## Why it matters

**The suite is testing a configuration no customer runs.** Every TimescaleDB background policy — compression, retention, continuous-aggregate refresh — is launched by a background worker. On CI those launches routinely fail outright, so a large part of what the product does in the field is exercised only by the tests' explicit foreground `run_job` calls, never by the scheduler that actually drives it in production.

**And it makes failures intermittent instead of reproducible.** Whether a given job launches depends on a free slot at that instant, so a test that races the scheduler passes or fails by luck of slot availability. That is precisely how #1862's compression flake behaved: it failed twice on CI in two unrecognizably different ways, and could not be reproduced locally at all until `max_worker_processes` was raised on the same rig, after which the background run fired every time.

[...]

PR #2144, opened 2026-08-09
#2136: synthetic scale test — volume→duration→telemetry→alert end to end

erikdarlingdata:
Per Erik: nothing closes without testing. This is the validation half of #2136, live against a scratch store:

- **1x → 4x volume** through a throwaway hypertable's compression job (parked policy created in one transaction, deterministic `run_job` — the #1888 discipline, immune to the #2143 scheduler-race class), with a discarded warm-up run so first-run costs can't invert the measurement.
- **Pin 1 — the capacity claim**: `job_stats.last_run_duration` is measurable after `run_job` (failed loudly with a diagnosis if TimescaleDB ever stops accounting foreground runs) and grows with volume. Monotonicity, not a ratio: runner jitter owns the constant factor; the direction is the model.
- **Pin 2 — the telemetry**: the V56 `store_metrics` series carries both readings, in order, matching the measured durations exactly.
- **Pin 3 — the alert**: `alter_job` shrinks the schedule interval to half the measured 4x duration, then the REAL `ReadJobCadenceReadingsAsync` feeds the REAL evaluator, which fires **Critical** under `storejob:<id>` — no fabricated readings anywhere in the chain.

With this merged, both named halves of #2136 plus its validation are done; I'll close the issue on merge.


erikdarlingdata:
CI's first execution earned its keep: the loud-premise assert fired — foreground `run_job` does **not** update `job_stats.last_run_duration` (that accounting lives in the scheduler path). Production telemetry was never at risk (it reads scheduled runs), but the test's mechanism was wrong. Fixed in 6c9d00a7: each measurement arms the parked job with `next_start => now()`, polls `total_runs` until the real scheduler runs it (90s deadline with a diagnosis pointing at CiClusterWorkerSizingTests), then parks again. Strictly better — the duration now flows through the exact accounting production reads.

erikdarlingdata:
Round 2 taught the next lesson: `total_runs` increments at run **start**, and `last_run_duration` reads NULL mid-flight — the 4x run was caught in flight (d4=0ms) while the 1x run just happened to finish inside one poll tick (d1=98ms, a real measurement, so the scheduler mechanism itself is proven). Fixed in de026d60: poll `last_successful_finish`, which only advances at completion. Two CI rounds, two real TimescaleDB accounting semantics pinned into the test's comments — this is why the scale test exists.

PR #2146, opened 2026-08-09
#2138: machine-first structured_remediation for MCP consumers

erikdarlingdata:
The build-out of the MCP-first design posted on the issue (Erik: "mcps are going to look at this more than people. what would help them most?").

## What agents get

`analyze_server` and `get_analysis_findings` (both SKUs) now emit `structured_remediation` alongside the prose `remediation_command` — per force-plan target:

- **verdict**: `eligible` + **named** blockers (`parameter_sensitivity_cofired`, `secondary_replica_evidence`) — the same decisions the comment prose carries, as data;
- **evidence**: regression factor, latest/best cpu-per-exec, the PSP flag;
- **join keys**: database, query_id, plan ids, both plan hashes, replica_role — enough to pivot straight into `analyze_query_store_plan` / `get_plan_xml`;
- **split artifacts**: `force_sql` / `unforce_sql` / `verify_sql`, each self-contained with its own USE — `verify_sql` asks both post-force questions (did the force stick: `is_forced_plan`/`force_failure_count`/failure reason; and the per-interval cost since). All advisory; the read-only MCP never executes.

## The architectural point

`FactRemediation.ForcePlanBlockers` is **the** policy gate: one implementation that fills the blockers agents inspect today and that the Phase 1+ bot consults tomorrow — "never auto-force a flagged target" becomes a testable data contract, not a promise in comments.

Built at **read time** from persisted targets, deliberately never persisted itself: no DTO mirror to forget (the exact #2140 catch class), one source of truth.

## Pins

Gate arms (clean primary → eligible; PSP flag → named blocker; Secondary/Geo Secondary role → named blocker; Primary role → eligible; both → stacked), split artifacts including verify content, null/non-force → null, and a wire-shape pin serialized under `McpHelpers.JsonOptions` — snake_case exists only by attribute, so a record-property rename fails a test instead of silently breaking agent consumers.

Part of #2138 — leave open (stats classifier + phases 1–2 remain gated on Erik).


erikdarlingdata:
Both findings fixed in 2e5f100e: (1) wire-level assertions in both suites per the established remediation_command pattern — Lite now plants a FORCE finding with a PSP-flagged target and asserts the verdict crosses the wire intact (ineligible + named blocker + split artifacts + evidence) after the store round-trip, which doubles as end-to-end proof the #2140 DTO mirror carries the flag through persistence; Darling's envelope pin gains the field with present-but-null. (2) All four McpInstructions tool-guide rows now mention structured_remediation — agreed that the instructions doc is the first thing an agent reads, so the field being invisible there defeated the PR's premise.

Issue #2148, opened 2026-08-10
[BUG] Regression on 3.4.0 - no CPU Data on Elastic Pool


TrudAX:
### Component

Lite

### Performance Monitor Version

3.4.0 (from the release section, but it still displays 3.3.0)

### SQL Server Version

Azure Elastic Pool

### Windows Version

Windows Server 2022

### Describe the Bug

CPU % graph do not display any data after installing 3.4.0

see for example this picture, I download a new version at 9.40am, after that - no display for CPU load

<img width="894" height="534" alt="Image" src="https://github.com/user-attachments/assets/6c984c1e-970d-4b8e-bbd6-5d933289e135" />

### Steps to Reproduce

Install update 3.4.0 (instead of 3.3.0)
CPU % graph are not displayed now 

### Expected Behavior

CPU % should display some data 

### Actual Behavior

No data in CPU %

See also collection heath info

<img width="1958" height="1081" alt="Image" src="https://github.com/user-attachments/assets/d92b5644-6ab0-4349-ba3f-92405732eea5" />

### Error Messages / Log Output

```text

```

erikdarlingdata:
Thanks — and thank you for including the Collection Health screenshot, because it shows something bigger than the CPU chart: **your whole collection loop stalled ~12 minutes after the upgrade**. Look at the timestamps: your screen refreshed at 10:36, but the newest 'Last Run' anywhere is 9:52–9:55, and most collectors' 'Last Success' froze at 9:38–9:39 — the minute you upgraded. The CPU chart is just where you noticed it first: no collector has landed a row since ~9:55.

**Working root-cause hypothesis (3.4.0 regression, likely specific to many-database servers like elastic pools):** 3.4.0 introduced a Query Store history backfill worker. It runs in the same background loop as collection, one slice per tick — and on first contact after the upgrade, *every database in your pool* has a pending backfill tail to drain. If a slice runs long against a big Query Store (a busy elastic pool qualifies), it blocks the next collection sweep for that entire duration. That mechanism matches your timeline exactly.

**To confirm it, two things would nail it:**
1. The log file for today from the `logs` folder under Lite's data directory (zip and attach) — it will show what the background loop was doing after 9:55 (backfill slice lines, and how long each took).
2. Roughly how many databases are in the monitored pool?

[...]

Issue #2147, opened 2026-08-09
[BUG] When download lite 3.4.0, it still display 3.3.0


TrudAX:
### Component

Lite

### Performance Monitor Version

3.4.0

### SQL Server Version

Azure elastic pool

### Windows Version

Windows Server 2022

### Describe the Bug

Download the latest lite release 3.4.0, it still display 3.3.0

### Steps to Reproduce

<img width="886" height="755" alt="Image" src="https://github.com/user-attachments/assets/0e66b86e-6253-4f3a-ad89-2a59c34ebe17" />

Download the latest lite release 3.4.0, it still display 3.3.0

### Expected Behavior

Should be 3.4

### Actual Behavior

Still 3.3

### Error Messages / Log Output

```text

```

erikdarlingdata:
Thanks for the report — you've hit a known stamping defect in the shipped 3.4.0 package (duplicate of #2113, same root cause).

**What happened:** when the 3.4.0 release was cut, the project carried four hand-maintained version properties; only `<Version>` was bumped, so the shipped binaries are still STAMPED 3.3.0.0 — and the window title/About box read that assembly stamp. **The zip you downloaded IS functionally 3.4.0** — every 3.4.0 feature and fix is in it; only the displayed version string is wrong.

**Already fixed in source:** the version is now single-sourced (`<Version>` alone; AssemblyVersion/FileVersion/InformationalVersion derive from it at build time), so every build since — including current nightlies — stamps correctly, and the next release cannot recur this.

Closing as a duplicate of #2113. If the mis-stamped 3.4.0 assets get re-uploaded with corrected stamping, this thread will get a note — that's a release-management call for the maintainer. Sorry for the confusion, and thanks for taking the time to file it with a screenshot.

erikdarlingdata:
Update: the v3.4.0 release archives have been rebuilt with correct version stamping and re-uploaded (same code, fixed metadata — binaries now identify as 3.4.0.0). If you re-download [PerformanceMonitorLite-3.4.0.zip](https://github.com/erikdarlingdata/PerformanceMonitor/releases/tag/v3.4.0) it will display 3.4.0 and the update nag ends. Checksums on the release page are current.

Issue #2113, opened 2026-08-07
[BUG] PMLite 3.4.0 package installs binaries stamped 3.3.0


SalmanRajwani:
### Component

Lite

### Performance Monitor Version

3.4.0 package; installed binaries report 3.3.0

### SQL Server Version

N/A

### Windows Version

Windows 11

### Describe the Bug

After installing the Performance Monitor Lite 3.4.0 package, the package metadata reports 3.4.0 but the installed EXE/DLL binaries still report 3.3.0.

### Steps to Reproduce

1. Install `PerformanceMonitorLite-3.4.0-full.nupkg` / 3.4.0 Lite setup.
2. Check package metadata / `sq.version`.
3. Check file version metadata for `PerformanceMonitorLite.exe` and `PerformanceMonitorLite.dll`.

### Expected Behavior

The 3.4.0 package should install binaries stamped as 3.4.0.

### Actual Behavior

The package metadata reports 3.4.0, but the installed app binaries report 3.3.0 / 3.3.0.0.

### Error Messages / Log Output

No application log output. This was verified from package metadata and file version metadata.

### Additional Context

This does not appear to be an old shortcut target. The shortcut points to the current install folder.

erikdarlingdata:
Confirmed, and thank you — this was verifiable straight from the tags: the 3.4.0 release bump updated `<Version>` (which names the package) in each app project but missed the three hand-pinned stamp properties (`AssemblyVersion` / `FileVersion` / `InformationalVersion`), which stayed at 3.3.0. Every earlier release had moved all four together; this one didn't, and the same skew exists in the Darling projects too.

Two things worth knowing on your side:

1. **Your installed code is genuinely 3.4.0.** The package contents are built from the 3.4.0 tag — only the version *stamps on the binaries* lie. Nothing to reinstall, no behavior difference.
2. The fix (#2116) deletes the pinned properties so all stamps derive from the one `<Version>` line at build time — this class of skew structurally can't recur. Correctly-stamped binaries ship with the next release.

erikdarlingdata:
Correction to my "no behavior difference" above — review turned up a real consequence: the in-app update check compares the entry assembly's version (the stale 3.3.0.0 stamp) against the latest release tag, so your install would keep offering the 3.4.0 update you already have. Harmless to accept (it reinstalls the same code), annoying until the next release ships the fixed stamps. Everything else stands.

erikdarlingdata:
[...]

PR #2149, opened 2026-08-10
#2148: one wedged ladder step must never stop all collection

erikdarlingdata:
Fix for the live 3.4.0 field regression (TrudAX, Azure elastic pool): all collection stopped permanently ~12 minutes after upgrading — every collector's last success frozen at the upgrade minute, last runs frozen shortly after, 40+ minutes of silence at screenshot time.

## Root cause class

Lite's collection ladder runs its steps **sequentially**, and every step's exception armor was intact — but armor bounds *throws*, and nothing bounded a **hang**. Two steps could hold the loop without bound: the Query Store backfill tick (new in 3.4.0, and the prime suspect on the reporter's timeline — first contact after upgrade queues a backfill tail for every database in the pool) and the connection check (un-timeboxed Azure token/network paths). One wedge = every chart silent, permanently, with nothing in the log.

## The fix

Both steps now run under **AbandonableStep** (new, in PerformanceMonitor.Common) — the ladder's own scheduled-analysis idiom extracted and made reusable: `Task.WhenAny` against a deadline + an in-flight guard cleared only when the task TRULY ends.

- **Abandoned** (deadline elapsed): the loop keeps collecting; ERROR logged naming #2148.
- **Quarantined**: a wedged run is never overlapped by a relaunch (no stacking hung slices).
- **Self-restoring**: the moment the wedged task actually dies, the step runs again — a guard that never released would turn one hang into a permanently dead step, which is the bug again with extra steps.
- Deadlines (backfill 180s, connection check 90s) are generous multiples of healthy behavior (a tick is one 30s-capped slice per server), so an abandonment line is always a defect signal, never jitter.
- Shutdown reads as **Cancelled**, never a false ERROR.

## Pins

All five outcomes; the quarantine (second run skipped while wedged, never overlapped); the release (step runs again when the wedge ends); abandoned-then-faulted is observed (no UnobservedTaskException) and still releases; synchronous delegate throws contained; cancellation semantics.

## Notes

- This makes the failure **survivable and diagnosable**; the reporter's log (requested on the issue) will confirm which step wedged and whether the backfill needs its own deeper fix (e.g., off-ladder execution). Kept minimal deliberately — it's the safe 3.4.1 candidate.
- Flagging for Erik: this + #2145 + the #2113 stamping fix look like a **3.4.1 patch release**; the mis-stamped 3.4.0 assets (#2147) could be re-uploaded at the same time.

Closes nothing automatically — #2148 stays open pending the reporter's log confirmation.


erikdarlingdata:
Both findings taken in 68034746 — and both were right to insist on:

1. **Darling parity**: `RunQueryStoreBackfillLoopAsync` now wraps each server's slice in a per-SERVER `AbandonableStep` (keyed by ServerId, 300s deadline vs the 60s statement cap), so one wedged server is abandoned loudly and quarantined while the rest of the fleet keeps backfilling — strictly better than the Lite shape, where the whole tick shares one guard. The changelog claim is now true for both apps.

2. **Late faults**: `RunAsync` gains `onLateFault`, wired at all three call sites as 'this is the wedge's own exception (#2148)'. It fires only for faults the awaited path did not already return, so ordinary failures never double-log; the microsecond race at the abandonment flag can cost the callback (never double it), which loses one log line, not correctness — documented on the parameter. A throwing callback still releases the guard, pinned.

Three new pins cover the callback arms.

erikdarlingdata:
Round-2 main finding taken in 3774ae51 — correct on both counts: the shared tick-level guard was a narrower rerun of the very symptom this PR fixes (one wedge → all Lite servers' backfill quarantined forever) AND a false-abandonment risk scaling with fleet size. Lite's guards now live inside the tick, per server, in exactly the Darling loop's shape; the ladder-level wrapper is gone because the tick is bounded by construction.

The secondary point (migrating both apps' hand-rolled scheduled-analysis guards onto AbandonableStep so there aren't three copies of the pattern) is real but out of scope for the 3.4.1 candidate — noting it on #2148 as follow-up.

erikdarlingdata:
Round-3 finding taken in a2cbcedf: the changelog entry now describes the shipped architecture — per-server quarantine in both apps, late-fault surfacing, and the connection-check treatment — instead of the first-commit tick-level shape. Agreed on the reasoning: this entry is the field-diagnosis reference for exactly this incident class, so it describing a superseded design would have cost someone an hour.

Issue #2151, opened 2026-08-10
[BUG] in update from v 3.3 to 3.4


sm2nashalj-commits:
### Component

Lite

### Performance Monitor Version

3.3

### SQL Server Version

2022

### Windows Version

11

### Describe the Bug

the program did not update from v 3.3 to 3.4, even after downloading and installing new update or even reinstating,, it stock to version 3.3 and keep notify to upgrade every time the app is starting

### Steps to Reproduce

help/about/check for update/download/install the update

### Expected Behavior

update to v 3.4

### Actual Behavior

remain in v 3.3

### Error Messages / Log Output

```text

```

erikdarlingdata:
Thanks for the report — your install almost certainly DID update to 3.4.0. This is a duplicate of #2113 (also #2147): the shipped 3.4.0 package's binaries are mis-STAMPED as 3.3.0, because the release was cut with only one of four version properties bumped. Your symptom is the nastiest consequence of that: the update checker reads the same wrong stamp, so a correctly-installed 3.4.0 reports itself as 3.3, nags you to upgrade, you reinstall, and the loop repeats — nothing you did wrong, and nothing actually failing to install. Every 3.4.0 feature is in the binaries you have.

The fix (single-sourced version stamping) is already merged, so the next release breaks the loop permanently. Closing as a duplicate of #2113 — sorry for the confusing loop, and thanks for taking the time to file it.

erikdarlingdata:
Update: the v3.4.0 release archives have been rebuilt with correct version stamping and re-uploaded (same code, fixed metadata — binaries now identify as 3.4.0.0). If you re-download [PerformanceMonitorLite-3.4.0.zip](https://github.com/erikdarlingdata/PerformanceMonitor/releases/tag/v3.4.0) it will display 3.4.0 and the update nag ends. Checksums on the release page are current.

PR #2152, opened 2026-08-10
One-off: restamp workflow for the v3.4.0 archives (#2113/#2147/#2151)

erikdarlingdata:
Erik-authorized (this morning, on the back of #2151 — the third user report from the 3.4.0 stamp): adds a workflow_dispatch-only workflow that rebuilds the three PLAIN archives from `restamp-3.4.0` (= the v3.4.0 tag + only the single-source stamp fix cherry-picked, verified version-properties-only) and clobbers them + merged checksums onto the existing v3.4.0 release.

- **Stamp-verification gate** before any upload: the job throws if the published binaries don't read 3.4.0.
- **Checksums merged**, not replaced: rebuilt assets get new hashes; untouched assets keep their existing lines.
- **Velopack chain untouched by design** (delta/full nupkgs, Setup.exe, releases.*.json) — regenerating the updater feed blind risks breaking updates; that stays the release-cutter's call.
- Targets main only because `workflow_dispatch` requires default-branch registration. **Revert after the run.**

Build steps mirror nightly.yml verbatim (same publish shapes, archive layouts, pg-runtime staging), version fixed at plain 3.4.0.

### DIFF
--- a/.github/workflows/restamp.yml
+++ b/.github/workflows/restamp.yml
@@ -1,0 +1,139 @@
+# ONE-OFF (#2113/#2147/#2151): rebuild the v3.4.0 plain archives from the restamp-3.4.0 branch
+# (= the v3.4.0 tag + ONLY the single-source version-stamp fix cherry-picked) and clobber them onto
+# the existing v3.4.0 release, so downloads stop identifying as 3.3.0 and the update-nag loop ends.
+#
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Closing — the branch gate correctly only allows dev→main, and there's a cleaner path: push-triggered workflows run from any branch without default-branch registration. The workflow now lives on restamp-3.4.0 with an on:push trigger scoped to that branch.

Issue #2153, opened 2026-08-10
[BUG] Installed 3.4.0 but About shows: 3.3.0


ZedZipDev:
### Component

Lite

### Performance Monitor Version

3.4.0

### SQL Server Version

Azure SQL DB

### Windows Version

W11

### Describe the Bug

After the 3.4.0 installation the About window shows 3.3.0 and Check Update again prompted to update 

<img width="430" height="349" alt="Image" src="https://github.com/user-attachments/assets/793326bc-4e8f-46e2-ba51-6bee281ff9e8" />

### Steps to Reproduce

Install, run, open About

### Expected Behavior

3.4.0

### Actual Behavior

3.3.0

### Error Messages / Log Output

```text

```

ZedZipDev:
and also: 
1. Last 24 hrs after installation again Last 4 hrs
2. If I click on Custom Range, the Calender appears but I do not click it and go back to the current setting the Calender  does not disappear.
3. Using query I see deadlock appeared on Friday but PML does not show it in Deadlocks.

erikdarlingdata:
Thanks for the detailed report — four things in here, taking them in order:

**1. About shows 3.3.0.** Known defect, duplicate of #2113: the 3.4.0 release was cut with mis-stamped binaries (the code IS 3.4.0 — every feature is there — only the version metadata is wrong, which also confuses Check for Update into re-prompting). The plain zip on the release page was rebuilt with correct stamping this morning, but you installed via **Setup.exe**, and the installer packages still carry the old stamp — those get refreshed with the upcoming 3.5.0 release, which will end the update-nag loop for installer users. Until then: the nag is safe to ignore, and reinstalling won't change what About shows.

**2. Time range reset to "Last 4 hrs" after (re)installation.** Likely a side effect of the reinstall cycle the version nag pushed you into. If the time range resets across a NORMAL app restart (no reinstall), that's a separate bug — could you confirm which it is?

**3. Custom Range calendar staying open when you navigate back without picking a date.** Genuine UI bug, reproduced from your description — filing it separately so it doesn't get lost in this thread; it'll be linked here.

[...]

PR #2155, opened 2026-08-10
#2154: close the DatePicker dropdowns when leaving Custom Range

### LINKED ISSUE #2154: [BUG] Custom Range calendar stays open when navigating back to a preset range without selecting
Reported by @ZedZipDev in #2153 (item 2 of their comment): open the time-range picker, click Custom Range so the calendar appears, then — without clicking a date — switch back to a preset range. The calendar does not disappear.

Expected: leaving Custom Range dismisses the calendar. Component: Lite, 3.4.0, Windows 11.

Likely a visibility toggle bound only to the Custom Range selection edge, not to deselection — needs the picker's hide path wired on any non-custom selection.

erikdarlingdata:
Fix for #2154 (reported by ZedZipDev in #2153): open the time-range picker, click Custom Range, open a date picker's calendar, then switch back to a preset **without selecting a date** — the calendar stayed floating on screen.

**Why**: a WPF DatePicker's calendar dropdown is a `Popup`, which lives outside the visual tree's visibility. The selection-changed handler collapses the pickers when leaving Custom Range, but collapsing a picker does not close its already-open popup — the orphaned calendar is the popup outliving its owner's visibility.

**Fix**: set `IsDropDownOpen = false` on both pickers alongside the collapse. Two lines plus the comment explaining the popup/visibility split so it doesn't get 'simplified' away.

No test: this is WPF popup behavior with no seam short of UI automation — the comment carries the regression knowledge. Closes #2154 on merge (explicitly, as always).


erikdarlingdata:
Parity catch confirmed and taken in 48fed89d — Darling's Viewer had the structurally identical handler with the identical bug; same two-line fix, same explanatory comment, changelog updated to say both apps.

Issue #2170, opened 2026-08-10
[FEATURE] Make MaxConcurrentServerSweeps configurable (hardcoded N=4 already bottlenecks well below the 500-server goal)


strikes223:
### Which component(s) does this affect?

- [ ] Lite
- [x] Darling
- [ ] Documentation

### Problem Statement

First off, thank you to Erik and everyone who works on this project — Darling has been a huge upgrade for our fleet monitoring, and the amount of care in the collector design, migration history, and even the log-severity nuance in #1597 really shows. We really appreciate the work that goes into this.

We're running Darling 3.3.0 on a dedicated Azure VM (8 vCPU / 32 GB RAM, SSD-backed storage, same Azure region as the monitored fleet) watching a 56-server fleet (mix of SQL Server and Azure SQL Database). The host is not resource-constrained — CPU, memory, and disk I/O all have comfortable headroom during the exact windows when the issue below occurs, so this isn't a "the box is too small" problem.

We're intermittently seeing "collection stale" warnings on the Fleet Health Overview for multiple servers, even though every individual collector reports `HEALTHY` via `get_collection_health` (zero errors, recent `last_success` timestamps, well within the classifier's own 4-hour/1.5x-cadence STALE floor). Digging in, we traced this to `MaxConcurrentServerSweeps`, which is hardcoded to `4` and not exposed via `darling.json` or any CLI flag. With 56 servers cycling through that fixed gate, 1-minute-cadence collectors are queued long enough (observed gaps of ~2m30s–3m) to cross the Fleet Overview's much tighter freshness check (roughly 2x the fastest collector's cadence, ~2 minutes) — even though nothing is actually failing and the underlying data is only a couple of minutes stale.

This was already flagged as a known future limitation by the maintainer in #1597 ("Deliberately not done" section):

> `MaxConcurrentServerSweeps` is **not** raised — that trades directly against the memory bounds won in #1557/#1581 and is the wrong lever for an alignment problem. Separately worth tracking: N=4 doesn't survive the 500-server goal (125 waves), so the gate will eventually need to scale with fleet size against the existing working-set guard.

We're hitting this well before the 500-server figure discussed there, at 56 servers — which may be useful signal that the practical threshold is lower than expected, at least for fleets with a lot of 1-minute-cadence collectors in play.

### Proposed Solution

[...]

Issue #2164, opened 2026-08-10
[PERF] QS collector drain-mode passes cost the full byte budget every cycle — make the budget a knob, skip already-shipped plans by hash


erikdarlingdata:
### Component
Shared collectors (QueryStoreCollector), both apps

### Field evidence (dogfood, 2026-08-10)
A 4-core multi-tenant box post-resize-restart (all plans recompiled) + a freshly restored tenant catalog: every database hit `MaxTextBytesPerDatabase` (64MB, compile-time const) on every pass, at 44–109s of SQL time per database, ~380–427s per full server pass, repeating every cycle. The staged shape (#2134) worked exactly as designed — bounded, resumable, no wedge — but the drain tax is large and highly visible (60–100s `PerformanceMonitorLite` statements in sp_WhoIsActive on a prod primary).

### Why passes cost this much
1. The per-pass ceiling is generous and fixed: `MaxTextBytesPerDatabase = 64 * 1024 * 1024` (QueryStoreCollector.cs:327). 64MB of QS text/plan extraction ≈ 45–110s on small hardware because `CONVERT(nvarchar(max), qsp.query_plan)` decompresses plan XML server-side — that's the dominant CPU term.
2. **No cross-pass plan memory**: `query_plan_text` ships once per plan *per pass* (`ROW_NUMBER() OVER (PARTITION BY qsp.plan_id ...) = 1`). A hot plan re-decompresses and re-ships every cycle; the store dedups on arrival — after the server already paid the extraction. In a recompile storm, most of every 64MB is plans the store has.

### Proposed levers
- **(a) Budget knob**: make the per-database text byte budget a store setting (per-server override ideal), default 64MB, floor ~4MB — the #2107 store-backed knob pattern. Small boxes drain in 15–25s bites instead of 60–110s ones; same total work.
- **(b) Known-hash skip (the big win)**: `qsp.query_plan_hash` is available *before* the CONVERT. Maintain the recently-shipped hash set per (server, database) in collector state, pass it in (temp table / TVP), and extend the ship condition to `ROW_NUMBER = 1 AND qsp.query_plan_hash NOT IN (SELECT h FROM #known)`. Recompile storms then ship stats rows but almost no plan payload. Caveat to design around: `query_plan_hash` is a plan-shape hash, not a full-XML digest — distinct XML can rarely share a hash; acceptable for perf telemetry but should be a documented tradeoff (or refresh hashes older than N days).

[...]

PR #2172, opened 2026-08-10
Collector memory bounds become operator knobs — V59 text budget + sweep width (#2164, #2170)

erikdarlingdata:
## What

**V59** puts two compile-time constants on the store control plane, live-reloaded, clamped on read:

| Knob | Was | Default | Clamp |
|---|---|---|---|
| `query_store_text_budget_mb` | `QueryStoreCollector.MaxTextBytesPerDatabase` = 64 MB | 64 | [4, 256] |
| `max_concurrent_sweeps` | `DarlingWorker.MaxConcurrentServerSweeps` = 4 | 4 | [1, 16] |

Defaults reproduce the old constants exactly — an upgraded store behaves identically until a dial moves. Both appear in the Viewer's Settings window beside the plan-capture and backfill toggles, with a hint naming their interaction.

**Why one rung for both:** peak transient memory is approximately `sweeps × budget`. Shipping them separately invites an operator to raise one without seeing the other.

**Budget path:** `CollectorContext.TextByteBudgetOverride` (bytes; null = keep the definition's own `PerItemTextByteBudget`, which is what Lite passes so its behavior is untouched), consulted at the single enforcement site in `QueryStoreCollector`'s read loop. Darling supplies it through a live provider, same shape as `capture_plans`, so a store reload takes effect next cycle without rebuilding the runner. Guarded `> 0` so a corrupt value can never mean "ship nothing".

**Sweep path:** a `SemaphoreSlim` cannot be resized, so the gate is constructed at the clamp ceiling (16) and immediately narrowed to the configured width. Widening `Release`s permits back; narrowing absorbs the surplus in the background as in-flight bodies finish — so a knob change never blocks the collection loop and never cuts a running collection short. Reconciled at the top of a sweep, right after the config-version reload.

## Why (field)

2026-08-10, cross-region 4-core production primary: 64 MB per database per pass is a **minute** of the monitored server holding one collector query open draining to a client a network hop away (ASYNC_NETWORK_IO-dominated), and tick passes sat at 400–500s indefinitely — they did not converge even after the backfill was stopped, because nothing bounds the cost below the budget. Separately, #2170's reporter has a 56-server fleet on a host with headroom where 4-wide serialization makes the Fleet Health screen report staleness while every collector reports HEALTHY.

This PR makes both bounds tunable. The remaining structural fix — not re-shipping plan XML the store already holds — is #2164's second half and follows in its own PR; it is what makes the backfill safe to re-enable.

## Testing

`CollectorMemoryKnobTests`: V59 rung identity, defaults pinned **against the constants they replace** (so an upgrade can't silently re-tune deployments), both clamps across bounds including 0/negative/over-ceiling, the clamp ceiling and the semaphore's construction ceiling pinned equal by contract, `ApplyToConfig` round-trip, and the context-override precedence Lite depends on. Ladder-tip pins moved 58→59 suite-wide; the V58 rung pin now states the V59 flag false so it keeps testing V58 rather than drifting into "the newest rung". Collectors, Service, Viewer, and test projects all build clean.

Addresses #2164 (budget half) and #2170. Both stay open until their remaining halves land — explicit close discipline applies regardless.


erikdarlingdata:
Review round taken — both findings were real bugs, thank you:

1. **Budget never reached backfill slices** — correct and important, since backfill is the *heavier* of the two paths over the same link, so a knob that only bounded the tick would have left the PR's own field motivation unaddressed. `QueryStoreBackfill` now takes the same live `Func<int>` provider and sets `TextByteBudgetOverride` on its context.

2. **Permit-stealing race** — also correct, and the failure mode is worse than it first looks: the gate would sit *below* the configured width indefinitely, i.e. the knob would silently not work in exactly the direction operators care about. Replaced the per-call absorb loop with `absorbed`/`desired` counts under a lock plus a single absorber that re-reads the target both before and after every wait, handing a granted permit straight back when the width grew while it was parked. Pinned by a test that holds every permit (so the absorber is genuinely parked), narrows, widens, releases everything, and asserts the gate lands at the configured width retaining nothing.

3. **Orphaned doc comment** — already fixed in `5edad09f` before this round: my insertion of `ReconcileSweepGate` had displaced `RunQueryStoreBackfillLoopAsync`'s summary onto it. The `NoMemberCarriesTwoStackedSummaryBlocks` pin caught it, and its own guidance was right — the fix was moving the block back, not deleting it.

erikdarlingdata:
Third finding taken — you were right that the startup path was the uncovered one, and the fix is to remove the window rather than shorten it: `SemaphoreSlim` accepts an initial count below its max, so the gate is now **born at the configured width** with the ceiling as its cap, and the permits it was never issued seed the absorbed count. No absorber runs at startup at all, so there is no interval where more than `max_concurrent_sweeps` bodies can launch — which matters most in precisely the scenario you named, a restart with many servers simultaneously due before the #1581 stagger spreads them. Pinned by a theory across 1/4/16 that also asserts the gate can still be widened back to the ceiling later (Release past MAX throws, so that is what makes born-narrow safe rather than a one-way door).

erikdarlingdata:
Taken — and it's the sharpest of the four, because it's a diagnostic that would actively mislead in the exact scenario the knob exists for: an operator raises the width to 12, then reads `fleet concurrency limit 4` in the log they're using to judge whether it worked. Now reports `EffectiveSweepWidth` (ceiling minus the *desired* absorb, under the gate lock) — deliberately the target rather than the momentarily-larger real count during a narrow, since the target is the number to act on. Round 4 was 5/5 green before this; this is the last change.

PR #2173, opened 2026-08-10
Split per-database collection timing into open + drain — the measurement the next QS fix needs (#2164)

erikdarlingdata:
## Why this instead of the fix I said I'd build

I was queued to build #2164's second half (skip re-shipping plan XML the store already holds). Then I measured the first half on a production server and it overturned the premise:

| | rows shipped | `sql:` time |
|---|---|---|
| 64 MB budget (10 passes) | 2,530–8,626 | 358–480 s |
| 12 MB budget (2 passes) | 698, 1,233 | **383 s, 352 s** |

**5x less text, ~7x fewer rows, and the clock did not move.** So the cost is upstream of shipping, which indicts the hash-skip too — it targets `CONVERT(nvarchar(max), qsp.query_plan)` in the *final select*, i.e. the same term the budget already proved isn't dominant. Building it would have been a second guess dressed as a fix.

The blocker to knowing more is that our own log blends the batch: `ExecuteReaderAsync` blocks through the non-rowset `SELECT … INTO #pm_qs_slice`, so the aggregate and the final select are indistinguishable in `sql:`.

## What this does

Times the open separately from the drain, exploiting the fact that ADO.NET returns the reader only when the first **rowset** is available:

- **open** = every preceding non-rowset statement (for query_store, the `#pm_qs_slice` aggregate) + the final select's time-to-first-row. Nothing client-side shortens this.
- **drain** = row streaming, which the byte budget and the network path do govern.

The per-database line becomes `sql:Xms = open:Yms + drain:Zms`. Zero means not measured (Lite doesn't) and suppresses the split rather than printing `open:0ms`, which would read as "the aggregate was free". Cleared before each open so a faulted read can't log the previous item's split as its own.

## What it buys

After one nightly on the dogfood box, the field question becomes answerable instead of arguable: if those 350–480s passes are ~95% open, the fix is narrowing the server-side aggregate (bound its input, narrow the interval span, or run it less often on expensive servers) and no amount of payload trimming matters. If they're mostly drain, the budget knob was right and the link is the problem. Either way the next PR starts from a number.

## Testing

`StatementSplitTimingTests` pins the contract: zero means unmeasured (so an unmeasured host is never read as instant), drain is the remainder and never negative even under clock skew between the two watches, and the query_store read's own signal resets must not clobber the host's measurement — which if broken would make the whole instrumentation silently log zero. Collectors, Service, and test projects build clean.

Groundwork for #2164; no behavior change to collection itself.


erikdarlingdata:
Both taken, and the second one's fix is where the first one belonged:

1. **Watermark time was landing in drain.** Correct, and it matters more than a rounding detail — for query_store the refresh is a monitor-store read, and on the clamp/adaptive path a store write too, so on a catch-up cycle it could be a large chunk of what I'd have read as "streaming". Now its own measured phase: `sql:X = wm:A + open:B + drain:C`. Not cleared alongside the open reset, since it ran for the same item and clearing it would hand its milliseconds straight back to drain.

2. **The tests re-implemented the formula.** Fair, and provably so: a test that duplicates the arithmetic could not have caught finding 1, because my copy and the log line's copy were wrong in the same way. The subtraction now lives in `CollectorContext.DrainMsFrom` with one definition that both the log line and the tests call, and the pins cover watermark exclusion, the zero clamp under stopwatch skew, and that wm + open + drain sums to the total so nothing is silently unattributed. The tautological reset test is gone rather than reworded.

PR #2175, opened 2026-08-11
Forced Plan Failing alert — page when Query Store can't reproduce a forced plan (#2157)

### LINKED ISSUE #2157: [FEATURE] Alert on Query Store forced-plan force failures (force_failure_count deltas)
### Component
Darling (Lite alert engine could follow)

### What
The Query Store collector already captures, per plan: `is_forced_plan`, `plan_forcing_type`, `force_failure_count`, `last_force_failure_reason` (QueryStoreCollector.cs ~931-933, 1291-1294). Nothing evaluates them. When a forced plan stops being reproducible — classic trigger: database restored/migrated to a different server where the forced plan references structures that no longer resolve, or an engine version/compat change — every execution attempts the force, fails, and silently falls back to the optimizer's plan. The query keeps running (often slower); the only witness is `force_failure_count` climbing with a `last_force_failure_reason`.

### Proposed alert
Per (server, database, query_id, plan_id) where `is_forced_plan`:
- **Warning** when `force_failure_count` RISES between consecutive collections (delta, not level — the counter is cumulative, persists through restores, and a nonzero standing value is history, not news).
- Include `last_force_failure_reason`, `plan_forcing_type` (MANUAL vs AUTO both matter — AUTO = automatic plan correction's forces failing), and the delta in the alert body.
- Handle counter resets (unforce/re-force cycles) as re-arm, not negative-delta noise.
- Standing-condition semantics with recovery, like the other Darling alerts.

### Motivation (field case)
2026-08-10 tenant-database migrations: databases restored onto different monitored instances carry their QS content — including forced plans — with them. A force that can't reproduce on the target is exactly the silent-regression shape this alert catches. Baseline captured during the move: 8 AUTO-forced plans on the migrated database, all `force_failure_count = 0` pre/post-restore — the alert would have confirmed the clean landing instead of a hand-run query.

erikdarlingdata:
## What it catches

When a forced Query Store plan can't be reproduced, **the query keeps running** — on whatever plan the optimizer picks. Nothing in the product witnessed that: the operator's mitigation was silently not in effect, and the only trace was `force_failure_count` climbing inside Query Store.

The new **Forced Plan Failing** alert fires per plan on a *rise* in that counter, carrying database, query/plan ids, MANUAL vs AUTO forcing, the engine's own failure reason, and how many failures are new.

## The design decisions worth reviewing

**A rise, never a level.** The counter is cumulative *and it travels with a restored database* — restore one elsewhere and its Query Store arrives carrying every historical failure. Level-based firing would page forever about failures that happened on hardware the operator may not own anymore. So the STORE computes the delta (mirroring how the database-state adapter returns only deviating rows) and the engine never sees a level.

Corollaries, both pinned: a counter that **drops** is an unforce/re-force reset and is silence, not a negative delta. A plan with only **one** sample is omitted — one cycle of delay, because "new" isn't knowable from a single observation, and firing on first sight is exactly how you'd alert on every restored database's imported history.

**Warning for every rise, no Critical tier.** A failing force isn't an outage. Grading one Critical would need evidence about which reasons or rates correlate with harm, and I don't have it; the token doc says what evidence would justify it later.

**Per-PLAN keys.** Two failing plans in one database are independent conditions that resolve independently — a server- or database-keyed cooldown would swallow the second entirely. Pinned by a test with two plans in one database.

**The resolution says "no longer failing", not "fixed"**, because the counter also stops when the force is removed or the query stops running, and the alert can't distinguish those.

**Enabled with no toggle.** The alert is silent by construction on a healthy fleet (it needs a rise), so there's nothing for a switch to protect anyone from. Notably a `darling.json`-only flag would NOT have worked: the store reload swaps the whole `Alerts` object, so it would silently reset on the first `config_version` bump — if this ever needs to be configurable it belongs in the store, with the ladder that implies.

## Field motivation

2026-08-10 tenant migrations: six databases restored onto new hosts arrived carrying **51 forced plans** between them. Verifying they still reproduced took three hand-run baseline/delta passes across the wave. This alert is that check on a collection cadence.

## Testing

`AlertEngineTests`: the disabled gate skips the READ (not just the fire), two plans in one database both alert with reason/forcing/delta in the body, cooldown suppresses the second fire and a per-plan resolution follows when the counter stops, and an excluded database never alerts (case-insensitive, parity with the other database-scoped families). Four `IAlertReadAdapter` fakes and four `IAlertEngineSettings` fakes updated; Alerting, Darling service, Lite, and both test projects build clean.

Store SQL is shape-for-shape across Postgres and DuckDB so the apps can't disagree about what counts as new. Remaining for a follow-up: SQL-shape pins on the two adapter queries (the delta comparison, the forced filter, the two-hour bound).

Fixes #2157.


erikdarlingdata:
Both taken, and the first one is the more embarrassing of the two:

1. **Leaked internal key in the recovery text.** Real bug — operators would have read `forceplan:Sales:11:22 no longer failing to force` in toasts, emails, webhooks and history rows. The active set is now keyed by the plan key but VALUED with the plan's identity, so the resolution names it the way the firing message does. Worth stating plainly why my test missed it: I asserted `Message.Contains("22")`, which the leaked key satisfies — the assertion was compatible with the bug. It now asserts the key prefix is **absent** and that `Sales`, `query 11`, `plan 22` are present.

2. **The unquoted `INTERVAL 2 HOUR`.** Good catch on the reasoning as much as the syntax: nothing exercises that SQL against a real DuckDB yet, so a parse failure would have been swallowed by the catch-and-log path and this alert would have failed silently every sweep — which is precisely the failure class it was built to surface. Switched to `INTERVAL '2 hours'`, the form every other DuckDB query in the repo already proves parses. I'm treating 'match the form the codebase demonstrably runs' as the rule here rather than trusting my read of DuckDB's grammar.

The follow-up SQL-shape pins noted in the PR body would have caught #2 mechanically; that stays on the list.

PR #2176, opened 2026-08-11
Lite gets the Query Store backfill off switch (#2167 parity)

### LINKED ISSUE #2167: [FEATURE] Query Store backfill needs an off switch (global service toggle, hot-applied)
### Component
Both apps (Darling first — fleet-operational need)

### Why now
Field event 2026-08-10: a freshly restored tenant catalog put the #2058 backfill into sustained drain against a cross-region 4-core production primary — 64MB slices every cycle, dominated by ASYNC_NETWORK_IO (server holding the query open while the remote client drains), stacked on budget-bound tick collection. The operator's correct call was "stop the backfill on this deployment until the tuning lands" — and there is NO way to do that today: the loop runs unconditionally for every server; the only related knob (capture_plans) is global and guts plan collection everywhere, which is not an acceptable trade.

### Ask
`config_service.query_store_backfill_enabled boolean NOT NULL DEFAULT TRUE` (new migration rung), read into DarlingConfig via the live store-reload path so a flip applies next cycle without a restart, gating RunQueryStoreBackfillLoopAsync (skip iteration + log state transitions). Viewer settings toggle beside the capture-plans checkbox. Lite parity via its app settings.

Per-server granularity is the natural follow-up but the global switch is the emergency brake and ships first. Related: #2164 (budget knob + known-hash skip), #2165 (tick/backfill serialization).

erikdarlingdata:
## What

Completes #2167. Darling got the switch as a store column in #2168 (V58); Lite now gets it as a Settings checkbox — "Fill Query Store history gaps in the background", default on, persisted to `settings.json` as `query_store_backfill_enabled`.

## Two details that aren't obvious

**The switch is checked BEFORE the due-time stamp.** If it were checked after, a disabled backfill would still consume its own schedule, and flipping it back on would wait out an interval that elapsed while it was off. Checked first, re-enabling runs on the next due tick.

**Read live, not captured.** The due-check reads `App.QueryStoreBackfillEnabled` each pass, so the checkbox takes effect without restarting Lite — the same responsiveness Darling gets from its store reload. A captured bool would have quietly required an app bounce, which is the kind of thing nobody notices until they're mid-incident wondering why the switch did nothing.

**Transition-logged once**, not once per idle tick, seeded `true` to match the default so a deployment that never touches the switch says nothing about it in the log.

## Why the two apps differ in surface

Darling's is a store column because a headless service has no window to click, and its operator may be nowhere near the box. Lite's is a checkbox because it has a settings window and a single local user. Same behavior, appropriate surface each — noted in both places so the asymmetry reads as deliberate.

## Testing

Lite builds clean. The behavior is a two-branch gate on a settings bool in a background loop; there's no seam to unit-test that wouldn't just be asserting `if (!flag) return;` back to itself. The Darling half's store/probe/gate plumbing is pinned in `BackfillSwitchTests` from #2168.

Closes #2167 (its Darling half shipped in #2168).


erikdarlingdata:
Taken — and the diagnosis of how it happened is exactly right. I modeled the wiring on `AnalysisEnabled` (App static + settings.json key + checkbox) and inherited its **section** along with its **pattern**, so a data-collection control ended up under Automated Analysis, reading as an analysis-engine toggle in the one place nobody would look for it. The App field was wedged between the two analysis flags for the same reason.

Moved to the Collection Control section, beside Pause Collection — the other "should collection do this" affordance — which is also where Darling puts its equivalent, next to plan capture. Field re-homed out of the analysis block too, with a comment recording why, so a future copy-the-pattern edit doesn't drift it back.

erikdarlingdata:
Fair hit, and worse than a nit given the context: I fixed the XAML placement, wrote a commit message about not letting copy-the-pattern drift the grouping, and then landed the C# field under `/* System tray settings */` — ahead of the actual first tray setting. Same error, same PR, one commit apart.

Now under its own `/* Collection settings */` header with the tray header restored to `MinimizeToTray`. The general lesson I'm taking: when relocating something, read the destination's surroundings rather than anchoring on the nearest textual match — both instances came from pattern-matching an insertion point instead of looking at what was already there.

erikdarlingdata:
Round-3 build failed, and it was NOT this PR — but it blocked it, so the fix rides along in `e22684ff` (filed separately as #2177).

Two `AnomalyDetectorTests` thin-baseline tests fail **deterministically for CI runs between 03:39 and 04:00 UTC**. My push landed at 03:55. The seed helpers write ~21 minutes of samples starting at whatever time-of-day `_analysisStart` inherits from `DateTime.UtcNow`; when that start falls within 21 minutes of midnight the span crosses a date boundary, so each intended 'day' contributes **two** dates — 4 instead of 2 — which clears the Full tier's 3-distinct-day minimum, makes a deliberately-thin baseline trustworthy, and routes the detector down the z-path. `baseline_low_quality` then reads 0 where the test requires 1.

Fixed by flooring the seed start to the hour. Deliberately NOT midday-anchoring (the #1972 discipline used elsewhere in the suite) because the Full tier buckets by hour AND day-of-week — the seeds have to keep `_analysisStart`'s hour and weekday to land in the bucket the analysis window reads. Starting at :00 preserves both and makes a 21-minute span unable to leave the hour. Applied to all seven seed helpers, not just the two that failed, since they all share the shape.

Worth noting: a rerun after 04:00 UTC would have passed and hidden this permanently.

Issue #2166, opened 2026-08-10
[FEATURE] Database State alert: per-state notify mode - fire OFFLINE/RESTORING once instead of re-firing every cooldown


gotqn:
### Which component(s) does this affect?

- [ ] Lite
- [x] Darling
- [ ] Documentation

### Problem Statement

The Database State alert (#1986) re-fires every cooldown for as long as a database stays deviated from its expected state. For the integrity-failure states (SUSPECT / RECOVERY_PENDING / EMERGENCY) that repetition is correct - you want to keep being nagged until it's fixed. 

But for a clean, deliberate OFFLINE (and RESTORING) it's noise: a database intentionally taken offline - routine maintenance, a log-shipping secondary flickering RESTORING, or a deliberate "park it offline for a while" retention step - produces an identical WARNING every cooldown for the entire time it's parked (Occurrences in the dozens/hundreds for a single intended action). 

The tool can't know intent, so a deliberate offline is treated like an accidental one and repeats indefinitely. We want to be told once (so a human can judge whether it's a false positive) without being reminded every ~15 minutes for weeks.

### Proposed Solution

Add a per-state notify mode to the Database State alert - each state configurable as off / once (edge-triggered: fire on the transition into the state, then stay quiet until it changes or resolves) / re-fire (today's cooldown-repeat) - with sensible defaults:

- OFFLINE, RESTORING → once
- SUSPECT, RECOVERY_PENDING, EMERGENCY → re-fire
- recovery to expected → resolution, as today

Minimum viable version if a full matrix is too much: just make OFFLINE/RESTORING edge-only (fire once, no cooldown re-fire) while the integrity states keep re-firing. 

Expose it through the same Settings / get_alert_settings/update_alert_settings surface as the existing master toggle and expected-state editor. It composes cleanly with the existing per-database expected-state (a DB we deliberately park can be set expected = OFFLINE to go fully silent yet still alert if it's brought back online or goes SUSPECT).

### Use Case

We retire databases with a "soft delete": set the database OFFLINE, leave it ~a month, and drop it if nobody objects. During that month it sits OFFLINE against an expected state of ONLINE, so the alert fires on the transition and then re-fires every cooldown for ~30 days - one intended action generating hundreds of identical alerts. 

### Alternatives Considered

[...]

PR #2182, opened 2026-08-11
Chosen database states alert once, not forever — edge-trigger OFFLINE/RESTORING (#2166)

erikdarlingdata:
## What changes

The Database State alert re-fired for as long as a database stayed deviated. @gotqn's case: a soft-delete workflow parks a database OFFLINE for a month, which produced **hundreds of identical alerts for one intended action**.

Now OFFLINE / RESTORING / RECOVERING / STANDBY are **edge-triggered** — alert on the transition, stay quiet until the state changes. SUSPECT / RECOVERY_PENDING / EMERGENCY are untouched and keep re-firing on the cooldown.

The split is the whole idea: those first states are usually ones somebody **chose**, and repetition tells the operator nothing new. Nobody parks a database in SUSPECT, so there the repetition *is* the signal — there's a test pinning that an already-announced integrity state must keep firing, because the failure mode worth guarding is a real corruption quietly stopping its nagging.

## Why the memory is persisted, not in-memory

This is the part that decided the design. The case is a database parked for **weeks**. In-memory edge state would go quiet, then re-announce every parked database on the next service restart — worse than the cooldown-repeat it replaces, and it would look like a regression to exactly the person who asked for this.

So V60 adds `last_alerted_state` / `last_alerted_at` to `config.database_state_expected` — the table that already holds this alert's per-database config, already keyed `(server_id, database_name)`.

**Rejected alternative:** `config_edge_trigger_watermarks` already does restart-surviving edge memory, but it's keyed `(server_id, metric_name)` and `metric_name` feeds alert history and mute-rule matching. Per-database use would mean smuggling a compound key into a label column — fine today, a trap later.

## The composition property @gotqn identified

Pinned by test: a database announced OFFLINE that turns **SUSPECT** fires again, at Critical. Going quiet for a parked state must not mean going blind, and the new state doesn't inherit the old one's quiet treatment.

## Scope

**Darling only**, matching where the issue was filed. The engine change is **inert for Lite by construction**: with no persisted memory the adapter reports empty, `current != empty` is always true, and Lite behaves exactly as before. Lite's `SaveDatabaseStateAlertedAsync` is a documented no-op rather than a fake in-memory cache — a cache would go quiet and then re-fire on restart, which is the worst of both. Lite parity follows separately, same as #2167 → #2176.

## Behavior-change warning

This alters a default. Anyone relying on the repeat-nag to track deliberate offlines will see it go quiet after the first alert — called out in the changelog in those terms.

## Testing

Four engine pins: an already-announced chosen state stays silent with no cooldown involved; a first observation fires **and records the state** (recording is what makes the silence survive a restart); an already-announced SUSPECT still fires at Critical; and OFFLINE→SUSPECT fires again at Critical. Plus the full V60 ladder — rung identity, probe sentinel, gate arm mapping a fully-migrated store to 60 and a V59 store to 59, and the suite's ladder-tip pins moved. Alerting, Darling service, Viewer, Lite, and both test projects build clean.

Addresses #2166 (Darling half).


erikdarlingdata:
Both findings confirmed and fixed in 5e2a74b. Both were real, and the first one was worse than what this PR set out to fix — thank you.

**The upsert.** Walked the pending path and it lands exactly as described: no row → INSERT → `expected_state` takes the state being alerted on → the database stops deviating → the recovery block reads it as recovered and fires a false resolution on a still-corrupt database → silent forever. My doc comment's justification ("the seed logic would have written anyway") was simply wrong: `SeedDatabaseStateExpectedSql` explicitly refuses to baseline SUSPECT/RECOVERY_PENDING/EMERGENCY while they're critical, so this write was doing behind its back the thing it declines to do in front. Now UPDATE-only.

Worth stating why UPDATE-only loses nothing, since "skip the no-row case" sounds lossy: a row is absent only when a database was first observed in an integrity state, and `RepeatsAreNoise` is false for all three of those, so `alreadyAnnounced` never consults the memory for them — they keep alerting on the cooldown, which is the intended behavior. The parked-database case that *needs* the memory always has a row, because it was baselined while the database was still healthy.

**The permanent memory.** Also correct, and the OFFLINE → restore → OFFLINE-again sequence is the exact workflow the issue describes, so this would have bitten the reporter specifically. Added `ClearDatabaseStateAlertedAsync` to the interface, both stores and every stub, called on the falling edge. One deliberate choice: it clears even when `suppressed`, because suppression governs whether operators are *told* about a transition, never whether the engine keeps accurate state — leaving a stale memory behind under suppression would swallow the next real episode.

**On "none of the new tests catch this."** That was the most useful part of the review, because it named the structural reason rather than just the miss: the stub state store counted calls while the adapter supplied states, so nothing crossed the round trip the edge trigger actually depends on. `FakeStateStore` now models what the store would HOLD (save writes, clear removes), and `DatabaseState_RepeatEpisode_OfTheSameState_FiresAgainAfterRecovery` feeds that memory back in as `LastAlertedState` for the third evaluation — park, recover, re-park in the same state. It fails if either direction of the trip is missing, which the previous harness could not do. `DatabaseState_AlertedStamp_IsAnUpdate_NeverAnInsert` pins the SQL so the upsert cannot come back.

Verified: Alerting, Darling.Tests, Lite and Lite.Tests all build clean (the interface addition needed the Lite no-op and three stubs).

erikdarlingdata:
Confirmed and fixed in 5be13e7. Third real one, and the diagnosis of *why* was the valuable part.

You're right that this alert can't follow the convention documented at lines 87-89. For the other `_active*` flags, losing the set on restart risks only a missed or duplicate **resolution** notice, because the persisted thing that gates re-firing is a monotonic count or time — a stale value is naturally superseded by the next observation. Here the persisted thing is a sticky state *string* whose only path to being cleared was that ephemeral set, so the invariant didn't carry over and I shouldn't have assumed it did.

Fixed by taking your suggestion literally: the clear is now **derived from the store**. `ClearRecoveredDatabaseStateAlertsSql` runs beside the existing `SeedDatabaseStateExpectedSql` and `PruneDatabaseStateExpectedSql`, on the same connection, and clears the memory for any database the store now shows back at its expected state. It can't have the gap because it asks rather than remembers, and it heals a stale memory regardless of how it got stale — restart, crash, or somebody editing the table by hand.

I kept the engine-side clear as well, and want to be explicit that this is two layers rather than accidental duplication: the engine clear is the **immediate** path, so a recovery inside one process doesn't wait for the next cycle's sweep, and the store sweep is what **owns the invariant**. If they ever disagree, the store wins by construction.

Three details that took some thought:

- It reads the same **effective** state the deviation query does — `CASE WHEN is_in_standby THEN 'STANDBY' ELSE state_desc END` — so a log-shipping secondary expected at STANDBY actually clears. Comparing against raw `state_desc` would have left exactly those rows stuck, which is a nasty variant of the same bug.
- It clears the `(ignore)` sentinel too. An operator silencing a database shouldn't leave a memory behind that outlives the silence.
- **One** sample at expected is enough, where the deviation rule needs two. Clearing is the safe direction — it can only cause an extra alert, never a missed one — and a flap can't exploit it, because a flap doesn't survive the two-sample deviation test to alert in the first place.

On testing: you were right that the harness test can't reach this, and I don't think any harness test can — the restart gap is *defined* by there being no process state, so anything driving the engine proves only the path a running process takes. So it's a gated-live test (`DatabaseStateAlertMemoryLiveTests`) that seeds five databases — recovered, still-deviating, deviating-differently, standby-effective-state, and `(ignore)` — runs the real statement against a real store with nothing in memory, and asserts a second sweep changes **zero** rows, since this now runs on every evaluation of every server.

Worth noting these three statements had no test coverage at all before this, so that's a small net improvement beyond the fix.

erikdarlingdata:
Fixed in 91a3ecfa rather than accepted, because I think accepting it would contradict the property this PR claims as its reason for existing.

Your framing is exactly right: the shared cooldown key predates this PR, and this PR is what turns it from a delay into a silence. Before, every deviation re-fired every cooldown, so a transition suppressed by the previous state's clock re-announced on the next tick. Now that a chosen state goes quiet indefinitely there's nothing to retry it — the suppression lasts the full cooldown window — and the state it lands on is SUSPECT, which the PR explicitly designs to never go quiet. "Cooldown windows are typically much shorter than an outage" is true and still not a good enough answer for the integrity case.

So each state now carries its own clock, per your second suggestion. One consequence worth naming: recovery has to clear **every** state's clock for that database, not one, or a stamp left behind would rate-limit a later episode against a cooldown that started before the recovery — the same class of stale-state bug as the previous two findings, one layer down.

Two tests, because the fix can fail in both directions:

- `DatabaseState_TransitionToADifferentState_IsNotSuppressedByThePriorStatesCooldown` — both evaluations run back to back so no wall-clock time passes, which is precisely the condition that broke it. OFFLINE fires and stamps; the flip to SUSPECT must still fire, at Critical.
- `DatabaseState_SameState_StillRateLimitsItself_WithinOneCooldown` — proves the cooldown was **re-keyed and not disabled**. An integrity state repeats deliberately, so its own cooldown is the only thing between it and an alert per evaluation.

You were right that none of the existing tests could have caught this: they all start from a fresh harness where the key was never stamped, so `CooldownElapsed` returns true trivially. Same structural blind spot you identified on the state store — the tests never crossed the boundary where the state actually lives. That's four findings on this PR where the diagnosis of *why the tests missed it* was more useful than the bug report.

erikdarlingdata:
Fixed properly in the latest commit rather than left documented, because you're right that this one is on me in a way the earlier findings weren't: the old code did an **exact-key** `TryRemove`, so no collision was possible, and my per-state re-key is what introduced the prefix sweep. "The blast radius is bounded" is a fine thing to say about an inherited gap and a weak thing to say about one I'd just created.

Your second suggestion is the one I took, slightly generalised. There's no delimiter a `sysname` cannot contain, so rather than hunt for a safer separator I stopped encoding a compound key as a string: `_lastDatabaseStateAlert` is now keyed by a `(Server, Database, State)` tuple, and recovery matches two of the three parts structurally. No string parsing anywhere in the path, so the bug class is gone instead of narrowed.

That needed `CooldownElapsed` to take a key type parameter — one line, every existing string-keyed caller infers unchanged, and the other nine families are untouched. I preferred that to giving this family a private copy of the cooldown helper, which is the kind of duplication that drifts.

The test pins your exact scenario: `Foo` and `Foo|Bar` both alerting, `Foo` recovers, `Foo|Bar` must stay quiet. Worth noting how it's observable, since it wasn't obvious — it has to run through an **integrity** state, because SUSPECT is the one case not edge-suppressed, so its cooldown is the only thing keeping it silent on the second pass and a wrongly-cleared clock makes it fire. Under the string-prefix version that test fails.

Five findings on this PR, and every one was real. The pattern I'll take away is that all five lived at a boundary the tests didn't cross — process memory vs store, one state vs another, one database's key vs another's.

erikdarlingdata:
Right on both counts — fixed in the latest commit.

**The V60 justification was false, not just imprecise.** I'd copy-pasted the V59 wording and it claimed the viewer names the new columns and would 42703 on a V59 store. It doesn't: the expected-state editor touches only `expected_state` and `is_user_override`, so nothing in the viewer reads either V60 column. Your read of how it got there is exactly what happened.

The bump is still correct, so the comment now gives the actual reason rather than being deleted: the standing `RequiredStoreSchemaVersion == StorageVersion.SchemaVersion` invariant, plus the point that it's the **service** — the alert engine's edge memory and the deviation query — that would fail against a V59 store, which is precisely why the viewer must not report V59 as good. Your last sentence is why I wrote it out at length instead of trimming it: someone deciding later whether this gate can be relaxed would have been reasoning from a false premise, which is worse than no comment.

**The stacked summary** was mine too — `CooldownElapsed` already had a one-line summary and I added a second above it instead of extending it. Merged.

That pin has now caught me three times in this change set, always the same way: writing a doc block for a member without checking whether it already had one. So rather than keep relying on CI to find it, I added a scripted stacked-summary scan over every file I touch as a pre-push step — it's what confirmed the other five files are clean this time. Worth noting the pin is doing real work; all three would have shipped as silently wrong documentation, since XML docs take the last block and tooling looks fine.

erikdarlingdata:
Fixed — you're right that it's a real regression and not just a theoretical one, since the old string key made this an O(1) remove and the tuple key turned it into a scan of every server's entries, once per recovered database.

Hoisted to a **single** pass over the cooldown map against a `HashSet` of the recovered names, and skipped entirely when nothing recovered (which is the common case — recovery only fires when a database actually returns to its expected state). So it's O(tracked entries) per evaluation rather than O(recovered × tracked). Correctness is untouched: still per-state removal, since a single remove per database would leave its other states' stamps behind to rate-limit a later episode against a pre-recovery cooldown, and still matching on two tuple parts so `Foo|Bar` survives `Foo` recovering.

I did consider your secondary-index suggestion and decided against it, for what it's worth: `(Server, Database) -> HashSet<State>` would get it to O(states-for-this-db), but it's a second structure that has to stay in sync with the first, and it would need the same clearing discipline on every path that touches the primary map — which is precisely the class of "two places to remember" bug that produced three of the earlier findings on this PR. The factor I removed is the one that was actually new; the remaining single scan is the same order the surrounding recovery code already pays. If the entry count ever gets large enough to matter, the index is the right answer and this comment is the note explaining why it wasn't yet.

That should be all seven findings on this PR closed.

erikdarlingdata:
Correct, fixed. And this one stings a little, because it's the same collision class the tuple key had just eliminated for `'|'` — reintroduced one line later by a comparer choice.

Verified your reading against the code rather than taking it on faith: the comment at 1231-1235 does state per-database keys are ORDINAL to match the stores' case-sensitive expected-state joins, `current` is `StringComparer.Ordinal`, `active` is `StringComparer.Ordinal`, and my `recoveredSet` was the only outlier. On a case-sensitive collation with both `Foo` and `foo`, recovering `Foo` would have cleared `foo`'s per-state stamps and reset the only quiet mechanism an integrity state has. Now `Ordinal`, with a comment saying why so the next person hoisting this doesn't widen it again.

While I was there I checked the other two `OrdinalIgnoreCase` uses in the method, since a blanket "make it all Ordinal" would have been wrong: the excluded-databases filter is deliberately case-insensitive (user-facing list, consistent with the other alerts) and the already-announced check compares **state tokens**, not database names. Both stay.

Eight findings on this PR now, every one real. The through-line is worth naming: each lived at a boundary — process memory vs store, one state vs another, one database's key vs another's, and now one *casing* of a database name vs another. The store-derived clear, the tuple key and this comparer are all the same lesson about identity having to mean exactly one thing on both sides of a boundary.

erikdarlingdata:
Rebased onto dev after #2183 merged — the only real conflict was `CHANGELOG.md`, where both PRs added an entry under `### Changed`; kept both, #2164's first since it landed first. Every other file was disjoint.

Rebased rather than merged so the nine commits stay linear and each review finding remains its own reviewable step.

Re-verified after the rebase rather than assuming it was mechanical: V60 is still the ladder tip (`StorageVersion.SchemaVersion = 60`, both `V60Sql` references intact), dev's `QueryStorePlanXmlState.cs` is present so the rebase didn't drop #2183's work, all seven affected projects build with zero errors (Collectors, Alerting, Darling.Service, Darling.Viewer, Darling.Tests, Lite, Lite.Tests), and the stacked-summary guard is clean across every file this branch changes.

erikdarlingdata:
Real bug, fixed. And it's the same failure mode this PR exists to remove, reached from a direction I hadn't considered: an operator's *temporary* mute silently became irreversible.

You've also identified the one place where this memory and the cooldown next to it must **disagree**, which is worth stating in the code rather than leaving as a subtlety. The cooldown is rate limiting — it applies whether or not anyone was told, which is exactly why the line above it is stamped even when muted. This memory means "the operator has been told about this state," and under a mute they have not. So the stamp is now skipped when muted while the cooldown stamp stays.

I took your first option rather than gating `alreadyAnnounced` on delivery, because delivery success is a different question again — the existing comment notes the stamp is deliberately written *after* `FireAsync` so a failed fire retries next cycle rather than being silenced. Tying the memory to delivery outcome would entangle "nobody was told because they muted it" with "nobody was told because SMTP was down," and those want different handling. Muted is knowable up front; delivery failure isn't.

Cost of the fix is a repeat inside the mute, which is invisible by definition and is precisely the pre-#2166 cooldown behaviour for a muted alert.

The test drives the whole sequence instead of just asserting the stamp is absent, using the harness's injected clock: muted fire must be marked muted **and** leave the memory empty; then with the mute removed and the cooldown elapsed, the same still-deviating state must produce a real deliverable alert and stamp the memory. Under the old code that second half is silent, which is the regression.

You were right that no existing test touched mute interaction — that's now three findings on this PR whose root cause was a boundary the tests never crossed, so I'll take the hint about where to look first next time.

erikdarlingdata:
Both fixed — a doc comment running into the preceding closing brace, and a double blank before the private helper, in each of the two state stores. Real inconsistencies I introduced; every other member in both files uses a single blank-line separator.

Small process note since it nearly bit me: both files are CRLF, and my first pass anchored on bare LF and matched nothing. The assertion on anchor count caught that before anything was written rather than after, which is exactly why I anchor-assert scripted edits in this repo instead of trusting a replace to have done something. Verified afterwards that CRLF counts are unchanged (114 and 243) with zero bare LF introduced, and that no triple-newline runs remain in either file.

That's every review finding on this PR closed — nine substantive plus these two. Green across all five checks and CLEAN on merge state as of the last run; this one's ready pending whatever CI says about this push.

erikdarlingdata:
Considered and declining this one, with a specific reason rather than on cost grounds — and merging, since every substantive finding is closed and all five checks are green.

The enumerate-the-known-tokens approach would leak entries. `DatabaseStateTokens` defines ONLINE, OFFLINE, RESTORING, RECOVERING, RECOVERY_PENDING, SUSPECT, EMERGENCY and STANDBY — but the cooldown key is built from `db.StateDesc` **verbatim**, and `sys.databases.state_desc` also yields **COPYING** and **OFFLINE_SECONDARY** (Azure SQL geo-replication states) which that list does not contain. A database in one of those would create a cooldown key that removal-by-known-token could never find, leaving a stale stamp behind to rate-limit a later episode against a pre-recovery cooldown.

That is precisely the failure class three earlier findings on this PR shared: stale state on one side of a boundary silently suppressing a real alert. Trading a rare scan for a new instance of it isn't a good deal, and guarding it would mean keeping the token list exhaustive forever against a vocabulary Microsoft controls — a coupling I'd rather not create.

Worth noting the scan is also narrower than it reads: entries only exist for databases that have actually **fired**, so the dictionary holds deviating databases, not all servers × all databases. And it runs only on a recovery event, which is by definition rare — recoveries are the transition out of an alerting state, not a per-cycle occurrence.

If entry volume ever becomes real, the right fix is the secondary index from your earlier comment — `(Server, Database) -> HashSet<State>` — which is exhaustive by construction rather than by maintained list. This thread is the note explaining why neither is in yet.

Ten findings on this PR, all ten legitimate. Six were genuine bugs (the SUSPECT baseline, the permanent memory, the restart-gap clear, the cooldown-borrowed-across-states silence, the delimiter collision, the permanent mute), and the diagnosis of *why the tests missed them* was consistently more valuable than the report itself. Thank you.

PR #2183, opened 2026-08-11
Stop re-shipping plan XML the store already holds (#2164)

### LINKED ISSUE #2164: [PERF] QS collector drain-mode passes cost the full byte budget every cycle — make the budget a knob, skip already-shipped plans by hash

erikdarlingdata:
## What does this PR do?

Closes #2164.

Query Store collection was re-shipping execution-plan XML the store already had. In a three-hour fleet window it shipped **202,790 plan-XML rows, of which 197,113 (97.2%) were for plans the store had already held for over an hour.**

The #1556 dedupe lands each plan once per *pass* — but it re-lands it on every pass, forever. Three measurements narrowed it to that:

1. **Payload volume is not the cost.** Dropping the text budget 64 MB → 12 MB on a production server moved 5.3x less text and ~7x fewer rows, and the batch clock did not move (383s/352s against a 358–480s baseline band).
2. **Server-side aggregation is not the cost either.** The #2173 open/drain split says **open is 3–6% of a pass, drain is 94–97%.**
3. **Drain tracks LOB bytes PER ROW, not total bytes.** db-B at ~180 KB/row drained at 256 KB/s (706 ms/row); db-M at ~41 KB/row drained at 1,080 KB/s (38 ms/row).

Given (3), *not fetching* a plan is worth far more than fetching less of one — which is why a smaller budget bought nothing. So collection now carries a per-database watermark on `plan_id` (monotonic within one database's Query Store, per Erik) and narrows the plan-text CASE to plans above the highest one whose XML actually stored.

Absent, malformed or expired renders **no predicate at all**, so the conservative path is byte-identical to the pre-change query. That matters because absent is what a first run, a restarted host and a broken store all look like.

- [ ] Lite
- [x] Darling
- [ ] Lite Tests
- [x] Darling Tests
- [x] SQL collection scripts
- [x] Documentation

Lite shares the collector and is unaffected in behavior: it never sets `CapturePlanXml`, so the predicate never renders and it writes no watermark. Both Lite projects build.

## How was this tested?

**Against four real production Query Store catalogs, before merge.** I rendered the actual SQL out of `BuildPayloadBody` by reflection — not a hand-written approximation — and ran both shapes with the watermark at each database's MEDIAN `plan_id`, so both sides of the predicate were non-empty. 15-minute window, 45s command timeout, read-only apart from the payload's own `DROP TABLE IF EXISTS #pm_qs_slice`.

| server / database | max plan_id | watermark | rows | xml rows | at-or-below-wm WITH xml | elapsed |
|---|---|---|---|---|---|---|
| db-F-01 / db-F | 2,078,278 | none | 884 | 884 | 375 | 3689 ms |
| db-F-01 / db-F | | 1,039,139 | 884 | **509** | **0** | **2272 ms** |
| multi-34 / db-G | 636,897 | none | 213 | 213 | 77 | 479 ms |
| multi-34 / db-G | | 318,448 | 213 | **136** | **0** | **236 ms** |
| multi-33 / db-H | 2,589,659 | none | 651 | 474 | 156 | 910 ms |
| multi-33 / db-H | | 1,294,829 | 651 | **318** | **0** | **657 ms** |
| multi-02 / db-J | 3,993,757 | none | 244 | 244 | 69 | 487 ms |
| multi-02 / db-J | | 1,996,878 | 243 | **174** | **0** | **258 ms** |

- The SQL **parses and runs** on real catalogs; the predicate splices inside the CASE without breaking the ROW_NUMBER gate.
- **Row counts are identical** in every case. Only the XML column changes. (244 → 243 on db-J is a live-window race between the two runs, not a dropped row.)
- **`at_or_below_wm_with_xml` = 0 in every watermarked run.** The predicate suppresses exactly the plans at or below the watermark and nothing above it. That is the correctness assertion.
- Elapsed fell **28–51%** at only the median plan_id.

XML megabytes fall less than XML row count (88.78 → 69.31 MB, 22%, versus 42% fewer rows) because the older plans below the median carry smaller XML than the newest ones — consistent with the per-row LOB finding.

SQL Server versions: the four instances span the production use2 fleet (2019 and 2022 box/RDS). Unit coverage is 25 tests in `QueryStorePlanWatermarkTests`, all driven through the collector's **public** surface (`BuildPerItemQuery`, `BuildBackfillPerItemQuery`, `ReadItemAsync`) over a real `DbDataReader` generated from the collector's own `PayloadColumns` — so no production visibility was widened for tests, and a column added to the collector cannot silently shift the ordinals the read loop depends on.

## Three bugs the testing found in my own first cut

Worth listing, because each was silent and two were shipped-code-shaped:

1. **The watermark leaked onto the BACKFILL path.** `BuildBackfillPerItemQuery` passes no database name, so it fell through to `CurrentDatabaseName` and applied. Backfill digs into intervals *older* than anything collected, whose rows reference plans compiled long ago and therefore numbered *below* the live watermark — so it would have suppressed essentially every plan the backfill exists to fetch. Silently: runtime stats still ship, so a filled range would look complete while carrying no plan XML at all.
2. **The watermark advanced on budget-cut passes.** Rows ship ordered by `last_execution_time`, **not** `plan_id`, so a cut drops an arbitrary set of plan_ids off the tail of the window — including ids *below* the highest one that stored. Advancing past them suppressed their XML on every later pass despite it never having shipped once. It now does not advance at all on a cut pass; the cut is already resumable on the time watermark, so that costs one repeated fetch and nothing else.
3. **Reset detection was unsound in the common case.** The tempting test — "the highest plan_id seen this pass is below the standing watermark" — is **TRUE in any ordinary window where no new plan compiled**, which on a steady workload is most windows. It would have dropped the watermark constantly and defeated the entire optimization. Caught by the quiet-window test failing. Removed: exact detection needs the server's live `MAX(plan_id)`, which this payload does not carry.

## The refresh horizon, and why it is one day

The watermark expires, and its stamp dates the **last full fetch** rather than the last advance. Re-stamping on advance would push the horizon out every time a new plan compiled — so on any database that keeps compiling, which is the busy ones where a stale plan matters most, the horizon would never fire and the watermark would effectively be permanent.

One day rather than seven, because the redundancy being removed is per-pass: a 15-minute cadence re-ships the same plan ~96 times a day, so a daily full fetch already eliminates ~99% of it and a weekly one adds almost nothing. What a longer horizon *does* buy is 7x the exposure on all three things the horizon guards:

- an in-place XML rewrite (memory-grant feedback and friends editing a plan that keeps its id — 0 of 38,420 plan_ids changed their plan hash in a day of fleet data, but the horizon means that does not have to be load-bearing),
- **reset recovery** — clearing Query Store restarts plan_id at 1, so every new plan sorts below a stale watermark; with no detection this horizon is the whole recovery mechanism, and a week-long plan-XML blackout on a database is not an acceptable price for a saving that rounds to nothing,
- the dormant-plan gap below.

## Known gap, documented at the predicate

`plan_id` is monotonic in **compile** order, which is not the same as "we have stored it". A plan compiled before monitoring began, dormant through every collected window, then executed again, arrives below the watermark and has its XML suppressed until the horizon expires. Bounded to one day by the above.

The exact fix is a **store-derived** watermark — the host asking its own plan dimension for the lowest plan_id missing XML, which cannot be wrong by construction — but that needs host plumbing on both products, so it is deliberately not in this PR.

## Checklist

- [x] My code builds with zero warnings (Collectors, Darling.Service, Darling.Tests, Lite, Lite.Tests — all zero errors, zero warnings)
- [x] I have tested my changes against at least one SQL Server version (four production instances)

The watermark is inlined into the SQL as a parsed `long`, never operator input, because the body nests inside `sp_executesql` on three paths and threading another parameter through all of them buys nothing.

_(Database names are relabelled `db-*` — this repo is public and the originals are client-identifying. Server names and every measurement are unchanged.)_


erikdarlingdata:
Good catch, and it's now moot in the best way — the redesign in dd3a1008 removed `StateKeys` from this collector entirely, so Lite issues **no** extra `collector_state` read at all.

You were looking at the same seam two other tests were, from the Lite side. `CollectorStateContractTests.TheOnlyCollectorDeclaringStateIsDefaultTraceEvents` and `QueryStoreBackfillTests.StateIdentity_IsTheWorkersOwn_NotTheDefinitions` both failed on that `StateKeys` override, and reading them showed the backfill had already solved this exact problem: per-database Query Store state is owned by the **host** under its own `collector_state` name, never declared by the definition. `QueryStoreBackfillState`'s doc calls that "the seam that lets both stay true."

Which also means my version was heading somewhere worse than one wasted DuckDB lookup. A declared *prefix* isn't a real key — the host would have persisted watermarks under the **definition's** name while the load still never matched a dynamic per-database key, so the watermark would have resolved absent forever and collection would have quietly kept paying full price with every test green. Your "the result is always discarded there" observation was one step from that: the value was going to be discarded on Darling too.

So the watermark now lives in `QueryStorePlanXmlState` (beside `QueryStoreBackfillState`, shared across SKUs so a row written today decodes the same after an upgrade), the definition declares nothing, and the Darling runner loads under `query_store_plan_xml` only when plan capture is on and routes query_store's pending state to that owner. Lite needs no equivalent for the reason you identified: it never captures plan XML.

The rendered SQL is byte-identical across that commit — I re-verified the predicate still lands inside the CASE and still never renders for backfill or with plan capture off. The PR description's "Lite is unaffected in behavior" is now literally true rather than approximately true.

erikdarlingdata:
Correct, and it's a fair catch against the sibling pattern — `QueryStoreBackfillState`'s `hole:` keys do get an explicit `DeleteCollectorStateKeyAsync` and these don't. Filed as #2188 rather than folded in here, since it's orthogonal to whether the watermark works and this PR is otherwise ready.

One thing I noted on the issue that makes the naive version of the fix worse than nothing, in case it saves someone time: the enumerated database list is not the same as "every database that exists." A database can be absent from a cycle's enumeration because it's offline, excluded by configuration, or failed its probe — none of which mean it was dropped. Pruning on a single absence would delete a live watermark and force a full plan-XML refetch for that database, which is the safe direction but self-defeating on precisely the servers that have excluded or intermittently-offline databases. So it wants either absence across consecutive enumerations or a check against `sys.databases` directly.

Bounded in the meantime: one row per database name ever seen, ~100 bytes each, never read once orphaned.

Issue #2187, opened 2026-08-11
[BUG] install-darling.ps1 creates an unstartable service when installed under a user profile


erikdarlingdata:
### Component

Darling

### Describe the Bug

Split out of #2185 (reported by @jovon44). `install-darling.ps1` will happily create a working-looking service that can never start, when the install directory sits under a user profile.

The mechanism:

- The installer deliberately creates the service under the **`NT SERVICE\<serviceName>` virtual account** rather than LocalSystem, and that choice is correct and documented — the bundled PostgreSQL refuses to run with administrative privileges.
- That account is not the installing user, not SYSTEM, and not Administrators.
- A directory under `C:\Users\<someone>\Desktop` (or anywhere in a profile) grants read access to approximately those three and nothing else. The installer's ACL work is scoped to `darling.json` and its `.bak-*` copies — it never touches the install tree, so `pg-runtime` keeps whatever the profile location gave it.
- The service therefore cannot read `initdb.exe`'s dependencies, and the bootstrap dies with `0xC0000135` STATUS_DLL_NOT_FOUND before writing any output.

In #2185 the content root was `C:\Users\username\Desktop\PerformanceMonitorDarling-3.2.0\`, which is a completely reasonable thing for someone to do with a zip they just downloaded — extract it where it landed and run the installer. Nothing tells them it cannot work.

### Expected Behavior

Refuse, or warn unmissably, when the install root is under a user profile path (`$env:USERPROFILE`, `C:\Users\...`), naming the reason: the virtual service account cannot read a user profile, so the service will install and then fail to start. Suggest a machine-scoped path.

Two candidate behaviors, and I lean toward the first:

1. **Hard refuse** with the explanation and a suggested path. The install cannot work, so proceeding is never right.
2. Warn and additionally grant the service account read+execute on the install tree. This makes it work, but it also means the product starts ACLing directories inside somebody's profile, which I would rather not do silently.

### Additional Context

Worth checking the same question for `C:\ProgramData\PerformanceMonitorDarling` (machine-scoped, fine) and for network/UNC install paths, which have the same "service account is a different principal" problem for a different reason.

Pairs with the exit-code-decoding half. Between them, the reported failure becomes either impossible or self-explaining.

erikdarlingdata:
[...]

Issue #1738, opened 2026-07-27
Runtime advance has no direction check: a package with an OLDER PostgreSQL major replaces a working newer runtime and the store cannot start (hit on DARLING01)


erikdarlingdata:
`TryAdvanceRuntimeAsync` treats "the package's PostgreSQL major differs from the extracted runtime's" as a reason to update, without checking which direction the difference goes. A package carrying an OLDER major replaces a working newer runtime, and the store then cannot start: the data directory is on the newer major and no older postmaster can open it.

This is not hypothetical. I did it to DARLING01 tonight, with the merged #1718 code, and it took the store down for about seven minutes.

## What happened

DARLING01 runs a PostgreSQL 18.4 store (`PG_VERSION` 18). A deploy placed a `pg-runtime.zip` beside the service that was a PostgreSQL **17** bundle. Verbatim from `darling-service_20260726.log`:

```
17:17:35 [WARN] The extracted runtime is PostgreSQL 18 and the shipped package is PostgreSQL 17 - this host has never had its runtime updated, and the update proceeds.
17:17:36 [WARN] The package ships a different Postgres runtime than the one extracted on this host - rescuing the current runtime to C:\PerformanceMonitorDarling\pg-runtime-prev\pgsql and extracting the new one.
17:17:39 [WARN] Could not determine the store's PostgreSQL major (data directory: 18, bundled runtime: unreadable) - skipping the runtime version check. The store starts normally.
17:17:40 [CRIT] Managed Postgres bootstrap failed: pg_ctl status reported exit code -1073741515 for C:\ProgramData\PerformanceMonitorDarling\pg - the data directory is not usable.
```

The store stayed down until the runtime was restored by hand from `pg-runtime-prev`. No data was lost, because nothing reached `pg_upgrade` - the failure is at bootstrap, before any data directory is touched.

## Two distinct gaps

**1. No direction check.** The no-stamp branch compares majors and, on any difference, logs "this host has never had its runtime updated, and the update proceeds". A LOWER package major is never a valid update. It cannot be a rollout, because the data directory has already moved forward; `pg_upgrade` only goes up. The correct behavior is to refuse, loudly, and leave the runtime alone: the store keeps running on the runtime it has, which is the one that matches its data.

[...]

Issue #2193, opened 2026-08-11
[BUG] Pre-existing analyzer warnings on dev make a zero-warning build gate unreadable (CA1068, CA2022, CS0628)


erikdarlingdata:
### Component

Build / Repo hygiene

### Describe the Bug

Noticed while gating #2190 (PR #2192), which required a zero-new-warnings build. The change itself introduced none, but the baseline on `dev` is not clean, and these three classes are what a "0 warnings" gate has to be read around every time:

**1. `CA1068` in `PerformanceMonitor.Common/AbandonableStep.cs(75,46)`** - `RunAsync(Func<Task>, TimeSpan, CancellationToken, Action<Exception>?)` should take the `CancellationToken` last. This one is a real signature decision rather than a mechanical fix: `AbandonableStep` is the shared primitive #2148 extracted for Lite's tick and Darling's fleet loop, so reordering touches call sites in both apps.

**2. `CA2022` in `Darling/Darling.Tests/NpgsqlRootCertificateValidationTests.cs(135,25)`** - inexact read with `SslStream.ReadAsync(Memory<byte>, CancellationToken)`. A partial read here would make the #2117 TLS chain test assert against a short buffer, so this is the one with actual test-correctness weight. `ReadExactlyAsync` is the usual answer.

**3. `CS0628` x5 in `Lite/Services/LocalDataService`** - `GetTimeRange`, `GetTimeRangeServerLocal`, `TimeQuery`, `ToDouble` and `ToInt64` are declared `protected` on a `sealed` type, so the accessibility is dead. Mechanical (`protected` to `private`), but it lands in a file parallel sessions touch often, which is why it did not ride along with an unrelated PR.

### Expected Behavior

`dotnet build` on the affected projects reports 0 warnings, so a warning appearing in a PR build is unambiguously that PR's.

### Actual Behavior

On `dev` at 7a30c9bd:

- `dotnet build Darling/Darling.Tests/Darling.Tests.csproj -c Debug` reports `2 Warning(s)` (CA1068 from the referenced `PerformanceMonitor.Common`, CA2022 from the test project itself).
- `dotnet build Lite/PerformanceMonitorLite.csproj -c Debug` reports `10 Warning(s)` (CA1068 again, plus the five CS0628 counted across both the real and the WPF temp csproj).

### Additional Context

[...]

PR #2194, opened 2026-08-11
Decode the loader status instead of printing a raw Win32 number (#2186)

### LINKED ISSUE #2186: [BUG] Managed Postgres bootstrap reports a raw Win32 status and an empty Output on loader failures
### Component

Darling

### Describe the Bug

Split out of #2185 (reported by @jovon44), which is the second time a managed-Postgres bootstrap failure has surfaced to a user as a raw Win32 status code with no diagnosis.

The bootstrap reports:

```
Managed Postgres bootstrap failed: initdb failed (exit code -1073741515) for C:\ProgramData\PerformanceMonitorDarling\pg. Output:
```

Two things make this worse than it needs to be:

1. `-1073741515` is `0xC0000135` = `STATUS_DLL_NOT_FOUND`. The service knows this — `DarlingManagedPostgres.cs` already names that exact code in the #1738 comment on `EnsureDataDirectoryMajorAsync`, where it deliberately converts "the binaries could not run" into a message naming the runtime to restore. The initdb path never got the same treatment.
2. `Output:` is EMPTY and always will be for this class of failure, because the process died in the loader before writing anything. So the one field an operator would look at is guaranteed blank precisely when the failure is a load failure — which reads as "no information available" rather than "this is a loader failure."

Net effect: a user whose install cannot possibly work gets a number, an empty field, and a follow-on message about a missing credential file that sends them looking at `darling.json`, which is not the problem.

### Expected Behavior

When `initdb` (or any bundled Postgres binary) exits with a loader status, say so and name the two real causes: the MSVC runtime DLLs missing from `pg-runtime\pgsql\bin` (which the packaging step bundles, so this means a damaged/partial extract), or the service account being unable to read the install tree. Worth decoding at least `0xC0000135` (DLL not found) and `0xC0000139` (entry point not found), and stating explicitly that an empty `Output:` is expected for a loader failure rather than leaving it looking like missing data.

### Actual Behavior

Raw signed exit code plus an empty output field, and the operator's attention gets pointed at the downstream credential-file message instead.

### Additional Context

The decoding already exists in the same file for the sibling case, so this is about applying it at the bootstrap path rather than inventing anything. Pairs with the installer-side half.

erikdarlingdata:
Closes #2186.

## The report

A managed-Postgres bootstrap failed with this, and nothing else:

```
Managed Postgres bootstrap failed: initdb failed (exit code -1073741515) for C:\ProgramData\PerformanceMonitorDarling\pg. Output:
```

`-1073741515` is `0xC0000135` = `STATUS_DLL_NOT_FOUND`. Windows killed `initdb.exe` in the loader, before a line of its own code ran, which is also why `Output:` is empty and always will be for this class of failure: the one field an operator reads is guaranteed blank exactly when the failure is a load failure. It reads as "no information available" rather than "this is a loader failure", and attention went to the follow-on missing-credential message and `darling.json`, neither of which was the fault.

## Verifying the issue's claims first

Both held. `InitializeClusterAsync` interpolated the raw signed code (`DarlingManagedPostgres.cs:891` on dev), and the #1738 sibling at line 1103 names `STATUS_DLL_NOT_FOUND` **in a comment only** - there was no decoder anywhere in the file to reuse, so this adds one rather than moving one.

One claim was worth testing rather than assuming: that the empty `Output:` really is guaranteed. A test drives a genuine loader failure through the real `RunToolAsync` and pins it - exit code `-1073741515`, captured output length `0`.

## Before / after, verbatim

Produced by a real Windows loader failure driven through the real `EnsureRunningAsync`, not a hand-written string: a scratch `pg-runtime` whose `initdb.exe` is a binary whose app-local DLL cannot be resolved (a doctored `PATH`), which Windows kills with `0xC0000135` and zero bytes on both streams. Nothing touched DARLING01 or the repo's `pg-runtime.zip`.

**Before (dev):**

```
initdb failed (exit code -1073741515) for C:\...\repro\pg. Output:

```

**After (this branch):**

```
initdb failed (exit code -1073741515 = 0xC0000135 STATUS_DLL_NOT_FOUND) for C:\...\repro\pg.
Windows killed initdb.exe in the LOADER: it never ran a line of its own code. This is a Windows failure, not a PostgreSQL one, and it is why the captured output is empty - for a loader failure an empty Output is EXPECTED, not missing information.
Two causes account for nearly all of these:
  (1) The Microsoft Visual C++ runtime is missing from C:\...\repro\pg-runtime\pgsql\bin. Packaging bundles vcruntime140.dll, vcruntime140_1.dll and msvcp140.dll there so the box needs no prerequisite - if any of the three is absent the install tree is a partial or damaged extract, so redeploy the package, or install the Microsoft Visual C++ 2015-2022 x64 redistributable.
  (2) The service account cannot read the install tree. The service runs as the virtual account NT SERVICE\PerformanceMonitor Darling, which is neither you nor Administrators, so an install under a user profile (Desktop, Downloads, anywhere below C:\Users) is unreadable to it - reinstall to a machine-scoped path such as C:\PerformanceMonitorDarling.
Two checks that tell them apart:
  (a) Run "C:\...\repro\pg-runtime\pgsql\bin\initdb.exe" --version from an elevated prompt. Failing there too means (1). Succeeding points at (2) - and note that these tools re-execute themselves under a restricted token that drops the Administrators group, so a tree readable only VIA Administrators still fails the real run even when it runs by hand.
  (b) Event Viewer > Windows Logs > Application, at the time of the failure: an Application Error or SideBySide entry usually names the exact module that could not be loaded.
Output:
(none - Windows ended the process before it could write anything, which is expected for this failure rather than missing information)
```

The leading clause is deliberately unchanged: existing field reports and the issue tracker are searchable by `initdb failed (exit code`, so the fix must not rename what operators paste into search.

The restricted-token caveat in check (a) is in there because the field thread's most confusing evidence was `initdb --version` succeeding by hand while the service run failed. PostgreSQL's frontend utilities re-execute themselves via `get_restricted_token()` / `CreateRestrictedProcess` with the Administrators and Power Users groups removed, so a tree readable only *via* Administrators fails the real run while passing the by-hand check ([paquier.xyz](https://paquier.xyz/postgresql-2/postgres-utilities-restricted-token/), [PG source](https://doxygen.postgresql.org/restricted__token_8c.html)). Verified before shipping the claim, and it is attached only to tools that actually do this (initdb, pg_ctl, pg_upgrade) - never to `vacuumdb`.

## What changed

`DarlingToolExitCode` decodes any exit code at or above `0xC0000000` - the NTSTATUS error range, which no real program exits with deliberately - so an *unlisted* status is not left as opaque as `-1073741515` was. It separates loader statuses (`0xC0000135`, `0xC0000139`, `0xC000007B`, `0xC0000142`, `0xC0000022`) from crashes (`0xC0000005`, `0xC0000374`, `0xC0000409`); a crash must not send anyone hunting for missing DLLs, which would be the same wrong-direction error pointed somewhere new.

Wired into every bundled-binary failure, per the issue's "or any bundled Postgres binary": initdb (bootstrap and upgrade), pg_ctl start/status/stop/reload, pg_upgrade and its `--check`, and the post-upgrade analyze.

Two messages also stopped blaming the wrong thing, which the decode made visible:

- `pg_ctl status` said "the data directory is not usable" for *every* unexpected code. That is pg_ctl's own exit 4 talking; on a Windows status it pointed an operator at deleting a healthy store to fix a missing DLL. The verdict is now conditional.
- `pg_upgrade --check` reported "the clusters are not compatible" on any non-zero exit. Only pg_upgrade can reach that verdict, so on a Windows status it was invented. `NOTHING has been changed` still holds either way.

And the #1738 refusal asserted "its binaries did not run" while discarding the one piece of evidence it held - the probe's own exit code died in a local. `ReadRuntimeMajorAsync` now carries it out, and the refusal quotes and decodes it.

## Tests

Watched red first against an inert seam (the new API returning today's behavior): **16 failed, 9 passed**, then green. The 9 that passed under the inert seam are the ones asserting today's *correct* behavior is preserved - ordinary exit codes stay bare, real output passes through untouched - so the red was the decode, not the harness.

- `DarlingToolExitCodeTests` - the decode, both diagnosis families, the empty-output note, that ordinary codes (initdb's 1, pg_ctl status's 3) are left completely alone, and the live `RunToolAsync` pin above.
- `DarlingManagedPostgresTests` - the shipped messages: the field report rebuilt from its own numbers, the conditional data-directory verdict, and that the start failure diagnoses *before* it points at a server log that a loader failure guarantees is absent. Plus a wiring pin at the source, because three correct builders that no throw site calls is precisely what #1738 already was, and behavioral coverage cannot reach it - reproducing it needs a bundled Postgres that dies in the Windows loader, which is not something a CI runner can be asked to arrange.

Full `Darling.Tests`: **4256 passed, 0 failed**, 236 skipped (all gated). Service project builds **0 Warning(s)**. The two warnings in a full rebuild (`AbandonableStep.cs` CA1068, `NpgsqlRootCertificateValidationTests.cs` CA2022) are in files this branch does not touch.

## Not in scope

The installer half is #2187.


erikdarlingdata:
Heads up from the other work stream: this PR's `Darling Linux build` failure is **not yours**. It's a repo-wide break I've fixed in #2199.

The Dockerfile built from the floating `mcr.microsoft.com/dotnet/sdk:10.0` tag, that tag advanced to an image shipping SDK **10.0.400**, and `global.json` requests **10.0.302** with `rollForward: latestPatch` — which only rolls inside the 3xx band. So the in-container publish dies with `A compatible .NET SDK was not found`, surfacing as a bare `exit code: 155` with no compiler diagnostics, which reads like a code failure and isn't one.

Your PR is the control that proved it: an unrelated change failing the identical step the same way. dev's own Linux build passed at 18:22 and would fail on its next run.

#2199 pins the build stage to `sdk:10.0.302`. Once that merges, rerun this job and it should go green with no change on your side.

erikdarlingdata:
#2199 is merged, so the SDK pin is on dev now. Your `Darling Linux build` was already re-running when I checked; if it's still red, rebase onto dev to pick up the pinned `sdk:10.0.302` and it'll clear — a plain rerun only helps if the job re-resolves the base, which it doesn't for the container step.

Confirmed the fix works before merging it rather than after: `Darling Linux build` went green on #2199 while it was still red on this PR, which isolates the cause to the floating tag.

erikdarlingdata:
## Status: green on everything this PR can affect, blocked by a repo-wide CI break it did not cause

Final sha `c1e75ebf` (merge of `origin/dev`, resolving a whole-file `CHANGELOG.md` conflict - see below).

`statusCheckRollup`:

| check | result |
|---|---|
| build (Windows, full test suites) | **success** |
| Darling PostgreSQL tests | **success** |
| Check pull request target branch | **success** |
| Claude Auto Review | **success** |
| Darling Linux build | **failure** - #2200, not this branch |

`mergeable: MERGEABLE` (`mergeStateStatus: BLOCKED`, i.e. awaiting the failing required check).

### The Linux failure is #2200, filed separately

It fails before compiling anything:

```
Requested SDK version: 10.0.302
Install the [10.0.302] .NET SDK or update [/src/global.json] to match an installed SDK.
10.0.400 [/usr/share/dotnet/sdk]
```

`Darling/Dockerfile:16` uses the floating tag `mcr.microsoft.com/dotnet/sdk:10.0`, which rolled to SDK 10.0.400; `global.json` pins `10.0.302` with `rollForward: latestPatch`, so the 4xx band is correctly refused. Four things say it is not this branch:

- sha `9a60f423` on this same PR **passed** this exact job; `c1e75ebf` adds only a `CHANGELOG.md` line plus a dev merge, and neither `global.json` nor `Darling/Dockerfile` differs between them.
- dev's own Build passed at 18:22 UTC and the next run failed at 18:38, with no repo change in between.
- Re-running the failed job reproduces it identically, so it is not a flake.
- In the same run, the Windows `build` job and `Darling PostgreSQL tests` both pass - only the container publish fails.

#2200 has the diagnosis and the suggested one-line fix (pin the Dockerfile tag to the band `global.json` names, which is what #1758 intended and the one place still floating). **This PR is unmergeable until that lands, through no change of its own.**

### On the CHANGELOG conflict

It came back as a single whole-file conflict rather than the usual both-added-an-entry kind: dev's copy is CRLF, and my original edit had silently normalized the file to LF, so both sides looked entirely rewritten. Resolved by taking dev's file wholesale and re-inserting the entry and link-ref with CRLF preserved. `git diff origin/dev HEAD --stat -- CHANGELOG.md` is now **2 insertions**, and dev's new #2190 entry sits intact right below the new one.

### Also filed

#2197 - the follow-on missing-credential CLI message still reads as a first run even after a bootstrap has already failed. That is the half of #2185's misdirection that lives in `DarlingCliCommands.cs`, deliberately left out of this PR: different file, different surface, and it was under concurrent edit.

### Verified locally on the merged tree

Full `Darling.Tests`: **4265 passed, 0 failed**, 236 skipped (all gated). Service project builds **0 Warning(s)**; the two warnings in a full rebuild are in files this branch does not touch and are tracked in #2193.

erikdarlingdata:
## Fully green

Merged the current `dev` (which carries #2199's container SDK pin, the fix for the repo-wide break described in my previous comment). Final sha `8cda1626`, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

`statusCheckRollup`:

| check | result |
|---|---|
| build | **SUCCESS** |
| Darling PostgreSQL tests | **SUCCESS** |
| Darling Linux build | **SUCCESS** |
| Claude Auto Review | **SUCCESS** |
| check-branches | **SUCCESS** |

#2200 is closed as a duplicate of #2199.

### CHANGELOG byte check

Flagged by a sibling lane: `CHANGELOG.md` is CRLF **and carries 6 bare CR bytes**, so an editor pass silently normalizes them and turns a 2-line entry into a ~5,350-line diff. Verified this branch against `origin/dev` on both measures, before and after the merge:

- `git diff origin/dev HEAD --stat -- CHANGELOG.md` → **2 insertions**, unchanged.
- Byte census: this branch has **6 bare CR, 0 bare LF**; `origin/dev` has **6 bare CR, 0 bare LF**. Identical. The file is 2,118 bytes larger, which is exactly the entry plus the link-ref.

The earlier whole-file conflict was this hazard, caught and resolved by taking dev's copy and re-inserting with a script that splits and joins on CRLF (which preserves the bare CRs, since they sit inside the split line strings rather than at the boundaries) rather than saving the file from an editor. `file CHANGELOG.md` alone would not have caught it - it reports "CRLF" while saying nothing about whether the 6 bare CRs survived.

Not merging, per instructions.

erikdarlingdata:
## Green again on top of #2198

Merged `origin/dev` at `44fd7288` (which carries #2198). Final sha `9558744f`, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

`statusCheckRollup`:

| check | result |
|---|---|
| build | **SUCCESS** |
| Darling PostgreSQL tests | **SUCCESS** |
| Darling Linux build | **SUCCESS** |
| Claude Auto Review | **SUCCESS** |
| check-branches | **SUCCESS** |

Full `Darling.Tests` locally on the merged tree: **4271 passed, 0 failed**, 236 skipped (all gated) - up 6 from the previous round, which is #2198's new `DarlingInstallLocationTests`.

### The CHANGELOG conflict was one line this time, and that is the point

Last round it came back as a ~5,350-line whole-file conflict. This round it was a **single 5-line hunk** at the link-ref block - `[#2186]` on my side against `[#2187]` on dev's - because the byte-precise resolution held and there was no longer a competing normalization for git to see. Resolved keep-both-sides, ours then theirs, so both link-refs survive alongside every one already there.

Verified on both measures, as before:

- `git diff origin/dev HEAD --stat -- CHANGELOG.md` → **2 insertions**.
- Byte census: this branch **6 bare CR / 0 bare LF**; `origin/dev` **6 bare CR / 0 bare LF**. Delta is **2,118 bytes**, identical to the previous round, which is exactly my entry plus link-ref and nothing else.
- Zero conflict markers remain, and both prose entries are present - mine and #2198's "The installer refuses an install directory the service could never read".

Resolution was scripted, not editor-saved: split and join on CRLF, which leaves the 6 bare CRs untouched inside the line strings.

Not merging, per instructions.

erikdarlingdata:
## Green on top of #2202

Merged `origin/dev` at `fbd9a930` (carrying #2202). Final sha `20669f56`, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

`statusCheckRollup`:

| check | result |
|---|---|
| build | **SUCCESS** |
| Darling PostgreSQL tests | **SUCCESS** |
| Darling Linux build | **SUCCESS** |
| Claude Auto Review | **SUCCESS** |
| check-branches | **SUCCESS** |

Full `Darling.Tests` locally on the merged tree: **4273 passed, 0 failed**, 238 skipped. Skips are up 2 from the previous round, and that is expected rather than coverage lost: `a1916a9d` converted `RunTool_SurfacesAWindowsStatusAsTheSignedFieldValueWithNoOutput` from an early `return` to `Assert.SkipUnless`, so on non-Windows it now reports as skipped instead of passing vacuously - and this run counts it once per target.

### Carrying `a1916a9d`

This merge picks up the review commit pushed onto the branch, and both halves are improvements over what I wrote:

- **`{ExitCode}` stays numeric.** I had replaced the structured-logging value with the decoded string, which costs structured sinks a numeric field to filter and aggregate on. The decoded meaning now rides its own `{ExitCodeMeaning}` field. I checked this landed on **all three** sites I had degraded - the pg_ctl stop warning, the pg_ctl reload critical, and the post-upgrade analyze warning - so there is no straggler left behind.
- **`Assert.SkipUnless` over a bare `return`.** A guard that returns early is indistinguishable from a passing test in the report; skipped says what actually happened.

### CHANGELOG

One hunk again, this round on the entry block (`[#2186]` against `[#2189]`) rather than the link-refs. Resolved keep-both-sides, ours then theirs.

- `git diff origin/dev HEAD --stat -- CHANGELOG.md` → **2 insertions**.
- Byte census: this branch **6 bare CR / 0 bare LF**; `origin/dev` **6 bare CR / 0 bare LF**. Delta **2,118 bytes** - the same figure for the third round running, which is exactly my entry plus link-ref.
- Zero markers remain; all five link-refs (`#2186`, `#2187`, `#2189`, `#2190`, `#2171`) present exactly once each.

One note in case it looks like an artifact: there is a blank line between #2189's entry and the rest of the list. That is **dev's own formatting**, not something the resolve introduced - `origin/dev` has it at the same place. I left it alone rather than tidying it, since reformatting dev's line would have broken the 2-insertion property that makes this verifiable.

Not merging, per instructions.

Issue #2191, opened 2026-08-11
[BUG] planwm/backfill state pruning does not reach Azure SQL DB


erikdarlingdata:
Split out of #2188 while implementing it, so the residual is recorded rather than implied by a code comment.

#2188 prunes the per-database `collector_state` rows query_store accumulates (`planwm:` under `query_store_plan_xml`, `done:` / `hole:` under `query_store_backfill`) by anti-joining against the newest `database_states` snapshot for the server. That snapshot is an unfiltered `SELECT ... FROM sys.databases`, which is the only list in the store that answers "does this database still exist" without confusing a dropped database for an offline, excluded, or unprobeable one.

`DatabaseStateCollector.AppliesTo` is `!target.IsAzureSqlDb`, so an Azure SQL DB server has no such snapshot. The prune statement is guarded on the snapshot existing (an unguarded anti-join against an empty snapshot deletes every row), so on Azure it correctly no-ops — and those orphans therefore accumulate unpruned.

Azure SQL DB does write these rows: `QueryStoreCollector.ReadAsync` (the per-database Azure path) and `ReadItemAsync` (the on-prem enumerated path) both route through `ReadRowsAsync`, which holds the watermark write-back, and Darling sets `CapturePlanXml` on both.

### Why the obvious fix is wrong

The per-cycle Azure database list from `DarlingCollectorRunner.GetAzureDatabaseListAsync` is NOT authoritative for existence, in three separate ways:

- it filters `state_desc = N'ONLINE'`, so a paused or offline database is absent while still existing;
- it applies the per-server excluded-database filter;
- on a master-access error it falls back to the connection's own single database (#1506), which as a prune input would delete every other database's state on the logical server.

Pruning against it would delete live watermarks on exactly the servers that have excluded or intermittently-offline databases — the same self-defeating shape #2188's issue text warns about for the on-prem enumeration.

### What a fix would need

An authoritative unfiltered `sys.databases` name list read from the logical master, used only when the master read actually succeeded (never on the fallback path). That is reachable — the Azure enumeration already opens a master connection — but getting it without either a second master connection per cycle or reshaping the existing enumeration query is a design call, which is why it is not folded into #2188.

### Severity

Cosmetic, same as #2188: an orphaned row is never read again, and the accumulation is bounded to one ~100-byte row per database name ever seen. Worth doing on a logical server with churny database provisioning, not urgent anywhere.

erikdarlingdata:
[...]

Issue #1506, opened 2026-07-13
[QUESTION] Connection is broken: need PM restart ?


ZedZipDev:
### Which component is your question about?

Lite

### Performance Monitor Version

3.1.0

### Is your question about how it works, or the results?

How the tool works

### What's your question?

My situation: the PM->Azure SQL Db is broken 1-2 times per 24h. I run the PM and then check it next morning: the connection was broken. 
I have 2 notes:
1. It will be good to have in the settings and in the code: to execute app/ps1 in case of connection is broken. In this case the app/ps1 can be used at least for notifications/emailing about problem. In my case I can re-run gh action to activate access.
2. Also, I restore connection but after it I need to restart the PM because it does not restore functionality

### Additional Context

Azure SQL DB

erikdarlingdata:
Thanks for the detailed report — both notes were on point, and note 2 turned out to be a real bug. Here's where each one stands.

## 2. Having to restart after the connection recovers — this was a bug, and it's fixed

Your description ("I restore connection but after it I need to restart the PM because it does not restore functionality") was exactly right, and it pointed straight at the cause.

When your public IP rotates, Azure SQL rejects the connection with error **40615** ("client with IP address … is not allowed to access the server"). Performance Monitor was misreading that firewall rejection as a *permanent* "this login can't read `master`" condition and caching it for the life of the process. So even after you re-added the firewall rule and the server was reachable again, it kept running against that stale cached state — and only an app restart cleared it. Error **40613** ("database not currently available, retry later") had the same problem, even though the retry helper elsewhere already treated it as transient.

The fix (#1508):

- Firewall/transient errors (40615, 40613) are no longer treated as permanent — they're handled like any other retryable connection error.
- The internal "can't read master" fallback now expires on a timer **and** is cleared the moment a server comes back online, so recovery is automatic — no restart.
- A related silent case is also fixed: if the connection had no explicit database set, those collectors could quietly report "OK" while collecting zero rows. That now surfaces a real error telling you to set a database, instead of looking healthy.

The same fix is applied to the new headless collector service. It'll be in the next release.

## 1. Running something when the connection breaks — coming, via webhooks + built-in alerts

Your underlying goal here — get notified (and ideally re-trigger your GitHub Action) when the connection drops — is a good one, and Lite was the odd app out: it only showed a tray balloon, while the full Dashboard already sent alerts. Two things are being added:

[...]

PR #2195, opened 2026-08-11
Prune query_store per-database state for dropped databases (#2188)

### LINKED ISSUE #2188: [BUG] planwm: collector_state rows are never pruned for dropped databases
### Component

Darling

### Describe the Bug

Follow-up from review on #2183 (the #2164 plan-XML watermark). Not a correctness bug — an orphaned row is simply never read again — but a gap against the cleanup pattern the sibling state already establishes.

`QueryStorePlanXmlState` writes one `collector_state` row per database, keyed `planwm:<databaseName>` under the `query_store_plan_xml` owner. Nothing ever deletes them. `QueryStoreBackfillState`'s `hole:` keys, by contrast, get an explicit `DeleteCollectorStateKeyAsync` once resolved.

So a dropped or renamed database leaves its watermark row behind forever. On a server with churny database lifecycles — dev/test, or multi-tenant provisioning that creates and drops databases routinely — those accumulate with no pruning path. `collector_state` is a keyed registry rather than a hypertable (pinned as such by `CollectorStateContractTests`), so it has no retention policy to catch this either.

### Expected Behavior

Prune `planwm:` keys for databases that are no longer present, most naturally during the eligibility/enumeration pass that already produces the live database list for the server.

### Additional Context

One design caution for whoever picks this up, including me: the enumerated list is NOT the same thing as "every database that exists." A database can be missing from a cycle's enumeration because it is offline, excluded by configuration, or failed its probe — none of which mean it was dropped. Pruning on a single absence would delete a live watermark and cost a full plan-XML refetch for that database on its next pass.

That is the safe direction (a refetch, never missing data), so this is not dangerous, but it does make naive pruning self-defeating on exactly the servers that have excluded or intermittently-offline databases. Worth requiring absence across some number of consecutive enumerations, or checking against `sys.databases` directly rather than against the filtered enumeration.

Bounded either way: at most one row per database name ever seen, ~100 bytes each.

erikdarlingdata:
## What does this PR do?

Closes #2188.

Query Store collection writes per-database rows into `collector_state` and never deleted any of them for a database that was dropped or renamed:

| key | owner | written by | which app |
|---|---|---|---|
| `planwm:<db>` | `query_store_plan_xml` | the #2164 plan-XML watermark | Darling |
| `done:<db>` | `query_store_backfill` | the #2022/#2058 backfill worker | **both** |
| `hole:<db>` | `query_store_backfill` | the #2022/#2058 backfill worker | **both** |

`collector_state` is a keyed registry rather than a hypertable — pinned as such by `CollectorStateContractTests` — so no retention policy caught them either. Both apps now retire them on the query_store cycle.

## The design is the input, not the delete

The delete itself is four lines. What this PR actually is, is the choice of which list answers *"does this database still exist"*.

**It is `database_states`** — the newest snapshot for the server, an unfiltered `SELECT ... FROM sys.databases`.

**It is NOT query_store's own enumeration**, which the issue text flagged and which is the tempting place to put this. That list screens out, in `OnPremDatabaseListQueryText` alone: non-ONLINE databases, AG secondaries, the excluded-database filter, a vendor/system name screen, anything failing `HAS_DBACCESS`, plus any database whose per-database probe threw.

Pruning on that absence would delete **live** state on exactly the servers that keep databases parked or excluded — and the only symptom would be a silent refetch, which is the safe direction and therefore reports nothing. So the load-bearing test in each app seeds a database that is present in `sys.databases` but that no enumeration would ever return, and asserts its state **survives**.

## Two guards, and the second is the one that makes it correct

```sql
AND   snapshot.newest IS NOT NULL
AND   s.updated_at < snapshot.newest
```

**Existing** is not the same as **current**. If `database_states` stops collecting for a server — a schedule change, a failing collector, anything — its newest snapshot freezes, and every database created after that instant is missing from it while being perfectly alive. Presence alone would prune such a database's watermark on *every cycle forever*, silently paying a full plan-XML refetch each time (the exact cost #2164 exists to remove) while logging that a live database is gone. A snapshot cannot judge a row written after it was taken. Both stamps are the service clock's naive UTC, so this holds regardless of the two collectors' relative cadences.

The NULL check is now belt-and-braces rather than independently load-bearing: `<` against a NULL `MAX` is already NULL, so the freshness guard covers the empty snapshot too. The red-watch below reports that honestly — removing it leaves the behavioural test green and only the source pin red. It stays because "no snapshot prunes nothing" is a promise worth reading off the statement instead of deriving from three-valued logic. An unguarded anti-join against an empty snapshot deletes **every** row, and two ordinary configurations reach that state: **Azure SQL DB**, where `DatabaseStateCollector.AppliesTo` is `!IsAzureSqlDb` so the snapshot never exists (residual filed as **#2191**, with the reason that path's own database list cannot substitute — it filters `state_desc = N'ONLINE'`, applies exclusions, and falls back to a single database when master is unreachable), and any server whose snapshots have aged out of raw retention.

## Both apps, from one shared list

The first cut of this shipped a **wrong** parity claim, caught in review before merge: Lite writes no `planwm:` (it never sets `CapturePlanXml`) but its own backfill worker writes `done:` and `hole:` per database and only ever deletes a hole it *services or expires* — so a dropped database's markers were kept forever in Lite's DuckDB too, the identical defect.

The key set now lives in the shared `PerformanceMonitor.Collectors/QueryStorePerDatabaseState.cs` and both hosts iterate it, at the same point in the same cycle. Lite runs the `planwm:` statement too, against rows it never writes: one no-op delete buys the guarantee that enabling plan capture there later cannot quietly create an orphan class this forgot about.

`EveryPerDatabaseKeyPrefixQueryStoreOwnsIsAccountedFor` reflects over both state classes' own `*KeyPrefix` consts and fails if a new one appears in neither shared list. Its message names the trap rather than just the omission: adding a *server-scoped* key to `PrunableKeys` to silence it would have that key deleted on every cycle, since both prunes test a key by rebuilding it as `prefix + databaseName`.

## The two edges the issue asked about

**A database dropped then RE-CREATED with the same name** — the only shape here that could cost data, since Query Store restarts `plan_id` at 1 and every new plan sorts below the stale watermark.

I verified which mechanism covers this rather than assuming: **#2183 ships no reset detection.** It was written, found unsound, and removed — the tempting test ("the highest plan_id seen this pass is below the standing watermark") is TRUE in any ordinary window where nothing new compiled. What bounds it is `QueryStorePlanXmlState.RefreshAfter`, the one-day horizon measured from the last FULL fetch. `RecreatedDatabase_IsBoundedByTheRefreshHorizon_NotByResetDetection` pins that explicitly, including the one-second-inside case where the stale watermark *does* still apply, so the bound is a checked fact rather than a PR claim. The prune strictly improves on it — a drop observed between cycles removes the row outright — but cannot *be* the guarantee, because a drop and recreate inside one cycle is never observed as an absence.

**Racing an in-flight collection cycle** — a database that still exists is never targeted, since the predicate is absence from `sys.databases`. `Prune_RacingAnInFlightCycle_CannotLoseTheWatermark` drives the adversarial case deliberately anyway (load state, prune the key underneath it, write back) and shows the upsert restores it — and that the restored row, stamped now, is then protected by the freshness guard, so the two cannot fight on the next cycle.

- [x] Lite
- [x] Darling
- [x] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [x] Documentation (CHANGELOG)

No migration — `collector_state` is V44 and unchanged, and Lite's twin is created unconditionally at startup.

## How was this tested?

**Live Postgres 18.4 + TimescaleDB 2.28.1** on a throwaway cluster for Darling, **real DuckDB** via `SharedDuckDbFixture` for Lite. Both drive the production methods, not re-implementations.

**Watched red, with rebuilt test DLLs.** Eleven mutations of the shipped code, each reverted after:

| mutation | result |
|---|---|
| Darling: drop the snapshot-freshness guard | RED `Prune_LeavesStateWrittenAfterTheSnapshot` |
| Darling: match the anti-join by prefix instead of equality | RED `Prune_RetiresOnlyDroppedDatabases` |
| Darling: ungate the prune from `AppliesTo` | RED `BothHostsPruneOnTheQueryStoreCycle` |
| Darling: drop the key-prefix filter | RED `Prune_RetiresOnlyDroppedDatabases` |
| Darling: read the OLDEST snapshot | RED `Prune_RetiresOnlyDroppedDatabases` |
| Darling: unscope the snapshot from the server | RED `Prune_WithNoDatabaseSnapshot_RetiresNothing` |
| Darling: make the write-back unable to restore a pruned row | RED `Prune_RacingAnInFlightCycle_CannotLoseTheWatermark` |
| shared: forget the backfill prefixes | RED `EveryPerDatabaseKeyPrefixQueryStoreOwnsIsAccountedFor` |
| Lite: drop the freshness guard / prefix filter / equality / call site | RED (4 separate runs) |
| Darling: drop the `IS NOT NULL` guard | **green** behaviourally, red on the source pin — reported above, not hidden |
| log the count but not the keys | RED `Prune_RetiresOnlyDroppedDatabases` — the `RETURNING` diagnostic |
| add a new per-database key prefix nobody pruned | RED `EveryPerDatabaseKeyPrefixQueryStoreOwnsIsAccountedFor` |

Two more parity-shaped changes went in after that round. The drift guard now **discovers** the query_store state classes by reflection rather than checking a hand-written pair, since a third state class would have been invisible to it — the same silent-omission shape the guard exists to catch. And Lite's anti-join moved from `NOT IN` to `NOT EXISTS` to match Darling's NULL semantics exactly: one NULL anywhere in a `NOT IN` list makes the whole predicate NULL, so that prune would silently stop retiring anything. Fail-safe, and therefore exactly the divergence that sits undetected for a release.

Final suite totals after merging `origin/dev` three times as it moved: **Darling 4509 passed / 0 failed** (12 environment-gated skips), **Lite 2189 passed / 0 failed**.

The live cases assert on one pass that a dropped database loses all three prefixes; that a **parked (OFFLINE)** database keeps its state; that `AppArchive` is retired while `App` survives (the prefix-overlap trap, which a `starts_with` anti-join would have gotten wrong); that both `done:` and `hole:` have a survivor as well as a casualty; and that a non-database-keyed key, another collector's key, and a neighbouring server's rows are untouched. A second pass over the clean store is a no-op.

## Checklist

- [x] My code adds zero warnings. Measured with `--no-incremental` on both sides, because an incremental build silently reports 0 by not recompiling: this branch and clean `dev` each produce the SAME two (`CA1068` in `PerformanceMonitor.Common/AbandonableStep.cs`, `CA2022` in `Darling.Tests/NpgsqlRootCertificateValidationTests.cs`), and Lite the same 10. Neither file is touched here.
- [x] I have tested my changes against at least one SQL Server version (n/a — this is store-side only; exercised against a live Postgres store and a real DuckDB instead)


erikdarlingdata:
Two things on this one, both external to your change:

1. **#2199 is merged** — it pins the container SDK image to `10.0.302` to match `global.json`. Tonight's floating `sdk:10.0` tag advanced to an SDK 10.0.400 image and broke every container build repo-wide with a bare `exit code: 155`. Rebase onto dev and that step clears.
2. This PR currently shows **CONFLICTING**, so it needs a rebase regardless.

Also worth flagging since it overlaps: I filed #2188 from the review on my own #2183, and the caution I noted there applies to this implementation — the enumerated database list is not the same as "every database that exists." A database absent from one cycle's enumeration may be offline, excluded by config, or have failed its probe, none of which mean it was dropped. Pruning on a single absence deletes a live watermark and forces a full plan-XML refetch, which is the safe direction but self-defeating on exactly the servers that have excluded or intermittently-offline databases. Worth requiring absence across consecutive enumerations, or checking `sys.databases` directly.

erikdarlingdata:
## Final rollup

**Scope grew by one app during review, and that was the important finding.** The first cut declared this Darling-only because Lite never sets `CapturePlanXml` and so writes no `planwm:` row. True, and irrelevant to two thirds of the defect: Lite's own backfill worker writes `done:` and `hole:` per database and only ever deletes a hole it *services or expires*, which a dropped database can never do. Verified in Lite's source (`RemoteCollectorService.QueryStoreBackfill.cs`) before acting, not taken on report. Ported to DuckDB, with the key set moved into a shared `QueryStorePerDatabaseState` that both hosts iterate at the same point in the same cycle.

**Second review finding, and the one that makes this correct rather than usually-correct:** the snapshot must be NEWER than the row it judges. Existing is not current — if `database_states` stops collecting, its newest snapshot freezes, and every database created after that instant is missing from it while being perfectly alive. Presence alone would prune those live rows on every cycle forever, paying a full plan-XML refetch each time and logging that a live database was dropped.

### Gates

| gate | result |
|---|---|
| Watched red, rebuilt test DLLs | 9 mutations across both apps; **8 red**, 1 documented green (below) |
| Darling suite (live PG 18.4 + TimescaleDB 2.28.1) | **4475 passed, 0 failed**, 12 env-gated skips |
| Lite suite (real DuckDB) | **2179 passed, 0 failed** |
| Build warnings | **0 added** — `--no-incremental` on branch and clean `dev`, identical sets both sides |
| CHANGELOG | `[Unreleased]` → Fixed, link-refs `[#2188]` / `[#2191]`; **4 added / 0 deleted** lines, all 6 of the file's bare CR bytes preserved through the conflict resolve |
| `origin/dev` merge | merged after dev moved; both suites re-run green afterwards |

The one mutation that stays green is dropping `AND snapshot.newest IS NOT NULL`: the freshness guard subsumes it, since `<` against a NULL `MAX` is already NULL. Only the source pin catches its removal. Reporting it rather than claiming nine of nine — the predicate stays because "no snapshot prunes nothing" is worth reading off the statement instead of deriving from three-valued logic.

### Not done, deliberately

**#2191** — Azure SQL DB never collects `database_states` (`DatabaseStateCollector.AppliesTo` is `!IsAzureSqlDb`), so the prune correctly no-ops there and those orphans remain. Filed rather than folded in, because that path's own database list cannot substitute: it filters `state_desc = N'ONLINE'`, applies exclusions, and falls back to a single database when master is unreachable (#1506). Pruning against it would be exactly the trap the issue text warns about. Both call sites are gated on the same `AppliesTo`, so the boundary is stated in code rather than emergent from the guard.

Not merged, and #2188 left open, per instruction.

erikdarlingdata:
## CI green on the final commit

All five checks pass against `5f5665f2` — verified by `headSha`, not by the checks list, since the list happily shows a green run from an earlier push:

| check | |
|---|---|
| build | pass (17m8s) |
| Darling PostgreSQL tests | pass (3m30s) — the job that actually runs this PR's live prune tests |
| Darling Linux build | pass |
| review | pass |
| check-branches | pass |

`mergeable=MERGEABLE`. `origin/dev` moved three times during this work and was merged each time, with the CHANGELOG resolved byte-precisely on every one (all 6 of the file's bare CR bytes intact, total diff 3 added / 0 deleted lines).

Final local totals after the last merge: **Darling 4509 passed / 0 failed** (12 environment-gated skips), **Lite 2189 passed / 0 failed**.

Not merging, and leaving #2188 open, per instruction.

erikdarlingdata:
Merging on this rollup. All six review threads reconciled against the FINAL head 5f5665f2: the Lite-coverage parity gap and the NULL-unsafe NOT IN were both fixed in the last commit (verified in source - the anti-join is NOT EXISTS with Darling-matching semantics; the two 20:26-timestamped threads were anchored to the previous push), the stale-snapshot residual fails SAFE by construction (a frozen database_states MAX protects rows rather than exposing them - the guard's direction is the right one), and the one genuinely outstanding item, Lite logging a count where Darling names the pruned keys, is diagnostic-only and tracked as a follow-up issue rather than a fourth CI cycle tonight. The reflection-discovered state-class guard and the headSha-verified check reads are both the right pattern; the jq-not-on-PATH silent no-op warning is going into project memory.

Issue #2189, opened 2026-08-11
[BUG] RESTORING gets baselined as a database's expected state, alerting forever once it goes ONLINE


erikdarlingdata:
### Component

Both

### Describe the Bug

Found by dogfooding on the 52-server production monitoring box. In the last 24 hours it fired **636 Database State alerts from only 5 distinct (server, database) pairs** — roughly 127 identical repeats each:

```
636 fires / 5 distinct (server,detail) pairs

multi-22       Database: db-A  Expected: RESTORING  Current: ONLINE  -> x134
psi            Database: db-B  Expected: RESTORING  Current: ONLINE  -> x130
multi-53       Database: db-C  Expected: RESTORING  Current: ONLINE  -> x128
multi-44       Database: db-D  Expected: RESTORING  Current: ONLINE  -> x125
multi-49       Database: db-E  Expected: RESTORING  Current: ONLINE  -> x119
```

(Database names are relabelled `db-A`..`db-E` — this repo is public and the originals are client-identifying. Server names, counts and every number are unchanged.)

Every one is a database that was **observed mid-restore** during a consolidation, had `RESTORING` learned as its accepted baseline, and has been "deviating" ever since by being healthy.

`SeedDatabaseStateExpectedSql` refuses to baseline a critical first observation, but its exclusion list is only `SUSPECT`, `RECOVERY_PENDING`, `EMERGENCY`:

```sql
AND (CASE WHEN ds.is_in_standby THEN 'STANDBY' ELSE ds.state_desc END)
    NOT IN ('SUSPECT', 'RECOVERY_PENDING', 'EMERGENCY')
```

`RESTORING` is every bit as transient as `RECOVERY_PENDING` — it is a database in the middle of an operation, not a steady state anyone would choose as "expected." Baselining it inverts the alert permanently: the database is now paged about for being ONLINE, forever, and the only escape is an operator noticing and re-baselining by hand.

The guard's own doc says onboarding a server mid-outage "must not learn the bad state as expected." A restore in progress is that same situation; the list just doesn't include it.

### Steps to Reproduce

1. Add a server, or let one be swept, while a database is `RESTORING` (a restore, a log-shipping cutover, a consolidation).
2. Let the seed baseline it.
3. Restore completes; database goes `ONLINE`.
4. It now deviates from `RESTORING` forever and alerts every cooldown.

### Expected Behavior

[...]

PR #2196, opened 2026-08-11
Stop learning a mid-restore state as a database's expected one (#2189)

erikdarlingdata:
The baseline seeder refused to learn SUSPECT, RECOVERY_PENDING and EMERGENCY, but
not RESTORING or RECOVERING, which are equally transient. A database swept while a
restore was running therefore had RESTORING recorded as its ACCEPTED state, and
every healthy observation afterwards counted as a deviation - permanently, with no
exit but an operator noticing and re-baselining by hand.

Found by dogfooding rather than reasoning: the 52-server production monitor fired
636 Database State alerts in 24 hours from 5 databases, every one reading
"Expected: RESTORING, Current: ONLINE", all of them consolidation restores observed
at the wrong moment. The seed guard's own doc says onboarding mid-outage must not
learn the bad state as expected; a restore in progress is that same situation and
the list just did not include it.

Both transient states are now excluded from seeding, so the database stays pending
until it settles and the seeder learns what it settles into.

Rows already poisoned are repaired automatically, because the seed cannot do it -
it is insert-if-absent by design, so it will never correct an existing row. A
maintenance statement runs beside the existing seed and prune and drops any
auto-baseline naming a transient state. Deliberately NOT a migration rung: this way
it is self-healing for any future path that poisons a baseline, needs no schema
change, and is idempotent.

Ordering is load-bearing: the repair runs BEFORE the seed, so the correct steady
state is re-learned in the same cycle. Repairing afterwards would leave the
database with no baseline until the next pass.

Two things it must not touch, both pinned: is_user_override rows, because an
operator who deliberately expects RESTORING means it, and STANDBY, because a
log-shipping secondary lives there permanently and only superficially resembles the
transient states.

Lite gets the identical change - same exclusion list, same repair before its seed -
since both SKUs seed baselines the same way and a fix in one would leave the other
generating the same flood.

Tested live rather than in the harness, because the statement IS the behaviour:
seeds six baselines covering guessed-transient, steady, STANDBY and
operator-override, asserts exactly three rows deleted and which three survive, and
asserts a second pass deletes zero since this runs on every evaluation of every
server. A second test pins that the seed excludes every state the repair deletes -
if those lists ever disagree the pair oscillates, writing and deleting the same row
forever.

## How was this tested?

Live-gated test against a real Postgres store, because the fix is entirely SQL — there is no C# behaviour to exercise at harness level. It seeds six baselines (guessed RESTORING, guessed RECOVERING, guessed SUSPECT, steady ONLINE, steady STANDBY, operator-override RESTORING), asserts **exactly three** rows are deleted and names which three survive, then asserts a second pass deletes **zero** since this runs on every evaluation of every server.

A second test pins the invariant that matters most: the seed's exclusion list must cover everything the repair deletes. If they ever disagree the two statements oscillate — the seed writes a row the repair removes next cycle, forever, churning the table and re-announcing the database.

Builds clean (zero errors) across Darling.Service, Darling.Tests, Lite and Lite.Tests.

**Not verified by me:** that the five poisoned baselines on the production box actually clear. That needs the deploy, and I'll confirm the alert rate drops from ~720/24h rather than assuming it.


erikdarlingdata:
The `Darling Linux build` failure here is the repo-wide SDK image drift, not this change — fixed in #2199, which needs to merge first.

Diagnosis trail, since `exit code: 155` with no compiler output is misleading: the floating `sdk:10.0` tag now ships SDK 10.0.400, `global.json` pins 10.0.302 with `rollForward: latestPatch` (3xx band only), so the in-container publish can't find a compatible SDK. Verified it isn't mine three ways — `global.json` is byte-identical to dev and absent from this diff, the runner-side publish of the same csproj succeeded in the same job seconds earlier, and PR #2194 (an unrelated change) fails the identical step identically.

Everything else here is green, including a genuine `review` pass — 3m52s and zero findings, which per the duration check is a real review rather than the skipped-but-green pattern I hit on the Studio repo earlier today.

erikdarlingdata:
Repo-side session here (the one running the #2186/#2187/#2188 lanes today). Gating this PR found the review's parity finding CONFIRMED in source on head: ViewerDataService.DatabaseStates.cs line 39 - the Viewer's independent DatabaseStateSeedSql still excludes only ('SUSPECT', 'RECOVERY_PENDING', 'EMERGENCY'), so the Viewer's own seed path can still learn RESTORING/RECOVERING as expected and re-poison exactly the baselines the repair statement clears - the 636-alert flood recreated through the other door. The seed/repair oscillation invariant your live test pins also cannot see this copy, since it checks the service-side pair. Two more from the review threads: the CHANGELOG cites [#2189] with no bottom link-def (renders broken), and the Lite test nit reads as covered by the existing DatabaseStateExpectedStoreTests - your call. This branch is yours, so I am not pushing to it: fix round on your side (Viewer NOT IN list + mirror the repair there if the Viewer evaluates deviations, plus the link-ref), or say the word here and I will take it on your branch with a clear commit. I hold the merge either way until the Viewer copy matches - parity drift between the three seed sites is precisely the failure mode this repo bleeds from. If a third seed copy exists anywhere else, worth one grep while you are in there: the fix that ends this class is hoisting the exclusion list to ONE shared constant all three sites reference, if the SQL composition allows it.

erikdarlingdata:
All three fixed, and the first one was a genuine hole — thank you.

**The Viewer's third copy.** You're right, and the framing is what makes it serious: that file's own doc says it mirrors the service's seed "so the editor and the alert agree on what 'expected' means," and my previous commit broke that invariant. It isn't a display inconsistency — the editor is a **write** path, so opening the override screen while a database was mid-restore would baseline RESTORING again and reproduce the exact storm this PR fixes, via the editor instead of the sweep. And it would show a poisoned `Expected: RESTORING` to the one person most likely to assume the screen had already fixed it.

Same exclusion list there now, with the repair before its seed for the same reason as the service. I also grepped for further copies rather than assuming three was the total — the service adapter, Lite, and the Viewer are all of them.

**The changelog link.** Correct, the reference block jumped `[#2138]` to `[#2190]`, so it would have rendered unresolved. Added between them.

**The Lite tests.** Fair, and your pointer to `DatabaseStateExpectedStoreTests` was the useful part — in-process DuckDB, no live gate, so these are cheap. Three added: a mid-restore first observation stays pending then baselines what it settles into; an already-poisoned guessed baseline is repaired **and re-seeded in one call** (pinning the ordering claim, not just the end state); and an operator override on a transient state survives.

That third one needed a direct-insert helper, which is worth noting because it's a real gap in the test surface: `SetDatabaseStateExpectedAsync` always stamps `is_user_override = true`, so there was no way to express "the product guessed this" — and that distinction is precisely what the repair keys on.

One limitation stated plainly: `Lite.Tests` is `net10.0-windows` and doesn't run on this machine, so those three are CI-verified only. The Darling live test and all four affected projects build clean locally.

erikdarlingdata:
Closing this in favour of #2202, which implements the same issue better. We collided — I opened this at 18:24 and #2202 landed at 19:07 — and theirs is the one that should ship. Recording why, since the differences are instructive rather than cosmetic:

**It removes the bug class I just got caught by.** #2202 puts the state list in a shared `DatabaseStateTokens.NeverBaselinedSqlList` and pins by test that *both* Darling seed sites interpolate it. I hardcoded the same list in three separate SQL strings, which is exactly the third-copy divergence the reviewer found in this PR an hour ago — I fixed the instance and left the mechanism.

**It heals rather than deletes.** Mine DELETEs a poisoned baseline and waits for the seed to re-learn; #2202's `HealDatabaseStateBaselineToOnlineSql` relearns ONLINE directly when the effective state is ONLINE. Fewer moving parts and no window where the row is simply absent.

**It closes a route I never touched.** `DatabaseStateResetToCurrentSql` — the editor's "reset to current" button — records whatever it sees with **no state filter at all**, so an operator pressing it mid-restore re-poisons the baseline. My PR left that open. That's not a detail; it's a permanent re-entry path for the same bug.

**It reasoned harder about what NOT to heal.** I swept SUSPECT/RECOVERY_PENDING/EMERGENCY along with the transient states. #2202 deliberately leaves auto-inferred OFFLINE and STANDBY alone, with a real argument: a STANDBY secondary that turns up truly ONLINE is *news* (log shipping is broken), and an auto-baselined OFFLINE database brought up for maintenance and re-parked would otherwise deviate forever against a baseline it never had. It also matches the **effective** state rather than `state_desc`, because a standby reports `state_desc = ONLINE` with `is_in_standby` — matching the raw column would have re-baselined every secondary and then alerted it forever, which is this same bug recreated for the family the alert works hardest to keep quiet.

Two things from here worth confirming survive in #2202, both of which its design may already make moot:

1. **Idempotency** — my live test asserted a second pass changes **zero** rows, which matters because this runs on every evaluation of every server. A heal-to-ONLINE is naturally idempotent once healed, so this is probably inherent, but worth an explicit assertion.
2. **Repair-before-seed ordering** — I ran the repair first so the correct steady state is re-learned in the same cycle rather than leaving a gap. If #2202 heals in place, the ordering question dissolves.

Nothing here needs porting. Closing and deleting the branch.

Issue #2185, opened 2026-08-11
[BUG]


jovon44:
### Component

Darling

### Performance Monitor Version

3.2.0

### SQL Server Version

SQL Server 2022 CU 25

### Windows Version

Windows 2019

### Describe the Bug

Good morning,
I am trying to get your Performance monitor tool and keep getting the following error after install. 
darling.json could not be read: darling.json uses the managed Postgres (postgres.managed = true), but the credential file for the 'admin' role (C:\ProgramData\Performance MonitorDarling\pg-admin- credential.dpapi) does not exist yet. Start the Performance Monitor Darling service once - its first run initializes the bundled Postgres and provisions the least-privilege roles and their credentials.
here is what's in the log file:
8-11 10:07:22.798 [INFO ] [Lifetime] Application started. Hosting environment: Production; Content root path: C:\Users\username\Desktop\PerformanceMonitorDarling-3.2.0\
2026-08-11 10:07:22.879 [INFO ] [DarlingWorker] Loaded configuration from C:\Users\Username\Desktop\PerformanceMonitorDarling-3.2.0\darling.json: 2 server(s)
2026-08-11 10:07:22.894 [INFO ] [DarlingWorker] Initializing managed Postgres cluster in C:\ProgramData\PerformanceMonitorDarling\pg (first run)
2026-08-11 10:07:23.096 [CRIT ] [DarlingWorker] Managed Postgres bootstrap failed: initdb failed (exit code -1073741515) for C:\ProgramData\PerformanceMonitorDarling\pg. Output: 

### Steps to Reproduce

installed app from powershell  .\install-darling.ps1

### Expected Behavior

service and app to install properly

### Actual Behavior

darling.json could not be read: darling.json uses the managed Postgres (postgres.managed = true), but the credential file for the 'admin' role (C:\ProgramData\Performance MonitorDarling\pg-admin- credential.dpapi) does not exist yet. Start the Performance Monitor Darling service once - its first run initializes the bundled Postgres and provisions the least-privilege roles and their credentials.

### Error Messages / Log Output

```text
8-11 10:07:22.798 [INFO ] [Lifetime] Application started. Hosting environment: Production; Content root path: C:\Users\username\Desktop\PerformanceMonitorDarling-3.2.0\
[...]

Issue #1823, opened 2026-07-29
[QUESTION] Regarding installing the windows service and integrated security


ghauan:
### Which component is your question about?

Darling

### Performance Monitor Version

3.3.0

### Is your question about how it works, or the results?

Setup / installation

### What's your question?

My apologies if this is already covered in the documentation (I tried to find the answer there). My question is about the installation of the service and which user account is used for logging into the service.

From what I can see in the documentation, it uses the Virtual NT SERVICE account.

I have not yet tried installing it, as I want to make sure I understand this correctly first. I did run the preflight --test-connection, and that worked fine, but I assume that uses my admin account since I am logged into the VM with it.

So my question is: if the service uses the NT SERVICE account, which account will be used when connecting to the monitored servers? I expected to be able to specify my own AD service account for this.

Is this something I may have misunderstood or missed in the documentation?

### Additional Context

Windows SErver 2025 and SQL Server 2025 (while testing)

erikdarlingdata:
You have it exactly right, on both counts.

For any server with `"auth": "integrated"` in darling.json (the default), the service connects to your monitored SQL Servers as whatever Windows account the service logs on as. There is no separate place to configure a Windows credential for monitoring. And yes, your preflight ran as you: `--test-connection` from a console connects as the logged-in user, so it proved your config and reachability, not what the service account can do.

The NT SERVICE virtual account in the docs is the default because it is the least-privilege choice, and it is the right one for SQL auth monitoring. For integrated auth to remote servers it is usually the wrong identity: a virtual account reaches the network as the computer account (`DOMAIN\YourCollectorVM$`), so you would end up granting the machine account on every monitored server. Your plan to use your own AD service account is the standard setup, and a gMSA is even better if you have them.

Sequence for a fresh install:

1. Run `install-darling.ps1` as normal. It creates the service under the virtual account.

2. Stop the service and change the Log On account: Services.msc > PerformanceMonitor Darling > Log On > This account > `DOMAIN\svc-account`. Services.msc also grants the "Log on as a service" right automatically. From a prompt instead (then grant that right yourself via secpol.msc or GPO):

   ```
   sc config "PerformanceMonitor Darling" obj= "DOMAIN\svc-account" password= "ThePassword"
   ```

   A gMSA works with an empty password: `obj= "DOMAIN\gmsa-name$" password= ""`. Keep the account out of the local Administrators group: the bundled PostgreSQL refuses to run with administrative privileges.

[...]

Issue #1802, opened 2026-07-28
performance darling service acccount


ESLCH:
### Which component is your question about?

Full Dashboard

### Performance Monitor Version

performancemonitor darling 3.2.0

### Is your question about how it works, or the results?

How the tool works

### What's your question?

Hi,
how to change service account to connect to sql Server instance, because local service account is forbidden
i'm using an another one AD service account 

### Additional Context

windows server , SQL Server version >= 2019 , on-premise

erikdarlingdata:
The Darling service connects to your monitored SQL Servers as whatever Windows account the service runs under, for any server entry with `"auth": "integrated"` in `darling.json`. So "changing the service account" is a Windows-side change plus a SQL-side grant, with one file permission step in the middle that will bite you if you skip it.

**1. Stop the service, change the Log On account**

Services.msc > PerformanceMonitor Darling > Log On tab > This account > `DOMAIN\your-svc-account`. Doing it through Services.msc also grants the account the "Log on as a service" right automatically.

Or from an elevated prompt (note the space after `obj=` and `password=`, `sc` requires it):

```
sc config "PerformanceMonitor Darling" obj= "DOMAIN\svc-account" password= "ThePassword"
```

With the `sc` route you have to grant "Log on as a service" yourself (secpol.msc or GPO), which is why Services.msc is the easier path. A gMSA works too: `obj= "DOMAIN\gmsa-name$" password= ""`.

**2. Grant the AD account on each monitored instance**

```sql
USE master;
CREATE LOGIN [DOMAIN\svc-account] FROM WINDOWS;
GRANT VIEW SERVER STATE TO [DOMAIN\svc-account];
GRANT ALTER ANY EVENT SESSION TO [DOMAIN\svc-account];

/* Only if the instance hosts Availability Groups */
GRANT VIEW ANY DEFINITION TO [DOMAIN\svc-account];

/* Optional: SQL Agent job monitoring + failed-job alerts */
USE msdb;
CREATE USER [DOMAIN\svc-account] FOR LOGIN [DOMAIN\svc-account];
ALTER ROLE SQLAgentReaderRole ADD MEMBER [DOMAIN\svc-account];
```

**3. The step people miss: re-grant the service's own files**

The service deliberately locks its files down to SYSTEM, Administrators, and the account it was running as. After you switch accounts, the new account is not on those ACLs and the service will fail to read its config or write its store. One-time fix from an elevated prompt, before starting the service:

```
[...]

PR #2198, opened 2026-08-11
Refuse an install location the service account can never read (#2187)

### LINKED ISSUE #2187: [BUG] install-darling.ps1 creates an unstartable service when installed under a user profile

erikdarlingdata:
Fixes #2187. Split out of #2185 (reported by @jovon44).

`install-darling.ps1` created a working-looking service that could never start when the install directory sat under a user profile. A fresh install in such a location is now refused, before anything is installed.

## The mechanism, verified against the script

Every claim in the issue body checked out:

| Issue claim | Verified |
|---|---|
| The service is created under the `NT SERVICE\<serviceName>` virtual account, not LocalSystem | `install-darling.ps1:127` — `sc.exe create ... obj= "NT SERVICE\$serviceName"` |
| The installer's ACL work is scoped to `darling.json` and its `.bak-*` copies, never the install tree | the 4b loop iterates `@($configPath) + darling.json.bak-*` only; pinned by a new test |
| A profile directory grants access to about three principals and nothing else | measured on this box, below |
| `C:\ProgramData\PerformanceMonitorDarling` is machine-scoped and fine | measured, below |

Measured `icacls` on this machine — this is the whole bug in six lines:

```
C:\Users\edarl        NT AUTHORITY\SYSTEM:(OI)(CI)(F)
                      BUILTIN\Administrators:(OI)(CI)(F)
                      NADAFACCIA\edarl:(OI)(CI)(F)

C:\ProgramData        NT AUTHORITY\SYSTEM:(OI)(CI)(F)
                      BUILTIN\Administrators:(OI)(CI)(F)
                      CREATOR OWNER:(OI)(CI)(IO)(F)
                      BUILTIN\Users:(OI)(CI)(RX)
                      BUILTIN\Users:(CI)(WD,AD,WEA,WA)
```

And a directory created under each, which is what the install tree actually inherits:

```
C:\Users\edarl\pm2187-profile   NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
                                BUILTIN\Administrators:(I)(OI)(CI)(F)
                                NADAFACCIA\edarl:(I)(OI)(CI)(F)

C:\pm2187-machine               BUILTIN\Administrators:(I)(OI)(CI)(F)
                                NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
                                BUILTIN\Users:(I)(OI)(CI)(RX)
                                NT AUTHORITY\Authenticated Users:(I)(M)
                                NT AUTHORITY\Authenticated Users:(I)(OI)(CI)(IO)(M)
```

The profile directory has no `BUILTIN\Users`, no Authenticated Users, and **no `CREATOR OWNER`** — so the virtual account cannot read the program files it is pointed at, and cannot read back even what it writes there itself. A machine-scoped directory inherits `BUILTIN\Users:(RX)`, which every service account is a member of. That is exactly why `C:\PerformanceMonitorDarling` works and a Desktop does not.

The network half rests on Microsoft's own statement rather than a guess: ["Services that run as virtual accounts access network resources by using the credentials of the computer account in the format `<domain_name>\<computer_name>$`"](https://learn.microsoft.com/en-us/sql/database-engine/configure-windows/configure-windows-service-accounts-and-permissions#virtual-accounts) — a different principal from the operator who typed the UNC path, and a mapped drive letter belongs to a logon session a service never shares.

## The fix

Option 1 from the issue — hard refuse — with one deliberate exception.

**Fresh install** under the machine's profile root, or on a UNC / mapped-drive path: refused. The guard runs after the elevation and service-exe checks and **before** the pre-flight, the `darling.sample.json` copy, the Event Log source, `sc create`, the ACL hardening, and `Start-Service`, so a doomed location costs nothing and leaves nothing behind. Verbatim output (rendered by running the real block out of the shipped script):

```
ERROR: This folder is under a user profile:

  C:\Users\username\Desktop\PerformanceMonitorDarling-3.2.0

The service runs as the unprivileged virtual account 'NT SERVICE\PerformanceMonitor Darling' - never LocalSystem,
because the bundled PostgreSQL refuses to run with administrative privileges. That account is not you,
not SYSTEM, and not Administrators, and a user profile grants access to about those three and nobody
else. So the service installs cleanly and then cannot read its own program files: the bundled
PostgreSQL's initdb.exe dies at exit code -1073741515 (0xC0000135, STATUS_DLL_NOT_FOUND) before it can
write a word of output, because the DLLs sitting beside it are unreadable (#2185).

Move the extracted folder to a machine-scoped local path and run this script again from there:

  C:\PerformanceMonitorDarling

That is the documented location, and a folder created there inherits read + execute for BUILTIN\Users,
which the service's virtual account is a member of. Your darling.json can move with it.

Nothing was installed or changed.
```

The UNC variant swaps in the computer-account explanation:

```
ERROR: This folder is on a network location (UNC):

  \\fileserver\deploy\PerformanceMonitorDarling

The service runs as the unprivileged virtual account 'NT SERVICE\PerformanceMonitor Darling' - never LocalSystem,
because the bundled PostgreSQL refuses to run with administrative privileges. That account reaches the
network as the COMPUTER account rather than as you, so a share that opens for you is not open for it;
and a mapped drive letter belongs to YOUR logon session, which a service does not share and cannot see
at all. Either way the service installs cleanly and then cannot read its own program files.
```

**Upgrade** of a service that already exists there: same diagnosis, but it asks instead of refusing (`exit 4` on no, default no).

```
WARNING: a FRESH install would be refused in this location.
[...same diagnosis and fix as above...]
Service 'PerformanceMonitor Darling' already exists, so this is an upgrade, and this folder may have been made
to work by hand (granting the service account read + execute on the tree). That is the only reason
this is a question. If it has NOT been, the service will stop working the moment it restarts.
Point the service at this folder anyway? [y/N]:
```

That asymmetry is not new policy — it is the one this script already applies to the logon account, whose upgrade path deliberately touches only `binPath` so a re-homed domain account or gMSA survives (#1802, #1823). Option 2 in the issue (granting the service account read+execute on the tree) is exactly the customization an operator may already have applied here, and refusing outright would strand a deployment that runs today. A fresh install in the same folder is still refused.

Option 2 itself is **not** implemented, and a test now pins that: the installer's only `Set-Acl` targets remain the credential files it has always hardened. The product does not reach into somebody's profile and change permissions.

Two details worth calling out:

- The profile root comes from `ProfileList\ProfilesDirectory`, not a hardcoded `C:\Users`. It is relocatable, and a literal would stop matching on precisely the box that moved it. `$env:USERPROFILE` is checked as well, for a profile redirected outside that root.
- `\\?\C:\...` is the long-path prefix on a local path, not a share, and is excluded from the network test.

**One residual, measured and accepted rather than found later.** `C:\Users\Public` sits under the profile root and is refused, but an install there would actually work — it grants `NT AUTHORITY\SERVICE:(OI)(CI)(IO)(M,DC)`, which every service account holds:

```
C:\Users\Public BUILTIN\Administrators:(OI)(CI)(F)
                CREATOR OWNER:(OI)(CI)(IO)(F)
                NT AUTHORITY\SYSTEM:(OI)(CI)(F)
                NT AUTHORITY\INTERACTIVE:(OI)(CI)(IO)(M,DC)
                NT AUTHORITY\SERVICE:(OI)(CI)(IO)(M,DC)
                NT AUTHORITY\BATCH:(OI)(CI)(IO)(M,DC)
```

It is deliberately not carved out, and the reason is in the script: nobody installs a Windows service into the shared documents profile, an exception would amount to documenting it as a reasonable place to install, and the refusal is not a dead end — it names `C:\PerformanceMonitorDarling`. One rule, no exceptions, and the single location it costs is one nobody wants.

## Tests

`Darling/Darling.Tests/DarlingInstallLocationTests.cs`, 6 tests. Ordering, verdict, profile-root source and the never-ACL-the-tree decision are pinned structurally, as this project pins its other PowerShell invariants. The **path boundary is executed**, not parsed: `C:\UsersData` is not under `C:\Users`, and a prefix test that says it is would refuse an install that works — a worse failure than the one the guard exists to catch, and one no source-parsing assertion can see. The extracted function is run under `powershell.exe` (Windows PowerShell 5.1), which is what an operator on Windows Server 2019 — the reporter's OS — actually gets, so 5.1 compatibility of the guard is pinned at the same time.

Watched-red, one mutation per property, test DLL rebuilt first and the script restored byte-identical afterwards:

```
RED (good)    M1 ordering: register the Event Log source BEFORE the guard
RED (good)    M2 verdict: a fresh install warns instead of refusing
RED (good)    M3 boundary: prefix test without the separator (C:\UsersData reads as under C:\Users)
RED (good)    M4 UNC: drop the \\?\ long-path exclusion
RED (good)    M5 profile root: hardcode it instead of reading ProfileList
RED (good)    M6 rejected option 2: ACL the install tree instead of refusing

--- restored; full class green again ---
Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6
script restored byte-identical: True
```

## Gates

- Build: `Build succeeded. 2 Warning(s), 0 Error(s)` — both pre-existing (`AbandonableStep.cs` CA1068, `NpgsqlRootCertificateValidationTests.cs` CA2022), neither in a file this PR touches.
- Full `Darling.Tests` with a rebuilt DLL: `Failed: 0, Passed: 4232, Skipped: 236` (skips are the gated-live PG suites).
- The script parses clean under both PowerShell 7 and Windows PowerShell 5.1.
- CHANGELOG entry under `[Unreleased] / Fixed` with the `[#2187]` link-ref. Applied byte-surgically: the committed file carries mixed CRLF/lone-CR terminators, and a normal whole-file write normalizes them into a 5000-line diff that would conflict with every other lane touching this file. The diff is 2 insertions.
- `Darling/README.md` gains a **The install location has to be machine-scoped** section, since the manual `sc create` path can still put the service somewhere unreadable and the installer cannot guard that.

## What is NOT verified here, and why

The end-to-end *observation* — creating a scratch service under a profile path, watching it fail, then watching the fixed installer refuse — is scripted and ready but **not run**: it needs elevation, and UAC on this box is `ConsentPromptBehaviorAdmin=5` with `PromptOnSecureDesktop=1`, so it needs a physical click that an autonomous run cannot supply (two `-Verb RunAs` attempts returned "The operation was canceled by the user"; WinRM loopback is not configured either). Everything above is measured or executed; the service-start observation is the one thing deduced from the measured ACLs rather than watched.

The kit is staged at `C:\pm2187-verify\` and self-checks for elevation. From an elevated PowerShell:

```
C:\pm2187-verify\verify-2187.ps1
```

It stages an identical published tree under a profile path and under `C:\`, creates one scratch service per location under its own `NT SERVICE\` virtual account, starts both, and reports what each process managed to do — the probe is a framework-dependent exe that must read its own app-local DLLs to reach managed code, and it launches a child exe out of a `pg-runtime\bin` subdirectory the way `DarlingManagedPostgres` launches `initdb.exe`. Then it runs the fixed installer from a profile path and from a machine path and prints both outcomes, and deletes every scratch service and directory it made. Scratch names only (`PM2187Profile`, `PM2187Machine`); it never touches `PerformanceMonitor Darling` or DARLING01. Transcript lands in `C:\pm2187-verify\verify-2187.log`. Delete `C:\pm2187-verify` when done.

## Scope

The installer half only. The service-side diagnosis of the same failure — decoding the `initdb` exit code so an already-broken install explains itself — is #2186's, as the issue lays out.


erikdarlingdata:
Final CI status on `b2f1876` — all green, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

| check | result |
|---|---|
| build | SUCCESS |
| Darling Linux build | SUCCESS |
| Darling PostgreSQL tests | SUCCESS |
| check-branches | SUCCESS |
| review | SUCCESS |

Local gates: `Build succeeded. 2 Warning(s), 0 Error(s)` (both pre-existing, in files this PR does not touch), full `Darling.Tests` post-merge `Failed: 0, Passed: 4241, Skipped: 236` with a rebuilt DLL, and the script parses clean under PowerShell 7 and Windows PowerShell 5.1.

Not merging and not closing the issue — #2187 stays open until the elevated end-to-end run in the "What is NOT verified here" section above has been done.

erikdarlingdata:
Merging on this rollup. The one review nit (WMI-unavailable mapped-drive detection failing open) is tracked as a follow-up issue rather than another CI cycle - the exposure is drive-letter-mapped shares on WMI-restricted images only, UNC stays lexically caught, and the failure mode is exactly the pre-#2198 behavior on that fringe. The elevated end-to-end observation gap is honestly stated in the PR body; the refusal logic itself is fully measured (ACL inheritance on this box) and mutation-tested, and the staged verify kit at C:\pm2187-verify gives Erik a one-click confirmation path.

Issue #2203, opened 2026-08-11
Lite has no persisted alerted-state memory, so every edge-triggered database state still repeats forever


erikdarlingdata:
Parity gap left open by #2166 / PR #2182, filed so it is tracked rather than living only in a changelog caveat. Surfaced again while fixing #2189.

## What is missing

#2182 made the four "somebody chose this" database states (OFFLINE, RESTORING, RECOVERING, STANDBY) edge-triggered: they alert on the transition and stay quiet until the state changes. The memory that makes that work is persisted per database in `config.database_state_expected.last_alerted_state` / `last_alerted_at` (V60), and the engine compares against it:

```csharp
var alreadyAnnounced =
    DatabaseStateTokens.RepeatsAreNoise(db.StateDesc)
    && string.Equals(db.LastAlertedState, db.StateDesc, StringComparison.OrdinalIgnoreCase);
```

Lite's `config_database_state_expected` has no such columns (`Lite/Database/Schema.cs`), `LiteAlertStateStore.SaveDatabaseStateAlertedAsync` / `ClearDatabaseStateAlertedAsync` are no-ops, and Lite's deviation read returns no `LastAlertedState`. So in Lite `alreadyAnnounced` is always false, every deviation reads as new, and behavior is exactly pre-#2166: **a database parked OFFLINE for a month produces an alert every cooldown, forever.** That is the original reported complaint, still live in Lite.

This is deliberate and documented (`AlertEngine.cs`: "A host that does not persist the memory (Lite today) reports empty here"), not an oversight. It just has no issue.

## Why it is worth doing rather than leaving

While fixing #2189 the review turned up a scenario where the missing memory turns a mild annoyance into the same flood shape #2189 was about:

1. A database is OFFLINE when first observed, so the seed learns OFFLINE as its baseline (nobody chose it, but it is a steady state, so it is legitimately learned).
2. Somebody brings it up for an hour of maintenance and re-parks it.
3. In Darling, `RepeatsAreNoise(OFFLINE)` is true and the persisted memory edge-triggers it: **one** alert.
4. In Lite, the same sequence alerts **every cooldown for as long as it stays parked**.

#2189 deliberately does NOT heal an auto-baselined OFFLINE row (healing it would be a worse bug - see that PR), so this path stays reachable in Lite until the memory lands.

## Shape of the fix

Mirror V60 in DuckDB and wire the three pieces Darling already has, which are the same three every new store-backed field needs:

[...]

PR #2204, opened 2026-08-11
Lite gets the persisted alerted-state memory (#2203, the #2166 Lite half)

erikdarlingdata:
#2182 made the four chosen database states edge-triggered, but only in Darling.
The memory an edge trigger needs is persisted per database, Lite's
config_database_state_expected had no such columns, and LiteAlertStateStore's two
methods were documented no-ops. So alreadyAnnounced was always false in Lite and a
database parked OFFLINE for a month still alerted every cooldown forever - the
complaint #2166 was filed about, still live in the SKU most likely to hit it.

Four pieces:

- Schema v53: last_alerted_state + last_alerted_at, nullable, plus an
  ADD COLUMN IF NOT EXISTS migration so existing stores get them. NULL means never
  announced, which is what a first observation, a fresh store and a recovered
  database all look like - all three must be free to alert.
- The two store methods write for real. Save is an UPDATE and never an upsert, the
  constraint #2166 established: an INSERT would have to invent expected_state (NOT
  NULL) from the state being alerted ON, so a database first observed SUSPECT would
  get SUSPECT as its accepted baseline, stop deviating, read as recovered while
  still corrupt, and never alert again. The seed refuses to baseline the integrity
  states for exactly that reason and this must not do it behind the seed's back.
- The deviation read returns the value, so the shared engine can use it here.
- A STORE-DERIVED recovered-clear running before the seed, which is the part worth
  reviewing. The engine also clears on the falling edge it witnesses, but that path
  is reachable only through an in-memory active set that empties on restart - so a
  restart landing between an alert and the recovery left the memory sticky forever
  and silently swallowed the next parking. Darling needed the same fix during
  #2182's review; porting the no-op without porting this would have shipped the bug
  I already fixed once. One sample at expected is enough where deviation needs two:
  clearing can only cause an extra alert, never a missed one, and a flap cannot
  exploit it because a flap never survives the two-sample test to alert at all.

Five tests against a real in-process DuckDB, no live gate: the value round-trips to
the deviation read; a recovered database has the memory cleared BY THE STORE with
nothing held in memory (the restart-gap invariant); (ignore) clears it too, since a
silenced database should not keep a memory outliving the silence; and the engine's
immediate clear forgets the announcement while leaving the baseline intact -
clearing the expected state instead would re-baseline and silence a real deviation.

Deliberately unchanged: the integrity states keep repeating on the cooldown.

One self-inflicted note: the schema comment initially used double quotes inside a
C# verbatim string and broke the literal. Caught by the compiler, and I then
scanned every file I touched for the same mistake.

## How was this tested?

Five tests in `DatabaseStateExpectedStoreTests`, which runs against a **real in-process DuckDB** with no live gate — so unlike the Darling live tests these run anywhere CI runs:

| test | what it pins |
|---|---|
| round-trip | the value reaches the deviation read, so the engine's `alreadyAnnounced` can be true in Lite |
| recovered-clear | the memory is cleared **by the store**, with nothing held in memory — the restart-gap invariant |
| `(ignore)` clear | a silenced database does not keep a memory that outlives the silence |
| immediate clear | the engine's falling-edge call forgets the announcement and leaves the baseline intact |

`Lite`, `Lite.Tests` and `PerformanceMonitor.Alerting` all build with zero errors.

**Limitation, stated plainly:** `Lite.Tests` is `net10.0-windows` and cannot run on this machine, so those five are CI-verified only. I have not exercised them locally and am not claiming otherwise.

## Coordination

Claimed on #2203 before starting, after checking for an existing `fix/2203-*` branch — there was none. This is deliberately scoped to the alerted-state memory and does **not** include the #2189 seed/heal work, which is #2202's and still open.


erikdarlingdata:
**Green and CLEAN on all five checks, and I am deliberately not merging it.** Review was genuine rather than skipped — 3m12s with real token spend and no workflow-validation skip, zero findings.

The reason to hold: this PR and **#2202** (held "per instruction") share **three files**:

```
CHANGELOG.md
Lite.Tests/DatabaseStateExpectedStoreTests.cs
Lite/Services/LocalDataService.DatabaseStates.cs
```

Merging mine first would put #2202 into conflict in all three and force a rebase onto work that is being held on purpose. #2202 is also the older PR and the one fixing a **live** production flood (still 721 alerts/24h on the box as of the last health check), so it has the stronger claim to going first.

**Proposed order: #2202, then I rebase this onto dev and it lands clean.** The overlap in `LocalDataService.DatabaseStates.cs` is adjacent rather than semantic — #2202 changes the seed's exclusion list and adds the heal; mine adds the alerted-state memory columns to the read and a store-derived clear before the seed — so the rebase should be mechanical, and if it isn't I would rather find that out in my rebase than in theirs.

Nothing needed from anyone until #2202 moves. I'll rebase and re-verify when it does.

erikdarlingdata:
Rebased onto dev now that #2202 has landed (`0dc238f4` -> `d6d1d164`). Three files touched, two real conflicts, and one of them turned out to be a genuine ordering question rather than a textual one.

**`LocalDataService.DatabaseStates.cs` — the clear now runs AFTER your #2189 heal, not before the seed.**

The heal *rewrites* `expected_state`. A database whose illegitimate `RESTORING` baseline was just healed to `ONLINE` is, from the clear's point of view, a database that has arrived back at its expected state — and its `last_alerted_state` is a memory of a deviation that no longer exists. Running the clear first would leave that dead memory sitting there for a cycle. Ordering against the *seed* is immaterial either way, since a row the seed just inserted has a NULL memory, which the clear skips. So the sequence is now seed -> heal -> clear -> prune -> read, and every write the read depends on has happened before it.

I convinced myself the one-cycle version was harmless before deciding to move it — in that window the stale memory reads `ONLINE` while `expected_state` is also `ONLINE`, so it can't suppress anything, because there's no deviation to suppress. Moving it is about not leaving a meaningless row behind, not about a live bug. Worth saying out loud in case you read the ordering differently.

**`CHANGELOG.md`** — kept both entries, yours first. I also updated two claims that your entry and mine jointly made stale: #2166's "Darling only for now; Lite keeps its current behavior until parity lands" and your #2189 entry's "which remains Darling-only until parity lands". Parity lands in this same release, so both now point at #2203 instead. Flagging that I edited a sentence inside your entry — revert it if you'd rather word it yourself.

**`DatabaseStateExpectedStoreTests.cs`** auto-merged cleanly, and I walked my five tests against the heal: all of them baseline at `ONLINE` and deviate from there, so `expected_state IN (NeverBaselinedSqlList)` is never true for them and the heal is a no-op in every one. Your tests don't touch the two new columns, so the clear is a no-op in yours.

Two other notes:

- `DatabaseStateTokens.Ignore` is now used unqualified in my block, matching how your `NeverBaselinedSqlList` reads. The deviation read a few lines down still has `'(ignore)'` as a bare literal — not touching it here, but it's the same one-spelling-each argument your class exists for, if you want it.
- Unrelated, and deliberately NOT folded into this PR: `CHANGELOG.md` on dev has four lines terminated `\r\r\n` (the #2062/#2063/#2064 block, around line 106). Any tool that round-trips the file normalizes them and produces a phantom four-line diff — it did exactly that to me here, and I rebuilt my commit from dev's bytes to keep it out. Someone should fix those four lines on purpose sometime; it shouldn't ride along inside a feature PR.

Lite schema stays at **v53** — #2202 added no rung, so the collision I flagged didn't happen. Re-verifying CI now.

erikdarlingdata:
Merged (`2e38447b`), #2203 closed explicitly since the PR targeted dev.

On the bot's one observation — Darling's `HealDatabaseStateBaselineToOnlineSql` nulls the memory columns inline in the same UPDATE, Lite's heal doesn't — the split is deliberate, and worth writing down because it's a coordination artifact rather than a style preference. The heal is #2189's statement, which merged about four minutes before I rebased on top of it; folding two columns into somebody else's just-landed UPDATE to save a statement isn't a trade I wanted to make silently. The end state is identical for the reason the bot traced: nothing reads between heal and clear, and heal only fires when the effective state is already `ONLINE`, which is exactly the condition the clear tests.

The one way the split is genuinely weaker: it depends on ordering. Move the clear back above the heal and Darling still forgets the memory while Lite doesn't. That's an argument for the inline version if you (@ the #2189 author) want to fold it in — I'd rather you make that call on your own statement than have me quietly widen it. Either way the ordering is now load-bearing enough that I documented it in the comment above the clear, so a future reorder has to argue with it rather than trip over it.

Issue #1760, opened 2026-07-27
Flaky: CompressionJobSelfHeal detection test false-fails on the nightly despite its settle-wait


erikdarlingdata:
`TimescaleSupportTests.CompressionJobSelfHeal_DetectionQueryValid_AndRearmSucceeds_AgainstDevPostgres` intermittently fails on the nightly's live-Postgres job. It has already been hardened once for this exact class and failed anyway, so the existing wait is insufficient rather than absent.

## Evidence

Nightly run `30234946444`, dev at `8849ce97`:

```
attempt 1   Darling PostgreSQL tests   FAILURE
  Failed  TimescaleSupportTests.CompressionJobSelfHeal_DetectionQueryValid_AndRearmSucceeds_AgainstDevPostgres [54 ms]
  Assert.DoesNotContain() Failure: Filter matched in collection

attempt 2 (rerun, NO code change)   Darling PostgreSQL tests   SUCCESS
```

Green on a rerun of the identical tree is what makes this a flake rather than a defect. Everything else in that run passed, including both gated store-upgrade tests.

## The assertion and why it is timing-dependent

```csharp
using (var arm = new NpgsqlCommand("SELECT alter_job($1::integer, next_start => now() + interval '1 hour')", connection))
{ ...; await arm.ExecuteNonQueryAsync(ct); }
await WaitForSettledNextStartAsync(connection, jobId, ct);

var healthy = await TimescaleSupport.ReadStuckCompressionJobsAsync(connection, DateTime.UtcNow, null, ct);
Assert.DoesNotContain(healthy, s => s.JobId == jobId);
```

The detection query reported a job as stuck that the test had just deliberately settled into a healthy state. The test's own comment documents the class:

> ...asserting immediately after `ApplyCompressionPolicy` raced that window and intermittently false-failed on a slow CI runner. Deterministically settle the job into the healthy state the assertion is actually about... then wait for the catalog to reflect a non-`-infinity` `next_start`.

So `WaitForSettledNextStartAsync` was added precisely to close this, and the failure happened anyway. Two candidates, and they want distinguishing rather than guessing:

1. **The wait returns before the state it is waiting for is actually visible** — e.g. it polls only `next_start` while `ReadStuckCompressionJobsAsync` also keys off another column (`last_run_status`, `last_finish`, a `job_stats` row that has not materialised yet). Then the wait is satisfied while the detection query still sees a stuck-looking job.
[...]

PR #2207, opened 2026-08-11
Assert compression-job health from the settled snapshot (#2206)

### LINKED ISSUE #2206: [BUG] Compression self-heal live test asserts health from a re-read, not the observation its wait validated
### What happened

Tonight's nightly on dev ([run 31534394223](https://github.com/erikdarlingdata/PerformanceMonitor/actions/runs/31534394223)) failed with **1 failure out of 4,507**:

```
Darling.Tests.TimescaleSupportTests.CompressionJobSelfHeal_DetectionQueryValid_AndRearmSucceeds_AgainstDevPostgres [FAIL]
  Assert.DoesNotContain() Failure: Filter matched in collection
  TimescaleSupportTests.cs:1559
```

The 16:05 nightly on the same day was green, and **nothing touching Timescale or compression has landed since it** — the four merges in between (#2202, #2194, #2204, #2195) are database-state baselines, managed-Postgres error decoding, Lite's DuckDB alert memory, and per-database state pruning. This is not a regression; it is the test's own race, and this is its third appearance in the same class.

### Root cause

`WaitUntilDetectorReportsHealthyAsync` returns the moment ONE poll finds the job unflagged. Leg (1) then issues two further queries and asserts health from a **fresh read**:

```csharp
await WaitUntilDetectorReportsHealthyAsync(connection, jobId, ct);   // proves: healthy at instant T
var observed = await ReadObservedJobIdsAsync(connection, ct);
Assert.Contains(jobId, observed);
var healthy = await TimescaleSupport.ReadStuckCompressionJobsAsync(...);  // a DIFFERENT observation
Assert.DoesNotContain(healthy, s => s.JobId == jobId);                   // <-- line 1559
```

So the assertion is not the observation the wait validated. The helper's own doc comment claims "settled-according-to-the-wait IS settled-according-to-the-assertion" — and that is true only for leg (3), which ends with the wait and asserts nothing afterward. Leg (1) breaks it by re-reading.

Anything that moves the job out of the healthy state in that gap fails the assertion, and the test itself documents such a state: from scheduler pickup to completion, `job_stats` reads `next_start = -infinity` with status Running, which the detector flags and is **correct** to flag (leg 3's comment says exactly this).

Note the previous two fixes each closed a narrower version of this. #1760 polled `next_start <> '-infinity'` directly — one of the detector's two arms, and the very value the caller had just written, so it was satisfied on its first poll and waited for nothing. The follow-up replaced that with polling the detector, which made the *first* observation deterministic and left the second one racing. The guarantee stops at the moment the helper returns, and the caller kept reaching past it.

### Fix

Have the helper return the flagged list from the poll that satisfied it, and have leg (1) assert against that snapshot instead of re-querying. Then settled-according-to-the-wait and settled-according-to-the-assertion are the same observation by construction, which is what the contract claimed.

The observability assertion (`Assert.Contains(jobId, observed)`) can keep its own read and is safe there: `StuckCompressionJobsSql` filters on `proc_name` alone, so it returns every compression job whatever state it is in. Flagging is the C# predicate applied on top, and that is the only part that moves.

erikdarlingdata:
Fixes #2206.

## The failure

Tonight's nightly on dev ([run 31534394223](https://github.com/erikdarlingdata/PerformanceMonitor/actions/runs/31534394223)) went red on **one test out of 4,507**:

```
CompressionJobSelfHeal_DetectionQueryValid_AndRearmSucceeds_AgainstDevPostgres [FAIL]
  Assert.DoesNotContain() Failure: Filter matched in collection
  TimescaleSupportTests.cs:1559
```

Not a regression from the four merges since the last green nightly (#2202, #2194, #2204, #2195) — none of them touch Timescale or compression. I checked that before looking at the test.

## Why it raced

`WaitUntilDetectorReportsHealthyAsync` returns the moment one poll finds the job unflagged. Leg (1) then ran two more queries and asserted health from a **fresh read**, so the thing asserted was never the thing the wait validated. The test documents a state that lands squarely in that gap: from scheduler pickup to completion, `job_stats` reads `next_start = -infinity` with status Running — and the detector flags that, correctly.

The helper's doc comment claims "settled-according-to-the-wait IS settled-according-to-the-assertion." True for leg (3), which ends with the wait. Leg (1) broke it by re-reading.

## The fix

The helper returns the flagged list from the poll that satisfied it, and the assertion reads that snapshot. The two statements become the same observation by construction.

The observability assertion keeps its own read, and that one genuinely cannot race: `StuckCompressionJobsSql` filters on `proc_name` alone, so it returns every compression job whatever state it is in. Flagging is the C# predicate applied on top of those rows, and that is the only moving part. Keeping it as a separate read also preserves what it exists to prove — that the production SQL compiles and sees this job, so `DoesNotContain` can't pass vacuously against a query that was swallowed by the failure-isolated wrapper.

## Worth flagging for review

**This is the third fix in this class, so the pattern matters more than the patch.** #1760 polled `next_start <> '-infinity'` — one of the detector's two arms, and the very value the caller had just written, so it was satisfied on its first poll and waited for nothing. The follow-up replaced that with polling the detector, which made the first observation deterministic and left the second one racing. Both times the helper's guarantee stopped at the moment it returned and the caller reached past it. That's why the fix here is to make the guarantee *consumable* rather than to widen a window or add a retry: there is no longer an "after" for the caller to read.

I could not run this test locally to prove the fix — it is `net10.0-windows` and gated on `DARLING_TEST_PG`, so a live TimescaleDB fixture is required and CI is the only place it runs. What I verified locally is that `Darling.Tests` compiles clean. The real proof is this PR's `Darling PostgreSQL tests` check plus the next nightly, and I'll watch both.

### DIFF
--- a/Darling/Darling.Tests/TimescaleSupportTests.cs
+++ b/Darling/Darling.Tests/TimescaleSupportTests.cs
@@ -1570,6 +1570,11 @@ LIMIT 1", connection))
         var observed = await ReadObservedJobIdsAsync(connection, ct);
         Assert.Contains(jobId, observed);
 
+        /* Deliberately tautological, and kept for what it documents rather than what it can catch: `healthy` is
+           the snapshot the wait already found clean, so this cannot fail today. It states the property leg (1)
+           exists to assert, at the place a reader looks for it, and it fails loudly if the helper is ever
+           changed to return something other than the satisfying poll's own result. The load-bearing check is
+           the wait's bounded loop, which fails carrying the detector's reason string. */
         Assert.DoesNotContain(healthy, s => s.JobId == jobId);
 
         /* (2) The #1586 REGRESSION GUARD: the production re-arm runs the real alter_job against TimescaleDB and
--- a/Darling/Darling.Tests/TimescaleSupportTests.cs
+++ b/Darling/Darling.Tests/TimescaleSupportTests.cs
@@ -1538,24 +1538,38 @@ LIMIT 1", connection))
            CORRECT to flag -infinity — so asserting immediately after ApplyCompressionPolicy raced that window
            and intermittently false-failed on a slow CI runner. Deterministically settle the job into the
            healthy state the assertion is actually about: give it a real FUTURE next_start (via the same
-           alter_job the self-heal uses), then wait for the catalog to reflect a non-(-infinity) next_start. */
+           alter_job the self-heal uses), then wait for the detector itself to report it healthy.
+
+           The wait's RESULT is what the assertion below reads, and that is the whole point rather than a
+           convenience: a wait that only proves "healthy at some instant", followed by a fresh read, asserts
+           against a DIFFERENT observation than the one it validated, and the job can leave the healthy state
+           in the gap between them (the scheduler picking it up reads next_start = -infinity with status
+           Running mid-run — see leg 3). That gap is this test's third flake in the same class, after #1760
+           polled a copy of one detector arm and after the wait was introduced to close it; it survived
+           because the helper's guarantee stops at the moment it returns. Consuming the returned snapshot
+           makes settled-according-to-the-wait and settled-according-to-the-assertion the same observation
+           by construction, which is what the helper's contract claimed all along. */
         using (var arm = new NpgsqlCommand("SELECT alter_job($1::integer, next_start => now() + interval '1 hour')", connection))
         {
             arm.Parameters.Add(new NpgsqlParameter { Value = jobId });
             await arm.ExecuteNonQueryAsync(ct);
         }
-        await WaitUntilDetectorReportsHealthyAsync(connection, jobId, ct);
+        var healthy = await WaitUntilDetectorReportsHealthyAsync(connection, jobId, ct);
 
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Good catch, and it was the right thing to flag — addressed in `b6885321`.

You're right that the assertion is now tautological, and I'd rather say so at the line than let a reader mistake it for independent verification. The comment now states three things: it can't fail today, it's kept because it names the property leg (1) exists to assert at the place a reader looks for it, and it becomes load-bearing again if the helper is ever changed to return something other than the satisfying poll's own result. The check actually doing work is the wait's bounded loop, which fails carrying the detector's own reason string.

On the wider point behind your note: I considered keeping a second independent read so the assertion stayed meaningful, and rejected it, because that IS the bug. Any read after the wait is a new observation the wait never validated, and the state genuinely moves — the scheduler picking the job up reads `next_start = -infinity` with status Running mid-run. There's no version of "independent verification after settling" that isn't the same race with a smaller window. The verification has to happen inside the settle loop or not at all, which is why the loop is where the teeth are now.

PR #2209, opened 2026-08-11
Decide "first run" vs "bootstrap already failed" from the store's own files (#2197)

### LINKED ISSUE #2197: [BUG] The missing-credential CLI message always reads as a first run, even after a bootstrap has already failed
Split out of #2186 (which came out of @jovon44's report in #2185), and deliberately not fixed there because it is a different file and a different surface.

### Component

Darling

### Describe the Bug

When a managed bootstrap fails, the operator's *last* message is usually not the bootstrap error - it is whatever CLI verb they run next. Those verbs say:

```
The 'admin' role credential (C:\ProgramData\PerformanceMonitorDarling\pg-admin-credential.dpapi) does not exist yet. Start the PerformanceMonitor Darling service once so its first run initializes the store, then re-run this command.
```

That is correct advice for a genuine first run, and a dead end for the case that actually produces it in the field: the service HAS been started, its bootstrap failed, and starting it again will produce the same failure. In #2185 this is the message the reporter led with, and it is what sent them looking at `darling.json` - which was never the problem. #2186 fixed the upstream half (the bootstrap now explains a loader failure instead of printing a raw Win32 status); this is the half that decides where the operator looks.

The same wording appears in several verbs in `Darling/PerformanceMonitor.Darling.Service/DarlingCliCommands.cs` - the per-role message around line 486, and the managed-store-credential variants (roughly lines 2451, 2950, 3307, 3595).

### Expected Behavior

Keep the first-run advice, and add the branch it is missing: if the service has already run, the reason the store was never initialized is in its log, and that is where to look. Naming the log path would be better than naming the log. Cheap version - one extra sentence in the shared message. Better version - notice that the service log exists (or that the data directory exists without a credential beside it) and lead with the "already tried, look at the log" phrasing in that case, since the two situations want opposite advice.

### Actual Behavior

Every missing-credential message reads as a first run, including when the operator's problem is a bootstrap that has already failed repeatedly.

### Additional Context

Not fixed inside #2186 because `DarlingCliCommands.cs` was outside that change's blast radius and under concurrent edit at the time. Nothing about it is hard; it is a wording change plus, optionally, one existence check.

erikdarlingdata:
Closes #2197.

## The report

When a managed bootstrap fails, the operator's *last* message is rarely the bootstrap error - it is whatever CLI verb they run next. In #2185 that message is what @jovon44 led with, and it is what sent them to `darling.json`, which was never the fault. #2194 fixed the upstream half (a loader status is now decoded instead of printed as a bare Win32 number); this is the half that decides where the operator looks.

## Verifying the issue's claims first

Both held, and one was worth checking rather than assuming.

The issue lists the per-role message at `DarlingCliCommands.cs:486` and the managed-store variants at roughly 2451, 2950, 3307 and 3595 - all five confirmed on dev, all five independent string literals. What the issue could not have known is the mechanism that makes the sharper branch *reachable* for the role-credential verb: `InitializeClusterAsync` writes the store's own `pg-credential.dpapi` **before** it runs initdb (`DarlingManagedPostgres.cs:874`, ahead of `RunToolAsync` at 890), precisely so a crash after a successful initdb cannot orphan a cluster nobody has the password for. So in @jovon44's exact failure - initdb killed in the loader - that one file is on disk and the role credentials are not, which is a decisive on-disk fingerprint of "already tried".

Grepping for the wording also turned up a **sixth** copy the issue does not name: `PerformanceMonitor.Darling.Viewer/ViewerSettings.cs:303`, the managed-mode derivation, whose text the main window renders. Same absence, same operator, same dead end. It is fixed here too rather than left to drift.

## Before / after, verbatim

From live runs of the built binaries against scratch data directories - `origin/dev` (`ab024191`) built in a throwaway worktree for the before, this branch for the after. Nothing touched a real install or DARLING01. (The console renders the messages' em dashes as `-` in this code page; the source is unchanged in that respect.)

**Before (dev) - a genuine first run, `--print-viewer-connection`:**

```
The 'viewer' role credential (C:\...\state-firstrun\store\pg-viewer-credential.dpapi) does not exist yet. Start the PerformanceMonitor Darling service once so its first run provisions the least-privilege roles and their credentials, then re-run this command.
```

**Before (dev) - the #2185 state, the store credential already on disk from a bootstrap whose initdb died:**

```
The 'viewer' role credential (C:\...\state-failed\store\pg-viewer-credential.dpapi) does not exist yet. Start the PerformanceMonitor Darling service once so its first run provisions the least-privilege roles and their credentials, then re-run this command.
```

Byte-identical to the first run apart from the path. That is the bug: the one state where "start the service once" is guaranteed not to help gets exactly that advice.

**After - genuine first run (the advice is unchanged, plus the sentence it was missing):**

```
The 'viewer' role credential (C:\...\state-firstrun\store\pg-viewer-credential.dpapi) does not exist yet. Start the PerformanceMonitor Darling service once so its first run provisions the least-privilege roles and their credentials, then re-run this command. If you have ALREADY started it, its first run never got this far and starting it again will not either - the reason is in the service log (C:\ProgramData\PerformanceMonitorDarling\logs\darling-service_yyyyMMdd.log), not in darling.json.
```

**After - the #2185 state:**

```
The 'viewer' role credential (C:\...\state-failed\store\pg-viewer-credential.dpapi) does not exist, and this is NOT a first run: C:\...\state-failed\store\pg-credential.dpapi is already there, and the service writes that one immediately before it runs initdb. The service has already run against this store and its bootstrap stopped before it got this far, so starting it again is not the fix. Read the newest service log (C:\ProgramData\PerformanceMonitorDarling\logs\darling-service_yyyyMMdd.log) and work the FIRST error in it - a bundled Postgres tool that Windows killed is decoded there in words rather than left as a bare exit code. Nothing in darling.json produces this.
```

**Before / after - the store-credential family, `--enable-mcp` against a cluster whose credential is gone:**

```
before: The managed store credential does not exist yet - start the PerformanceMonitor Darling service once so its first run initializes the store, then re-run this command.

after:  The managed store credential (C:\...\state-mcp-failed\store\pg-credential.dpapi) does not exist, and this is NOT a first run: the cluster in C:\...\state-mcp-failed\store\pg is already initialized. The service has already run against this store and its bootstrap stopped before it got this far, so starting it again is not the fix. Read the newest service log (...) and work the FIRST error in it - a bundled Postgres tool that Windows killed is decoded there in words rather than left as a bare exit code. Nothing in darling.json produces this.
```

That one is worse than misleading, it is a closed loop: `EnsureRunningAsync` only runs initdb when the data directory has no `PG_VERSION`, so on a cluster that already exists no number of restarts ever produces the missing credential. These four verbs also never named the file's path before; they do now.

## What changed

`DarlingStoreBootstrapEvidence` answers one question - has a bootstrap been ATTEMPTED against *this* store - and builds the message in whichever of the two voices the answer earns. Evidence, in the order it settles the question: an initialized cluster (`PG_VERSION`), `pg.log`, the store's own credential (with the reason that file is decisive quoted into the message), a role credential, then a non-empty data directory.

**What it deliberately does not look at is the load-bearing part.** The tempting signal is the service's log directory: it exists, therefore the service has run. But that is machine-global while the question is about one store, so a box that has monitored happily for a year would tell somebody configuring a *second* store that their bootstrap had failed - this same defect pointed somewhere new. Every path read here sits under or beside `postgres.dataDirectory`. The cost is that a bootstrap dying before it writes anything (a missing `pg-runtime.zip`, say) still reads as a first run, which is why the first-run branch carries its own "if you have ALREADY started it" sentence rather than being left as it was. An empty data directory an operator pre-created is not evidence either, for the same reason.

The lead clause is unchanged in both voices (`<subject> does not exist`), because the old wording is already searchable in field reports and the issue tracker - the branch changes what follows it, never what somebody pastes into a search box.

The Viewer half mirrors it in-file rather than taking a project reference: `ViewerSettings.cs` already duplicates the DPAPI entropy and the role/credential names under a documented sliver rule, so the three new file names join that rule and are pinned against the service's constants by a test.

## Tests

Watched red first against an inert seam (the evidence probe returning today's answer, "no evidence", on both sides): **9 failed, 27 passed**, then green. The 27 that passed under the seam are the ones asserting today's *correct* behavior survives - the first-run voice, the two non-evidence cases, and the wiring pin - so the red was the split, not the harness. The test project was rebuilt for the red run and the green run both; a stale test DLL is exactly what would have made this look green from the start.

- `DarlingMissingCredentialMessageTests` (in the CLI test family): the probe including the two ways it must NOT fire, both message voices, the shared searchable lead, and end-to-end runs of `--print-viewer-connection` and `--enable-mcp` in both states - because a correct builder nothing calls is precisely what the sibling #1738 defect already was. Plus a source-parsed wiring pin: neither old sentence survives anywhere in `DarlingCliCommands.cs`, and the call counts are 1 + 4.
- `ViewerSettingsTests`: the viewer's own split, and the sliver pin that walks each service-side file name through the viewer's probe - a rename on the service side would otherwise turn the viewer's sharper branch off silently.

Full `Darling.Tests`: **4297 passed, 0 failed**, 242 skipped (all gated). A `--no-incremental` rebuild of the Service and Viewer projects reports **1 warning each**, the pre-existing `AbandonableStep` CA1068 from Common tracked as #2193 - this branch adds zero.

## Not in scope

The installer half of #2185 is #2187; the loader-status decode is #2194.


erikdarlingdata:
CI green on head `e42bfdc8` (`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` - confirmed right after opening, since a conflicting PR silently never runs its workflows at all). Every check run below is attached to that SHA, not to an older push:

| check | result | duration |
|---|---|---|
| build | pass | 4m15s |
| review | pass | 4m0s |
| Darling PostgreSQL tests | pass | 3m19s |
| Darling Linux build | pass | 1m35s |
| check-branches | pass | 3s |

Local, for the record: full `Darling.Tests` **4297 passed, 0 failed**, 242 skipped; `--no-incremental` rebuilds of the Service and Viewer projects each report 1 warning, the pre-existing `AbandonableStep` CA1068 from Common tracked as #2193 - zero added by this branch. The CHANGELOG insert was scripted byte-precisely: `git diff --stat` is exactly 2 insertions, and the committed blob carries the same 6 bare CRs and 0 bare LFs as the one on dev.

Not merged, and #2197 left open, per instructions.

erikdarlingdata:
Took the review's wording-drift note directly on the branch: the Viewer's NOT-a-first-run copy now carries the decoded-in-words clause - the reviewer is right that the operator reading the Viewer's message is precisely the one who benefits from knowing the log decodes the failure rather than dumping a bare exit code. The two pre-existing warnings on the rebuild are dev's #2193 pair in untouched files; 18/18 on the affected test families with a rebuilt DLL. Merging off the fresh rollup.

Issue #1140, opened 2026-06-17
[FEATURE] Add a stable, machine‑readable dedup fingerprint (and involved‑object list) to alert payloads - esp. deadlocks & blocking


gotqn:
### Which component(s) does this affect?

- [x] Full Dashboard
- [x] Lite
- [ ] SQL collection scripts
- [ ] Installer
- [ ] Documentation

### Problem Statement

Downstream automation (Logic Apps → Azure DevOps tickets, SIEMs) needs to tell distinct incidents apart and collapse recurrences of the same one. Today the webhook/alert payload exposes server + metric + summarized facts, but no stable per‑incident key - so consumers must reconstruct one from truncated query text / volatile wait‑resource IDs, which is fragile (and Logic Apps can't even hash). 

### Proposed Solution

Please emit a precomputed fingerprint.

1. A dedupKey/fingerprint field: a stable hash the app computes server‑side, where it has the full data:

    - Deadlocks / Blocking: hash of the sorted set of fully‑qualified objects involved (database.schema.object), across all participating sessions/databases. (Don't use raw query text — it varies by literals; don't use bare hobt/page IDs — they change on index rebuild.)
    - Long‑Running Query: the query_hash.
    - Long‑Running Job: the job name (+ instance).
    - Volume Free Space: the drive/volume.
  
2. A human‑readable involvedObjects list (e.g. SalesDB.dbo.Orders, SalesDB.dbo.LineItems) — for the ticket body. For multi‑database deadlocks, list all databases/objects (one incident, not split).

### Use Case

I have build a Logic app in order to create tickets in Azure DevOps Board. If a deadlock in database X and specific objects occurs, I want to create a ticket. If the deadlock occurs again, I want to increase a custom field counter, not to create separate ticket. If a developer opens the Board, it will be easier to focus on issues with high occurrences.

<img width="1420" height="1124" alt="Image" src="https://github.com/user-attachments/assets/41769085-1e59-4fbc-8a75-89f71d2b4189" />

### Additional Context

**Why a precomputed hash and not just names:** consumers can dedup on one stable field without parsing/normalizing/sorting; names alone are ambiguous and order‑sensitive, and several automation platforms have no hashing primitive.

**Scope**: deadlocks + blocking are the priority; query/job/disk keys are a nice follow‑on. Keep the existing LowDiskAlertGate/rolling‑count gating - this is about identifying which incident, not whether to fire.

gotqn:
[...]

PR #2213, opened 2026-08-12
Darling monitors PostgreSQL and Amazon Aurora PostgreSQL

erikdarlingdata:
## What does this PR do?

Teaches Darling to monitor **PostgreSQL and Amazon Aurora PostgreSQL** alongside SQL Server. A monitored
server declares `"engine": "postgres"` and is collected by seven PostgreSQL collectors instead of the T-SQL
ones, into the same store, on the same naive-UTC contract and the same `server_id` identity. A mixed fleet
is one store, one viewer, one MCP endpoint — nothing is partitioned by engine.

Every collector definition declares the engine it targets and is never dispatched against the other one, so
a PostgreSQL target is never sent T-SQL and a SQL Server target never sees `pg_stat_statements`. A store
monitoring only SQL Server gains seven empty tables and nothing else changes.

**What gets collected**, and why each earns its place:

| Collector | Source | Cadence | Why |
|---|---|---|---|
| `pg_wait_stats` | `aurora_stat_system_waits()` | 1 min | **Aurora only.** Core PostgreSQL has no cumulative wait counters in ANY version — `pg_stat_activity.wait_event` is an instantaneous sample, so there is no `dm_os_wait_stats` equivalent to read elsewhere |
| `pg_statement_stats` | `aurora_stat_statements()` | 1 min | **Aurora only.** Per-query-shape totals, plus the storage-vs-cache I/O split and per-statement peak memory that stock PostgreSQL does not expose |
| `pg_wraparound_stats` | `pg_database`, `pg_class` | 5 min | XID and MultiXact freeze headroom. Highest-consequence signal PostgreSQL has, no SQL Server counterpart: run out and the server stops accepting writes |
| `pg_xmin_horizon` | `pg_stat_activity`, `pg_replication_slots`, `pg_stat_replication`, `pg_prepared_xacts` | 1 min | Why vacuum reclaims nothing. Five unrelated causes, identical symptom, completely different fixes — so it attributes the specific holder rather than reporting a number |
| `pg_replication_slots` | `pg_replication_slots` | 1 min | An abandoned slot retains WAL without bound AND pins the vacuum horizon. Two independent ways to take a server down |
| `pg_autovacuum_stats` | `pg_stat_user_tables`, `pg_class` | 60 min | **Writers only**, per database. Stores each table's OWN computed threshold beside its dead-tuple count, honouring per-table `reloptions` — without the threshold the count is not actionable |
| `pg_io_stats` | `pg_stat_io` | 1 min | I/O by `(backend_type, object, context)` rather than by file. PostgreSQL 16+, valid on a standby |

Three of the seven are outage predictors rather than performance metrics, and they **alert** — graded against
the target's own settings (wraparound against that cluster's `autovacuum_freeze_max_age`, not a constant),
delivered through the same deliverer, history and mute rules as every SQL Server alert. That is deliberate:
PostgreSQL's most damaging failures are quiet, slow, fully predictable days ahead, and nothing in the engine
raises its hand about them.

**Also here:** the engine-execution seam (`ITargetProvider`), per-database fan-out for PostgreSQL, fault
classification so an operator-actionable condition records a non-fatal skip with an explanation instead of
logging `ERROR` every cycle, and a read surface + MCP tool per collector.

Permissions are one `GRANT pg_monitor`. Nothing is created on the monitored instance — no Extended Events
sessions to provision, no server setting to bootstrap.

### Two defects found while writing the runbook, worth calling out for review

1. **`config.config_monitored_servers` had no `engine` or `port` column.** That registry is authoritative for
   the server list once seeded, so a PostgreSQL entry round-tripped as `"sqlserver"` (the property default)
   and the service opened a `SqlConnection` to port 5432 — on the FIRST start, since the seed is immediately
   followed by the load that replaces the file's list. Fixed by V68 plus the seed/read paths and
   `add_servers`. A new property-driven test fails when any round-trip-critical `MonitoredServer` property
   lacks a column.
2. **`MapProbedSchemaVersion`'s newest arm was V60.** `RequiredStoreSchemaVersion` is
   `StorageVersion.SchemaVersion` and the connect-time gate refuses a store below it, so a fully-migrated
   store would have been rejected as "older than this viewer". V68 gets an arm.

Also fixed seven stale literal pins in test files that the PostgreSQL rungs had been silently breaking since
V60, four of which are now asserted as the invariant the test's own name states rather than a literal.

- [ ] Lite
- [x] Darling
- [ ] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [x] Documentation

`PerformanceMonitor.Collectors` is shared with Lite, so Lite rebuilds — but no Lite behaviour changes. Lite
has no PostgreSQL target, and the alerting deliberately rides a separate `IPostgresAlertReadAdapter`
consulted only for PostgreSQL targets rather than extending the shared `IAlertReadAdapter`, precisely so Lite
is not left implementing three methods that can only return empty. `AlertEngine` is untouched.

## How was this tested?

**Please read this section rather than the checklist — the honest answer is more specific than the boxes.**

Verified against **live Amazon Aurora PostgreSQL 16.11 and 17.7** (stage):

- Every collector's generated SQL executes on both majors, returning the expected shape and real data
  (`pg_stat_io`: 37 rows on 16.11, 25 on 17.7, carrying Aurora-specific enum values the community docs do not
  list, e.g. `aurora cache receiver process`).
- Every naive-declared column comes back carrying **no timezone**. This caught a real defect:
  `pg_replication_slots.inactive_since` and the four autovacuum timestamps are `timestamptz`, and Npgsql 10
  refuses to write a `DateTime(Kind=Utc)` into `timestamp without time zone`, so it would have failed at
  COPY time in production. Every probed instance is `TimeZone=UTC`, which is exactly what hides the wrong
  cast form.
- Aurora's `pg_stat_io` write side really is NULL, not zero — backends there do not write data files. Every
  counter column is nullable on purpose and the read reports whether writes are TRACKED, so absent writes
  cannot be misread as no writes.
- Each MCP reader's SQL executed on real Aurora with synthetic `VALUES` tables substituted for the store
  tables, asserting the arithmetic: counter-reset clamping, NULL handling, single-sample → 0, and the ratio
  ranking (a 10k-row table at 50× its threshold outranks a 50M-row table at 2.5×).
- `pg_autovacuum_stats` is gated off standbys because `pg_stat_user_tables` reads fine on an Aurora reader and
  reports **all zeros** — measured on 17.7, same cluster and tables: writer 13,654,458 dead tuples, reader 0.
  Ungated, a replica target would have produced a confident report of perfect autovacuum health.

Every migration rung was diffed against the ladder generator, since V1 is generated from the collector
catalog and a hand-written rung must match it column-for-column.

**Not tested, and I would rather say so than have it found in review:**

- **`Darling.Tests` and `Lite.Tests` have never executed for this branch.** They target `net10.0-windows` and
  reference the WPF Viewer, so they compile on macOS and cannot run there. ~200 new xUnit assertions are
  unexecuted. Local substitutes were used instead — the ladder-generator diff, standalone console harnesses
  referencing the real projects, live SQL probes, and source-scanning reproductions of the reflection-based
  MCP parity test and the `DocCommentHygieneTests` stacked-summary scan (the latter caught a real violation
  in this branch). CI is the first real run.
- **The service has never been pointed at a PostgreSQL target.** Nothing has done connect probe → collector
  dispatch → binary COPY into a store → MCP read back out. The queries are proven; the wiring around them is
  not. `docs/postgres-first-target-runbook.md` is the procedure for closing that gap, with a proof point at
  every step, and it says plainly that it has not been executed.

`dotnet build -c Debug` → **0 errors, 16 warnings, all pre-existing on dev** (the one warning this branch
introduced, a CA1859 on a new field, is fixed).

## Checklist

- [ ] My code builds with zero warnings (`dotnet build -c Debug`) — **0 errors and zero NEW warnings; 16
      pre-existing warnings remain, unchanged from dev.** Not ticking a box I cannot honestly tick.
- [x] I have tested my changes against at least one SQL Server version — *read as PostgreSQL here: live
      Aurora 16.11 and 17.7. No SQL Server behaviour changes; the shared-catalog changes are engine-gated.*


erikdarlingdata:
Not touching this PR — it's yours — but flagging something about its review before anyone merges on the strength of a green check.

**The review ran for real and posted nothing.** Run 31582450960: 09:20:25 → 09:26:04, conclusion `success`, **no** `workflow validation failed` skip signature, and `total_cost_usd: 3.53` of genuine token spend. Then: `/pulls/2213/reviews` → 0, `/pulls/2213/comments` → 0, and no issue comments on the thread at all.

So this is not the green-but-skipped case we already know about (where the action self-skips because the workflow differs from the default branch and reports success anyway). This is a new variant: the review **executed**, spent five and a half minutes and three and a half dollars on 14,666 changed lines, and its output did not land. Either it genuinely found nothing to say across eight schema rungs and a new engine target, or the posting step failed silently.

Worth knowing because the discriminator we've been using — skip signature plus token evidence — says "genuine review" here, and it would have been reasonable to read that as reviewed. The check that actually catches it is cruder: **did any comment appear?**

I'd want the review output before this lands, given what's in it: `StorageVersion.SchemaVersion` 60 → 68, `PgSchemaGenerator`, and both workflow files. Not least because the rung range matters to me directly — my #2211 map migration takes V69 behind it, and a rung numbered inside your range would be silently skipped rather than rejected, which for the map table means every read returns empty and the cutover reads as "plan not yet collected" while holding nothing.

Suggest re-running the review workflow on this head and checking a comment actually appears. Happy to leave that to you.

erikdarlingdata:
## Full review: three dimensions plus the first live run of the feature

Repo-side session. Review ran as three parallel deep passes (engine seam/dispatch, storage/migrations V61-V68/data contract, collectors' SQL/alerting/read surfaces - every finding below was verified in source at head f4dfe261 before inclusion) **plus the thing the PR body says nobody had done: the service was pointed at a real PostgreSQL target.** Local rig, this branch's build: PostgreSQL 18.4 store (BYO, fresh), stock PostgreSQL 18.4 target on a second cluster, `darling_monitor` with `pg_monitor`, seeded activity, service run in console mode through multiple sweep cycles.

**Verdict: the foundation is genuinely strong and the feature is currently non-functional against a live target.** The defects cluster in exactly the layer the honesty section predicted - "the queries are proven; the wiring around them is not."

---

### The live run, first ever

**What passed, with the runbook's own proof points:**
- Pre-flight is excellent: `[PASS] e2e-pg-target: PostgreSQL 18 (server_version_num 180004), writer, not Aurora - 5 of 7 PostgreSQL collectors apply (skipped: pg_statement_stats, pg_wait_stats)` - exactly the promised shape, Aurora detection correct on stock PG.
- Store migration clean: `Postgres store ready (schema v68, 67 migration(s) applied)`, all 20 CAGGs ready, collection loop started.
- The per-database path (the ONE engine-neutral branch) works end to end: `pg_autovacuum_stats => 0 rows (sql:14ms, pg:0ms)`, SUCCESS - and the zero is HONEST (the target's own autovacuum had cleaned my seeded 25k dead tuples before the sweep; `n_dead_tup=0` confirmed on the target).

**What failed, verbatim:**

```
collector_name         | status  | count     (260s window)
pg_autovacuum_stats    | SUCCESS |     1
pg_io_stats            | ERROR   |     2
pg_replication_slots   | ERROR   |     2
pg_statement_stats     | SUCCESS |     2   <- Aurora-only, this target is not Aurora
pg_wait_stats          | SUCCESS |     2   <- same: fake success, no skip, no explanation
pg_wraparound_stats    | ERROR   |     1
pg_xmin_horizon        | ERROR   |     2
database_config        | SUCCESS |     2   <- T-SQL on-load collectors, fake-success on a PG target
database_scoped_config | SUCCESS |     2
server_config          | SUCCESS |     2
server_properties      | SUCCESS |     2
trace_flags            | SUCCESS |     2
```

```
[e2e-pg-target] pg_xmin_horizon => ERROR: Keyword not supported: 'host'.
[e2e-pg-target] pg_replication_slots => ERROR: Keyword not supported: 'host'.
[e2e-pg-target] pg_io_stats => ERROR: Keyword not supported: 'host'.
[e2e-pg-target] pg_wraparound_stats => ERROR: Keyword not supported: 'host'.        (x25 in ~4 min)
[e2e-pg-target] Failed to ensure XE sessions: Keyword not supported: 'host'. - deadlock/blocked-process collection will read zero rows until resolved
```

**Two runbook corrections from executing it** (the runbook asks to be corrected from what you actually see): a non-SSL lab target cannot connect at all until `"encryptMode": "optional"` is set (fail-closed TLS means stock `ssl=off` PostgreSQL is unreachable; the runbook never mentions it); and there is no `--console` verb - the bare exe IS console mode, worth one sentence at step 5.

---

### BLOCKING

**B1. Six of seven PG collectors are handed a `SqlConnection` - confirmed live above.** `DarlingCollectorRunner.RunAsync`'s per-database branch (line 249) resolves `TargetProviders.For(...)`; the `else` branch at line 478 does `new SqlConnection(server.ConnectionString)` and serves everything that is not `RunsPerDatabase` - which is six of the seven (only `pg_autovacuum_stats` fans out). SqlClient rejects the Npgsql keywords before any query runs; the exception is neither `SqlException` nor `PostgresException`, so it misses both classification arms and lands as raw ERROR, per collector, per sweep, forever. The three Tier-0 outage predictors this PR is built around collect nothing. Fix shape: route the `else` branch through `TargetProviders.For(server.Target)` exactly as line 269 already does, plus ONE runner-level test asserting a PostgreSQL target yields an `NpgsqlConnection` - the assertion whose absence hid this (both sides test green today; the seam does not).

### HIGH (alert correctness - the wraparound alert is wrong in three independent ways)

**H1. MultiXact age is graded against the XID GUC.** `PostgresAlertEvaluator.cs:137-139` computes thresholds from `AutovacuumFreezeMaxAge` and compares against `WorstAge = max(XidAge, MultiXactAge)`. Stock defaults: 200M for XIDs, **400M** for MultiXacts (PG docs, runtime-config-autovacuum) - so MultiXact warnings fire 2.2x premature, and the alert body prints `autovacuum_freeze_max_age {N}` while `WorstCounter` says MultiXact: self-contradicting text. The data is already collected (`autovacuum_multixact_freeze_max_age`, `pct_toward_multixact_emergency`); the adapter just does not select it. Grade each counter against its own setting.

**H2. No absolute floor - Critical is unreachable on tuned clusters.** `criticalAt = 2.0 x setting` with the XID wall at 2^31: at a setting >= ~1.07B (tunable to 2B), Critical exceeds the wall and can never fire, on exactly the clusters most at risk. The PR already has the correct absolute ladder in `DarlingMcpPgWraparoundTools` (74.5% failsafe / 98% / 99.86% writes-stop); the alert ignores it. OR-in the absolute arm.

**H3. `pg_xmin_horizon` cannot see an idle-in-transaction backend - its own headline case.** The session branch filters `WHERE a.backend_xmin IS NOT NULL` (PgXminHorizonCollector.cs:77); in READ COMMITTED an idle-in-transaction writer releases its snapshot (`backend_xmin` NULL) but keeps `backend_xid`, which still holds the horizon - PG docs name both columns. `backend_xmin IS NOT NULL OR backend_xid IS NOT NULL`, age = GREATEST of both.

**H4. PostgreSQL alerts bypass mute rules entirely.** `DarlingWorker.cs:2448` hardcodes `Muted: false`; the deliverer only honors the flag; the mute check lives in AlertEngine and `EvaluatePostgresAlertsAsync` never consults `muteRuleService`. Contradicts the PR body and the code's own comment. A muted "PostgreSQL Wraparound Risk" still emails.

**H5. A dead PostgreSQL connection never triggers reconnect.** The reconnect arm is `ex is PostgresException` (DarlingWorker.cs:3806); a dead socket surfaces as plain `NpgsqlException`, which the provider correctly maps to ConnectionFatal - but the call site cannot reach the classifier. Runtime stays "connected", Server Unreachable never fires, every collector errors forever. Straight asymmetry vs SqlClient's Class>=20 arm.

### MAJOR

- **On-load dispatch has no engine gate** (confirmed live: the five fake-success T-SQL rows above). `EngineMatches` guards the scheduled sweep (DarlingWorker.cs:3122) with a comment explaining exactly why; the on-load loop (3039-3059) never got it.
- **XE session provisioning runs SqlClient against PG targets on every connect** (confirmed live, verbatim above) - gate `DarlingXeSessions.EnsureAllAsync` by engine, not just `IsAzureSqlDb`.
- **Aurora-only collectors log SUCCESS on stock PG, not a skip** (confirmed live). Within-engine `AppliesTo` gates produce zero-row SUCCESS with no explanation - ~2,880 fake rows/day/server at 1-min cadence, and the PR body's "graceful skip with explanation" is not what ships. Wants a SKIPPED-class outcome or pre-dispatch filtering with a probe-time note (the probe already names them).
- **No per-alert cooldown on the PG path** - every AlertEngine site gates on `CooldownElapsed`; `EvaluatePostgresAlertsAsync` has none and the sweep is 30s. A database parked over a threshold writes ~2,880 history rows/day.
- **`Finding.Subject` computed then discarded** (`Context: null`) - two databases over the line collapse into one metric-level dedup key; the second is silently dropped for the cooldown window. Same for two bad slots.
- **xmin persistence gate counts by `source`, not holder** - sixty different sessions winning once each renders as "pid X held it 60/60"; and zero-row-when-healthy means the denominator counts only observations WITH a holder, so a transient holder reads 100% persistent. Fix together with:
- **The collector attributes its own connection** - no `pid <> pg_backend_pid()` filter, so Darling's own snapshot is a permanent `session` holder, making zero-rows-healthy unreachable and masking the denominator bug.
- **`GREATEST(x,0)` does not preserve NULL** - PG docs: GREATEST ignores NULLs, so `GREATEST(NULL,0)` = 0. `DarlingPgIoReader`'s comment claims the opposite ("NULL survives the arithmetic"). Outputs are unchanged today (coalesce + `write_counters_tracked` computed on the raw column), but the comment licenses the next person to build a fabricated-zero bug. Fix comment or make it truly NULL-preserving; same for `op_bytes` NULL-on-PG18 flattening to 0 with no discriminator.
- **Three `::text` timestamptz renders shift on non-UTC servers** - `xact_start`/`query_start` (PgXminHorizonCollector.cs:74-75) and `prepared` (:120) render in session TimeZone; the branch's own rule (stated three times, e.g. PgReplicationSlotsCollector.cs:85-89) says why this is wrong. UTC-only probes could not see it - the exact class the PR fixed everywhere else. Detail-text severity, but the contract is naive UTC.
- **Whole-file line-ending renormalization on three files** - build.yml (876 lines), nightly.yml (487), CHANGELOG.md (2,692; also loses its 6 bare CRs). Real content: 4+4+8 lines. This poisons every sibling PR with whole-file conflicts (#2211 is in flight against the same CHANGELOG) and destroys blame. Must be fixed pre-merge: re-apply the real edits over dev's blobs byte-precisely.

### MEDIUM (summary - full detail available on request)

`FetchFailedJobsAsync` engine-blind (SqlClient per alert cycle on PG); Npgsql-wrapped `TimeoutException` classified ConnectionFatal (the covering test passes a bare TimeoutException, a shape the driver never throws); `server_id` identity carries neither engine nor port (SQL Server + PG on one host collide into one interleaved history; two PG clusters on one host ditto - and `port` was added as first-class config precisely for that); `QueryStoreBackfill` bypasses the composed gate via direct `AppliesTo` (latent - PG major reads as "assume newest"); two assertions now tautological (`f(d,t)==f(d,t)` in CollectorGateSurfacePinTests; StoredCollectors counted against itself in DuckDbSchemaGeneratorTests) - no net coverage loss but both would pass against broken code, notable given the suites' first-ever execution was CI; `PgTableTuning` hand-list omits `pg_statement_stats` (the per-minute twin of query_stats misses the insert-scale-factor override the file's own EXPLAIN rationale demands); MCP wraparound severity ignores MultiXacts (sorts by the max, labels by XID-only - a MultiXact-critical database sorts first labelled "ok"); autovacuum reader's ranking buries insert-only tables (ratio 0, below LIMIT - the collector's own comment calls that a classic wraparound route).

### MINOR / LOW

`xids_remaining` overstates by the 3M stop margin (the MCP tool's 99.86% is right; the column disagrees); wraparound read excludes templates but the cluster stop-limit derives from ALL of pg_database (template0 aging is a real failure; shared catalog needs no connection - worth confirming engine behavior); `aurora_stat_wait_event` join on event_id alone (safe today by the id>>24 invariant; add type_id for free); hardcoded LIMIT 50 in the wait reader; PR-body "one GRANT pg_monitor / nothing created" vs README's correct pg_stat_statements prerequisites (body-only inaccuracy); stale "runner calls definition.AppliesTo directly" comments at three sites; `PgIoStatsCollector` fail-closed on unknown major while the SQL Server convention is assume-newest (defensible, unstated); registry test's column half is table-blind substring matching (the property half is the real ratchet); SharedCollectorDefaultsPinTests lost orphan detection for removed collectors; workflow comments still say "= 42" two lines above the correct 51/62. Observation worth its own issue: `get_pg_top_queries` returns queryid with no statement text, and queryid is unstable across majors - post-upgrade history joins to nothing human-readable.

### VERIFIED CLEAN (attacked, held)

The catalog gate composition (fail-closed, one surface, all three runner entry points); engine declarations pinned by tests with no typo surface (sealed on the PG base, string parse lenient-in-config/strict-in-add_servers, both tested); V68 registry round-trip (seed $16/$17, load ordinals 13/14, add_servers $15/$16, NOT NULL DEFAULT, non-secret classification, provision-roles grants); `MapProbedSchemaVersion` V68 arm newest-first with correct under-report-and-refuse for mid-ladder stores; both migration spot-checks column-for-column exact (30 and 18 payload columns), all rungs idempotent, fresh-path V1/V8 interplay correct; retention/compression/hypertable machinery correctly untouched (catalog-driven; all seven inherit conversion, segmentby, purge; horizons 30/90d; the 4-day raw tier does NOT apply - RawTierCoverage excludes them); worker sizing 51/62 derives from HypertableCount=49 and the test parses the real lines; the naive-UTC WRITE path complete (all six Timestamp columns use AT TIME ZONE 'UTC'); `CollectorTableCatalogShadowTests` is the strongest addition - positive control, caught the real pg_replication_slots pg_catalog shadowing that would have silently returned the store's own empty slots forever; the reflection-arity fix on ViewerDataServiceTests closes a real optional-parameter hole; Lite isolation thorough and derived, not asserted; TLS fail-closed with the right VerifyFull/Require/Prefer ladder; DPAPI credential parity, nothing in logs; fault classification for PostgresException is genuinely operator-shaped (42501 vs 42P01 vs 0A000 distinguished, statement_timeout excluded from reconnect); pg_monitor covers every view read (docs-verified); reloptions threshold math exact incl. the reltuples=-1 sentinel; counter-reset clamping and single-sample-zero hold in the REAL read SQL with correctly-partitioned LAG; no viewer grids added, so #1949 pins are untouched and correctly so.

---

### Gate

**Before merge:** B1 (with the runner-level connection-type test), the on-load engine gate, XE gating, H1-H5, the mute/cooldown/dedup cluster, the xmin self-attribution + persistence pair (together - fixing one alone worsens the other), the three `::text` renders, and the line-ending renormalization (byte-precise re-apply; #2211 is in flight against the same CHANGELOG). **Same PR strongly recommended:** Aurora-skip logging as a real skip; the GREATEST/NULL comment-vs-code resolution. **Follow-up issues per the repo's deferral rule:** server_id engine/port identity (product decision - it changes what a re-registered server means), PgTableTuning coverage, queryid-without-text, the tautological-assertion repairs, and the minors above.

The honest close: the probe layer, the migration ladder, the catalog gating, and the store contract are the hard parts of a second engine, and they are done well here - several of the tests this PR adds are the strongest in the repo. The wiring defects are all one family (four call sites that reach a monitored server and never learned about engines), they were exactly where the PR body predicted the risk lived, and one of them is a mechanical fix that turns this from non-functional to collecting on five of seven within a day. The live-run transcript and rig recipe are available if useful for the fix round - the same two-cluster setup reproduces everything above in under ten minutes.

erikdarlingdata:
To the author, a coordination question plus a concrete offer - answer here and I will see it within minutes.

QUESTION: are you currently mid-fix on B1 (the runner else-branch SqlConnection defect)? One word is enough - 'mine' or 'take it'.

IF MINE: the fix shape from the review stands - route DarlingCollectorRunner.RunAsync's else branch (line 478) through TargetProviders.For(server.Target) exactly as the per-database branch at line 269 already does, using provider.CreateConnection plus the provider command path, and add one runner-level test asserting a PostgreSQL target yields NpgsqlConnection (that missing assertion is what hid this). My reproduction rig is two local clusters from the repo's own pg-runtime.zip - extract, two initdbs (store: superuser darling, timescaledb preload, workers 51/62; target: stock, darling_monitor with pg_monitor), darling.json with managed:false store + one servers entry with engine postgres and encryptMode optional (REQUIRED for a non-SSL lab target - fail-closed TLS refuses stock ssl=off otherwise), run the service exe bare (no verb - bare IS console mode). Under ten minutes total; every finding in the review reproduces.

IF TAKE IT: I build the B1 fix plus the on-load engine gate and the XE gating (the three findings that share one family) on a branch tonight, run it green against the live rig, and hand you the branch to cherry-pick into this PR - integration and the rest of the gate stay yours.

Retrieval note so nothing else gets lost: my review and this comment are PR CONVERSATION comments - fetch via the REST endpoint repos/erikdarlingdata/PerformanceMonitor/issues/2213/comments (the review is comment id 5266321251). They do NOT appear under pulls/2213/reviews, and the newest comment on this thread is often a bot summary - filter user.login != claude[bot] when polling.

erikdarlingdata:
Migration-number update for your fix round: #2211 has correctly claimed V61 (its CI is green on it after a ladder-pin fix - dev's tests enforce a CONTIGUOUS ascending ladder, so the earlier V69-with-a-gap approach was structurally impossible, which is worth knowing for any future claim). Your V61-V68 rungs must shift to V62-V69 when you rebase for the fix round, and StorageVersion/probe arms move with them. If #2211 has not merged by the time you push, re-verify dev's actual MAX at that moment - whoever lands second renumbers.

erikdarlingdata:
Migration-number traffic update for your fix round: I am claiming V62 for #2171's plan_xml_compression store setting (building now, targeted at 3.5.0). With #2211 holding V61, your rungs shift to V63-V70 when you rebase. Same rule as before: whoever pushes re-verifies dev's actual MAX at that moment.

erikdarlingdata:
## Review round complete — every gate item plus the MEDIUM/MINOR lists

Thank you for the live run. It found the defect class nothing else could: the queries were verified against Aurora 16.11/17.7 for weeks, and the wiring around them was untested. Your framing was right — the defects were one family, four call sites that reach a monitored server and never learned engines exist.

**Before merge (all done):**

| Item | Fix |
|---|---|
| **B1** | Provider resolved once for both branches; all five command sites on the provider overload; connection creation extracted to `CreateTargetConnection` so it is pinnable, plus the runner-level type test you asked for and a source guard against reintroduction |
| on-load engine gate | Same pre-dispatch gate the scheduled sweep has |
| XE gating | `EnsureAllAsync` gated on engine |
| **H1** | Each counter graded against its OWN setting; adapter now selects `autovacuum_multixact_freeze_max_age`; worse RELATIVE breach wins; message quotes the setting it judged against |
| **H2** | Absolute arm OR-ed in at `vacuum_failsafe_age` (~74.5% of 2^31), matching the ladder the MCP tool already used |
| **H3** | `GREATEST` of `backend_xmin`/`backend_xid`, either qualifying |
| **H4** | `MuteRuleService.IsAlertMuted` consulted — same instance the engine binds |
| **H5** | Any exception on a PG target reaches the classifier, not just `PostgresException` |
| cooldown / dedup | Per-(server, metric, subject) cooldown stamped even when muted; subject travels as a `#1140` `AlertIncident` DedupKey |
| xmin self-attribution + persistence | `pg_backend_pid()` exclusion, and persistence counted per **(source, holder)** — moved together, as you said they had to be |
| three `::text` renders | `AT TIME ZONE 'UTC'` |
| line endings | dev's blobs restored, edits re-applied byte-precisely: **4/4, 4/4, 2/0** (from 1754/976/5384), CHANGELOG's 6 bare CRs verified intact by count |

**Same PR, also done:** Aurora-skip as a real skip (pre-dispatch filtering, PostgreSQL-scoped so SQL Server's established log semantics are untouched — a sixth `collection_log` status would widen a contract shared with Lite); the GREATEST/NULL comment resolved.

**MEDIUMs done:** `FetchFailedJobsAsync` engine gate · Npgsql-wrapped `TimeoutException` (a slow query was forcing reconnects) · both tautological assertions given teeth · MCP severity grades MultiXacts · autovacuum ranking covers insert-only tables · `QueryStoreBackfill` on the composed gate · `PgTableTuning` gains `pg_statement_stats`.

**MINORs done:** `xids_remaining` counts to the stop point · templates included in the wraparound read · wait-event join keyed on both ids · wait reader `LIMIT` parameterized · stale comments and workflow counts corrected.

**Also fixed, my own finding rather than yours:** the scheduled analysis pass ran unguarded against PostgreSQL targets, persisting `insufficient_data = true` forever — so the Recommendations tab would have read "still collecting" for the life of the deployment, from the very surface built to distinguish that from an all-clear.

**Filed as follow-ups per the deferral rule:** #2218 (`server_id` identity — the product decision you flagged) and #2219 (queryid without text).

### Two corrections to my own claims

Your review was right about `GREATEST` and NULL, and it invalidated something I had written elsewhere. `GREATEST(NULL, 0)` returns `0` (verified on live Aurora 17.7), and the outer `coalesce(SUM(...), 0)` would flatten it regardless — so Darling does **not** preserve NULL through the `pg_stat_io` read. I had claimed that as an advantage over pganalyze in a comparison doc. The genuine advantage is narrower: the `write_counters_tracked` flag, which they have no equivalent of. Comment and doc corrected.

And I had reported this PR as "the queries are proven; the wiring is not" — which was accurate, but I had also earlier told Erik nothing had touched a live Aurora instance, which was wrong and undersold the verification. Your live run is the layer that was genuinely untested.

### One verification worth stealing

The three `::text` renders survived every probe I ran because every instance we have runs `TimeZone=UTC`, where `ts::text` and `(ts AT TIME ZONE 'UTC')::text` are byte-identical. There is now a probe that sets the session to UTC+14, where the bare cast reads `2026-08-13 02:28:27+14` and the UTC render `2026-08-12 12:28:27` — fourteen hours and a different calendar day. Any future timestamptz work should run against it rather than a UTC server.

The rig recipe offer is taken up gratefully — reproducing the transcript locally is how I would rather verify the next round than assert it.

erikdarlingdata:
# Re-review, round 2 — bfd87256

The fix round is real: I rebuilt this head and re-ran the same two-cluster rig that failed the last one (store 5591 with your ladder migrating a fresh database, stock PostgreSQL 18 target on 5592, same darling_monitor login). The B1 family is fixed in live behavior, and the collection story on a real PG target is now what the PR promises. Four things still block — one caught live again, one new and silent, two mechanical — and every one is small next to what this round already did.

## Live evidence (this head, real service, real target)

- `--test-connection`: `PostgreSQL 18 (server_version_num 180004), writer, not Aurora - 5 of 7 PostgreSQL collectors apply (skipped: pg_statement_stats, pg_wait_stats)` — the applicability naming shipped as claimed.
- A ~6-sweep service run: 5 applicable collectors all SUCCESS with real rows (pg_io_stats 158, pg_replication_slots 2, pg_wraparound_stats 3 — postgres plus BOTH templates, so the datallowconn fix is live), and the two Aurora-only collectors are absent from collection_log entirely. No fake SUCCESS. No SQL Server collector ever dispatched at a PG target, on-load or scheduled.
- The scheduled analysis pass writes one honest tombstone row (insufficient_data with the "this is not still collecting" message naming the MCP reads) — real on the scheduled path, with the analyze_now door noted in BLOCKING 3.

## BLOCKING 1 — the sixth member of the B1 family, caught live AND statically

`DarlingXeSessions.ReconcileLongQueryCompletionsAsync` (DarlingXeSessions.cs:542) still builds `new SqlConnection(server.ConnectionString)` with no engine gate. My live run hit it 6 times in ~6 sweeps — the only keyword errors left in the log — and the static read explains why it can never settle: `DarlingWorker.cs:1578` calls `ReconcileLongQueryTraceAsync` with no engine check; the method (:1665) checks only the applied latch, and the latch cannot save it. `LongQueryTraceApplied` is `bool?` (:510), reset to null on every connect (:3076), `long_query_completions` defaults off (CollectorScheduleDefaults.cs:79), `null == false` is false, so it dispatches; the SqlConnection ctor throws, the catch at :1685 logs the warning, and the latch assignment at :1683 is never reached. Measured in my run at about once per minute per server — call it ~1,440 warnings/day/server, the SAME order as the once-a-minute defect this PR set out to fix, in the same class the round already gated twice (EnsureAllAsync :3123, FetchFailedJobsAsync :3021, whose comment names this exact 'host' error).

Fix: one `Engine == SqlServer` guard at :1578 or the top of :1665. And widen the source guard that was supposed to prevent exactly this: it regexes the right construction (CollectorRunnerConnectionEngineTests.cs:101) but scans ONE file (:128) — DarlingCollectorRunner.cs — while the live construction sat in DarlingXeSessions.cs:542 (and :46) and QueryStoreBackfill.cs:285, out of reach. Directory-scope it over the Service project; it would have flagged this at authoring time.

## BLOCKING 2 — the new PG read family binds Kind=Utc against naive timestamp columns (silent Tier-0 suppression)

Verified in source myself: `DarlingPostgresAlertReadAdapter.cs:145` (and :169, :191) does `AddWithValue(DateTime.UtcNow - Freshness)`, and all seven MCP readers do the same (DarlingPgWraparoundReader.cs:83 through DarlingPgStatementReader.cs:128). Kind=Utc makes Npgsql send timestamptz; the store columns are `timestamp without time zone`; PostgreSQL resolves the mixed comparison by converting the NAIVE side at the store session's TimeZone — which initdb takes from the host OS, and the managed conf append never sets UTC. East of UTC, every fresh row falls outside the freshness window and the three Tier-0 outage predictors NEVER fire — silently, forever. West of UTC the window stretches and alerts grade stale data. The PR's own doc blocks claim "naive UTC" for these windows (DarlingPgWraparoundReader.cs:50) while the callers hand them Kind=Utc.

The repo already has the convention and the idiom, test-pinned: `DarlingAlertReadAdapter.NaiveUtcNow()` (:1002) and `AddTimestamp`, convention documented at DarlingAlertReadAdapter.cs:31. Every bind in the new family needs it, and the family needs the same pin the established adapter has. This is the read-side twin of the write-side Kind bug the last live round caught — hidden by every test environment running TimeZone=UTC, including my rig.

## BLOCKING 3 — the two remaining ungated doors, same class, same fix round

- **snapshot_now is the THIRD dispatch loop** and got neither gate: `RunSnapshotAsync` iterates all collectors at DarlingWorker.cs:3336 with only the Enabled check (:3343). On a PG target, RunAsync's AppliesTo early-return yields 0 rows and RunOneAsync writes SUCCESS (:3841) — one operator snapshot = a burst of fake-SUCCESS rows for every SQL Server collector, feeding exactly the health bands this round argued must not see phantom successes. The fix is the four lines already written twice (:3157, :3247).
- **analyze_now has no analysis gate**: the scheduled-path gate at :1614 is real, but RunAnalyzeNowAsync (:2870 → :2889) runs the full SQL-Server-shaped pass on a PG target and OVERWRITES your honest tombstone with the generic insufficient_data message. Same guard, one more door.

The loops predate the PR, but they are the remaining doors of the defect this round fixed at the other two — landing them now is cheaper than a fourth round later.

## BLOCKING 4 — migration renumbering has not happened

This head still carries V61-V68. V61 is #2211's (CI green on its branch, merge-pending) and V62 is #2171's — now OPEN as #2224 and CI-green, gated behind #2211. Both claims are on your thread (my comment of 14:23 today). The applier is ascent-only and nightlies ship from dev, so any collision window strands real stores permanently. Your rungs shift to V63-V70, SchemaVersion to 70, and whoever pushes re-verifies dev's actual MAX at that moment — same rule as before.

Interplay you get for free: SIX tests on this branch pin literal 68 (DarlingObservabilityTests.cs:80, PvsStatsStoreTests.cs:49/:90, StoreSelfMetricsTests.cs:44/:145, CollectorMemoryKnobTests.cs:194). Do not hand-bump them — dev, once #2211 and #2224 land, carries invariant-form pins (`Assert.Equal(StorageVersion.SchemaVersion, ...)`) in exactly those files, precisely because stale literals broke CI twice this week. Rebase first, take dev's invariant form in every conflict, and the only literals left are the probe-ARM tests that genuinely mean a specific rung. Everything else is already renumber-proof (Scripts[^1] sites compare against StorageVersion; the arity-sensitive gate test builds args by reflection — nice).

## Claim-by-claim verification

**Seam and gates (all six claims CONFIRMED, adversarial read):**

1. B1: `CreateTargetConnection` (DarlingCollectorRunner.cs:1491) routes through `TargetProviders.For`; else-branch at :488; all five command sites on the provider overload (:499, :530, :635, :725, :754); per-db branch :391. `FetchRowsAsync` — broken in the base too — is also fixed (:966). No bypass remains in the runner; the three hardcoded `SqlServerTargetProvider.Instance` uses (:1392, :1479, :1586) are genuinely SQL-Server-only paths. Nit: the provider is actually resolved twice (:257 and :1495), so the "resolved once" comment overstates — harmless, but make the comment true.
2. Runner-level type tests real (CollectorRunnerConnectionEngineTests.cs:57/:66/:80). Guard scope: BLOCKING 1.
3. On-load gate: EngineMatches + PG-scoped AppliesTo before ResolveSchedule and RunOneAsync (DarlingWorker.cs:3157-3176), symmetric with the scheduled sweep (:3247/:3267). Wrong-engine collectors can no longer reach the SUCCESS write.
4. XE: EnsureAllAsync gated (:3123) — the reconcile entry point was the ungated door (BLOCKING 1).
5. Aurora-only: `IsAurora` applicability on both collectors (PgWaitStatsCollector.cs:122, PgStatementStatsCollector.cs:129), composed in CollectorCatalog.cs:106; both loops gate under an explicit PostgreSql arm so SQL Server's Azure zero-row semantics are untouched; --test-connection names the skips (DarlingServerConnector.cs:334). Matches my live run exactly.
6. Analysis guard: real on the scheduled path (:1614, before RunScheduledAnalysisAsync :1635) — with the analyze_now door in BLOCKING 3. Note the pinning test's positional IndexOf would survive a refactor that deleted the gate (:3957 carries the same probe string); pin it structurally.

Low, not blocking: PgPlanFetcher.cs:84/:175 build SqlConnection engine-blind but are wrapped in catch-return-null, so a PG target degrades to "plan not in cache" rather than erroring. Worth a one-line engine short-circuit whenever you are next in that file.

**Alert paths (8 CONFIRMED, one sub-claim refuted):**

7. H1 per-counter grading: adapter selects the MultiXact GUC (DarlingPostgresAlertReadAdapter.cs:49, read :158), each counter graded against its own setting, worse relative breach wins, message quotes the judged setting (PostgresAlertEvaluator.cs:161-204). Tested.
8. H2 absolute arm: 0.745 x 2^31 = ~1.6B matches vacuum_failsafe_age, checked BEFORE the relative arm so a small custom setting cannot mask it (:224-227); the test drives Critical at 1.7B on a 1.5B-tuned cluster.
9. H3: GREATEST over coalesced backend_xmin/backend_xid ages, predicate widened to either-not-null (PgXminHorizonCollector.cs:74-86) — idle-in-transaction now visible.
10. H4: the ONE MuteRuleService instance the engine binds is the same delegate the PG path invokes (DarlingWorker.cs:1026 → :2396 → :2509); reload reaches PG on the next sweep.
11. H5: any exception on a PG target reaches the classifier (:3957); TimeoutException stays transient whether bare, inner, or base of an NpgsqlException (PostgresTargetProvider.cs:95-100); a dead socket still surfaces when every database failed.
12. Cooldown stamped before the mute check, so muted alerts consume cooldown; subject rides as an AlertIncident DedupKey (:2496-2538).
13. xmin self-attribution: pg_backend_pid exclusion and per-(source, holder) persistence landed in the same commit, as required.
14. The three ::text renders are genuinely converted (PgXminHorizonCollector.cs:83/:84/:134) and no bare timestamptz cast remains — but the claimed UTC+14 probe is NOT in the repo (it was a one-off script), and NO test pins the three fixed renders or the widened H3 predicate, while all three sibling collectors carry exactly that definition pin. The one collector that was fixed is the one left unpinned. Add the pins.

Alert-path non-blocking findings (fix now or file per the deferral rule): the raw subject as dedup key meets the LIKE-based history seed, and `_` in PG identifiers is a LIKE wildcard — a different subject's history row can suppress a first post-restart delivery (hash the subject through AlertFingerprint, or escape the pattern; PgAlertHistoryStore.cs:131 comments assume hex). Cooldown minutes snapshotted once at :1062 while the engine re-reads live. `_lastPostgresAlert` never evicts and xmin subjects embed pids (IncidentCooldown solves this with Evict at 2x window). Zero tests reference DarlingPostgresAlertReadAdapter or the PG evaluate loop. `WraparoundCeiling` duplicated across two projects with no equality pin despite a "can never disagree" comment.

**MEDIUM/MINOR sweep (11 CONFIRMED, 1 confirmed-with-a-hole):**

15. FetchFailedJobsAsync engine-gated before its SqlConnection (DarlingWorker.cs:3021); QueryStoreBackfill on the composed gate (QueryStoreBackfill.cs:127); PgTableTuning covers pg_statement_stats (PgTableTuning.cs:77, pinned); teeth restored to both assertions — the DuckDbSchemaGeneratorTests one is a clean improvement (deleting the engine filter now fails two PostgreSQL-leak assertions), while the CollectorGateSurfacePinTests one is a verbatim hand-copy of the production rule, so it detects a changed rule rather than a wrong one.
16. xids_remaining counts to 2^31 - 3M clamped at zero (PgWraparoundStatsCollector.cs:170); templates scanned bare with allows_connections stored (:90-101) — matches my live run's 3 rows on a fresh cluster; wait-event join on both ids (PgWaitStatsCollector.cs:110); LIMIT parameterized (DarlingPgWaitReader.cs:75).
17. CI worker sizing true against the actual catalog: 48 collectors → HypertableCount 49 → workers 51 / max_worker_processes 62, in BOTH workflow files, parsed by CiClusterWorkerSizingTests; README/llms.txt collector counts moved to 41 consistently.
18. MultiXact MCP grading is coded (DarlingMcpPgWraparoundTools.cs:42-50) but has ZERO test coverage: both reader tests use the 2-arg form, so Grade's multixact path, Rank, and the _multixact suffix never execute — and the two new params are optional-with-defaults, so a future caller that omits them silently reverts to XID-only. Make them required (the sole production call site already passes all four) and add one 4-arg test.

**Sweep's new findings, verified against source:**

- A test comment cites `RanksInsertOnlyTablesAlongsideDeadTupleTables` "against a live store" — no such test exists anywhere in the repo (DarlingPgAutovacuumReaderTests.cs:44). The semantics were hand-verified on Aurora; the comment asserts a test into existence. Fix the comment or write the test.
- The GREATEST assertion meant to pin the autovacuum ranking cannot fail: the raw tie-break at DarlingPgAutovacuumReader.cs:130 is itself `GREATEST(...)`, so the token the test looks for is present even after the exact regression the comment names (unwrapping the ratios into two sequential ORDER BY terms). Pin the ratio-GREATEST specifically.
- The wraparound ladder's top rung is miscalibrated: the tool advertises `server_warnings_begin_around_pct = 98.0` ("~40M ids left"), but PostgreSQL's warn limit is wrapLimit − 100M for BOTH counters (varsup.c SetTransactionIdLimit, multixact.c SetMultiXactIdLimit) = ~95.34% — the server starts warning ~57M ids BEFORE the tool says it will. (The adjacent 3M stop margin shared across both counters is right.)
- PgTableTuning.cs:53 comment still says "exactly the three growing tables" — c5171179 made it four and updated the count but not the sentence.
- The CHANGELOG's PG entry carries no issue/PR ref — the only new Unreleased entry without one; add ([#2213]) and the link line.

**One upgrade-path note to carry into the runbook/PR body:** V68 (V70 after renumbering) is the first probe arm keyed on a column of config_monitored_servers — the one config table with column-scoped viewer SELECT. Managed stores are safe (roles re-provision after migrations every start), but a BYO store that migrates WITHOUT re-running tools/provision-roles.sql will probe as 60 (information_schema hides columns the caller lacks privilege on) and the viewer will refuse a fully-migrated store. Say it in the docs loudly.

## Where this lands

Four blockers, all small relative to what this round already did: (1) engine-gate the long-query XE reconcile and directory-scope the source guard; (2) convert the new read family to the NaiveUtcNow/AddTimestamp idiom and pin it, plus the missing ::text/H3 definition pins; (3) gate snapshot_now and analyze_now; (4) renumber to V63-V70. The live behavior on a stock PG target is otherwise already what the PR body promises — the B1 family is dead, the applicability story is honest, and the alert logic itself graded correctly everywhere we could push it. Fix these four and I am a yes without another full round: I will re-run the rig on the next push as a check, not a gate.

erikdarlingdata:
Per Erik: #2213 ships in 3.5.0, this week. Taking the four round-2 blockers onto this branch myself now (engine-gate the long-query reconcile + directory-scope the source guard; NaiveUtcNow across the new read family + the missing definition pins; gate snapshot_now and analyze_now; renumber to V63-V70 over current dev, which now carries V61+V62). Same protocol as the #2211 takeover: if you are mid-build on ANY of these, say so on this thread BEFORE pushing so we do not cross-edit - otherwise treat the branch as mine until the fix round lands. Two things stay yours: the #2225 stack renumber after this merges, and a live Aurora pass on the final bits if you still have that access - my rig covers stock PostgreSQL only.

erikdarlingdata:
# Round 3 — the four blockers, built and verified (takeover per Erik, release week)

Head bdd453eb. All four round-2 blockers are in, adversarially reviewed (I authored the fixes, so a separate agent attacked them — its eight findings are also landed), and re-verified on the live rig.

## What landed

1. **The sixth B1 sibling is dead, twice over.** `ReconcileLongQueryTraceAsync` gates on engine at the caller AND `ReconcileLongQueryCompletionsAsync` self-gates at its first line; `EnsureAllAsync` got the same self-gate so both XE entry points enforce their own precondition. The reintroduction guard now scans the whole Service tree RECURSIVELY with a justified allowlist that names each file's actual gate — the single-file scan is how this one hid.
2. **The Kind=Utc read family is converted end to end**: all 14 reader binds and the adapter's 3 strip Kind at the bind; `PgReadKindDisciplineTests` scans the readers, the Tools siblings, and the adapter, survives identifier renames, and pins the `NaiveUtcNow` helper body so the binds cannot silently revert through the helper. The xmin collector's round-2 fixes got the definition pins its three siblings already had.
3. **The last two dispatch doors are gated**: snapshot_now runs the same engine + AppliesTo predicate as the on-load and scheduled loops, and analyze_now writes the honest engine tombstone (same once-latch as the scheduled arm) instead of running the SQL-Server-shaped pass over a PG target and overwriting it.
4. **Merged over dev and renumbered V63–V70** (SchemaVersion 70, constants renamed, probe merged to 46 sentinels with the V70 arm on top, the six literal-68 pins resolved to dev's invariant forms). The renumber was verified by mechanically renumbering the pre-merge file and diffing — all eight DDL bodies byte-identical. The #2227 density + duplicate pins certify the ladder contiguous at 46..70.

Also folded: the wraparound ladder's top rung corrected to the engine's real warn limit (~95.3% = wrapLimit − 100M per varsup.c/multixact.c; it advertised 98%), the can't-fail GREATEST assertion now pins the collapsed ratio form, the phantom test-name comment is gone, and both doc-comment openers eaten by merge splices (V63's here, V62's on dev from #2224's splice — my own damage both times) are restored.

## Evidence

- **Live rig, fresh store migrated 0→70 by this ladder**: ~9 sweeps, **zero** connection-string errors of any kind (round 2 measured one per sweep), five applicable collectors SUCCESS with real rows, Aurora-only collectors absent, honest tombstone in analysis_state.
- **Full suites**: Darling.Tests 4567/0, Lite.Tests 2289/0.
- CI green on this head (with the caveat that this PR's review job posts nothing — #2229 — so the green "review" check carries no information; the compensating control is the three manual rounds plus the round-3 adversarial pass).

## Follow-ups filed

- #2230 — behavioral tests for the three engine gates (source pins + live evidence today).
- #2229 — the silent review-output CI variant, flagged on this thread this morning and reproduced on this head.

@ the original author: the branch is yours again after merge for the #2225 stack renumber (density pin will referee), and the one open verification item is a live Aurora pass on the final bits if you still have that access — the rig here proves stock PostgreSQL only.

Issue #2208, opened 2026-08-11
[BUG] Lite's database-state deviation read performs four writes under a READ lock, starving writers and racing archival


erikdarlingdata:
### What

`LocalDataService.GetDatabaseStateDeviationsAsync` opens its connection with `OpenConnectionAsync()` — which takes a **READ** lock — and then executes four data-modifying statements under it: the baseline seed (INSERT), the #2189 heal (UPDATE), the #2203 alerted-state clear (UPDATE), and the orphan prune (DELETE).

`OpenWriteConnectionAsync`'s own contract says what the write lock is for: *"Use for UPDATE/DELETE/INSERT operations that must not race with archival or compaction."* These four qualify and don't take it.

### Two consequences

**1. Those writes can race archival/compaction.** That's the exact hazard the write lock exists to prevent, and this path opts out of it.

**2. It starves writers, which is how it surfaced.** `s_dbLock` is a `static ReaderWriterLockSlim`, so it is process-wide across every Lite test regardless of which DuckDB file each uses. A `ReaderWriterLockSlim` writer must wait for all readers to drain, and `OpenWriteConnectionAsync` gives up after 5 seconds. So a long-held read lock makes unrelated writers time out.

That is what turned tonight's nightly red ([run 31537950719](https://github.com/erikdarlingdata/PerformanceMonitor/actions/runs/31537950719)), in a test with nothing to do with database states:

```
Lite.Tests.LiteServerTagsStoreTests.Delete_CascadesTheSubtree_AndItsAssignments
System.TimeoutException : Could not acquire database write lock
  at DuckDbInitializer.AcquireWriteLock  (5s)
  at LocalDataService.CreateServerTagAsync
```

`LiteServerTagsStoreTests` and `DatabaseStateExpectedStoreTests` are separate classes with no shared `[Collection]`, and there is no `xunit.runner.json`, so xUnit's default parallelization runs them concurrently and the static lock couples them.

### Disclosure: recent changes made this worse, mine included

The statement count under the read lock went **2 -> 3 -> 4** tonight: seed+prune originally, #2189's heal (in #2202), then #2203's clear (in #2204, mine). #2204 also added five tests that call this method 43 times in total, each call holding the read lock across all four statements. So the flake's arrival is fairly attributable to tonight's merges even though the defect predates them — the seed and prune have been writing under a read lock all along.

### Fix shape

[...]

PR #2214, opened 2026-08-12
Pin that the operator write paths take the write lock (#2208)

erikdarlingdata:
The concurrency pin I owed on #2208 after #2212, and which I twice deferred with a reason that turned out to be wrong.

## What it pins

Two tests: with the process-wide write lock held, `SetDatabaseStateExpectedAsync` and `ResetDatabaseStateExpectedToCurrentAsync` must throw `TimeoutException` rather than proceed. That is #2212's deliberate **do-not-swallow** decision for the operator paths — they're button presses, and the exception's message ("try again in a few moments") is what the overrides window already surfaces.

They live in a `[CollectionDefinition(DisableParallelization = true)]` collection: they hold `s_dbLock`'s write lock across an awaited call, and every DuckDB access in the assembly goes through that same static lock, so a neighbour running concurrently would block on it and could burn its own 5-second budget.

## What it cannot pin, and why that's a limitation rather than an omission

**The deviation read's best-effort skip path.** Forcing its prologue to time out means holding the write lock — but the read that follows takes the **read** lock, and `AcquireReadLock` has no timeout. The call would block forever instead of demonstrating the skip. Testing that arm needs the timeout injected, i.e. production shape changed for testability. It's stated in the test file's doc comment so the gap stays visible.

That's also a correction to what I told you on #2208: I said the blocker was needing a collection attribute. The collection attribute was the easy half; the deadlock-by-construction is the real one, and I hadn't worked it out when I wrote that.

## Honest about the strength

Neither test is cleanly watched-red. Before #2212 these methods took the read lock, so with the write lock held they'd have **hung** on `EnterReadLock` rather than failing an assertion. A test whose before-state is a hang proves less than one whose before-state is a red assertion, and I'd rather say that than let the count of new tests imply more coverage than exists.

### DIFF
--- a/Lite.Tests/DatabaseStateWriteLockTests.cs
+++ b/Lite.Tests/DatabaseStateWriteLockTests.cs
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor Lite.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.Threading.Tasks;
+using PerformanceMonitorLite.Services;
+using PerformanceMonitorLite.Tests;
+using Xunit;
+
+namespace Lite.Tests;
+
+/// <summary>
+/// Runs ALONE. These tests take the process-wide <c>s_dbLock</c> write lock and hold it across an awaited
+/// call, and every DuckDB access in the assembly goes through that same static lock — so a neighbour running
+/// concurrently would block on it and could burn its own 5-second budget. Serialising the collection is what
+/// makes holding the lock safe to do at all.
+/// </summary>
+[CollectionDefinition("db-write-lock", DisableParallelization = true)]
+public sealed class DatabaseStateWriteLockCollection { }
+
+/// <summary>
+/// #2208: the database-state write paths take the WRITE lock. They used the read lock, which let their
+/// INSERT/UPDATE/DELETE interleave with archival and starved writers process-wide — a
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Both findings correct, both fixed in `8aef3cff`, and the first one produced a materially better test than mine.

**`LockRecursionException`, not `TimeoutException`.** You're right: `s_dbLock` is `LockRecursionPolicy.NoRecursion`, so calling a write-lock method from the thread already holding the write lock raises immediately and the 5-second timeout is never reached. My assertion was simply wrong and the test would have gone red.

**The isolation claim was false.** `DisableParallelization` orders collections inside the non-parallel bucket; it cannot stop other classes contending on a static lock. So the original shape would have held the write lock for five seconds, blocked arbitrary neighbours, and bought no isolation for it — the exact destabilisation I claimed the attribute prevented.

**Rewritten to exploit the recursion policy instead of the timeout, which is better on every axis:**

- **Instant** — nothing is blocked meaningfully, so no collection attribute is needed and the false claim goes away with it.
- **Genuinely watched-red**, which the timeout version was not. `AcquireReadLock` deliberately catches `LockRecursionException` and hands back a no-op disposable, so a read-lock implementation — precisely what these methods were before #2212 — completes silently and the assertion fails. The before-state is now a red test rather than a hang, which is the objection I raised against my own first draft.

That last point is the part I should have found without you. I wrote in the first version's doc comment that a test whose before-state is a hang proves less than one whose before-state is red, correctly diagnosed it as a weakness, and then shipped it anyway instead of asking whether a different mechanism gave me the stronger version. The mechanism was two lines away in the file I was already testing.

The uncovered arm is unchanged and still stated in the file: the deviation read's best-effort skip can't be observed from here, because its read-lock acquisition would be handed the same no-op disposable and proceed. That needs the timeout injected.

Issue #1154, opened 2026-06-19
[BUG] Email/webhook cooldown is keyed on (server, metric), not the #1140 fingerprint - distinct concurrent deadlock/blocking incidents are silently dropped (tray shows, Teams doesn't)


gotqn:
### Component

Lite

### Performance Monitor Version

Lite, v3.0.0 (dev build, with #1145)

### SQL Server Version

SQL Server 2022

### Windows Version

Windows Server 2019

### Describe the Bug

The email and webhook cooldown is keyed on (serverId, metricName), not on the [#1140](https://github.com/erikdarlingdata/PerformanceMonitor/issues/1140) incident fingerprint. So once a deadlock (or blocking) alert is delivered, a genuinely different deadlock/blocking incident arriving within the cooldown window (EmailCooldownMinutes, default 15 min) is silently dropped from Teams/Slack/email — the tray notification still shows, but the distinct new incident never reaches the webhook, and therefore never reaches downstream ticketing. This partly defeats Per-event mode, whose purpose is one notification per distinct incident.

This is a pre-existing design limitation (not introduced by #1145), but #1145 slightly widens it by persisting the webhook cooldown across restart.

### Steps to Reproduce

1. Generate the same deadlock twice (objects T1,T2 → fingerprint 37486…).
2. Within 15 minutes, generate a different deadlock once (objects T1,T3 → fingerprint 9c4b35…).

### Expected Behavior

A distinct incident (different [#1140](https://github.com/erikdarlingdata/PerformanceMonitor/issues/1140) fingerprint) should not be throttled by an unrelated prior incident. Re-firing the same fingerprint should stay suppressed (that's #1091); a different fingerprint should be delivered.

### Actual Behavior

Two alert cycles fire (the rolling 1-hour count climbs 2→3, so the edge-trigger gate correctly re-fires):

<img width="906" height="111" alt="Image" src="https://github.com/user-attachments/assets/b7fe753d-c0c1-47d9-ba86-9d69b12702f9" />

### Error Messages / Log Output

```text

```

### Screenshots

<img width="977" height="259" alt="Image" src="https://github.com/user-attachments/assets/8084aa38-a00e-45fd-831f-9dea430aaec7" />

<img width="480" height="503" alt="Image" src="https://github.com/user-attachments/assets/3d984800-04d3-40d1-9203-72c3cd488aa9" />

[...]

PR #2221, opened 2026-08-12
A monotonic per-fingerprint occurrence total on alerts (#2216)

### LINKED ISSUE #2216: [FEATURE] Emit a monotonic per-fingerprint occurrence counter on alerts (the current count is a rolling-window gauge, so throttled consumers can't recover the true total)

- [ ] Lite
- [x] Darling
- [ ] Documentation

### Problem Statement

The deadlock/blocking alert value is a rolling 1-hour count (GetRecentDeadlocksAsync(hoursBack:1)), and webhook delivery is throttled per-fingerprint by the #1154 cooldown.

A downstream consumer therefore cannot reconstruct how many times an incident actually occurred: it's a gauge, not a counter, so summing delivered values over-counts overlapping windows and under-counts throttled/aged-out ones. 

Real example: a deadlock fingerprint occurred 5×; webhooks delivered at 14:41 (Occurrences 2) and 15:56 (Occurrences 1) while the re-fires between (Value 4) went to the tray only and then aged out of the 1-hour window. Our Azure DevOps ticket reads 3, not 5. The severity scales with the cooldown: at a ≥1-hour delivery interval, essentially every occurrence between deliveries is lost.

Image

Image

Image

Image

### Proposed Solution

Add a monotonic cumulative occurrence count per fingerprint to the alert payload (e.g. an Occurrences (total) fact / incident field) - total occurrences of that dedup key for the life of the active incident, independent of the rolling window and of how many deliveries the cooldown suppressed.

A consumer then SETs its count to that value on each delivered webhook and is exact regardless of throttling or window aging. The engine already keeps per-fingerprint state for the #1154 cooldown and the RollingCountAlertGate watermark, so the running total is already in reach. (Pairs with the dedup_key MCP filter — both surface per-fingerprint truth to downstream tooling.)

### Use Case

We turn each alert into an ADO work item and need its occurrence count to be trustworthy for triage ("this deadlock has hit 5× since 14:41") - especially at longer cooldowns, where the rolling gauge otherwise drops most of the signal. A cumulative counter makes the ticket accurate on the next delivered webhook without shrinking the cooldown (which would just restore alert spam).

### Alternatives Considered

_No response_

### Additional Context

 Darling (shared alert engine — Lite too) · Version: 3.4.0-nightly.20260812

erikdarlingdata:
Closes #2216.

## What @gotqn reported

The `Occurrences` fact on a grouped incident is a rolling-window **gauge** — it counts events inside the collectors' one-hour read window, so it rises as events arrive and falls as they age out. A consumer that only sees throttled deliveries (one per #1154 per-fingerprint cooldown) cannot recover the true total from a series of those readings: a 3 followed by a 3 is indistinguishable from *nothing happened* and *three happened while three aged out*.

## What this adds

Blocking and deadlock incidents now also carry **Total Occurrences** — accumulated per dedup fingerprint over the life of the incident — plus **Incident Since**, so a total that restarts can be told from one that continues. Both ride every surface that iterates the alert details (Teams facts, Slack fields, email, in-app dialog) and persist in the history row's context JSON.

## Decisions worth reviewing

**The existing `Occurrences` fact is deliberately untouched.** Downstream automation keys on fact *names*, so redefining that one from a gauge to a total would have silently changed the number every current consumer reads. The two answer different questions and both are emitted.

**Keyed per fingerprint, not per (server, metric).** A deadlock over one object set and a deadlock over another are separate incidents; pooling them would report each one's total as the sum of both. This is also what the issue title asks for.

**Two watermarks, not one.** `RollingCountAlertGate`'s watermark advances only when an alert *fires*, so an event arriving during a cooldown is still reported later. The accumulator's mark advances on every observation, so the same event is never counted twice. Sharing one mark would break whichever job lost the argument. Both decay as events age out, or a gauge falling and rising back to the same level would read as nothing new.

**Its own table in both stores, not columns on `config_edge_trigger_watermarks`.** Two independent reasons: the key is wrong (above), and Lite writes that row with `INSERT OR REPLACE` over a *partial* column list, which resets every unlisted column to its default — a counter living there would zero itself on every fired alert, i.e. exactly when it is read. There's a test pinning this so a later "simplification" fails loudly.

**Accumulation happens inside the context builders, via a pre-render hook.** Not a preference: the builders render each incident's facts in the same pass that produces the incidents, so a total attached to the finished context would never reach the alert anyone reads.

**V61 is `max(dev) + 1`.** #2211 and #2213 also claim 61+, which makes this a *loud* rebase conflict — the safe failure. A gapped rung merges quietly and then silently skips every rung the other branches still intend to fill, because the ascent test is `version <= current`. I'll renumber on rebase if either lands first.

## The honest bound

The total is a **lower bound** on occurrences, exact whenever the read window outlives the gap between deliveries — which is the reported case, since an occurrence arriving during a cooldown is still in the window when the next delivery observes it. An occurrence that both *arrives and ages out* between two deliveries cannot be recovered from a window gauge by any arrangement; that needs an event-identity watermark at the collector, which is a different feature. This is stated in the class remarks and the CHANGELOG rather than glossed, because a counter that quietly undercounts is worse than a gauge that is honestly a gauge.

Related: a row stranded by a host that died mid-incident is ignored once it is older than the read window. Without that rule its high mark would decay to the new window count, the recurrence would read as nothing new, and the alert would report a stale total under a stale start time — an undercount delivered with a confident timestamp. `RollingCountWindowHours` now replaces the two `hoursBack: 1` literals so the horizon cannot drift from the window.

## Testing

- **38 assertions verified locally by execution**, not just compilation. `Darling.Tests` is `net10.0-windows` and cannot run on macOS (`Microsoft.WindowsDesktop.App` is absent — confirmed, not assumed), so I ran the same assertions against the real `PerformanceMonitor.Alerting` build in a throwaway `net10.0` harness: the reported cooldown case, the aging-decay case, monotonicity across a sawtooth gauge, per-fingerprint separation, within-observation duplicates, the falling edge, staleness, a clock stepping backwards, the disabled horizon, all three renderer cases, and the context-JSON round trip. All passed.
- **Not executed locally, CI is the arbiter**: the four new `AlertEngineTests` cases (need the Windows xUnit host) and both store implementations. The Postgres path has no local runtime here — no `psql`, and the container daemon isn't running — so I hardened it instead: explicit array casts on the `unnest` binds, since an inference failure on that shape is a runtime error rather than a compile one, and a note on why the `<> ALL` prune is safe (the accumulator never admits a null fingerprint).
- Builds clean: Notifications, Alerting, Darling Storage/Service/Viewer, Darling.Tests, Lite, Lite.Tests.


erikdarlingdata:
Fixed — both findings were real, and the first one made this feature's own class doc false.

The accumulation now runs on **every sweep** that fetched rows, outside the `Fire` branch, and the builder's decorator is reduced to a pure lookup of what the sweep already computed. Your worked example is now a test in both directions: delivery-only observation gives 11, per-sweep gives 14.

Your framing of *why* it undercounts was sharper than mine, and I've taken it into the docs. The loss is one occurrence per event that **ages out** between two observations, because a retirement and an arrival cancel in the gauge. My original bound talked about events that arrive *and* age out between observations — the narrower claim, which is what let the delivery-cadence wiring look defensible. It wasn't.

Finding 2 is fixed by construction rather than by patching the cap: the sweep's grouping is uncapped, so a render budget can't evict occurrence state. A 14-fingerprint case pins it.

Two things that fell out of the fix and weren't in your review:

- The grouping is now **shared** between the counting path and the rendering path (`BlockingIncidents`/`DeadlockIncidents` over the same filter and grouper the builders use). Had the exclusion filter drifted between them, the two would key fingerprints differently and every delivered incident would read as a first contact — a quieter version of the same bug.
- Per-sweep observation would otherwise put a store write on every metric of every server on every sweep (~12k/hour on a 52-server fleet at 30s). `Changed` is now substantive-or-heartbeat: a flat gauge is not a write, but a live incident refreshes at half the staleness horizon, so it can never sit long enough to judge *itself* stale.

One change worth flagging rather than burying: the falling-edge clear at the `Cleared` branch is now conditional on `rows is null`, because an observed empty window already clears itself through the replace-the-set contract. Rows are null only when the gate is disabled — a fetch failure returns before reaching there — so turning the alert off still drops the counters.

erikdarlingdata:
Second round noted — and thank you for confirming the accumulation math independently; that was the part I could only verify outside the xUnit suite here.

On the double grouping: you're right that it happens, and I'm **not** fixing it in this PR. Filed as #2222 instead, with the options and the measurement that should decide between them.

The reasoning, so it's on the record rather than implied. Firing sweeps are rare by construction — edge-triggered plus cooldown-gated, so a server fires once per cooldown at most, and on that sweep the engine is already building a context, rendering details and sending email/webhook. The second parse is small against that. Meanwhile the fix widens a builder signature shared with Lite, in the exact seam this PR just spent a round repairing a correctness bug in, and I'd rather not thrash it for a cost nobody has measured yet.

Your unflagged note is the one I'd weight higher, and I agree with your framing: the unconditional per-sweep grouping is new steady-state load, not alert-time load. It's the price of the counter being correct — observing only at delivery time is what produced the undercount you caught — but it is a real change in what the collector host does when nothing is wrong, and it deserved saying out loud rather than living only in a class comment. It's now in #2222's framing too, with the honest note that it is the larger of the two costs.

Cheap mitigation already in place: the grouping is skipped entirely when the gate is disabled or the fetch failed, and the accumulator's substantive-or-heartbeat `Changed` keeps the per-sweep observation from turning into a per-sweep store write.

erikdarlingdata:
**Merge-order note.** #2224 (`fix/2171-plan-xml-compression`, opened after this) takes `Migration(62, …)` and `SchemaVersion = 62` with **no V61 in its ladder**, while dev is at 60 and this PR is the V61.

If #2224 merges first, dev sits at 62 with a hole at 61, every store stamps 62, and this PR's V61 is then `<= currentVersion` — silently skipped forever. `config.incident_occurrences` would never be created on an existing store, the accumulator would fall back to reporting the window count as the total (the exact behavior #2216 was filed about), and the viewer's probe would cap at 60 against a `RequiredStoreSchemaVersion` of 62, so the connect-time gate would refuse a healthy store.

So: **this should merge before #2224, or one of us renumbers.** Raised it on #2224 with the options; happy to move this to 62 (or 63) instead — whichever that lane prefers. Nothing to do here unless the answer is "you move".

erikdarlingdata:
Migration-number traffic, before this merges: this branch's V61 (incident-occurrence-counters) DUPLICATES #2211's V61, which was claimed on its thread first and is CI-green. V62 is taken by #2224 (open, gated). Rule stays what it has been: first of the V61 claimants to actually merge keeps 61; the other renumbers to max(dev)+1 at its own merge - and as of #2227 (density + duplicate pins) both wrong outcomes fail CI at rebase instead of shipping a stranded-store window. Nothing to do right now unless you merge second; then bump the rung and SchemaVersion together.

PR #2225, opened 2026-08-12
PostgreSQL blocking chains: the pg_blocking collector (stacked on #2213)

erikdarlingdata:
Stacked on #2213 — base is `feat/postgres-target-connection-seam`, so this diff is just the collector. Merge #2213 first; V69 depends on V61–V68 existing.

## What this adds

An eighth PostgreSQL collector, `pg_blocking` → `collect.pg_blocking_edges`. Rung **V69**, `SchemaVersion` **69**, 1 min / 30 d.

It stores an **edge list** — one row per (blocked, blocking) pair — rather than a rendered tree, so root blocker, chain depth and fan-out become ordinary SQL over the stored rows instead of string parsing. Both sides of every edge carry their own state, because that is what decides the remedy: an `idle in transaction` root is an application defect while an `active` root is a query-tuning problem, and the pid alone does not tell you which. Capturing only pids is the usual gap in homegrown PostgreSQL blocking monitoring — you get a number, go looking, and by then it is gone.

Runs on any PostgreSQL target **including standbys**, deliberately unlike `pg_autovacuum_stats`. That collector gates off replicas because `pg_stat_user_tables` reports zeros there; `pg_stat_activity` reports the standby's own backends, and recovery conflicts are blocking that happens nowhere else.

## The caveat is part of the feature

**This is a sample, not an event log.** SQL Server's blocked-process report is written by the engine when blocking crosses a threshold, so evidence exists whether or not anyone was looking. PostgreSQL records nothing unless something asks — so blocking shorter than the interval is never seen.

`get_pg_blocking` therefore reports its own capture counts in every response, sourced from `collection_log`, which can tell "no blocking" from "not collected" where the edge table cannot (both are an absence of rows). There is a distinct `not_sampled` status for a window the collector never ran in. An empty answer here should never be readable as an all-clear.

## Cost

`pg_blocking_pids()` takes ShareLock on the lock manager partitions per call, so it is evaluated only where `wait_event_type = 'Lock'` — as a `CASE` in the select list rather than a `WHERE` filter, so the gate bounds the cost while one query still returns the whole activity snapshot and the blocker's own state comes back without a second round trip. On a 5,000-connection instance the ungated form is the monitoring query that becomes the incident. Two tests guard this, and they guard *cost*, not correctness: the ungated version would collect perfectly correct data and hurt.

## Cycles — the thing the design note missed

The chain read finds a root **by absence**: a backend that blocks something and is not itself blocked. In a lock cycle every participant is blocked, so there is no root, so the entire cyclic component was **silently dropped** — zero rows from a capture that recorded real blocking. For a collector whose whole premise is that an empty answer must never mean "nothing happened", that was the one place the read did precisely that.

Cycles are rare but reachable: PostgreSQL's deadlock detector resolves them only after `deadlock_timeout` (1 s default), and a capture can land inside that window — after which the stored edges are the only evidence that will ever exist, because the engine kills a participant a moment later.

Fixed with a second read that finds participants by **reachability from themselves**, not by "this capture has no root" — the latter would miss a cycle sharing a capture with an ordinary chain. Reported alongside `chains`, always present so its absence can never read as "not checked", with a dedicated `cycles_only` status.

## Three defects found by running the real strings on real Aurora

None of these was reachable from the C# suite, which is the same lesson as #2213's review.

1. **`current_setting('track_activity_query_size')::int` throws.** `current_setting()` renders a memory GUC **with its unit** — the value comes back `'8kB'` on stage 17.7 and `'4kB'` on 16.11. The cast raises invalid-input-syntax, which fails the **whole collection**, every cycle, not just that column. Now `pg_size_bytes(current_setting(...))`, which parses every unit form. Pinned.
2. **The chain query needed `WITH RECURSIVE`, not `WITH`.** PostgreSQL scopes `RECURSIVE` to the entire `WITH` clause, not to the CTE that self-references, so it failed with `relation "chain" does not exist` — at runtime, on the first `get_pg_blocking` call, invisible at build time. Pinned.
3. **Three output columns had no alias** and came back all named `coalesce`. Harmless to the positional C# reader; unusable in psql, which is the tool anyone debugging this query will reach for. Every output column of both reads is now aliased.

## Two invariants nobody was checking

Both surfaced while adding the collector, and both are broader than it:

- **`EveryPostgresRung_IsIdenticalToTheGeneratedSchema`** diffs **all eight** PostgreSQL rungs against `PgSchemaGenerator`. A fresh store's tables come from V1's generated schema and an existing store's come from the rungs. The suite asserted the generator emits every table, and that each rung is well-formed — never that the two said the *same thing*. One drifted column type would have been a permanent, invisible divergence between the two populations of store, with reads that work against one and fail against the other.
- **`StoreSchemaProbe_ColumnCount_MatchesTheMapArity`** couples the probe SQL's column count to `MapProbedSchemaVersion`'s parameter count, which are read positionally. A sentinel without a parameter makes a fully-migrated store report one rung short (so the viewer refuses a healthy store); a parameter without a column throws `IndexOutOfRange` at connect time. Both are runtime-only against a live store, and the existing arity test builds its arguments from the signature, so it agreed with itself either way.

## Verification — all PASS on stage Aurora 16.11 and 17.7

- **`laddercheck`** (new harness): all 8 rungs identical to the generated schema.
- **`probe_blocking_collector_live.py`**: runs the **shipped** collector string against a **real two-deep blocking chain**, built with **advisory locks**. That is what keeps the probe strictly read-only while still producing genuine lock-manager blocking — `pg_blocking_pids()` reports advisory waiters and their `wait_event_type` is `Lock`, identical to a row or table lock, but nothing is written and every lock releases on disconnect. A syntax-only check would have exercised none of the parts that can be wrong, because a healthy instance returns zero rows. **17 checks**, including that the synthetic `backend_start`+pid identity is *recomputable* rather than merely present.
- **`probe_blocking_reader_sql.py`**: the reader text over synthetic edge sets with known answers — a 3-deep chain (depth 3, 1 direct victim, 3 total), a wide fan-out (depth 1, 3 direct), recurrence across two captures for one backend id, and the cycle. **21 checks**, including that neither acyclic scenario is *mis*-reported as a cycle: a false deadlock claim is worse than a missed one, because it prescribes lock ordering as the fix when that is not the fix.
- `pincheck` and the other five harnesses clean; full solution builds with 0 errors.

## Registration

Catalog, schedule defaults, worker dispatch, migration ladder, viewer probe sentinel + map arm, reader, MCP tool, web descriptor + read dispatch, Gemini tool registration, Lite's Darling-only tool list, the viewer coverage allow-list, and the CI worker sizing (**52/63**, derived from 50 hypertables) in both workflow files.

## Review notes

- `get_pg_blocking` collides by name with Lite's existing `get_blocking`. They are **not** twins and the pin list says so: Lite's reads an engine-recorded event with a graph, this reads periodic samples of an edge list.
- `pg_blocking_edges` is on the viewer allow-list rather than the Blocking tab. That tab is built on `blocked_process_report` XML — an engine-recorded event with a threshold — and rendering samples through a surface labelled as an event log would misrepresent coverage.
- The recursion depth cap of 32 in both reads is a cycle guard, not a chain-length opinion.


erikdarlingdata:
Stack note for the renumber round: when #2213 shifts to V63-V70 (per its round-2 review), this branch's rungs above it shift with it — renumber the whole stack in one pass and bump SchemaVersion once, at the top. #2227's density pin plus the existing duplicate pin will fail CI on any miss, so the check is mechanical now. Substantive review of this PR follows once #2213's four blockers land and it stabilizes underneath.

erikdarlingdata:
#2213 is merged (dev 922d99d5, SchemaVersion 70). This stack now rebases onto dev directly; its rungs renumber to max(dev)+1 at ITS merge - note that #2211 (QS plan map, back in build under my takeover) will also claim the next free rung, so whichever of the two merges second renumbers again. The #2227 density and duplicate pins referee both directions mechanically, so nothing needs pre-coordination beyond rebasing before merge. Substantive review of this PR starts once it is rebased onto the merged base.

PR #2227, opened 2026-08-12
Pin the migration ladder dense above the historical V45 hole (#2226)

### LINKED ISSUE #2226: [TEST] Pin the migration ladder as dense above V45, so a gapped rung fails at authoring time
Raised by a live near-miss between two open PRs (#2221 at V61, #2224 at V62 with no V61 while dev sat at 60).

## The failure a gap causes

`PgMigrations.MigrateAsync` ascends with `if (migration.Version <= currentVersion) continue;`. A rung at or below the store's stamped version is **silently skipped** — no error, no log, nothing to notice.

That is harmless for a hole nobody fills: V45 has been vacant for many releases, and `MigrationLadderPins.TheLadder_IsStrictlyOrdered_WithNoDuplicates` deliberately permits it ("a gap is harmless — the stamp comparison is `>`, not sequence arithmetic").

It is NOT harmless when a later branch fills the hole. Two branches off dev-at-60, one taking 61 and one taking 62, are compatible in exactly one merge order:

- 61 then 62 → dense, fine.
- 62 then 61 → every store stamps 62, and the 61 rung never runs on any of them. Its table is never created, whatever reads that table fails permanently, and if the viewer has a probe arm for it the connect-time gate refuses a healthy store while reporting a version mismatch that no upgrade can fix.

**A collision is loud** (rebase conflict on the same `Migration(N, …)` line). **A gap is quiet**, and quiet is the expensive one — it surfaces in the field, on stores that migrated during the window, and `DROP`/re-migrate is not an option on a production monitoring store.

## The pin

One assertion: every version above the historical V45 hole is contiguous.

```csharp
[Fact]
public void TheLadder_IsDenseAboveTheHistoricalGap()
{
    /* V45 is the one sanctioned hole (see TheLadder_IsStrictlyOrdered_WithNoDuplicates). Above it the
       ladder must be contiguous: a NEW gap is a rung some other branch can fill later, and a rung filled
       after a store has already stamped past it is skipped silently and forever. */
    var above = PgMigrations.Scripts.Select(m => m.Version).Where(v => v > 45).OrderBy(v => v).ToList();
    var expected = Enumerable.Range(above[0], above.Count).ToList();
    Assert.Equal(expected, above);
}
```

It would have failed on the gapped branch at authoring time, in the author's own `dotnet test`, instead of being caught by eye during merge coordination. It cannot false-positive on the sanctioned V45 hole, and it forces the max(dev)+1 discipline that is currently only a convention.

## Caveat worth stating

This makes two branches that both take max(dev)+1 **collide** rather than coexist — which is the point. A rebase conflict is a thirty-second fix by the second author; a silently-skipped rung is a field incident.

Also: it means a branch cannot pre-reserve a rung above an unmerged sibling's. That is a real ergonomic cost, and it is the correct trade — reserving a number is exactly how the gap gets created.

---

Filed as a follow-up per the repo's deferral rule; the near-miss it came from is #2221/#2224.

erikdarlingdata:
Closes #2226.

One new pin in `MigrationLadderPins`: every ladder version above the historical V45 hole must be contiguous. The applier ascends with `version <= currentVersion ? skip`, so a rung filled AFTER a store has stamped past it is skipped silently and forever — no error, no log, and no upgrade can repair the store. A collision is loud (rebase conflict, and the existing no-duplicates pin); a gap is quiet, and quiet is the one that surfaces in the field. This makes a gapped rung fail at authoring time in the author's own `dotnet test`, which converts the current merge-order coordination (a convention held together by PR-thread comments) into a build-time invariant: every branch takes max(dev)+1, re-verified at its own merge.

The sibling test's "density is deliberately NOT pinned" comment is updated to match — V45 stays sanctioned (a hole nobody fills is harmless), new gaps are this pin's job.

## Why now

The near-miss is live, and wider than #2226 knew: dev's MAX is 60, **#2211 and #2221 both carry V61 today** (qs-plan-fetch and incident-occurrence-counters — a duplicate the ordering pin will catch at whichever rebase happens second), #2224 carries V62, and #2213/#2225 carry V61-V68+. Whichever merge order actually happens, this pin plus the existing duplicate pin make every wrong outcome fail CI instead of shipping a stranded-store window through a nightly.

Effect on the in-flight PRs, stated plainly:

- **#2224 (V62, mine)** becomes mechanically gated: red until a V61 lands on dev, exactly matching the merge gate already written in its body. Endorsed — that gate should be CI, not prose.
- **#2211 / #2221 (both V61)**: first to merge keeps 61; the second renumbers at its own merge (the duplicate pin makes forgetting loud).
- **#2213 / #2225 (stacked)**: renumber to max(dev)+1.. at merge, as already agreed on the #2213 thread.

## Evidence

- On dev's ladder (46..60 dense above the hole): all three pins pass.
- Watched-red on the real hazard instance — the V62-carrying #2224 branch with no V61: the pin fails with `Expected: [..., 60, 61] / Actual: [..., 60, 62]` at the exact position. Restored after observing.
- Full-rebuild warning census unchanged: exactly the tracked pair (CA1068 x2, CA2022 x2).

Test-only change; no CHANGELOG entry per repo precedent for pin/infra commits.

### DIFF
--- a/Darling/Darling.Tests/MigrationLadderPins.cs
+++ b/Darling/Darling.Tests/MigrationLadderPins.cs
@@ -58,11 +58,27 @@ public sealed class MigrationLadderPins
     {
         /* The replay math above only holds when versions are strictly increasing and unique — a
            duplicated or out-of-order version would let two rungs disagree about what "already ran"
-           means. Density is deliberately NOT pinned: the ladder has one historical gap (V45) and a
-           gap is harmless (the stamp comparison is >, not sequence arithmetic). Cheap to pin,
-           catastrophic to debug from a half-migrated field store. */
+           means. The historical V45 hole stays sanctioned (a gap NOBODY fills is harmless — the
+           stamp comparison is >, not sequence arithmetic); new gaps are the next pin's job. Cheap
+           to pin, catastrophic to debug from a half-migrated field store. */
         var versions = PgMigrations.Scripts.Select(m => m.Version).ToList();
         Assert.Equal(versions.OrderBy(v => v).ToList(), versions);
         Assert.Equal(versions.Count, versions.Distinct().Count());
+    }
+
+    [Fact]
+    public void TheLadder_IsDenseAboveTheHistoricalGap()
+    {
+        /* #2226: a NEW gap is a rung some other branch intends to fill later — and the applier
+           ascends with `version <= currentVersion ? skip`, so a rung filled AFTER a store stamped
+           past it is skipped silently and forever: its objects never exist, readers of them fail
+           permanently, and no upgrade can repair the store. That exact window nearly opened between
+           two in-flight branches (one at V61, one at V62 while dev's MAX was 60; nightlies ship
+           from dev, so the window is real). Density makes the hazard fail at AUTHORING time, in the
+           author's own test run: every branch must take max(dev)+1, and two branches that both do
+           so collide loudly at rebase instead of coexisting into a field incident. V45 is the one
+           sanctioned hole, vacant for many releases and filled by nobody. */
+        var above = PgMigrations.Scripts.Select(m => m.Version).Where(v => v > 45).OrderBy(v => v).ToList();
+        Assert.Equal(Enumerable.Range(above[0], above.Count).ToList(), above);
     }
 }


erikdarlingdata:
Filed the issue, so I checked the implementation rather than just endorsing it.

The pin is right, and it passes on the real ladder: versions above 45 are 46–61, 16 rungs, contiguous, so `Enumerable.Range(above[0], above.Count)` matches. Confirmed independently against `PgMigrations` (60 rungs, 1–61, exactly one hole at 45) before filing, and the watched-red you got on #2224's branch is the other half of the proof.

You caught something I left broken in the proposal: the sibling test's comment said density is *deliberately* not pinned. Had that stayed, the two tests would have contradicted each other in the file where someone goes to learn the ladder's rules, and the older comment reads more authoritative. Updating it to "the historical V45 hole stays sanctioned; new gaps are the next pin's job" is the part that makes this land as one coherent rule instead of two tests arguing.

Your correction on #2226 also stands: the near-miss was wider than I described, because **#2211 carries V61 too** — both claimants are mine. I've flagged that on #2211 for whoever revives it, and corrected my #2224 comment, which had told that lane there was a single V61 claimant.

One thing I deliberately did not do, for the record: I didn't add this pin to #2221 myself, because landing a test that fails another lane's open PR while simultaneously asking that lane to renumber would have been forcing the outcome. Coming from your side, with the red demonstrated first, is the right way round.

Issue #2246, opened 2026-08-13
[BUG] FinOps provisioning status reports UNDER_PROVISIONED for every healthy server: the memory ratio compares Total to Target Server Memory, which converge at 1.0


erikdarlingdata:
Surfaced by #2231, where a user running both products side by side sees every server as `UNDER_PROVISIONED` in Darling and `RIGHT_SIZED` in the Full Dashboard.

## Measured

The rule in `ViewerDataService.FinOps.Utilization.cs`:

```csharp
else if (p95Cpu > 85 || memRatio > 0.95m)
    status = "UNDER_PROVISIONED";
```

with

```sql
CAST(total_server_memory_mb AS DECIMAL(10,2)) / NULLIF(target_server_memory_mb, 0) AS memory_ratio
```

Against a 42-server production fleet, latest `memory_stats` row per server:

```
servers=42   ratio_gt_0.95=42   ratio_eq_1=3
median=1.0000   min=0.9997   max=1.0002
```

**Every server trips it.** Not a threshold that is slightly too tight — the metric has no useful spread at all. Four decimal places of noise around 1.0.

## Why the metric cannot work here

`total_server_memory_mb` and `target_server_memory_mb` are the perfmon **Total Server Memory (KB)** and **Target Server Memory (KB)** counters. Target is the memory SQL Server wants; Total is what it currently commits. A warmed instance that is not under external pressure sits at Total == Target, so the ratio is 1.0 whenever things are *healthy*.

The trigger is therefore also backwards. Memory pressure appears as Total **below** Target — SQL Server wanting more than it has — which is a ratio *under* 1. As written, the rule reads the healthy steady state as starvation and would go quiet during actual pressure.

## Divergence from the Full Dashboard, which does not have this problem

`report.finops_utilization_efficiency` (`install/54_create_finops_views.sql`) computes the same-named ratio from a different pair:

| | numerator | denominator |
|---|---|---|
| Full Dashboard | `SUM(pages_kb)` from `sys.dm_os_memory_clerks` | `committed_target_kb` from `sys.dm_os_sys_info` |
| Darling | perfmon Total Server Memory | perfmon Target Server Memory |

Memory-clerk pages against the committed target normally lands well below 1, which is why the old verdict behaves sensibly. The CPU half is identical in both (24-hour window, p95 over `cpu_utilization_stats`), so CPU is not implicated.

## Second, separate divergence

[...]

Issue #2234, opened 2026-08-13
get_perfmon_trend: delta_value returns a fabricated 0 on ~38% of points while cntr_value advances


erikdarlingdata:
## Summary

`get_perfmon_trend` returns `delta_value = 0` on a large fraction of points while the cumulative `value` on those same points advances normally. Any rate derived from `delta_value` is wrong, and — because a spurious 0 is indistinguishable from a real idle interval — it is wrong in a way that reads as a finding.

## Observed

`prod-sql-use2-alpha-01`, counter `Batch Requests/sec`, `hours_back: 168`:

- 2,468 points returned
- **931 points (37.7%) had `delta_value == 0`**
- 0 points had a negative delta; 0 counter resets (`value` never decreased)
- Degrades sharply after the first day of the window:

| day | points | zero-delta |
|---|---|---|
| 08-06 | 683 | 2 |
| 08-07 | 493 | 122 |
| 08-08 | 225 | 192 |
| 08-09 | 319 | 123 |
| 08-10 | 230 | 126 |
| 08-11 | 217 | 156 |
| 08-12 | 224 | 149 |
| 08-13 | 77 | 61 |

A contiguous run where every `delta_value` was 0 while `value` advanced ~55,000 over ~70 minutes:

```
2026-08-11T13:06:47   20,056,930   delta 0
2026-08-11T13:12:33   20,060,956   delta 0
2026-08-11T13:19:10   20,066,328   delta 0
2026-08-11T13:28:31   20,072,557   delta 0
2026-08-11T13:36:14   20,077,796   delta 0
2026-08-11T13:45:12   20,083,510   delta 0
2026-08-11T13:51:40   20,087,811   delta 0
2026-08-11T13:59:17   20,092,702   delta 0
2026-08-11T14:08:34   20,100,946   delta 0
2026-08-11T14:17:02   20,111,780   delta 0
```

## Why this matters — it inverted a real result

I was testing whether a replica's CPU spikes were driven by request volume. Using `delta_value` as returned:

- "batches/sec p50 = 0.0" on four of seven days
- Pearson r between throughput and CPU = **−0.427** (negative)

Recomputing the rate myself from consecutive `value` readings, same data:

- batches/sec p50 = 10.3–12.2 every day, avg 11.1, p95 13.4
- Pearson r = **+0.304**

[...]

Issue #2233, opened 2026-08-13
[BUG] Every delta collector's 300s gap guard trips on ordinary sweep jitter — 21-45% of deltas silently written as 0


erikdarlingdata:
Reported against `get_perfmon_trend` (`delta_value` zeroed while `value` is fine). Reproduced, and the cause is **not** perfmon-specific and **not** an arithmetic error — it is one shared gap guard, and it affects every delta-based metric in the product including wait analysis.

## The measurement

`get_perfmon_trend('Batch Requests/sec')`, prod-sql-use2-alpha-01, 3h window, 38 intervals. Correlating each stored `delta_value` against the wall-clock gap to the previous collection:

```
ZEROED  n=17   gaps 300.9s – 441.7s
CORRECT n=21   gaps 149.4s – 293.6s
```

A clean split at **300 seconds**, with no overlap. And every non-zero delta equals the true difference of the adjacent `value`s **exactly** (verified all 21). So the delta computation is right; the value is being thrown away afterwards.

Same box, same window, `get_wait_trend('SOS_SCHEDULER_YIELD')`: 6/28 zeroed, gaps 302–357s, present 150–294s — **the same 300s boundary, at the same timestamps**. One cause, not two bugs.

| collector | samples | zeroed | rate |
|---|---|---|---|
| perfmon_stats | 38 | 17 | **45%** |
| wait_stats | 28 | 6 | **21%** |

## Mechanism

Every delta call site passes `maxGapSeconds: 300` — 36+ of them across `PerfmonStatsCollector`, `WaitStatsCollector`, `FileIoStatsCollector`, `QueryStatsCollector`, `ProcedureStatsCollector`, `LatchStatsCollector`, `SpinlockStatsCollector`, `MemoryGrantsCollector`, and both PostgreSQL collectors (`PgWaitStatsCollector`, `PgStatementStatsCollector`).

`CollectorDeltaCalculator.CalculateDeltaWithInterval` discards the delta when the gap exceeds it:

```csharp
if (maxGapSeconds > 0 && collectionTime.HasValue && previous.Timestamp.HasValue
    && (collectionTime.Value - previous.Timestamp.Value).TotalSeconds > maxGapSeconds)
{
    delta = 0; interval = 0;
    return (currentValue, collectionTime);
}
```

[...]

Issue #2205, opened 2026-08-11
Lite's query_store state prune logs a count where Darling names the pruned keys


erikdarlingdata:
Diagnostic parity residue from #2195, dispositioned at merge time. Darling's PruneOrphanedQueryStoreDatabaseStateAsync uses DELETE ... RETURNING s.state_key so the log line names exactly which per-database state rows were retired; Lite's DuckDB twin logs only the aggregate count across the owner/prefix loop. Correctness is identical (the NOT EXISTS anti-join and freshness guard match semantics exactly, pinned by the #2195 tests); the gap is incident forensics - on Lite you cannot tell FROM THE LOG which databases' watermark/backfill state was retired. Fix shape: DuckDB supports DELETE ... RETURNING; switch the loop to ExecuteReader, collect state_key values, log them the way Darling does. One file, one test asserting the log names the keys, watched red.

erikdarlingdata:
Shipped in #2236 (merged `7dfa9c50`).

`Lite/Services/RemoteCollectorService.QueryStoreBackfill.cs:557` now ends the prune with `RETURNING state_key` and reads the results back, so the log names exactly which per-database watermark/backfill rows were retired — the same forensics Darling's `PruneOrphanedQueryStoreDatabaseStateAsync` gives. The in-code note at :580 records why it's a reader and not a rows-affected count.

Pinned by `Lite.Tests/QueryStoreStatePruneTests.cs`, which states the defect in its own words: *"a COUNT where Darling names the retired keys, so on Lite you could see that three rows went without seeing WHICH databases' watermark and backfill state was retired."*

Closing: diagnostic parity restored, and the semantics the #2195 tests pin are untouched.

PR #2260, opened 2026-08-14
CHANGELOG: backfill the five shipped changes that had no entry

erikdarlingdata:
Docs only — no code touched.

Five changes merged to `dev` this week shipped without a CHANGELOG entry. Backfilled:

| issue | section | what shipped |
|---|---|---|
| [#2246] | Fixed | FinOps provisioning verdict rebuilt on measured signals; the shared `ProvisioningVerdict` replacing six copies |
| [#2233], [#2234] | Fixed | `DefaultMaxGapSeconds` 300 → 3600, the counter-reset interval invariant, and the real `sample_interval_seconds` on `perfmon_stats` |
| [#2150] | Fixed | `tempdb_stats` gated off Azure SQL Database (error 262 in tempdb, unfixable by configuration) |
| [#2201] | Fixed | `Get-NetworkPathKind` no longer fails open to "local" without WMI |
| [#2205] | Changed | Lite's Query Store prune names the retired state keys instead of counting them |

Each entry leads with the **measurement**, not the change, because that is the part that makes a threshold reviewable a year from now: 42 of 42 servers above a 0.95 ratio whose median is 1.0000; a 300 s guard against a 299 s median gap across 99,717 samples; `VIEW DATABASE PERFORMANCE STATE` in tempdb that no grant on Azure SQL DB can confer.

Also fixed an attribution I got wrong while drafting: #2246 was filed in-repo, not by a field reporter — it came out of the follow-up on the [#2150] field report, and now says so.

Link definitions added for #2246, #2233, #2234, #2150, #2201, #2205 and #2195.

Verified: CRLF preserved line-for-line (the only two LF lines in the file are pre-existing and untouched), no whole-file renormalization, diff is 17 insertions and 0 deletions, and the `3.4.0` headings below are outside the edited range.

[#2246]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2246
[#2233]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2233
[#2234]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2234
[#2150]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2150
[#2201]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2201
[#2205]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2205
[#2195]: https://github.com/erikdarlingdata/PerformanceMonitor/issues/2195


erikdarlingdata:
Good catch, fixed in `3f36448e`.

Cause: my insert appended `entry + blank` after each heading. `### Fixed`'s first entry sits **directly** under its heading, so that produced exactly one separator; `### Changed` already carried a blank line before its first entry, so the same append produced two. Only the Changed insert was affected.

Verified by scanning the whole Unreleased section for consecutive blank lines: one remains, at the very end, and `git show dev:CHANGELOG.md` confirms it predates this branch — it is the separator before `## [3.4.0]`. Left alone.

erikdarlingdata:
On the second half of that finding — **the `### Fixed` block does not have the problem.** I checked rather than assuming, since the same insert produced both.

Scanned the whole Unreleased section for consecutive blank lines. Fixed, from its heading down:

```
41  ### Fixed
42  - **The FinOps provisioning verdict said UNDER_PROVISIONED for e...
43  BLANK
44  - **Half of every delta collector's output was a zero that never...
45  BLANK
46  - **`tempdb_stats` failed forever on Azure SQL Database, and cou...
47  BLANK
48  - **The installer's mapped-drive refusal failed open when WMI wa...
49  BLANK
50  - **The service now names the `darling.json` servers it is not m...   <- pre-existing
```

One blank line throughout, including the boundary with the pre-existing entry at 50.

The asymmetry is the whole cause: `### Fixed`'s first entry sat **directly** under its heading, so appending `entry + blank` produced exactly one separator; `### Changed` already carried a blank line before its first entry, so the identical append produced two. Only Changed was affected, which is why only Changed was fixed.

Exactly one double blank remains in the section, at the very end, and `git show dev:CHANGELOG.md` confirms it predates this branch — it is the separator before `## [3.4.0]`. Left alone.

Issue #1902, opened 2026-07-31
The other ~124 live-test finally blocks still clean up on the body's connection, so a failing test can still report its teardown's error


erikdarlingdata:
Found by the sweep done for #1896, and deliberately left out of that PR on size grounds rather than principle.

## What #1896 fixed, and the shape it left standing

#1896 converted the sixteen `finally` blocks in `TimescaleSupportTests` and `DarlingSecuritySplitLiveTests` off the body's connection and onto `LiveStoreCleanup`, so a test that fails mid-body reports its own exception instead of whatever its teardown said afterwards.

Counting the same shape across the whole `live-postgres` collection turned up **148** `finally` blocks that do not go through `LiveStoreCleanup`. Classified:

| Count | Shape | Verdict |
|---|---|---|
| ~124 | `DELETE` of the test's own sentinel/server rows on the body's connection | the same defect, smaller blast radius — **this issue** |
| 21 | own-store classes that mint and drop their own database | out of scope by design (the `#1776 own-store` exemption) |
| 3 | file/process teardown, or cleanup already opening its own connection | not store state |

The ~124 live in roughly 55 classes, typically one or two each: `await DeleteTestRowsAsync(connection)`, `CleanServerRowsAsync(dataSource, ...)`, `CleanupAsync(connection, ...)`.

## Why it is the same defect

A throw from a `finally` replaces the exception already in flight, and the connection these run on is the one the body's failure may have closed — so the teardown fails *because* of the thing it then hides. #1896 demonstrated it end to end: with a body failure and a cleanup failure forced into one test, the pre-change shape reported the cleanup's `42883` and lost the body's exception entirely.

## Why it is genuinely lower priority

The blast radius is much smaller. These delete rows keyed to the test's own sentinel server id; they do not drop continuous aggregates, hypertables, retention policies or cluster-wide roles, so an abandoned one leaves a handful of rows rather than an object that changes compose's tier routing or flips the #1784 coverage gate. #1873's residue check does not see them either, precisely because they are rows and not relations — which cuts both ways: nothing is currently watching, but nothing is currently breaking.

What is lost is diagnosis. A failing test in one of these classes can still report the wrong error, which is the expensive-to-notice failure mode: the suite stays red and simply says the wrong reason.

## The work

[...]

PR #2261, opened 2026-08-14
Cover the snapshot_now engine gate behaviorally — the last of #2230's three

### LINKED ISSUE #2230: [TEST] Behavioral coverage for the three PostgreSQL engine gates (reconcile, snapshot_now, analyze_now)
## What

The three engine gates added in #2213's round-3 fix (the long-query XE reconcile early-return, the snapshot_now dispatch predicate, and analyze_now's honest-tombstone arm) are covered by source-scanning pins and a live rig run (9 sweeps, zero engine-blind dispatches), but none has a BEHAVIORAL test that drives a PostgreSQL runtime through the gate and asserts the short-circuit.

## What a test needs

- analyze_now is the tractable one: fabricate the worker (GetUninitializedObject idiom, see CollectorMemoryKnobTests.SweepGate tests) with a live _postgres, a ServerLoopState with a PostgreSql runtime, invoke RunAnalyzeNowAsync, assert the analysis_state row carries the engine tombstone (not the generic insufficient_data) and the CommandOutcome is the success shape.
- The reconcile and snapshot gates need the same fabrication plus a fake runner; the observable is absence (no connection attempt), which wants a counting seam.

From #2213 round 3 (reviewer finding 7); filed per the deferral rule during release week.

erikdarlingdata:
Closes #2230.

The third dispatch loop, and the one that got neither engine gate in #2213's first round. #2237 covered `analyze_now`, #2245 covered the reconcile early-return; this is `snapshot_now`.

## The regression is phantom SUCCESS, not a crash

Dispatched SQL Server collectors do not fail loudly against a PostgreSQL target — their own `AppliesTo` early-returns, yields zero rows, and `RunOneAsync` then writes **SUCCESS** to `collection_log`. One operator snapshot produced ~40 rows saying collection worked, on an engine where those collectors cannot mean anything. That reads as health, which is strictly worse than an error.

## Neither of the other two tricks was needed

`analyze_now` asserts a presence in `analysis_state`; the reconcile gate exploits a public precondition that throws when ungated. `snapshot_now` needs neither: the loop reports `collectorsRun` in its outcome JSON, and every run it makes writes itself to `collection_log`. The gate's effect is a count and a set of rows.

## One collector, two engines

Schedule overrides disable everything except `wait_stats` — SQL-Server-only, dispatched through the same loop — so the two arms differ in the **engine** and nothing else: same collector, same overrides, same store. One dispatch decision instead of 49, which also makes the SQL Server arm cheap enough to be the non-vacuity proof rather than a second slow test. The overrides go through the shipped `StoreConfigProvider.ResolveSchedule`, which the loop consults, rather than reaching past it.

| arm | `collectorsRun` | `collection_log` |
|---|---|---|
| PostgreSQL target | **0** | **empty** |
| SQL Server target, same overrides | **1** | `wait_stats` |

**Why absence rather than "not SUCCESS":** a gate that dispatched and then failed would also avoid SUCCESS, while having connected to a PostgreSQL host as SQL Server. The row must not exist at all.

**Why the SQL Server arm is load-bearing:** without it, the gate assertion passes just as well against a snapshot that stopped dispatching anything for some unrelated reason — "ran zero collectors" is also what a broken loop reports.

## Notes

- Real `DarlingCollectorRunner`, not a stand-in: a fake would have to reimplement the dispatch the test exists to observe, and the `collection_log` write that carries the assertion happens inside the real path.
- Unresolvable hosts (`.invalid`) with a one-second connect timeout, so the dispatched collector fails immediately. The **status** is not asserted — whether the attempt lands ERROR or a zero-row SUCCESS depends on how far it gets before the connection dies, and the fact under test is that it was dispatched at all.
- Its own synthetic hosts, separate from the analyze_now cases, because it asserts on EMPTINESS — sharing a `server_id` would make that depend on test ordering.
- Cleanup through `LiveStoreCleanup` on its own connection, per #1902.

Live-store gated on `DARLING_TEST_PG`, which the "Darling PostgreSQL tests" job sets. Test-only — no production code touched. Builds clean (0 warnings) with `-p:EnableWindowsTargeting=true`; **I cannot run `net10.0-windows` xunit locally, so CI is the only place the pass is observable** — watching it there rather than claiming it here.

### DIFF
--- a/Darling/Darling.Tests/PostgresEngineGateBehaviorTests.cs
+++ b/Darling/Darling.Tests/PostgresEngineGateBehaviorTests.cs
@@ -8,6 +8,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
@@ -29,12 +30,14 @@ namespace Darling.Tests;
 /// scan cannot tell a gate that returns early from one that falls through and happens to write the same
 /// text.</para>
 ///
-/// <para><b>Two of the three doors.</b> The analyze_now gate's observable is a PRESENCE — a row in
+/// <para><b>All three doors.</b> The analyze_now gate's observable is a PRESENCE — a row in
 /// <c>analysis_state</c> carrying the engine tombstone — so it is asserted directly. The reconcile gate
 /// looked like it needed a counting seam to observe an absence, and that was wrong: the belt gate inside
 /// <see cref="DarlingXeSessions.ReconcileLongQueryCompletionsAsync"/> is public and its own precondition,
 /// so an ungated call THROWS and a gated one returns — a difference an assertion can see with no seam, no
-/// live store, and no network. snapshot_now remains uncovered; see #2230.</para>
+/// live store, and no network. snapshot_now needed neither trick: its dispatch loop already reports how many
+/// collectors it ran, and every run it makes writes itself to <c>collection_log</c>, so the gate's effect is
+/// a count and a set of rows rather than something that has to be inferred.</para>
 ///
 /// <para><b>The regression it guards is specific and was real.</b> Clicking "Generate now" against a
 /// PostgreSQL target used to run the full SQL-Server-shaped pass, find nothing, and persist the GENERIC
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Taken in `c2ad1ca5` — you're right, and it is the house idiom: `ViewerControlPlaneStage3bTests.cs:599` plus the `DeadStore` constants in three MCP test files all use `Host=127.0.0.1;Port=1`.

The thing worth naming is that the two hosts had been conflated. **Identity** must stay the synthetic `.invalid` name — it is what `ServerIdFor` hashes, what the `servers.Find` lookup matches, and what the `collection_log` rows key on, so it has to be something no real server can collide with. The **connection** should go somewhere that refuses immediately. Those are different requirements and they now have different values, with the reasoning in the doc comment.

Also switched the SQL Server arm off integrated auth to SQL auth: the failure under test is the connect, and integrated auth on a Linux runner fails for a *platform* reason instead — same red, different cause, and asserting on it would have been asserting on the wrong thing.

Re-running CI. For the record, the previous run did execute both tests rather than skipping them — `SnapshotNow_AgainstASqlServerTarget_DispatchesTheSameCollector` in 1 s and `SnapshotNow_AgainstAPostgresTarget_DispatchesNoSqlServerCollector` in 149 ms, which is the timing split the design predicts: the SQL arm pays for a connect attempt and the PostgreSQL arm dispatches nothing at all.

Issue #2254, opened 2026-08-14
[BUG] Servers added to darling.json after the first seed are silently ignored, and --test-connection still validates them


erikdarlingdata:
Field report #2252. An operator added a second server to `darling.json`, confirmed it with `--test-connection` (both servers PASS), restarted the service, and the server never appeared. Nothing in any log said why.

## Mechanism

`StoreConfigProvider` seeds the server list only when the table is empty:

```csharp
if (await CountAsync(connection, "config_monitored_servers", cancellationToken) == 0)
{
    await SeedMonitoredServersAsync(connection, config, now, cancellationToken);
}
```

After the first successful start the store is authoritative (`DarlingWorker`: "The initial server set comes from config_monitored_servers WHERE is_enabled = TRUE (post-seed…)"). So a server added to the file afterwards is a permanent no-op, and there is no re-read path.

That is a reasonable design. The problem is that **two supported commands disagree and neither says so**:

| surface | reads | reported |
|---|---|---|
| `--test-connection` | `darling.json` **directly** | "Validating connectivity to 2 server(s)" — both PASS |
| the service / Viewer | `config_monitored_servers` | 1 server |

Both are correct for what they consult. Together they tell an operator the server is configured and reachable, then don't monitor it.

## Fix shape

Cheapest useful change, and it needs no schema or behaviour change: **on every start, compare the file's server list against the store and log the difference.** Something like

> darling.json lists 2 server(s); the store has 1. Ignoring 'dcvs-bd01' — the store is authoritative after the first seed. Add servers with the Viewer's Add Server dialog or the MCP add_servers tool.

Worth considering alongside:

- **`--test-connection` should say which servers are actually monitored.** It is the command an operator reaches for to answer "is my config right", and today it answers a question about the file that reads as a question about the service. Naming each server as `monitored` or `in file only` would end this class of report.
- Whether a `--reseed-servers` or `--sync-config` verb is worth having. Riskier (it invites clobbering store-side edits), so the diagnostic above is the better first move.

## Note

The same report also hit an unrelated DPAPI failure when adding the server from a remote Viewer, filed separately.

erikdarlingdata:
Fixed on `dev` in #2257.

[...]

Issue #2228, opened 2026-08-12
[BUG] N registrations silently resolving to one database get N identities and N copies of every incident: no DB_NAME() cross-check exists


erikdarlingdata:
## What

N registrations whose connection strings silently resolve to the SAME database get N distinct identities and N full copies of every collected row, with no warning anywhere. Surfaced by #2220: six Azure SQL DB registrations under one logical server, six `server_id`s each holding byte-identical deadlock graphs all naming the same source database, one real incident alerting six times.

## Mechanism, verified in source

- Identity is registration-derived, never connection-derived: `server_id = ServerIdHelper.GetDeterministicHashCode(server.Config.StorageName)`. Six registrations carry six identities regardless of where their connections actually land.
- Collectors stamp the live truth (`database_name = DB_NAME()` on the Azure per-database paths) but nothing COMPARES that answer to the registered catalog, at `--test-connection`, at connect, or at sweep time. A missing or wrong `Initial Catalog` (Azure defaults you into whatever the connection policy resolves) lands the sweep in the wrong database and every collector happily stores the wrong database's rows under the registration's identity.
- Database-scoped XE makes the #2220 evidence conclusive for this mechanism: a database-scoped session can only capture its own database, so six identical graphs naming Source-DB require six connections IN Source-DB.

## Fix shape

Two layers, both cheap because the data is already in hand:

1. **Sweep-time tripwire (the safety net):** the connector already round-trips the server on connect; compare `DB_NAME()` against the registered database and, on mismatch, log loudly to `collection_log` and raise the existing self-alert path - "registration 'Sibling-A' is connected to database 'Source-DB', not 'Sibling-A': check Initial Catalog". Alert once per registration per mismatch state, not per sweep.
2. **Registration-time collision check:** when a new registration's (host, actual DB_NAME()) pair matches an existing registration's, refuse or require an explicit override. This is the concrete half of what #2218 is already weighing (identity carries neither engine nor port; it also carries neither the server's own answer for the database).

Azure SQL DB is the primary exposure (per-database registrations are the normal shape there), but the tripwire is engine-generic.

## Relations

- #2220 (the field report; stays open until the reporter's `actually_connected_to` run confirms which side, but this defect is real under either outcome)
- #2218 (server_id identity design; this adds the DB_NAME() dimension to that decision)
[...]

Issue #2218, opened 2026-08-12
[DESIGN] server_id identity carries neither engine nor port, so two instances on one host collide


erikdarlingdata:
Deferred from the review of #2213 as a product decision rather than a fix, because it changes what re-registering a server *means*.

## The problem

`server_id` is derived from `ServerIdHelper.BuildStorageName(host, database, readOnlyIntent)` — host, database and read-only intent. It carries neither the **engine** nor the **port**, both of which #2213 made first-class configuration.

Two consequences:

- **A SQL Server and a PostgreSQL instance on the same host collide into one `server_id`.** Their histories interleave in the same tables under one identity. Collector tables are disjoint per engine, so the data itself does not mix, but `collect.servers`, `collection_log`, `analysis_state` and every alert key treat them as one server.
- **Two PostgreSQL clusters on one host collide too** — the normal shape for a dev box or a container host running two instances on 5432 and 5433. `port` was added as first-class config in V68 precisely so that configuration is expressible, and the identity ignores it.

## Why it is not a straightforward fix

Changing the identity derivation **re-keys every existing server**. A monitored server would get a new `server_id` on the next start, orphaning all of its collected history under the old id — the same class of event as [[#2158]]'s re-key drift. Options, none obviously right:

1. **Include engine + port in `BuildStorageName`.** Correct going forward, re-keys the entire installed base on upgrade.
2. **Include them only when non-default** (engine != sqlserver, port != 0). No re-key for existing SQL Server targets; PostgreSQL targets keyed correctly from the start. Asymmetric, and a SQL Server target that later gains an explicit port still re-keys.
3. **Leave the identity alone and reject the collision at registration** — refuse to add a second server whose storage name matches an existing one with a different engine or port, with a message saying why. No re-key at all; the collision becomes a configuration error instead of silent interleaving.

(3) is the smallest change and loses nothing that works today, but it makes a legitimate configuration unsupported rather than supported.

## What is not affected

The collector tables themselves: PostgreSQL and SQL Server collectors write disjoint tables, and #2213's engine gate means neither engine's definitions dispatch against the other's target. This is about identity and the shared observability tables, not data mixing.

---

Filed as a follow-up per the repo's deferral rule. Raised by the #2213 review; not fixed there because the answer is a product decision about what a re-registered server means, not a code correction.

erikdarlingdata:
## Recommendation, covering this, #2158 and #2228 together

[...]

Issue #2158, opened 2026-08-10
[BUG] Editing a server's Host/Database/ReadOnlyIntent re-keys its runtime server_id — registry PK orphaned, collected history split


erikdarlingdata:
### Component
Darling (Lite likely shares the class — verify ServerConnection.Id derivation)

### What happens
`MonitoredServer.StorageName` is a computed property over `(Host, Database, ReadOnlyIntent)` (DarlingConfig.cs ~1083), and `server_id = ServerIdHelper.GetDeterministicHashCode(StorageName)` — computed fresh at runtime (DarlingServerConnector.cs ~156) but written into `config.config_monitored_servers.server_id` only once, at seed/add (`ON CONFLICT (server_id) DO NOTHING`, StoreConfigProvider.SeedMonitoredServersAsync).

Edit any of those three fields on a registered server and the runtime id moves while the registry PK stays behind:

1. All NEW collected rows are stamped with the new hash → the registry row's `server_id` no longer joins to the server's own data (the column's doc comment says it exists to "JOIN collected data").
2. All PREVIOUSLY collected history remains under the old hash → trends, baselines, and retention-window reads see a hard discontinuity at the edit; the old data is orphaned under an id nothing resolves anymore.

### Field evidence (dogfood fleet)
9 of 52 monitored servers — exactly the use1 fleet, which was repointed replica→primary in June (`ReadOnlyIntent` true→false flips the `:RO` suffix in StorageName; hosts changed too) — now collect under `server_id`s absent from the registry. Registry `server_id` join against 10 minutes of `collection_log`: 43/52 match. Runtime behavior (alerts, viewer, collection) is unaffected because everything live rides ServerRuntime's freshly computed id — which is exactly why this went unnoticed for two months.

### Fix directions (for discussion)
- **(a) Migrate on edit**: when a connection edit changes StorageName, UPDATE the registry PK and re-key the server's collected rows (hypertable UPDATEs are expensive but this is rare and bounded per server).
- **(b) Stable identity**: stop deriving the persisted id from mutable fields — assign once, store, and let StorageName changes be a rename. Changes Lite/Darling cross-app same-server identity semantics; needs care.
- **(c) Minimum**: registry PK follows the edit (re-insert under new id, tombstone old) so the documented join holds going forward; history stays split but at least isn't silently mis-keyed.

Also worth a docs/UX note either way: the current behavior means "edit host" silently starts a new server history, which nobody expects.

[...]

PR #2263, opened 2026-08-14
Read server_id from the store instead of re-deriving it (#2218 step 1)

erikdarlingdata:
Step one of #2218, and a strict improvement on its own regardless of where the design lands. Prerequisite for #2158 and #2228.

## The finding that made this cheap

`config.config_monitored_servers.server_id` is that table's PRIMARY KEY and is authoritative once seeded. `ReadMonitoredServersAsync` selected **fifteen columns and not that one** — so the registry's own key was read past, and **twelve** downstream sites re-derived it from `host`/`database`/`read_only_intent`: every operator-command lookup, the reconcile, the self-alert stamps, the schedule resolution, and the connect-time stamp the collectors inherit.

So step 1 is not "allocate a surrogate." It is **stop discarding the one we have.**

## The change

`MonitoredServer` gains:

```csharp
public int? StoredServerId { get; set; }              // the store's value; null before the first seed
public int ServerId => StoredServerId ?? hash(StorageName);
```

That property is now the only place a monitored server's identity is decided.

| | before | after |
|---|---|---|
| `DarlingWorker.cs` derivations | **12** | **0** |
| `DarlingServerConnector` stamps | 2 | 0 |
| `DarlingMcpHostService` | 1 | 0 |
| mint/allocation sites | 3 | 3 (unchanged) |

**Behaviour is unchanged and no data moves.** The seed and the Viewer both write exactly the hash this used to recompute, so stored and derived are equal on every store that exists. What changed is that the derivation is the FALLBACK, in one place — so when identity stops being derivable, that property changes and the twelve call sites do not.

The store is preferred over the derivation rather than merely *agreeing* with it. That ordering is what makes a stored id which no longer matches its host keep working, which is the entire point of storing it.

`StoredServerId` is `[JsonIgnore]`: `darling.json` must not be able to pin an identity, or it becomes a second authority able to disagree with the registry — and disagree **silently**, since nothing downstream re-checks.

## Deliberately not done here

- **`StoreConfigProvider:203`**, the #2254 file-vs-store diagnostic, still compares a file-derived id against stored ids. That is the right question today and the wrong one later; it needs the observed-identity fingerprint. Noted on #2228 rather than half-solved.
- **`CadencePhaseOffset`** consumes the id purely as a *spreading function*. Its doc now says so, and says what breaks if ids ever become sequential: consecutive integers modulo a period do not spread, they line up.

## Tests

Pure: the fallback; **the seam** — a stored id survives an edit that changes the hash; the `JsonIgnore` refusal in three spellings.

Two source pins: no derivation left in the three converted files, and **minting is a closed set of three files**, so a fourth has to be argued for in review rather than appearing. A source pin because the failure it guards is a *new* call site added later — a fresh `GetDeterministicHashCode(config.StorageName)` agrees with the stored id on every store that exists right now, and only starts lying long after the commit that introduced it.

Live (scratch database): seed, assert the seed wrote the derivation — the "no migration needed" claim, asserted rather than assumed — then **re-key the row so its stored `server_id` disagrees with the hash of its own host**, and assert the loaded server carries the surrogate. That is the one thing production cannot produce yet; without it every assertion here would pass against the old code, since the two values coincide.

That test also asserts **all fifteen other projected columns**, each with a distinctive value, because adding a column to a positional reader is exactly how a silent mis-map happens: shift one ordinal and `auth` arrives in `username`, which no compiler and no id assertion would catch. `server_id` was appended LAST for the same reason.

## Notes

- Also corrected the V17 SQL comment, which asserted `server_id` **is** the hash. It is now the *allocation rule* rather than a definition anything re-derives. Inert for migrated stores (V17 does not re-run), and it removes a false positive from the new source pin without weakening it.
- Full solution builds clean, 0 warnings. Both source pins simulated locally against the real tree before shipping them — that is how the V17 false positive was found. **I cannot run `net10.0-windows` xunit locally, so CI is the only place the pass is observable.**

### DIFF
--- a/Darling/Darling.Tests/PostgresTargetConfigTests.cs
+++ b/Darling/Darling.Tests/PostgresTargetConfigTests.cs
@@ -307,6 +307,14 @@ public class PostgresTargetConfigTests
             ("ExcludedDatabases", "excluded_databases"),
             ("MonthlyCostUsd", "monthly_cost_usd"),
             ("AlertDeliveryModeOverride", "alert_delivery_mode_override"),
+            /* #2218. The one entry whose property name deliberately does not resemble its column: the
+               property says STORED because that is the whole point — it is the registry's value, and null
+               means "no store row yet", which is what makes MonitoredServer.ServerId fall back to the
+               derivation. It belongs in this list rather than beside Password's exemption because it really
+               does round-trip, through the table's own PRIMARY KEY. That column existed from V17 and this
+               read simply did not select it, which is the class of bug this test was written for: a property
+               and a column that fail to meet, with nothing failing to compile. */
+            ("StoredServerId", "server_id"),
         };
 
         /* Password is the deliberate exception: a plaintext dev password is never persisted, and is
--- a/Darling/Darling.Tests/ServerIdentityFromStoreTests.cs
+++ b/Darling/Darling.Tests/ServerIdentityFromStoreTests.cs
@@ -1,0 +1,310 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Npgsql;
+using PerformanceMonitor.Common;
+using PerformanceMonitor.Darling.Service;
+using PerformanceMonitor.Darling.Storage;
+using PerformanceMonitor.Notifications;
+using Xunit;
+
+namespace Darling.Tests;
+
+/// <summary>
+/// #2218 step one: a monitored server's <c>server_id</c> comes from the STORE, and is derived in exactly one
+/// place when there is nothing stored yet.
+///
+/// <para><b>What was wrong.</b> <c>config.config_monitored_servers.server_id</c> is that table's PRIMARY KEY
+/// and is authoritative once seeded — but the read selected fifteen columns and not that one, so the registry's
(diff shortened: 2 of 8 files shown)


erikdarlingdata:
Fixed in `b2802305` — and it was a real slip rather than a cosmetic one, worth naming because it is a repeatable mistake: my replacement anchor ended mid-line, so the sentence that followed it got appended to my new text and the line reached **194 characters** against the file's ~110. Wrapped.

erikdarlingdata:
Confirmed and fixed in `c4b2ad5b`. You were right on both counts, and the first one is a **bug that already exists** — better than the future one you were looking for.

Verified: `config = DarlingConfig.Load()` at line 138, no `StoreConfigProvider` anywhere in the file, and `config.Servers` feeding `serversById`. So `StoredServerId` is always null there and `ServerId` always falls back to the derivation, exactly as you said.

**The present-day bug.** The registry is authoritative after the first seed, so a server added through `add_servers` or the Viewer is in the store and **not in darling.json** (#2254/#2256). It was therefore missing from this map entirely — while every MCP tool resolved it fine, because `DarlingServerResolver` reads the store. Analysis and plan drill-down for those servers silently had no connection to fetch from. That predates #2218; it has been true since the Viewer could add a server.

**And my change made the second one worse than leaving it alone**, which is the part worth saying plainly: spelling the derivation differently got the site counted as converted and *source-pinned as such*, while it remained the one place that could disagree with the store. A pin asserting a property the code does not have is worse than no pin.

Fix for both is the same: key on the registry via `LoadViewAsync`. Store-unreachable falls back to the file — this host's documented posture is to start on `darling.json` when the store is down, and a fallback only reached when the store cannot answer cannot be the thing that disagrees with it. That path now logs a warning naming the consequence rather than degrading silently.

Issue #2150, opened 2026-08-10
[BUG] 3.4.0 query_store collector runs 37–100 min on Azure SQL DB (3.3.0 median: 4.8 s) — starves all other collectors


erikdarlingdata:
## Evidence (from the #2148 reporter's log — thank you, TrudAX)

Same two Azure SQL DB targets (elastic pool, edition 5, "version 12"), same day, before and after the 3.3.0 → 3.4.0 upgrade at 09:28–09:38 local:

| build | runs | median | max |
|---|---|---|---|
| 3.3.0 (00:00–09:28) | 198 | **4.8 s** | 31 s |
| 3.4.0 (09:38–12:00) | 6 | — | **99.8 min** |

Post-upgrade runs: 0.1 min (7,748 rows), 37.6 min (9,480), 46.1 min (11,845), 82.1 min (5,951), 0.1 min (19,873), 99.8 min (6,621). **Bimodal** — occasionally the old speed, usually catastrophic, on BOTH servers, at tiny row counts. `sql:` time ≈ wall time, so it's server-side execution, not storage writes.

Because Lite's collectors run sequentially, one 100-minute query_store cycle starves every other collector — this is the actual mechanism behind #2148's "all collection stopped" (the abandonment containment in #2149 protects backfill and the connection check; the LIVE collector pipeline is still sequential and still waits).

**Why the 30s CommandTimeout never fired:** rows trickle continuously, and SqlClient's timeout resets on every network read — a slowly-streaming result set never times out. Any fix should consider a wall-clock bound as defense in depth.

## Suspects

`QueryStoreCollector.cs` changed by ~958 lines between v3.3.0 and v3.4.0 (the #2134 staged-#temp rewrite with OPTION(RECOMPILE) both statements + interval pre-filter, plus #2058 backfill plumbing). The rewrite was validated hard on SQL Server (fixed the fleet's wedged-core catch-up), but Azure SQL DB elastic pools differ in exactly the ways that matter: resource-governed tempdb, DTU throttling, and a different optimizer surface for the query_store TVFs. The bimodality smells like a per-cycle plan flip on the staged aggregate despite the RECOMPILE hints — or the Azure arm taking a different, unvalidated path.

## Next steps

1. Determine which arm (regular vs Azure) edition-5 targets execute in 3.4.0, and diff the emitted SQL against 3.3.0's.
[...]

PR #2264, opened 2026-08-14
Bound one database's collection by wall clock, so it cannot starve the rest (#2150)

erikdarlingdata:
Step 3 of #2150 — the containment. The query itself (steps 1–2) stays open; this stops one database from taking the whole host down with it, which is the reported harm.

## The report

Two Azure SQL DB elastic-pool databases, same day, across the 3.3.0 → 3.4.0 upgrade:

| build | passes | median | max |
|---|---|---|---|
| 3.3.0 | 198 | **4.8 s** | 31 s |
| 3.4.0 | 6 | — | **99.8 min** |

A host's live collectors run one after another, so one 100-minute pass starves every other collector on that server. That is the actual mechanism behind #2148's "all collection stopped".

## Nothing already caught it, and nothing could have

`CommandTimeout` was **30 seconds the whole time.** It bounds the wait for a network read, and SqlClient **resets it on every read that arrives** — so a result set that trickles rows continuously never trips it, however long it takes in total. A 100-minute read under a 30-second timeout is documented behaviour, not a bug.

## The change

`ICollectorDefinition.PerItemWallClockBudget` — null for every collector but `query_store`, which declares **10 minutes**. Applied in **both** places a per-database pass can happen:

- the shared `EnumeratedCollectorDriver` loop (on-prem enumeration), and
- **each host's own per-database loop** (Azure SQL DB) — which is the path the report is on and does *not* go through the driver. Missing that would have shipped a fix that does not cover the reporter.

**Exceeding it abandons that database as a per-item failure and continues.** Nothing is lost: the watermark did not advance, so the range is re-read next cycle.

**And it converges rather than repeating.** The failure also feeds #2111's consecutive-failure count, so the catch-up window halves toward 15 minutes until a pass fits; a success resets it and the database returns to full width. Without that, this would be a bound that fires forever on the same impossible width.

## Why ten minutes, measured rather than chosen

| | |
|---|---|
| healthy, field | 4.8 s median, 31 s max across 198 passes |
| healthy, box | 375–524 ms for the staged query on a 212k-row Query Store |
| healthy, post-upgrade | 6 s for the two fast passes |
| **pathological** | **37.6 minutes at the low end** |

Ten minutes is ~19× the worst healthy pass, ~10× Darling's command-timeout default, and ~3.7× under the smallest pass this is meant to stop.

## Three design points

- **Classification is on the TOKENS, not the exception type.** Cancelling a SqlClient command does not reliably surface as `OperationCanceledException` — it commonly arrives as a provider exception, and which one depends on whether the cancellation landed during the open or the drain. So the budget arm is filtered on "did *my* source fire and the outer one not", and it sits **ahead** of the existing generic catch. **Shutdown deliberately wins the ambiguous case**, because misreading a service stop as a per-item skip would have the loop keep collecting through it.
- **The write is outside the budget.** Abandoning a flush already in flight would trade a slow cycle for a partially-written one.
- **A non-positive budget is treated as no budget**, so a misconfigured value degrades to today's behaviour rather than collecting nothing at all.

## Verification

`net10.0-windows` xunit cannot run on this machine, so I exercised the real driver through a scratch console harness: **27/27**, covering the abandon-and-continue path, shutdown propagation, token identity when no budget is set, the write's token, all four classifier combinations, the non-positive case, and which collectors declare a budget.

**That harness earned its keep.** The first cut formatted the budget as minutes only and reported a 150 ms budget as `0.0-minute` — a message naming no number, on the one line an operator has to work from. The shipped value is 10 minutes, so it would never have looked wrong in the field, **and the xunit assertion I had written (message *contains* "wall-clock budget") passed against it.** `DescribeBudget` now picks its unit, and the pin asserts the rendered number.

### DIFF
--- a/Darling/PerformanceMonitor.Darling.Service/DarlingCollectorRunner.cs
+++ b/Darling/PerformanceMonitor.Darling.Service/DarlingCollectorRunner.cs
@@ -309,6 +309,14 @@ public sealed class DarlingCollectorRunner
             {
                 cancellationToken.ThrowIfCancellationRequested();
                 attempted++;
+
+                /* #2150: THE path the field report is on — Azure SQL DB collects query_store per database
+                   here, not through the enumerated driver, so the wall-clock ceiling has to be applied on
+                   both. Null for every collector that declares none, in which case dbToken IS
+                   cancellationToken and this loop is byte-for-byte what it was. */
+                using var dbBudget = EnumeratedCollectorDriver.StartItemBudget(
+                    definition.PerItemWallClockBudget, cancellationToken);
+                var dbToken = dbBudget?.Token ?? cancellationToken;
                 try
                 {
                     /* The authoritative database_name for XE rows read on this path — see
@@ -393,11 +401,14 @@ public sealed class DarlingCollectorRunner
 
                     var sqlSlice = Stopwatch.StartNew();
                     List<TRow> batch;
-                    using (var dbConnection = await OpenDatabaseConnectionAsync(perDbProvider, server, databaseName, cancellationToken))
+                    /* dbToken, not cancellationToken (#2150): connect, execute and drain are the phases the
+                       budget bounds. The FLUSH below deliberately stays on cancellationToken — abandoning a
+                       write already in flight would trade a slow cycle for a partially-written one. */
+                    using (var dbConnection = await OpenDatabaseConnectionAsync(perDbProvider, server, databaseName, dbToken))
                     using (var dbCommand = CreateCollectorCommand(perDbProvider, dbPlan, dbConnection, perDbTimeout))
-                    using (var dbReader = await dbCommand.ExecuteReaderAsync(cancellationToken))
+                    using (var dbReader = await dbCommand.ExecuteReaderAsync(dbToken))
                     {
-                        batch = await definition.ReadAsync(dbReader, context, cancellationToken);
+                        batch = await definition.ReadAsync(dbReader, context, dbToken);
 
                         /* #1875: the payload path's probe-failure contract, on the path that used to
                            ignore it. blocked_process_report is the declaring collector that also runs per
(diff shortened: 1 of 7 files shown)


erikdarlingdata:
Both confirmed and fixed in `94793572` — same gap, both hosts, and you're right that it was on the one path that matters most here.

The contract said the ceiling covers "the watermark refresh, the command, and the whole drain", the shared enumerated path's `perItemWatermark` delegate honours that, and I threaded `dbToken` through only connect/execute/drain on the Azure loop. So the loop the field report is *actually* on was the single place the promise was not kept. Three round-trips now bounded in each host: `GetLastCollectedTimeForDatabaseAsync` and both `RecordQueryStoreBackfillHoleAsync` calls.

The part I checked rather than assumed, since a bounded **write** is a different question from a bounded read: **the hole records are safe to bound.** A budget expiry abandons the whole pass, so the watermark does not advance, the clamp is re-derived next cycle, and the hole is re-recorded — merged wider with any already pending, per `RecordQueryStoreBackfillHoleAsync`'s own contract — rather than lost. That is what makes this parity rather than a trade.

The payload flush stays on `cancellationToken` for the opposite reason, and that asymmetry is deliberate: abandoning a write already in flight trades a slow cycle for a partially-written one, where abandoning the hole record costs nothing because the pass that would have needed it did not happen.

erikdarlingdata:
Confirmed and fixed in `d9f5960d`, both hosts. Sharp catch — and the inconsistency is the tell: I ordered the **shared driver** correctly (`catch (OutOfMemoryException) { throw; }` ahead of the budget arm) and then wrote both per-database loops without it, in the same PR.

The root of it is the deliberate trade you identified: `ItemBudgetExpired` classifies on the **tokens** and never inspects the exception type, because a cancelled SqlClient command does not reliably arrive as an `OperationCanceledException`. The cost of that is the arm will claim *any* exception once the timer has fired — so its **position** behind the OOM rethrow is load-bearing rather than stylistic, which is exactly what I failed to carry across.

Verified the race rather than reasoning about it: expire the budget, then throw an unrelated `OutOfMemoryException` from the read, assert it propagates and is not reported as a per-item skip. Added as a driver test too, with the ordering's significance written down so the next edit does not undo it.

Harness now 29/29.

Issue #1836, opened 2026-07-30
Query Store collector never collects on Azure SQL Database


erikdarlingdata:
Spun out of #1833's triage (the reporter's other empty grid). Verified in source, not yet reproduced live on Azure.

`query_store_stats` is one of two database-scoped collectors that never opted into the Azure per-database connection path (`RunsPerDatabase`); #1833's fix handles the other one (`procedure_stats`), but Query Store is design-sized rather than a one-line override, because the collector is built on `BuildEnumerationQuery` + `BuildPerItemQuery` while the Azure per-database branch drives `BuildQuery`.

What happens today on Azure SQL DB:

- The collector runs once on the server entry's own connection (catalog defaults to `master` when the entry's Database field is blank).
- Its Azure enumeration cursors candidate databases and probes each via `QUOTENAME(@db) + N'.sys.sp_executesql'` (`QueryStoreCollector.cs:255-261`) — a cross-database three-part reference, which Azure SQL DB rejects for **every** database, from `master` or from any user database alike (the repo already documents this behavior class at `RunningJobsCollector.cs:148-150`).
- Every probe fails into an **empty CATCH**, the result list comes back empty, and the run logs `SUCCESS` with 0 rows.

So even with the Database field set correctly, Query Store data on Azure is never collected — and on a logical server with several databases, a correct fix collects each database's Query Store, which the single-connection shape can't do at all.

Sketch:

1. An Azure per-database `BuildQuery` variant that does the `sys.database_query_store_options` eligibility check inline against the current database (`actual_state IN (1, 2, 4) AND readonly_reason & 8 = 0`, same predicates as the on-prem probe) and returns the per-item payload for `DB_NAME()`.
2. `RunsPerDatabase => target.IsAzureSqlDb`, with the enumeration/per-item pair remaining the on-prem path.
3. The empty CATCH at `QueryStoreCollector.cs:260-261` should record probe failures somewhere visible even on-prem — a database that always fails its probe is currently indistinguishable from a database with Query Store off.

Darling runs the identical shared definition and has the same gap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
Shipped in #1844 (merged to dev; auto-close does not fire on non-default-branch merges). Query Store now runs per database on Azure SQL DB with the eligibility gate inline; live-Azure validation is tracked by #1848.

PR #2265, opened 2026-08-14
Scope an Azure SQL DB sweep to the database it registered (#2220)

### LINKED ISSUE #2220: [BUG] Blocking/Deadlock alerts cross-talk between Azure SQL databases sharing the same logical server (Darling)
### Component

Darling

3.3

### SQL Server Version

Microsoft SQL Azure (RTM) - 12.0.2000.8   Jul  2 2026 14:28:12   Copyright (C) 2026 Microsoft Corporation 

### Windows Version

Windows 11 23H2

### Describe the Bug

When blocking or a deadlock occurs in one Azure SQL Database, Darling generates near-identical "Blocking Detected" and "Deadlocks Detected" alerts on every other monitored database that shares the same Azure SQL logical server, even though Azure SQL Database engines are fully isolated per-database and one database's sessions cannot actually block or deadlock another's. This produces a large alert-noise fan-out: a single real incident in one database triggers what looks like the same incident across 10+ unrelated sibling databases.

### Steps to Reproduce

1. Register multiple Azure SQL Databases that share one logical server (e.g.,  myserver.database.windows.net:db1 ,  myserver.database.windows.net:db2 , ... each as a separate monitored connection with its own display name) with Darling.
2. Wait for real blocking or a deadlock to occur in one of the databases (e.g.,  db1 ).
3. Check the alert history ( get_alert_history  / alert feed) for the other sibling databases ( db2 ,  db3 , ...) registered under the same logical server.

### Expected Behavior

Only the database where the blocking/deadlock actually occurred ( db1 ) should generate a "Blocking Detected"/"Deadlocks Detected" alert, since Azure SQL Database sessions cannot span databases.

### Actual Behavior

Every sibling database registered under the same logical server host fires an alert with matching detail content (same blocked/blocking query text, same wait ranges, same deadlock victim SQL and involved objects) but a different per-alert  Dedup Key  and  server_id , all attributing the incident to the same source database (e.g., every sibling's alert says  Database: db1  even though the alert fired under  server_name = db2 ,  db3 , etc.). For deadlocks, the leaked alerts share the same underlying database GUID in "Involved Objects" across all siblings.

This looks like blocking/deadlock detection is being keyed off the shared logical server host string rather than the individual per-database connection, causing one real incident to fan out to every registered database sharing that host.

```text
(N/A — this is alert content duplication, not an error; happy to attach  get_alert_history  JSON samples showing the duplicate detail_text across sibling  server_name  values if helpful)
```

### Screenshots

_No response_

### Additional Context

• Monitoring Azure SQL Database (logical server with ~15 databases registered as separate connections
• Observed over a 7-day window: 100% of "Blocking Detected" cross-talk alerts on sibling databases pointed to the same source database ( horizon ); same pattern confirmed for "Deadlocks Detected".
• Worked around on our end for now using per-server mute rules ( database_pattern  /  query_text_pattern  matched against the leaking source's database name / GUID) on each sibling, but this only suppresses the symptom for the currently observed leak source — it doesn't fix the root cause, and a new leak source would need new mute rules.
• Did this work before? Unknown / first time we've closely audited alert content across sibling databases on a shared logical server.

erikdarlingdata:
Fixes #2220.

## The report was right, and the reporter found the mechanism before I did

One real deadlock in one Azure SQL Database produced near-identical alerts on every *other* monitored database on the same logical server. Azure SQL DB engines are isolated per database, so one database's sessions cannot block another's — and the stored data bore that out: byte-identical deadlock graphs and the same top query, counters within ~1%, under six unrelated `server_id`s. Not cross-talk. **The same rows, collected six times.**

Confirmed from source. `GetAzureDatabaseListAsync` → `BuildDatabaseListPlan` hops to **master**:

```sql
SELECT name FROM sys.databases WHERE state_desc = N'ONLINE' AND database_id > 0 {exclusion} ORDER BY name;
```

That returns **every online database on the logical server**, narrowed only by that registration's `ExcludedDatabases` — a denylist, not an allowlist. The caller sweeps all of them and stores every row under the one `server_id` of whichever registration ran the sweep. The single-database behaviour existed only as the fallback for a master-**access error**, so the safe path was the exceptional one.

On a logical server holding N separately-registered databases: **N registrations × N databases = N² collection, with every registration's history contaminated by its siblings'.** This is a data-correctness bug, not alert noise.

## Why it happened, which is the part worth keeping

Two parts of the product hold **incompatible ideas of what an Azure SQL DB registration is**, and both are deliberate:

- the enumeration assumes one registration = one **logical server** — the shape #857 was written for;
- identity assumes one registration = one **database**: `server_id` hashes `host[:database][:RO]`, so registering two databases on one server is the *supported* way to get two identities, and the Azure query_store path needs a per-database connection anyway (#1836).

Nothing reconciled them, so the second shape silently behaved like the first.

## The rule

A registration that **names** a database is a registration **of** that database, and sweeps exactly that one. Only a registration naming none — or naming `master`, where a catalog-less Azure connection lands — is a registration of the **server**, and only that one enumerates.

This also **subsumes #857's motivating case and improves on it**: a login granted access to one user database but not to master *has* a named database, so it now returns without probing master at all, rather than probing, failing, forming a verdict and falling back.

## Extracted to shared code rather than fixed twice

Both runners carried their own private copy of this predicate. A sweep-scoping rule that disagrees between Lite and Darling is the same class of defect as the one being fixed, so it now lives in `AzureSweepScope` and **both suites pin it**.

## Deliberately not done here

The master-probe throttle (`_azureMasterInaccessibleSince` / `IsMasterProbeThrottled` / `OnServerReconnected`) is now **unreachable for any registration that names a database** — which is #857's whole population. Removing that machinery is a separate change from a correctness fix a reporter is waiting on; its guard is left stating the precondition this path now satisfies by construction. Happy to do the cleanup as a follow-up.

## Verification

Rule checked against compiled code (12/12 in a scratch harness, since `net10.0-windows` xunit will not run on this machine), including the over-match cases — `mastermind`, `paymaster`, `master_archive` are real databases and are still swept — and that each call returns its own list, since both runners hand it straight to their per-database loop. Full solution builds clean, 0 warnings.

**What this does not fix:** existing stored data. Any store that has been collecting Azure SQL DB siblings already holds rows under the wrong `server_id`, and this change stops the contamination rather than unwinding it. Worth deciding separately whether that needs a cleanup path or just a note.

### DIFF
--- a/Darling/Darling.Tests/AzureSweepScopeTests.cs
+++ b/Darling/Darling.Tests/AzureSweepScopeTests.cs
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using PerformanceMonitor.Collectors;
+using Xunit;
+
+namespace Darling.Tests;
+
+/// <summary>
+/// #2220: which databases one Azure SQL DB registration's per-database sweep covers.
+///
+/// <para><b>The field report.</b> A single real deadlock in one Azure SQL Database produced near-identical
+/// "Deadlocks Detected" alerts on every OTHER monitored database sharing the same logical server — and the
+/// stored data matched: byte-identical deadlock graphs and the same top query, with counters within ~1%,
+/// under six unrelated <c>server_id</c>s. Azure SQL DB engines are isolated per database, so one database's
+/// sessions cannot block another's; the rows were not cross-talk, they were the same rows collected six
+/// times.</para>
+///
+/// <para><b>The cause, and why it is not a typo.</b> The enumeration read <c>master</c> unconditionally and
+/// swept every online database on the logical server, storing all of it under whichever registration ran the
+/// sweep. Two parts of the product hold incompatible ideas of what a registration IS, both deliberate: the
+/// enumeration assumes one registration = one LOGICAL SERVER (#857's shape), while identity assumes one
+/// registration = one DATABASE (<c>server_id</c> hashes <c>host[:database][:RO]</c>, and the Azure
+/// query_store path needs a per-database connection anyway, #1836). The second shape silently behaved like
+/// the first, N times over — N registrations of N databases is N² collection.</para>
+///
+/// <para>Pinned in BOTH suites against the shared implementation. A scoping rule that disagrees between Lite
+/// and Darling is the same class of defect as the one being fixed, and both runners previously carried their
+/// own private copy of it.</para>
+/// </summary>
+public sealed class AzureSweepScopeTests
+{
+    /// <summary>
+    /// THE FIX. A registration naming a database sweeps exactly that database — the reported case, where
+    /// fifteen registrations on one logical server each swept all fifteen.
+    /// </summary>
+    [Theory]
+    [InlineData("db1")]
+    [InlineData("AdventureWorks")]
+    [InlineData("Sibling-A")]
+    public void ARegistrationThatNamesADatabase_SweepsOnlyThatDatabase(string catalog)
+    {
+        Assert.Equal(new[] { catalog }, AzureSweepScope.OwnDatabaseOrEmpty(catalog));
+        Assert.True(AzureSweepScope.IsScopedToOneDatabase(catalog));
+    }
+
+    /// <summary>
+    /// A registration naming NO database is a registration of the logical server, so it must enumerate —
+    /// signalled by the empty list rather than by a separate flag, because the caller's next step is a list
+    /// either way. This is the behaviour #857 was written for and it is deliberately unchanged.
+    /// </summary>
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    public void ARegistrationThatNamesNoDatabase_StillEnumeratesTheServer(string? catalog)
+    {
+        Assert.Empty(AzureSweepScope.OwnDatabaseOrEmpty(catalog));
+        Assert.False(AzureSweepScope.IsScopedToOneDatabase(catalog));
+    }
+
+    /// <summary>
+    /// <c>master</c> counts as naming none, in any casing. A connection string with no
+    /// <c>Initial Catalog</c> lands in <c>master</c> on Azure SQL DB, so treating it as a named database
+    /// would scope such a registration to the one database holding none of the user's data — collecting
+    /// nothing and looking healthy while doing it.
+    /// </summary>
+    [Theory]
+    [InlineData("master")]
+    [InlineData("MASTER")]
+    [InlineData("Master")]
+    public void MasterIsNotADatabaseAnyoneRegisteredFor(string catalog)
+    {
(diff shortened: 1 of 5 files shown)


erikdarlingdata:
Both confirmed and fixed in `c4c844a0` — and this was a **regression I introduced**, not a nit, so thank you for catching it.

I read `hasFallback &&` as a redundant condition and dropped it. It was the opposite: in that branch `hasFallback` is **false**, so the guard existed to *disable* the throttle — precisely what the comment I deleted said. "Honouring the throttle would guarantee 15 minutes of failure without ever attempting to recover."

Left unconditional, and with the branch now reachable only when the registration names no database, one master failure would send every subsequent cycle straight into `FallbackDatabaseList` — which throws immediately without probing — for the whole recheck interval, where it used to retry every cycle. Recovery detection delayed by 15 minutes. Exactly as you described, in both hosts.

Removed the check from that path in both, restoring the old effective behaviour, with the reasoning written where the old comment was so it does not get re-added by the next person who reads the condition the way I did.

**One thing I did not do, deliberately.** The obvious follow-through is to delete the throttle as newly-dead code, and I checked before reaching for it: Lite's `IsMasterProbeThrottled` is directly tested by `AzureMasterFallbackTests` — six assertions carrying #857/#1506 behaviour — so retiring it is its own change with those tests, not a rider on a correctness fix.

Worth recording *why* it is obsolete now, since it is a different reason than this fix: the throttle's entire purpose was to stop re-probing master for a registration that **has** a fallback, and after this PR such a registration never probes master at all. So it is unreachable in production either way — but that is a deletion to argue on its own merits.

erikdarlingdata:
All five addressed in `45773906` — two fixed, one accepted as follow-up with your reasoning.

**`IsScopedToOneDatabase` was dead surface, and you caught the irony:** I wrote a named predicate arguing a call site should read as the rule rather than as a length check, then wrote both call sites as `ownDatabase.Count > 0` anyway. The reason is structural — both callers need the *list* to return, so the boolean twin would mean calling twice — so I dropped it rather than contorting the callers to justify it. The tests that existed only for it went with it.

**Both stale summaries rewritten.** You're right that they re-teach the pre-fix model: "try master enumeration first … and on a master-access error fall back … throttling re-probes per server" is now the *exceptional* path, not the default. Both now lead with the rule that governs, and Lite's also records that this covers #857's own case better than #857 did — a login with access to one database but not master no longer probes, fails and falls back; it simply never probes.

**`FallbackDatabaseList`'s `quiet` parameter — agreed, and taking it as follow-up on your own reasoning.** You're right it's now dead in both hosts (one caller, always the default), and right that it's the same now-unreachable machinery as `IsMasterProbeThrottled`. They should retire together, alongside the `AzureMasterFallbackTests` assertions that pin the throttle — six of them carrying #857/#1506 behaviour. Bundling a tested-behaviour deletion into a correctness fix a reporter is waiting on is how the fix gets held up arguing about the deletion.

Full solution still builds clean, 0 warnings.

erikdarlingdata:
Fixed in `aac1c26c` — and the irony is earned. The `sed` I used to customize the Darling header for Lite ate two words. In a PR whose whole point is that Lite and Darling must not disagree, that is the sentence to get right.

PR #2267, opened 2026-08-14
Seed inside the retention horizon these tests arm (#2223)

### LINKED ISSUE #2223: [TEST FLAKE] ArmedIntervalPurge re-hold test races the rebuilt daily CAGG's immediate first policy refresh
## What

`QueryStoreCorrectedRollupLiveTests.ArmedIntervalPurge_IsReHeldWhenItsCoverageListGainsAnEmptyConsumer_ThenReleasesOnBackfill` fails intermittently on identical bits. Measured on a local rig (repo pg-runtime PG 18.4 + TimescaleDB 2.28.1, real migrated store): 2 failures in 9 class-solo runs, plus 1-in-2 full-suite runs; the single test solo passed every time it was tried.

```
the backfill must carry the new consumer past L1's oldest bucket before the gate can release
(consumer 2026-08-05T00:00:00.0000000, L1 2026-08-03T00:00:00.0000000)
```

at `QueryStoreCorrectedRollupLiveTests.cs:561` (step 4, SELF-RELEASE).

## Suspected mechanism

Step 2 drops and rebuilds the daily CAGG, and the rebuild re-attaches a refresh policy whose **first check runs immediately** — the test already knows this: the step-2 guard was deliberately written shallow-or-empty because "under suite load the recent window can materialize before the strip lands" (#1564 note in the test body).

Step 4 then runs a manual `refresh_continuous_aggregate` over the full span. When the policy's immediate first refresh is still in flight at that moment, the manual refresh contends with it (Timescale serializes refreshes per CAGG), and the daily floor comes out at the policy's recent window (08-05) instead of the backfilled span (<= L1's 08-03). Same race the step-2 guard absorbed, one step later — step 4 has no equivalent tolerance, and can't have one as written, because its assertion is exactly "the floor reached the backfill."

Likely fix shape: rebuild the aggregate in step 2 **without** re-attaching the refresh policy (the test's own `EnsureAggregatesWithoutRefreshPoliciesAsync` name suggests that was the intent — worth checking whether the CASCADE rebuild path re-adds one behind it), or wait for/kill the policy job before step 4's manual refresh, so the only refresh in the story is the test's.

## Why filed now

Surfaced while shipping #2171 (whose branch does not touch this subsystem — the failure reproduces regardless of those changes). Filing rather than fixing inline since the right fix depends on #1877/#1869 intent, which that lane owns.

erikdarlingdata:
Fixes #2223.

## Root cause: retention, not the refresh policy

The earlier tracing on the issue correctly ruled out the suspected mechanism — refresh-policy contention doesn't produce the observed floor either way round. It's the **retention** policy, and not contention but actual deletion.

`TimescaleSupport.IntervalRetentionInterval = "7 days"`. All **four** tests in `QueryStoreCorrectedRollupLiveTests` that call `EnsureRetentionPoliciesAsync` seeded a hardcoded **nine** days of history. So each one armed a policy guaranteed to want to delete two days of its own fixture, then asserted against floors those deleted buckets defined.

In the re-hold test specifically: `l1Floor` is read once after step 2, and the step-4 backfill it is compared against runs ~20 lines later. If the armed retention job fires inside that window it drops L1's two oldest buckets, so the rebuilt consumer can only backfill to `now - 7d` while the assertion still holds the pre-deletion floor.

**The arithmetic is what identified it.** Reported failure: consumer `08-05`, L1 `08-03`, fixture seeded from `08-03` = `now.Date - 9d`, so `now.Date` = `08-12`. The consumer floor `08-05` is **exactly `now - 7 days`** — L1's retention horizon. Not the refresh policy's 3-day `start_offset`, which the earlier tracing looked at and was right to discard.

It also explains the observed frequency, which neither earlier theory did: it needs the job to fire inside that window. Solo runs passed every time (little elapsed time); full-suite runs failed about half (more wall clock, and four tests arming the same policy).

## Fix

The fixture depth is now **derived from the product constant** with two days of margin, so nothing a test seeds is ever eligible for deletion by a policy that test itself arms.

Derived rather than re-hardcoded to a smaller number, so that tuning the horizon moves every fixture with it instead of silently reintroducing this — the same pin-the-invariant lesson as today's schema-version pin conflicts, where dev had independently replaced brittle literals with invariant forms for exactly this reason. All four call sites updated, and the relationship is pinned by a new test rather than trusted as arithmetic.

**It was a class-wide contradiction, not one test's bug** — the other three retention-arming tests carried the same latent hazard and are fixed by the same change.

## What I can and cannot claim

Diagnosed entirely from source. I cannot run this suite — it needs live PostgreSQL + TimescaleDB, and both test projects are `net10.0-windows` — so CI is the only place a pass is observable, and **one green run cannot confirm a fix for an intermittent test anyway.**

What *is* verifiable, and is the actual claim: the contradiction is gone. The fixture is now provably inside the horizon it arms, by construction and by a pin. If the test still fails after this, the cause is something else and the arithmetic above is ruled out rather than merely re-rolled.

Left alone deliberately: the `RefreshWithRetryAsync` machinery and the step-2 shallow-or-empty guard. Both are correct for what they handle, and the earlier tracing already established that the refresh policy is not the culprit.

### DIFF
--- a/Darling/Darling.Tests/QueryStoreCorrectedRollupLiveTests.cs
+++ b/Darling/Darling.Tests/QueryStoreCorrectedRollupLiveTests.cs
@@ -51,6 +51,55 @@ public sealed class QueryStoreCorrectedRollupLiveTests
     /// closed form so the expectation is derived, not a magic number transcribed from a test run.</summary>
     private const long InflatedSum = ReCollections * (ReCollections + 1L) / 2L;
 
+    /// <summary>
+    /// L1's retention horizon in days, read from the product constant rather than transcribed (#2223).
+    /// </summary>
+    private static readonly int L1RetentionDays =
+        int.Parse(TimescaleSupport.IntervalRetentionInterval.Split(' ')[0], CultureInfo.InvariantCulture);
+
+    /// <summary>
+    /// How deep every retention-arming test in this class seeds its history — DERIVED from L1's horizon, with
+    /// two days of margin, so nothing a test seeds is ever eligible for deletion by a policy that test itself
+    /// arms.
+    ///
+    /// <para><b>#2223, and it was a class-wide contradiction rather than one test's bug.</b> All four tests here
+    /// that call <c>EnsureRetentionPoliciesAsync</c> used to seed a hardcoded 9 days while step 1 of each arms
+    /// L1's retention at <c>drop_after =&gt; INTERVAL '7 days'</c>. Each therefore armed a policy guaranteed to
+    /// want to delete two days of its own fixture, and then asserted against floors those buckets defined.</para>
+    ///
+    /// <para><b>How it failed.</b> In the re-hold test, <c>l1Floor</c> is read once and the backfill it is
+    /// compared against runs ~20 lines later. If the armed retention job fires inside that window it drops L1's
+    /// two oldest buckets, so the rebuilt consumer can only backfill to <c>now - 7 days</c> while the assertion
+    /// still holds the pre-deletion floor. The reported failure was consumer <c>08-05</c> against L1
+    /// <c>08-03</c> on a fixture seeded from <c>08-03</c> — the consumer floor was EXACTLY
+    /// <c>now - IntervalRetentionInterval</c>, which is what identified the mechanism. Intermittent because it
+    /// depends on the job firing in that window: solo runs passed every time, full-suite runs failed about half,
+    /// and more wall clock means more chances.</para>
+    ///
+    /// <para>Derived rather than re-hardcoded to a smaller number so that changing the product horizon moves
+    /// every fixture with it, instead of silently reintroducing this the next time it is tuned. The relationship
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Fixed in `b1e680de` — you're right, and it's a better catch than a style note.

`IntervalRetentionSpan` is exactly what I should have used, and `TimescaleContinuousAggregateTests` already pins it equal to the interval string, so it cannot drift from the SQL the policy is actually created with. My `Split(' ')[0]` reimplemented a typed conversion that was already there.

The consequence is worse than the duplication, and it is your point about the static initializer that makes it so: a format the parse could not handle — `"1 week"` — would either take out the whole class as a `TypeInitializationException` or produce a silently wrong seed depth. The second one **reintroduces #2223 through the exact mechanism added to prevent it**, which is the kind of fix that looks like a fix.

Issue #2229, opened 2026-08-12
[CI] Review workflow can execute fully (real token spend) and post nothing - green check with vanished output


erikdarlingdata:
## What

The PR review workflow has a failure mode where it EXECUTES genuinely and its output never lands anywhere. Distinct from the known green-but-skipped case (workflow-differs-from-default-branch self-skip, detectable by the skip signature): here the run completes with conclusion success, no skip signature, and real token spend, but /pulls/N/reviews, /pulls/N/comments, and the issue thread all show nothing posted.

## Evidence

- PR #2213, run 31582450960: 09:20:25 to 09:26:04, success, 3.53 USD of token spend against 14,666 changed lines, zero output anywhere (flagged on the PR thread at 10:36:45Z).
- Same PR, the 2026-08-12 evening head (5e9c6db9): review job pass in 4m39s, and again zero reviews, zero inline comments, zero issue comments from the bot.
- Other PRs the same day (#2224, #2227) received normal claude[bot] issue-comment reviews, so the posting path works in general.

## Why it matters

The discriminator in use for trusting a green review check (no skip signature + token evidence = genuine review) reads this variant as reviewed. The only check that catches it is did-any-comment-appear, which nobody automates. A large PR can merge on the strength of a review that was paid for and never seen.

## Suggested direction

Make the workflow fail (or at least warn-annotate) when the run completes without having posted at least one review/comment artifact - the job knows whether its posting step succeeded. Until then, reviewers should apply the did-any-comment-appear check by hand on large PRs.

erikdarlingdata:
A **third variant** from #2225, plus a near-miss that the suggested fix would not have prevented.

## All 15 reviews had empty bodies — every finding was inline-only

```
15 claude[bot] reviews on #2225, state=COMMENTED, body_len=0  (all of them)
15 inline comments  (carrying 100% of the substance)
 7 issue comments
```

Every one of the 12 real defects that review found on that PR arrived as an **inline** comment on a review whose body was empty. So the failure space has three shapes, not two:

| shape | reviews | inline | issue comments | reads as reviewed? |
|---|---|---|---|---|
| self-skip (known) | 0 | 0 | 0 | no — skip signature |
| executed, posted nothing (this issue) | 0 | 0 | 0 | **yes** |
| **executed, body empty, findings inline** | n | n | 0 | yes, and correctly so |

[...]

PR #2268, opened 2026-08-14
CONTRIBUTING.md sent contributors to a file that is deliberately private (#2243)

### LINKED ISSUE #2243: [CI] The review prompt points at a CLAUDE.md that does not exist on any branch
## What

`.github/workflows/claude-review.yml` opens its prompt with:

> Review this pull request. **Follow the conventions in CLAUDE.md and the T-SQL style guide it points to.**

Neither file exists.

```
$ git ls-files | grep -i claude
.github/workflows/claude-review.yml     # the only match, on any branch

$ git ls-tree -r origin/main --name-only | grep -i 'claude\.md'
(nothing)
```

There is no `CLAUDE.md` on `dev`, none on `main`, and no T-SQL style guide for it to point to. The repo's agent-facing documentation is `CONTRIBUTING.md` and `llms.txt`.

## Why it matters

Every review this repo runs is told, as its first instruction, to follow conventions in a file it cannot open. Two consequences:

1. **The conventions are not actually applied.** The house rules that do exist (T-SQL style, Lite/Darling parity, the no-missing-index-DMV rule that the prompt has to restate inline precisely *because* there is no file to carry it) reach the reviewer only insofar as the prompt repeats them.
2. **It burns a turn on every run.** The reviewer's first move is to look for the file. With `allowedTools` granting no file-reading tool at all (see #2229), that attempt is also a permission denial.

## Fix shape

Either direction is cheap; they are not equivalent:

- **Point the prompt at what exists.** Swap `CLAUDE.md` for `CONTRIBUTING.md` (and drop the dangling "and the T-SQL style guide it points to", or name the guide directly if one should exist). Smallest change, no new file to maintain.
- **Add a real `CLAUDE.md`.** Worth it if the intent was always a single agent-facing conventions file — it would also give the review a place to carry rules that currently have to be inlined in the workflow prompt, which matters because *editing that prompt costs a repo-wide review outage* until the next release (mechanism in #2229).

Note the second option has a side benefit: rules that live in `CLAUDE.md` can be changed without touching `claude-review.yml`, so they cost no drift window.

## Relations

- #2229 (the review can spend real money and post nothing; the same `allowedTools` list that blocks reading this file is that issue's root cause)

erikdarlingdata:
Partially fixes #2243 — the public-facing half, which turned out to be the more consequential one.

## The issue's premise needs correcting

There are **two** dangling `CLAUDE.md` references, not one, and `CLAUDE.md` is **not merely missing**:

```
# .gitignore, since the v1.0.0 initial release
# Internal development files (not for public release)
.internal/
.claude/
CLAUDE.md
AGENTS.md
```

It is **deliberately unpublished**, sitting with `.claude/`, `AGENTS.md` and the TODO files. So the issue's second option — "add a real `CLAUDE.md`" — is not available: committing it would publish a file intentionally kept out of a public repo. **The references are the bug, not the absence.**

I started down the add-the-file path and stopped when `git add` refused it. Worth recording, because the ignore entry is the thing that decides which fix is correct.

## What this fixes

`CONTRIBUTING.md:233` — *"The full T-SQL style guide is in [CLAUDE.md](CLAUDE.md)"* — is the worse of the two, because it is **public and outward-facing**: it tells contributors the full guide is one click away, in a file they can never see.

Fixed by putting the rules in the public document rather than pointing elsewhere. The T-SQL section now carries what was only implied:

- N-prefixed unicode literals, the two-space `ON` continuation, aligned `AND`/`OR` predicates, one term per line in `GROUP BY`/`ORDER BY`
- the three prohibitions that had **no public home at all** — no missing-index DMV recommendations, no full-text search, no ASCII-art headers
- collector-specific conventions an outsider cannot infer: why `OPTION(RECOMPILE)` is on collector queries and that a statement added to a batch needs its own, why migrations schema-qualify, and the comment-density this codebase actually holds

Every rule was read out of the shipped collectors and migrations before being written down.

## Not touched, deliberately

`.github/workflows/claude-review.yml` carries the other reference. Editing it costs a **repo-wide review outage** until the file matches the default branch — `claude-code-action` validates it server-side during the OIDC exchange and skips with `conclusion=success` when it differs. So it belongs with the #2229 `allowedTools` cure that has to ride a release anyway.

My recommendation, recorded on the issue: repoint that prompt at `CONTRIBUTING.md`, which after this actually holds the guide. Leaving #2243 open for that half.

Docs only — no code touched.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -249,13 +249,16 @@ Collector queries specifically:
   varies enormously between a first-run catch-up window and a steady-state minute, and a plan
   cached from one is wrong for the other. A statement added to an existing batch needs its own
   hint — one on a neighbouring statement does not cover it.
-- **Schema-qualify everything in migrations** (`collect.*`, `config.*`). The migrate session's
-  `search_path` resolves bare names to a different schema, so an unqualified `CREATE` can land
-  an object in the wrong one silently.
 - **Comments explain WHY, at length.** This codebase's comments carry measurements, issue
   numbers, and the failure the line prevents. A comment restating the code is noise; one
   recording "this threshold was 300s and the fleet's median gap is 299s, so it discarded half
   of every sweep" is what stops the next person undoing it.
+
+Darling's PostgreSQL store (not T-SQL — the Darling service stores to PostgreSQL/TimescaleDB):
+
+- **Schema-qualify every object in a migration** (`collect.*`, `config.*`). The migrate session's
+  `search_path` resolves bare names to a different schema, so an unqualified `CREATE` or `ALTER`
+  can land an object in the wrong one silently.
 
 ### C# Style
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,8 +242,6 @@ not obvious:
 - **Never suggest missing-index DMV recommendations.** `sys.dm_db_missing_index_*` output
   is not used in this project and changes proposing it will not be accepted.
 - **No full-text search.**
-- **No ASCII-art headers or banner comments.** Block comments carry the reasoning; dividers
-  do not.
 
 Collector queries specifically:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -230,7 +230,34 @@ ORDER BY
 OPTION(RECOMPILE);
 ```
 
-The full T-SQL style guide is in [CLAUDE.md](CLAUDE.md).
+A few more rules that come up in review, and the reasoning behind the ones that are
+not obvious:
+
+- **Unicode literals**: prefix with `N` (`N'ONLINE'`, not `'ONLINE'`).
+- **`ON` continues its `JOIN` at two spaces**, so the join graph reads down the left edge.
+- **`AND` / `OR` align their predicates** (`AND   d.state_desc = N'ONLINE'`), so a `WHERE`
+  clause reads as a list rather than as prose.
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Removed in `27aa3032` — you're right, and I checked rather than argued.

**8 of 60** `install/*.sql` files open with that 79-asterisk banner, and it frames a structured metadata block (Collector / Purpose / Collection Type / Target Table / Frequency / Dependencies / Notes) — documentation with content, not decoration. So the rule as I wrote it is simply false about this repo.

The narrower true statement is that **zero** modern shared collectors use the style — but "the legacy scripts do it and the current ones don't" is an observation about history, not a rule to hand a contributor.

Worth naming why this one slipped: every other rule in this change was read out of the shipped collectors and migrations before being written down. This one I brought with me from elsewhere and did not check — the exact failure the grounding was supposed to prevent. Dropped rather than narrowed, since the "comments explain WHY" bullet already carries the useful intent.

erikdarlingdata:
Fixed in `3d20045d` — correct on all counts, including your verification that `search_path` appears only in `PgMigrations.cs` and in no `install/*.sql`.

Moved under an explicit **"Darling's PostgreSQL store"** heading rather than rephrased in T-SQL terms, because there is no parallel SQL Server rule to state — the install scripts are already fully qualified and resolve by three-part naming, so inventing an equivalent would be padding. The T-SQL section now contains only T-SQL.

That is two rules in this small change that I placed by habit rather than by checking where they apply. The pattern is worth noting: the grounded rules were the ones I read out of the code first, and both mistakes were rules I already "knew".

PR #2269, opened 2026-08-14
Prune per-database state on Azure SQL DB (#2191)

### LINKED ISSUE #2191: [BUG] planwm/backfill state pruning does not reach Azure SQL DB

erikdarlingdata:
Fixes #2191.

## The issue was right that it was blocked, and #2220 unblocked it

#2191 asked for *"an authoritative unfiltered `sys.databases` name list read from the logical master, used only when the master read actually succeeded"* — and correctly rejected the per-cycle Azure enumeration, which filters `ONLINE`, applies the excluded-database list, and falls back to a single database on a master-access error.

**#2220 removed the need for all of that.** A registration that names a database now sweeps only that database, so its one legitimate per-database key comes from the connection string's own catalog: current by construction, nothing filtered, nothing that can go stale, no master read. The question changed from *"which databases still exist on the instance"* to *"which database does this registration own"*, and the second needs no snapshot.

## The shape, and what it deliberately does not have

```sql
DELETE FROM collector_state s
WHERE s.server_id = $1
AND   s.collector_name = $2
AND   starts_with(s.state_key, $3)
AND   s.state_key <> $3 || $4
RETURNING s.state_key
```

**No snapshot guard and no freshness guard, and that is the difference rather than an omission.** The on-prem statement has both because a *snapshot* can be empty or stale. Copying them here would be actively harmful: `database_states` is empty on every Azure server by design, so an anti-join against nothing deletes **every** row instead of none. Pinned as an absence for exactly that reason — with a companion pin that the on-prem statement **keeps** its guards, so a later edit cannot merge the two shapes into one.

A registration naming **no** database, or naming `master`, is still skipped. That is a registration of the logical **server**, whose legitimate database set is everything on it; a single-name prune there would delete every live watermark it has. The gate is the same `AzureSweepScope` predicate #2220 introduced.

## Why this was worth doing tonight rather than later

What it deletes in the field is mostly **#2220's residue** — before that fix every Azure registration swept each sibling database and wrote a watermark for it under its own `server_id`.

And this is the part that would not have resolved itself: **`collector_state` carries no retention.** Verified — zero references in `DarlingRetention`. It is state, not facts. So where the #2220 reporter's team reasonably chose to let the contaminated *collected rows* age out, these orphans would have persisted indefinitely. On an upgraded Azure server this fires once, names what it removed, and clears them.

## Parity

Both SKUs, since Lite supports Azure SQL DB too, iterating the same shared `QueryStorePerDatabaseState.PrunableKeys` set — so a prefix cannot be pruned on one target type and left orphaning on the other. Same log sentence in both, so an operator reading either SKU sees one wording.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -230,7 +230,34 @@ ORDER BY
 OPTION(RECOMPILE);
 ```
 
-The full T-SQL style guide is in [CLAUDE.md](CLAUDE.md).
+A few more rules that come up in review, and the reasoning behind the ones that are
+not obvious:
+
+- **Unicode literals**: prefix with `N` (`N'ONLINE'`, not `'ONLINE'`).
+- **`ON` continues its `JOIN` at two spaces**, so the join graph reads down the left edge.
+- **`AND` / `OR` align their predicates** (`AND   d.state_desc = N'ONLINE'`), so a `WHERE`
+  clause reads as a list rather than as prose.
+- **`GROUP BY` / `ORDER BY` put each term on its own indented line**, so adding one is a
+  one-line diff.
+- **Never suggest missing-index DMV recommendations.** `sys.dm_db_missing_index_*` output
+  is not used in this project and changes proposing it will not be accepted.
+- **No full-text search.**
+- **No ASCII-art headers or banner comments.** Block comments carry the reasoning; dividers
+  do not.
+
+Collector queries specifically:
+
+- **`OPTION(RECOMPILE)` on collector queries.** These run with parameters whose selectivity
+  varies enormously between a first-run catch-up window and a steady-state minute, and a plan
+  cached from one is wrong for the other. A statement added to an existing batch needs its own
+  hint — one on a neighbouring statement does not cover it.
+- **Schema-qualify everything in migrations** (`collect.*`, `config.*`). The migrate session's
+  `search_path` resolves bare names to a different schema, so an unqualified `CREATE` can land
+  an object in the wrong one silently.
+- **Comments explain WHY, at length.** This codebase's comments carry measurements, issue
+  numbers, and the failure the line prevents. A comment restating the code is noise; one
+  recording "this threshold was 300s and the fleet's median gap is 299s, so it discarded half
+  of every sweep" is what stops the next person undoing it.
 
 ### C# Style
 
--- a/Darling/Darling.Tests/AzureForeignStatePruneTests.cs
+++ b/Darling/Darling.Tests/AzureForeignStatePruneTests.cs
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.Linq;
+using PerformanceMonitor.Collectors;
+using PerformanceMonitor.Darling.Service;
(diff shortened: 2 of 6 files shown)


erikdarlingdata:
Fixed in `148caedf` — and you identified a gap rather than an asymmetry, which is the part worth conceding. My Darling tests assert the SQL **string**, so before this **neither** SKU exercised the statement actually running.

You were also right that the scaffolding was already there: extending the existing `Pruner` subclass against `SharedDuckDbFixture` cost almost nothing and now covers `starts_with(state_key, prefix)` plus `state_key <> prefix || own` executed by DuckDB.

Three cases, each carrying something the string assertions cannot see:

1. **Own database kept, siblings retired — including the prefix trap.** An inequality written as a prefix comparison would spare `PaymentsArchive`, because `Payments` is a prefix of it. The on-prem twin already pins that exact shape, so the Azure arm should too.
2. **Empty own-database short-circuits before the statement runs.** This is the arm's safety property, not an edge case: a registration naming no database is a registration of the logical *server*, so a single-name prune there would delete every live watermark it has — silently and totally.
3. **The watermark prefix is covered**, even though Lite writes none today (it never sets `CapturePlanXml`), because the shared prefix list is what stops a prefix being pruned on one SKU and orphaning on the other — and a Lite-only omission would be invisible from Darling.

The tests also seed **no** `database_states` snapshot on purpose: an Azure server never has one, and that this arm does not need one is the whole point.

PR #2286, opened 2026-08-15
feat: add opt-in GCF output format for MCP tool results

blackwell-systems:
### Summary

Adds an opt-in output format, `DARLING_OUTPUT_FORMAT=gcf`, that makes the MCP server return
[Graph Compact Format](https://gcformat.com) instead of JSON. A single call-tool filter
(registered once via `WithRequestFilters` / `AddCallToolFilter`) re-encodes each tool's JSON
result as a GCF generic wire: the repeated field names of the record arrays these tools return
(blocking pairs, wait stats, alerts, config rows, session/query stats, …) are factored into a
single header.

It hooks the SDK's call-tool filter pipeline in one place, so it covers **every tool with no
per-tool changes** — the ~60 data-read tools all flow through it.

### Measured on real tool output (compact baseline)

o200k tokenizer, GCF vs the server's **compact** JSON, every result verified to round-trip, on
eight real MCP payloads captured from a live service against a PostgreSQL target:

- **~26% fewer tokens across the encoded set**, from **54.7%** on the most uniform numeric records
  (`get_pg_index_bloat`) down to **11.3%** where a result is mostly one long unique string per row
  (`get_pg_top_queries`).
- The win tracks payload shape: uniform records with repeated keys win big; text-dominated results
  win least, or fall back to JSON via the never-grow guard.
- Two float-heavy payloads previously fell back to JSON on a precision-guard bug (it declined
  exactly-representable 16-digit doubles); this PR fixes that guard, so those encode too.

### Deliberately conservative — it can only ever help

- **Opt-in and additive.** Default output is JSON, unchanged unless the variable is set.
- **Never larger.** If the GCF wire is not smaller than the JSON for a given result, that result
  stays JSON.
- **Never lossy.** The wire is decoded and compared back to the parsed input; JSON integers are
  decoded as `long` (not `double`), so large ids/counts/durations are never float-rounded, and a
  non-integer is kept only when a double holds it exactly (a shortest-round-trip check, not a
  lossy decimal cast) and is finite. On any parse, encode, or mismatch the JSON is kept — a result
  is never dropped or garbled.
- **Single text block only.** A result carrying anything other than exactly one JSON text block
  (or an error) is left untouched.
- **Zero new footprint.** `BlackwellSystems.Gcf` is a zero-dependency package, pinned exact.

### Why GCF, beyond size

The harder question than size is whether a model reads the compact form as accurately as JSON.
GCF is designed and evaluated for that:

- **Generic-profile comprehension: 100% on every frontier model** (Claude Opus/Sonnet/Haiku,
  GPT-5.5, Gemini 2.5 Pro / 3.1 Pro / 3.5 Flash), 27 runs, 11 models, 4 providers; equal to or
  better than JSON on every model. https://gcformat.com/guide/benchmarks
- The grammar was reverse-engineered from attention-level research into how BPE tokenization
  shapes what a transformer can represent, then validated by controlled from-scratch training at
  1.3B scale, measured causally rather than by an LLM judge. Four papers (2026, under review):
  - GCF: A Token-Optimized Wire Format for Structured LLM Interactions — https://doi.org/10.5281/zenodo.20579817
  - Tokenizer-Attention Coupling — https://doi.org/10.5281/zenodo.20925910
  - Stranded Attention — https://doi.org/10.5281/zenodo.21158886
  - Developmental Atlas of Attention Head Specialization — https://doi.org/10.5281/zenodo.21205389

GCF is already in production across observability, developer tooling, and network automation:
[Chrome DevTools MCP](https://github.com/ChromeDevTools/chrome-devtools-mcp) (Google) merged it
as an experimental format; [Speakeasy](https://speakeasy.com) (customers include Google, Verizon,
Mistral AI) ships it in their `oq` CLI; [OmniRoute](https://omniroute.online) vendored it into its
gateway; [NetClaw](https://github.com/automateyournetwork/netclaw) benchmarked it against TOON on
real network data and replaced TOON; and many others across the MCP and agent ecosystem.

### The change

- `GcfOutput` (env gate + JSON→GCF encode with the never-grow and lossless guards) and
  `GcfCallToolFilter` (the call-tool filter), registered in `DarlingMcpHostService`.
- `BlackwellSystems.Gcf` package reference (central version pin).
- Unit tests: the encode + lossless round-trip against the parsed input, int64 preservation,
  exact-double vs precision-decline (including a 16-digit shortest-round-trip double accepted and a
  non-finite token declined), never-grow and invalid-JSON fallbacks, and the filter's
  enable / disable / error / multi-block behavior. Sample payloads are serialized compact to match
  the server's actual output.
- README (`Darling/README.md` MCP section) documents the variable.
- `dotnet build` clean; targets `dev` per the contribution flow.

GCF is open source and MIT-licensed. Format, spec, SDKs, and benchmarks: https://gcformat.com
Happy to adjust the variable name or extend beyond the single-text-block case; reproducible
numbers and harness on request.


erikdarlingdata (changes requested):
Thanks for this — it's careful work, and it's in the right place. I didn't want to review it from the diff alone, so I built a small console harness against `BlackwellSystems.Gcf` 0.2.0 and pushed realistic Darling payloads through your `GcfOutput.TryEncode` to see the actual wire. Most of what I went looking for, you'd already handled. One thing I found is blocking, and a few are cleanup.

## What's right

**The seam.** One call-tool filter registered once in `DarlingMcpHostService`, covering every tool with no per-tool changes, is exactly where this belongs. If I were adding it myself I'd have put it there.

**The guards, mostly.** Opt-in and default off; never-grow; fall back to JSON on exception at every step; only a lone text block re-encoded so an image alongside it can't be dropped; `IsError` left alone. Int64 survives exactly — I confirmed `9007199254740993` comes out intact rather than as a rounded double.

**Three things I specifically tried to break and could not:**

- **Sentinel collision.** A row whose string value is literally `-` or `~` gets quoted and round-trips byte-exact. That was my first guess at a silent corruption and it's handled.
- **Ragged rows.** Darling's serializers omit nulls on some paths, so a record array can have a field on later rows that the first row lacks. I built exactly that case, expecting the field to fall off a first-row-derived header. It doesn't — the header picks up keys absent from row 0 and the missing cells get `~`. Round-trips identical.
- **Key reordering** between rows normalizes to header order, which is semantically fine for JSON objects.

The package checks out too: MIT, and genuinely zero-dependency (net8.0 and netstandard2.0 groups, no dependencies in either), so the README claim is accurate.

And the wire is legible. `## events [12]{blocked_session_id,blocking_session_id,...}` followed by pipe-delimited rows is something a consuming model reads without trouble — that was my main worry going in, since JSON has universal priors and a new format does not, and I'm satisfied on the clean case.

## Blocking: "lossless" isn't, and the round-trip guard can't see it

The check is:

```csharp
if (Gcf.EncodeGeneric(Gcf.DecodeGeneric(wire)) != wire)
    return null;
```

That compares the wire against **itself**. It's a fixed-point test on the GCF codec, and it will pass for any wire the encoder produces, including one that has already lost data — decode gives back whatever the wire says, re-encode reproduces the wire, guard satisfied. The lossy step is upstream of everything it examines, on the JSON→native leg in `FromJson`:

```csharp
case JsonValueKind.Number:
    if (e.TryGetInt64(out var l))
        return l;
    return e.GetDouble();     // <- here
```

Any number that isn't an Int64 goes through a double. Measured, straight off the wire:

```
input  33.333333333333333     ->  wire  33.333333333333336
input  0.1234567890123456789  ->  wire  0.12345678901234568
```

Both produce a wire that passes the round-trip guard and gets returned to the caller. This is reachable rather than theoretical — Darling's tools return `numeric` values (percentages, ratios, average durations, cost figures), and a computed percentage is exactly the shape above.

I want to be fair about severity: **nobody triages a performance problem on the sixteenth significant digit**, and I'm not claiming operational harm. What I can't accept is the claim. The README says "a lossless round-trip of the JSON" and the PR says "the wire is decoded and must round-trip", and the guard is the entire reason a reviewer would trust this feature to sit in the output path of every tool. A guard that tests the wrong leg is worse than no guard, because it reads as proof.

**Save yourself a detour:** the obvious fix doesn't work. I tried putting a `decimal` into the native model directly, and GCF's generic profile emits it as a *quoted string* — `OrdersDB|"33.333333333333333"|...` — which preserves the digits but changes the JSON type from number to string. So `TryGetDecimal` before `GetDouble` trades a precision change for a type change. The generic profile appears to have no exact decimal number type.

Two ways out that I'd accept:

1. **Minimal and verifiable.** Decline payloads containing a number the model can't hold exactly, letting them stay JSON:

   ```csharp
   case JsonValueKind.Number:
       if (e.TryGetInt64(out var l))
           return l;
       throw new NotSupportedException("non-integer number");   // caught above -> JSON fallback
   ```

   Two lines, makes the losslessness claim true, and costs you the decimal-bearing results — which is honest about what the format can carry.

2. **Thorough.** Make the guard compare against the input: decode the wire, convert back to a JSON model, and check it semantically against a re-parse of `json`. More code, and it subsumes case 1 plus anything else either of us hasn't thought of.

Either way the README and PR wording should say what's actually guaranteed.

## The savings are payload-shaped, and the headline number is the good case

I measured 67% on a clean record array, close to your figure. But on a payload where real query text dominates — multi-line SQL with commas, quotes and pipes, which gets correctly quoted and escaped — it's **28%**:

```
clean get_blocking (short repeated query):   3555 -> 1175 chars   (67% smaller)
10 rows of real multi-line query text:       4236 -> 3069 chars   (28% smaller)
```

Your 48.5–51.4% comes from a `get_blocking` fixture whose `blocked_query` is one short repeated string. That's a legitimate shape, but the tools where token cost actually hurts — `get_top_queries_by_cpu`, `analyze_plan_xml`, deadlock XML — are precisely the text-heavy ones where factoring field names buys least. Worth stating in the README so nobody enables this expecting half their context back on the big results.

## The sentinels need documenting

The wire uses `-` for null and `~` for absent. Nothing in the wire, the README paragraph, or the tool descriptions says so. A model reading `resolved_at` = `-` will probably infer null, but "probably" is doing real work when the reader is an LLM during an incident. Since the point of this feature is to be read by a model, the meaning should be somewhere the model sees.

## Please drop the Dashboard version bump

`deprecated/Dashboard/Dashboard.csproj` goes `<Version>3.3.0</Version>` → `3.3.1` while `AssemblyVersion`, `FileVersion` and `InformationalVersion` stay at `3.3.0.0`/`3.3.0`, which leaves those fields disagreeing with each other in a deprecated project, for a change that has nothing to do with GCF.

If you added it to satisfy `check-version-bump`: that workflow is `on: pull_request: branches: [main]` with `if: github.head_ref == 'dev'`, so it doesn't run on a PR into `dev` at all. Nothing here needed a bump.

## Nits

- `using System.Linq;` in `GcfCallToolFilter.cs` is unused. Won't fail the build, but it's there.
- `Transform` assigns `result.Content` in place on the success path, so it mutates the `CallToolResult` the tool returned rather than producing a new one. Harmless as things stand — the tools build a fresh result per call — but the comment reads like it returns a transformed copy, and the in-place version is the one that would surprise someone later.

## One thing that isn't a defect, and is my call rather than yours

You're the format's vendor contributing your own package — that's fine, and stating it plainly is better than either of us pretending otherwise. What I'm weighing is dependency freshness in a monitoring product: `BlackwellSystems.Gcf` first published 2026-08-09, four releases in six days, and this PR pins 0.2.0, which went up 2026-08-15. A pre-1.0 package with a one-week history in the output path of every MCP tool is a different risk from the same code vendored or behind an interface. MIT and zero-dependency both help. I'm not asking you to solve this; I'm telling you it's part of my decision and it's separate from the quality of the work.

## Where this leaves us

Fix the losslessness guard (either option above) and the wording, drop the Dashboard bump, document the sentinels, and I'll take another look. Also note that **CI hasn't run on this** — only `check-branches` reported, because a fork PR needs maintainer approval for the workflows. I'll approve the run so you get build and test signal, since neither of us has seen this compile in CI yet.

The engineering instinct here is good — conservative guards, fail-safe fallbacks, the right registration point, and comments that explain why rather than what. The gap is that the feature measures tokens and claims losslessness, and only one of those was actually tested.

blackwell-systems:
Thanks for the depth here — the harness especially. You found a real one, and you were exactly right about where it lived: the guard tested the codec's fixed point, not the JSON→value leg, so it couldn't see the `GetDouble` loss in `FromJson`. Fixed, along with the rest.

**Losslessness guard (blocking).** Two changes:

- The fail-safe now compares the decoded wire against the **input model** (`ValuesEqual(DecodeGeneric(wire), native)`), not the wire against itself. Key order is compared order-insensitively, so the ragged-row and key-reorder cases you tried still pass; a value that survived only as a rounded double now fails the compare and falls back to JSON.
- `FromJson` declines any non-integer number a `double` can't hold exactly (via `TryGetDecimal` against `(decimal)d`). `33.5` and `0.25` still encode; `33.333333333333333` and a `ulong` past `Int64` stay JSON. I checked the decline also fires when the decimal is nested inside a record, not just at the top level.

**Wording.** The README now says the savings are payload-shaped (≈half on record arrays, less on text-dominated results like query text and plan/deadlock XML), documents the `-` (null) and `~` (absent) sentinels, and scopes the losslessness claim to what actually round-trips.

**Dashboard bump.** Reverted — you're right it didn't belong, and the check doesn't run on a PR into `dev` anyway.

**Nits.** Dropped the unused `using System.Linq;`. `Transform` now returns a new `CallToolResult` (carrying `StructuredContent`/`IsError`/`Meta`) rather than mutating the tool's result.

**On the dependency-freshness call** — that's yours to make and I won't argue it. Two things that might bear on it, though. First, the integration is built so the SDK isn't load-bearing on the hot path: every result is verified against its input and any mismatch or exception falls back to JSON, so a codec defect costs compression, not correctness. Second, its version number undersells it. `BlackwellSystems.Gcf` is one of a seven-language SDK family — Go, TypeScript, Python, Rust, Swift, Kotlin, and .NET — and .NET was the last one built, against a spec and a 281-fixture conformance suite the other six had already hardened over many release-cycles. All seven are gated by that same shared suite plus a cross-SDK differential fuzz, so every release the .NET package has shipped carries the ecosystem's accumulated fixes rather than rediscovering them; the low number reflects when it started, not how much is hardened into it. The 0.x API-stability question is still fair, and the exact-version pin here keeps that in your control.

erikdarlingdata (approved):
Re-ran the harness against this head rather than taking the fix on trust, since that is how the original defect surfaced. **The blocking issue is fixed, and the fix is the right shape.** Clearing my requested changes.

## Verified against the new code

The two things that matter most were that the defect is actually gone and that fixing it did not break the cases I had already tried to break it with. Both hold — 13/13:

```
lossy decimals now DECLINE
  PASS  33.333333333333333 / 0.1234567890123456789  -> falls back to JSON
  PASS  a lossy decimal nested in a record declines the whole payload
exact non-integers still encode
  PASS  33.5 / 0.25 / 0.5 encode, and the values appear on the wire
int64 boundary
  PASS  9007199254740993 exact (past 2^53)
  PASS  a ulong past Int64.MaxValue stays JSON
the three I tried to break it with STILL work
  PASS  ragged rows (a later row carrying a field row 0 lacks)
  PASS  literal '-' and '~' string values
  PASS  key reorder between rows (order-insensitive compare)
savings intact
  PASS  clean record array: json 8735 -> wire 2617 (71% smaller)
guards unchanged
  PASS  invalid JSON falls back;  tiny payload (never-grow) falls back
```

The guard change is exactly the correction that was needed — `ValuesEqual(DecodeGeneric(wire), native)` verifies against the input model, so a value that only survived as a rounded double now fails the compare instead of passing a fixed-point test that could never see it. And declining in `FromJson` via `(decimal)d != exact` is the better half of the fix: it fails *before* a wire exists rather than catching it afterwards.

Making the key comparison order-insensitive was the right call too — it keeps the header-order normalization from reading as data loss, which is what would otherwise have taken ragged rows and key reordering down with the real bug.

## Two small things, neither blocking

**The word "lossless" is still unqualified in two code comments.** The README now scopes it properly, but `GcfOutput`'s class comment still reads "Opt-in, lossless, and never larger than the JSON" and `GcfCallToolFilter`'s says "never larger, never lossy". That is the exact wording that sent me looking in the first place — worth matching the README's precision, since the code comment is what the next reader hits first.

**The remaining double-domain corner is real, and you already named it.** A non-integer outside the decimal range keeps its `double`, and the guard cannot see that particular loss because `native` itself holds the already-rounded value — the same structural blind spot as before, now narrowed to a corner instead of covering every decimal. Your comment says so plainly ("A token outside the decimal range is inherently double-domain"), which is the right way to handle it. Noting it only so it stays a known limit rather than becoming a surprise: it is unreachable for this codebase in practice, because Darling's numerics are int64 counts and microsecond durations plus decimal-range ratios and percentages, none of which carry 29+ significant digits.

Nits confirmed fixed: unused `using System.Linq` gone, and `Transform` now returns a new `CallToolResult` carrying `StructuredContent`/`IsError`/`Meta` instead of mutating the tool's own. Dashboard version bump reverted — the diff no longer touches that file at all.

## On the dependency call

That one is Erik's, not mine, and I am not going to pre-empt it — but your framing changes the input to it, so I will relay it accurately rather than paraphrase it away. "0.2.0, published the day before the PR" reads very differently once it is the seventh binding of a spec already hardened by six siblings against a shared 281-fixture conformance suite and a cross-SDK differential fuzz, versus a package written last week. That is a fair correction to how I characterised it, and the exact-version pin does keep the upgrade in this repo's control.

The 0.x API-stability question stands on its own, as you say. Worth adding one thing in your favour that I checked rather than assumed: the integration really is built so a codec defect costs compression and not correctness — every result is verified against its input and any mismatch or exception returns the original JSON, and `Transform` bails before touching a result that is an error or carries anything other than a lone text block. So the blast radius of the dependency being wrong is "results stay JSON", which is the property that makes the freshness question a judgement call rather than a risk.

Nice work on the turnaround, and thanks for taking the harness seriously rather than arguing with it.

erikdarlingdata:
Thanks for this, and sorry it sat — I read the code rather than the summary, and it is better built than most integration PRs I see. The filter is registered once so it needs no per-tool changes, it declines anything that is not a lone text block so an image alongside a result can never be dropped, and the never-grow guard means a bad case degrades to today's behaviour instead of a bigger payload.

The part I want to single out: **`TryEncode` verifies the decoded wire against `native` — the model the input JSON parsed to — not against itself.** That is the correct fail-safe and it is the one people usually get wrong.

Which makes the gap below worth naming, because it is the only place that reasoning slips.

## Findings

**1. `NumbersEqual` launders exactly the precision loss the guard exists to catch.** `FromJson` carefully keeps integers as `long` so large values are never float-rounded, and the comment on the fail-safe says a match means the JSON survived the round trip. But when one side is `long` and the other is `double` — which the comment says can happen, since an integer may decode as either — the comparison converts the `long` to `double` and compares. For any integer above 2^53 that conversion loses precision **on both sides identically**, so the values compare equal and the wire is accepted even though it no longer carries the input value. The guard is exact for `long`/`long` and inexact for precisely the case it was written to catch. Fix is one line: in the mixed branch require the double to be integral and to convert back to the same `long`, rather than comparing in double space.

I am not claiming a reachable data-loss bug today — I could not find a tool that emits an integer above 2^53 (durations are milliseconds, ids and counts are far below it, hashes serialize as strings). It is a latent hole in a guard whose entire value is being airtight.

**2. The strongest claim in the description is the one thing not tested.** `TryEncode_RecordArray_Is_Smaller_And_RoundTrips` asserts `Assert.Equal(wire, Gcf.EncodeGeneric(Gcf.DecodeGeneric(wire)))` — wire to wire. That proves GCF is stable under its own decode/encode; it says nothing about whether the wire faithfully represents the **input JSON**, which is the property the PR is selling and which the runtime actually enforces. A test that decodes the wire and compares to the parsed input (the `ValuesEqual` path), plus one that feeds a payload that must be *declined*, would cover the fail-safe itself. Right now the fail-safe is the least-tested code in the change.

**3. Lite gets nothing.** `Lite/Mcp/McpHostService.cs` is a second MCP host exposing the same tool names — `Lite.Tests/CrossAppMcpToolInventoryPinTests.cs` exists specifically to keep the two inventories from drifting. This lands on the Darling host only, so the two SKUs would answer differently for identical tool names. That is the asymmetry this repo tries hardest to avoid.

**4. An environment variable is the hardest possible switch for the deployment this targets.** Darling's primary mode is a Windows service running as a virtual service account. Setting a per-process env var there means editing service configuration or a machine-wide variable and restarting — while everything else about Darling is configured in `darling.json`. The feature is most awkward to enable exactly where Darling actually runs.

## The measurement I would want you to see

Your table compares GCF against the current **`WriteIndented = true`** JSON. I rebuilt your `Blocking(n)` shape and measured how much of that is indentation alone:

| rows | indented | compact | indentation's share |
|---|---|---|---|
| 10 | 2,977 | 2,297 | **22.8%** |
| 30 | 8,757 | 6,757 | **22.8%** |
| 100 | 29,082 | 22,462 | **22.8%** |

Bytes, not tokens — and I want to be straight that the **token** saving from dropping indentation is smaller than 22.8%, because BPE packs runs of spaces efficiently, so bytes overstate it here. But it is not nothing, it costs one serializer option, and it carries no new dependency, no new wire format, and no new failure mode. Any honest evaluation of GCF should be against **compact** JSON, since that is the free baseline. Against indented JSON, some unknown part of the win is being credited to the format for work a flag does.

## The dependency, stated plainly

You are the author of `BlackwellSystems.Gcf`. That is completely legitimate — format authors contributing integrations is how formats spread — but it should be on the record in the thread rather than inferred from the namespace.

What I verified on nuget.org: MIT, genuinely zero transitive dependencies, targets net8.0 and netstandard2.0. Also **first published 2026-08-09, six days before this PR, with five versions in eight days** (0.1.0 → 0.2.1), and pre-1.0, where a minor bump is allowed to break. This PR pins 0.2.0; 0.2.1 shipped the day after it was opened.

In this repo's favour: Central Package Management plus locked-mode `packages.lock.json` pins by content hash, so a republished version cannot silently change under us. That is a real mitigation and it is already in place.

Still, this adds a third-party package to the **service** assembly — shipped to every Darling install, including the ones that never set the variable — on a product that runs as a service on production database hosts. **Whether to take that dependency is Erik's call as the vendor, not something I will merge on his behalf.** I have flagged it for him with exactly the facts above; I am not going to pretend the code quality is the open question, because it is not.

## Where this stands

The PR is **CONFLICTING** and needs a rebase onto `dev` regardless — and note that while it is conflicting GitHub silently stops running the `pull_request` workflows, so the checks on the page are stale rather than passing.

If the dependency question comes back yes, findings 1–3 are what I would want fixed first. If it comes back no, I would still like to take the compact-JSON change on its own — that came out of your PR and it is a real improvement to every consumer, so it should be credited to you either way.

erikdarlingdata:
Thanks for the follow-up commits — the losslessness work is real. Verifying the decoded wire against the literals the payload was built from, rather than against values re-derived from the wire, is the right shape for that test and catches the failure mode that matters. The precision guards (declining 17-significant-digit decimals, declining `ulong` above `Int64`, keeping exact integers above 2^53 instead of letting them render as floats) are careful work, and the filter's conservatism is well judged: error results untouched, multi-block results untouched so nothing gets dropped, `StructuredContent` carried over, and a new result returned rather than mutating the tool's own.

There is a problem with what the tests establish, though, and it goes to whether the feature earns its keep.

## The benefit is measured against a baseline the product no longer emits

Every sample payload is built with `WriteIndented = true`, and `TryEncode_RecordArray_Is_Smaller_And_RoundTrips` asserts `wire.Length < json.Length` against that. But **#2350 shipped in 3.5.0 and made MCP tool results serialize compact** — the shared `McpHelpers.JsonOptions` no longer indents. So the "GCF is smaller" claim is demonstrated against pretty-printed JSON that the server stopped producing.

At runtime this is not a correctness bug: the filter passes the tool's actual output to `TryEncode`, which is now compact, and the never-grow guard means a wire that is not smaller returns null and the JSON is kept. That guard is doing its job.

But it makes the feature's value an open question rather than a demonstrated one, and the answer decides the PR:

- if GCF beats **compact** JSON by a worthwhile margin, this is worth having
- if it does not, `TryEncode` returns null on real payloads, the feature is inert, and the dependency is carried for nothing

Could you re-run the size comparison with the sample payloads serialized the way the server actually serializes them (no `WriteIndented`), and post the numbers? Worth noting from #2350's own measurement: the byte saving there was 23–36 % depending on payload shape, but the **token** saving was smaller than the byte saving, because BPE tokenizers pack runs of whitespace efficiently. Since the consumer here is a language model, tokens are the metric that matters — bytes are a proxy that flatters whitespace removal specifically.

## The dependency is the thing I need to weigh

`BlackwellSystems.Gcf` 0.2.0 is a pre-1.0 package published by the account opening the PR. I am not implying anything by that — but it is a third-party runtime dependency entering a product that holds database credentials and runs as a service on customer infrastructure, and the format's definition lives outside this repo. That is a different decision from accepting a code contribution, and it is mine to make rather than something CI can settle.

Practically, what would make it easy to say yes:

- what the package's dependency footprint is (ideally zero beyond the BCL)
- whether the wire format is specified somewhere durable, so a decoder can be written without the package
- the maintenance commitment behind a 0.x version

An alternative worth considering: if the encoder is a few hundred lines with no dependencies of its own, vendoring it under `PerformanceMonitor.Darling.Service/Mcp/` sidesteps the supply-chain question entirely and keeps the format pinned to a version this repo controls.

## Two mechanical things

**`Directory.Packages.props` shows +33/-32 for one real line.** The whole file is rewritten with different line endings; the only substantive change is the `BlackwellSystems.Gcf` entry. Please re-apply that as a single-line addition preserving the file's existing endings — as it stands the diff hides the one line that matters inside 32 lines of noise, which is exactly the change a reviewer should see clearly.

**CI has not actually run.** Only `check-branches` reports, which is the signature of the `pull_request` workflows not firing — `build` and `Darling PostgreSQL tests` are both required and neither has a result, so this cannot merge regardless of review. A rebase onto current `dev` and a push should get them running. Worth doing anyway: `dev` has moved a long way, including #2350 above.

## Where this stands

Not going into 3.5.0 — that is cut. No urgency on your side. Get the compact-JSON numbers and the rebase in, and answer the dependency questions, and I will give it a proper decision.

erikdarlingdata:
Ran the measurement I asked for. Numbers below are from **eight real MCP payloads** captured off a live
service (`/api/read/<tool>`), against a self-hosted PostgreSQL target with 40 tables and 160 indexes — not
synthetic fixtures. I extracted `GcfOutput.cs` from this branch and ran it against `BlackwellSystems.Gcf`
0.2.0, so this is your encoder and your guards, not a reimplementation.

### Tokens vs **compact** JSON (`o200k_base`)

Since #2350 the baseline is compact, and as I suspected the byte savings flatter whitespace removal — BPE
packs indentation cheaply, so the token column is the real one.

| tool | JSON tok | GCF tok | **token saving** | (byte saving) |
|---|---:|---:|---:|---:|
| `get_pg_index_bloat` | 4,663 | 2,112 | **54.7%** | 75.7% |
| `get_pg_wait_sampling` | 2,260 | 1,137 | **49.7%** | 66.7% |
| `get_pg_io_stats` | 967 | 683 | **29.4%** | 37.0% |
| `get_pg_index_usage` | 22,909 | 16,378 | **28.5%** | 33.2% |
| `get_pg_database_stats` | 1,565 | 1,241 | **20.7%** | 22.0% |
| `get_pg_top_queries` | 16,226 | 14,396 | **11.3%** | 9.7% |
| `get_pg_column_stats` | 2,925 | — | **DECLINED** | — |
| `get_pg_predicate_stats` | 4,398 | — | **DECLINED** | — |
| **total (encoded only)** | **48,590** | **35,947** | **26.0%** | |

So: **26% fewer tokens across what encodes**, not the ~50% in the PR description — that figure is against
indented JSON and is no longer the baseline. The spread tracks exactly what you'd expect: uniform numeric
records with repeated keys win big (bloat, wait sampling), and `get_pg_top_queries` wins least because it is
mostly one long unique string per row, which GCF cannot help with. 26% on the payloads a model actually
reads is worth having.

### Dependency footprint: zero beyond the BCL — confirmed

From the published nuspec, both groups are empty:

```xml
<group targetFramework="net8.0" />
<group targetFramework=".NETStandard2.0" />
```

103,851 bytes, shipping only `BlackwellSystems.Gcf.dll` for `netstandard2.0` and `net8.0`. Latest published
is 0.2.1 (this branch pins 0.2.0). That answers the question — nothing transitive comes with it.

### The two declines are a bug in the precision guard, and they are yours to fix

This is the part that matters more than the percentage. Both declines come from `FromJson`:

```csharp
var d = e.GetDouble();
if (e.TryGetDecimal(out var exact) && (decimal)d != exact)
    throw new NotSupportedException("number not exactly representable as a double");
```

**`(decimal)someDouble` keeps only 15 significant digits.** The tokens tripping it are 16-digit
shortest-round-trip doubles — values that are *exactly* representable as a double by construction:

```
token                 0.5029000043869019
double, "R" format    0.5029000043869019   <- round-trips exactly
(decimal)d            0.502900004386902    <- 15 digits, so the comparison fails
```

These are PostgreSQL `float4`/`float8` columns (`top_value_frequency`, `filtered_pct`) widened and
serialized shortest-round-trip. 10 such tokens in `get_pg_column_stats` and 31 in `get_pg_predicate_stats`
are enough to decline the whole payload each time. The guard rejects precisely the double-domain values it
is written to allow — **2 of 8 real payloads, 25%, inert for no reason.**

The check you want is a string round-trip, not a decimal cast:

```csharp
var d = e.GetDouble();
if (!double.IsFinite(d) ||
    !d.ToString("R", CultureInfo.InvariantCulture)
        .Equals(e.GetRawText(), StringComparison.Ordinal) &&
    e.TryGetDecimal(out var exact) && (decimal)d != exact)
    throw new NotSupportedException(...);
```

— or simply drop the decimal comparison and test `double.TryParse(raw) → "R" → equals raw`.

**Mirror hole in the same branch:** a token that overflows double, e.g. `1e400`, makes `TryGetDecimal`
return *false*, so the guard does not fire, `GetDouble()` yields `Infinity`, and it encodes. The fail-safe
does not catch it either — the decoded value is also `Infinity`, so `ValuesEqual` says they match, and the
wire silently carries `Infinity` where the JSON had a finite token. No collector emits that today, so it is
not urgent, but it is the same guard failing open while the other half fails closed. An explicit
`double.IsFinite` check covers it.

### Where that leaves the PR

26% on real payloads, zero transitive dependencies, opt-in and behind a never-grow guard plus a round-trip
fail-safe — that clears the bar I set, so I want this. Three things before I merge:

1. Fix the precision guard so those two payloads encode, and add a test pinning a 16-digit
   shortest-round-trip double (`0.5029000043869019` is a real captured value) as **accepted**.
2. Add the `double.IsFinite` check with a test pinning `1e400` as **declined**.
3. The `Directory.Packages.props` rebase / line-ending noise from my earlier review is still outstanding.

Reproduction for the numbers above, if you want to re-run them: capture payloads from `/api/read/<tool>`,
re-serialize with `WriteIndented = false`, encode via `GcfOutput.TryEncode`, and count with `o200k_base`.
Happy to hand over the eight captured payloads if that helps.

erikdarlingdata:
Re-ran the measurement against your fixes, same eight real MCP payloads, same method. **Both guards are
correct now and all eight encode.**

| tool | JSON tok | GCF tok | saving |
|---|---:|---:|---:|
| `get_pg_column_stats` | 2,925 | 1,305 | **55.4%** |
| `get_pg_index_bloat` | 4,663 | 2,112 | **54.7%** |
| `get_pg_predicate_stats` | 4,398 | 2,067 | **53.0%** |
| `get_pg_wait_sampling` | 2,260 | 1,137 | **49.7%** |
| `get_pg_io_stats` | 967 | 683 | 29.4% |
| `get_pg_index_usage` | 22,909 | 16,378 | 28.5% |
| `get_pg_database_stats` | 1,565 | 1,241 | 20.7% |
| `get_pg_top_queries` | 16,226 | 14,396 | 11.3% |
| **total** | **55,913** | **39,319** | **29.7%** |

**8 of 8 encoding, up from 6 of 8, and 29.7% up from 26.0%.**

The pleasing part: the two payloads the old guard declined are now the **best two results in the set**.
`get_pg_column_stats` and `get_pg_predicate_stats` are the most uniform data here — wide rows of repeated
keys and numbers — which is exactly the shape GCF is good at, so the guard was rejecting the cases that
most justified the feature. That is a better argument for the change than the average is.

The shortest-round-trip comparison is the right call over the decimal cast, and the `IsFinite` check closes
the other direction. Worth noting for anyone reading later that a single-value probe like `{"v":0.1}` still
declines — that is the **never-grow guard**, not precision, because one field is smaller as JSON than as
GCF. It caught me out when I was first characterising this, and it is the encoder behaving correctly.

`Directory.Packages.props` is a clean one-line addition now, and the dependency footprint question is
answered: the published nuspec has `<group targetFramework="net8.0" />` and
`<group targetFramework=".NETStandard2.0" />`, both empty — nothing transitive comes with it.

All three review points are addressed and the numbers clear the bar I set. Thank you for the careful
turnaround on the guard — you fixed both directions of it, not just the one that was failing.

One thing outstanding that is not yours: the full test suite has not dispatched on this branch, so only
`check-branches` has reported. I want to see it green before merging rather than merge on the strength of a
local measurement. I will sort out the run.

erikdarlingdata:
Merged. Thank you — this was a good contribution and you handled the review well.

For the record of what settled it: **29.7% fewer tokens across eight real MCP payloads**, measured against compact JSON with `o200k_base`, and 8 of 8 encoding after your guard fix. The two payloads the original guard declined turned out to be the best two results in the set at 55.4% and 53.0% — the most uniform data here, which is exactly what GCF is good at, so the bug was rejecting the cases that most justified the feature.

You fixed both directions of that guard rather than only the failing one, which is why I stopped worrying about it: the shortest-round-trip comparison for the 16-digit case, and `IsFinite` for the token that overflows to Infinity.

CI needed a maintainer click to dispatch — fork PRs from first-time contributors are gated here. That is on our side, not yours, and it will not apply once you have a merged PR. All six checks green, including the locked-mode restore, which was the one thing I actually wanted to see run.

Issue #2296, opened 2026-08-16
Two cross-region servers sit at 100% of their sweep budget — collection body exceeds the 60s cadence every cycle, so they collect at half rate


erikdarlingdata:
Found during the 2026-08-16 dogfood watch on `prod-sql-use2-monitor-01`. This is the only recurring warning left on that box after the decommissioned registrations were removed: **~50/hour**, steady across four hourly windows.

```
[prod-sql-use2-multi-01...] collection body has not completed after 62s of execution - skipping relaunch
[prod-sql-use2-multi-53...] collection body has not completed after 61s of execution - skipping relaunch
```

## It is not a fault, and that is the point worth recording

`get_collection_health` on `multi-01` reports **all 40 collectors HEALTHY, 0 errors, 0 yields, 0% failure rate**. Nothing is failing. The skip is back-pressure working exactly as designed — the previous body is still running, so relaunching would pile up.

The cause is arithmetic, visible in the per-collector averages:

```
procedure_stats   22,141 ms
query_store       16,590 ms
plan_correction   13,544 ms
query_stats        8,437 ms
                 ---------
                  ≈ 60.6 s   against a 60 s sweep
```

Four collectors consume the entire cadence on their own. So the body cannot finish inside one interval, every interval, and the relaunch is skipped every time.

## The consequence

**These two servers effectively collect at half the nominal cadence.** Nothing reports that as degraded — every collector is HEALTHY and the log line reads like a transient. A reader looking at collection freshness sees "Warning" status at worst, and the health tool sees nothing wrong at all, because from each collector's own point of view nothing is.

## Why these two specifically

Both registrations resolve to **us-east-1** endpoints (`prod-sql-use1-multi-01`, `prod-sql-use1-multi-53`) while the collecting box is in **us-east-2**. Every round trip pays cross-region latency, which inflates all four of the expensive collectors at once. The other 40 servers on the same box, collected in-region, do not produce this warning.

Nine registrations on this box currently point at use1 endpoints; only these two saturate, so latency is necessary but not sufficient — these are presumably also the busiest of the nine.

## Options, roughly in order of appeal

1. **Move them to the us-east-1 box.** `prod-sql-use1-monitor-01` exists, is built, and is in-region for them. It is currently blocked on the `darling_monitor` role, so this is gated on that credential rather than on any code.
2. **Give the sweep a longer interval for saturated servers**, so the body fits. Changes the data grain for those servers.
[...]

PR #2308, opened 2026-08-17
Serve a sweep_pressure verdict from get_collection_health so half-rate collection is visible (#2296)

erikdarlingdata:
Implements option 3 of #2296 — the "worth doing regardless" one. Options 1 (move the two servers in-region, gated on the use1 `darling_monitor` credential) and 2 (longer cadence for saturated servers, a data-grain decision) stay open on the issue; this PR does not close it.

## The gap

Two servers collect at half their configured cadence and nothing but a service-log warning says so. The mechanism: `RunDueCollectorsAsync` runs collectors serially and re-stamps `NextDue` from one `now`; when the body's own execution outlives the 60s cadence (multi-01's four heaviest collectors average ~60.7s combined), it finishes after the next due time, the watchdog skips the relaunch (~50 warnings/hour), and the delivered interval is a multiple of the configured one. Meanwhile `get_collection_health` reports every collector HEALTHY — correctly, because from each collector's own seat nothing is wrong. No counter of skipped relaunches exists anywhere, in memory or the store.

## The signal, and why this one

`get_collection_health` (both SKUs) now serves a top-level `sweep_pressure` block: `busy_ms_per_minute`, `busy_percent`, a verdict (`OK` / `AT_RISK` ≥ 75% / `SATURATED` ≥ 100%), the three heaviest contributors (attribution — "which collectors spend the budget" is the actionable half), and a `note` naming the consequence and the levers when the verdict isn't OK.

**Built from execution demand, not delivered-gap statistics, deliberately**: at fleet scale the delivered cadence stretches benignly — bounded sweep concurrency queues bodies, and the fleet-wide median gap already runs a multiple of the configured minute — so gap-based detection would flag all 52 servers and drown the two that matter. Execution demand (each collector's `avg_duration_ms` amortized by its own cadence) is the arithmetic behind the watchdog's own "has not completed after Ns of EXECUTION" distinction, and queueing cannot inflate it. This is also why the watchdog itself classifies Hang (running) apart from Queued (capacity).

No SQL change, no migration, no schema change — the roll-up is computed from the rows the tool already reads plus the shared `CollectorScheduleDefaults`, resolved exactly the way the health banding already resolves cadence (shipped defaults, for three-surface parity).

## Shape

- Decision in `SweepPressureClassifier` (`PerformanceMonitor.Common`), pure and static like `CollectorHealthClassifier` — callers resolve cadence, only the verdict lives there. Thresholds carry their rationale: 75% because the average smooths variance (intermittent overrun is how saturation looks before it's constant), 100% because the body mathematically cannot fit. The motivating server computes 101.2%.
- Both SKUs' tool projections are byte-identical twins, as the rest of this tool already is; `CollectorHealth`/`CollectorHealthRow` expose their existing cadence resolution internally rather than duplicating it.
- Tool description + MCP instructions row updated so a caller learns the field exists.

## Tests

`SweepPressureClassifierTests` — the same decision table pinned in Darling.Tests and Lite.Tests (the `CollectorHealthClassifier` twin-suite pattern): the motivating profile verbatim (22,141 + 16,590 + 13,544 + 8,437 ms → SATURATED at ~101%), a healthy profile, both inclusive band edges, on-load/zero-duration exclusion, own-cadence amortization (the pin that forbids charging a heavy daily collector at the fast cadence), and the empty window.

## Verification scope

Both test projects build clean on macOS (`EnableWindowsTargeting`). The Windows suites can't run locally, so the full decision table was executed against the real `PerformanceMonitor.Common` build via a throwaway net10.0 harness: 12/12 pass, including the motivating server's exact numbers. Not covered locally: the xUnit twins executing (CI's build job), and the tool end-to-end against a live store — the number it rolls up (`avg_duration_ms`) is already live-pinned by the existing collection-health tests.


erikdarlingdata:
Both passes' one finding fixed in 59eebce7 — Lite's McpInstructions row now carries the same sweep_pressure hint as Darling's, so the session-init guidance and the per-tool Description agree in both SKUs. Good catch on a file the diff didn't touch.

Issue #2160, opened 2026-08-10
[BUG] ScaleTest flaky on fast runners — 200k vs 50k rows can't outrun the fixed-cost floor (d4=217ms < d1=279ms)


erikdarlingdata:
### Component
Darling.Tests (CI flake)

### What
`StoreSelfMetricsTests.ScaleTest_JobDurationGrowsWithVolume_TelemetryRecordsIt_AndTheAlertFires_AgainstDevPostgres` failed on PR #2100 (a dependency bump that doesn't touch this code): `4x volume did not run longer than 1x (d1=279ms, d4=217ms)`.

At sub-300ms absolute durations the run is dominated by per-run fixed cost (scheduler dispatch, chunk discovery) plus cache warmth that GROWS across the two measured runs — 200k vs 50k rows isn't enough separation for the volume term to reliably win on a fast runner. The monotonicity premise is right; the signal-to-noise at this scale isn't.

### Fix
Widen the second seed to 500k rows (10x). At that separation the variable term is an order of magnitude above the observed fixed-cost floor, so monotonicity holds even with warm caches and runner jitter. Test-only change; the telemetry-recorded and alert-fires assertions are unaffected.

erikdarlingdata:
Fixed on dev by #2161 — second seed widened to 500k rows (10x); CI's live TimescaleDB run passed with the new separation.

PR #2345, opened 2026-08-19
Assert what the store scale test can prove, not TimescaleDB's throughput (#2266)

### LINKED ISSUE #2266: [TEST FLAKE] Two live-store tests fail on unrelated diffs and pass on re-run (scale-test noise floor, and a database-state heal race)
Two tests failed today on PRs whose diffs cannot reach them, and both passed on a re-run of the **same commit**. Filing rather than leaving them as re-run folklore.

## 1. `StoreSelfMetricsTests.ScaleTest_JobDurationGrowsWithVolume_TelemetryRecordsIt_AndTheAlertFires_AgainstDevPostgres`

```
10x volume did not run longer than 1x (d1=970ms, d10=863ms) — job runtime is not scaling
with volume, which invalidates the #2136 capacity model.
```

`StoreSelfMetricsTests.cs:404`. Failed on #2265, whose diff is the Azure sweep predicate plus two new test files.

**Root cause is in the assertion, not the product.** It compares two sub-second durations on a shared CI runner and requires the larger volume to be strictly slower. At ~900 ms the run-to-run noise exceeds the signal it is trying to measure — 1× happened to land 107 ms slower than 10×. The capacity model it guards is real; the *measurement* just has no margin.

Worth considering: require a ratio rather than an inequality (e.g. `d10 > d1 * 1.5`), scale the volumes until the difference clears the noise floor, or take a best-of-N. As written it will keep failing at a low rate forever, and each failure costs a re-run and a moment of "is that mine?".

## 2. `DatabaseStateExpectedStoreTests.RebaselinedByHandDuringAnOutage_HealsOnceTheDatabaseRecovers`

```
Assert.Empty() Failure: Collection was not empty
Collection: [DatabaseStateInfo { DatabaseName = "Payments", ExpectedState = "SUSPECT",
             LastAlertedState = "", StateDesc = "ONLINE" }]
```

`Lite.Tests/DatabaseStateExpectedStoreTests.cs:342`. Failed on the same PR one commit later — a commit that deleted an unused method and rewrote two doc summaries, so it cannot reach this test either. Passed on re-run.

I did **not** get to a root cause on this one. The expected state was still `SUSPECT` when the test expected it cleared, which reads like an ordering or visibility issue in the store rather than noise, so it may be a genuine race worth someone looking at properly rather than a timing tolerance.

## Why file it

Both cost a re-run today, and #2223 is already open for a third flake in the same suite family. A re-run that goes green teaches nothing durable, and the next person to see either failure will spend the same few minutes deciding whether it is theirs. Two of these in one afternoon on unrelated diffs is also the point at which the pattern matters more than the individual tests.

erikdarlingdata:
Closes #2266

That is the second and last open item on the issue.

## Where each item stood

**Item 2 (`DatabaseStateExpectedStoreTests` heal race) is already fixed on `dev`** and I re-verified it rather than redoing it: #2275 replaced the single-cycle assertion with `SweepUntilAsync`, and #2281 added the product-side logging for a silently-skipped maintenance cycle. The bounded retry cannot mask a regression — a genuinely broken heal never settles, all five cycles run, and the caller's own `Assert.Empty` fails on the final result exactly as before. Nothing to do there.

**Item 1 was still flaky.** #2287 added a diagnostic to it but deliberately changed no assertion, on the stated grounds that every candidate fix needed a measurement nobody had a rig for. This PR gets that measurement and acts on it.

## Root cause of item 1

`ScaleTest_JobDurationGrowsWithVolume_...` required `d10 > d1` between two sub-second compression-job durations. **The volume-dependent part of that measurement is ~10% of the measurement's own magnitude on the runner where it fails, so the assertion has no margin — and it is a benchmark of TimescaleDB's compression throughput on shared CI hardware, not a claim about this product.**

I found a TimescaleDB 2.29 / PG17 container on this machine and ran the exact sequence — same shipped `TimescaleSupport` SQL builders, same `PgMigrations`, same `StoreSelfMetrics.SweepAsync`, same seeds, same arm/poll/park scheduler helper — from a throwaway `net10.0` console harness, 15 consecutive times:

| | measured |
|---|---|
| `d1` (50k rows) | 23–39 ms |
| `d10` (500k rows) | 104–167 ms |
| absolute delta | **65–104 ms** |
| ratio for a 10x volume increase | 2.67 / 3.42 / 5.17 (min/median/max) |
| chunks after warm-up / 1x / 10x | 1 / 2 / 3, **every run** |
| per-day rows | exactly 2000 / 50000 / 500000, **every run** |

Two things fall out, and they correct the record on both earlier readings:

1. **The "both runs compressed nothing" hypothesis is refuted.** The chunks compress cleanly and the volumes land exactly as intended, 15 for 15. The byte-identical `d1=689/d10=689` pair also was never as improbable as it looked, because that pair is only ever *read* when the test fails, which selects for differences already near zero.
2. **The scaling is heavily sublinear** — 10x the rows buys ~3.2x the duration, about **85 ms** of absolute signal, because compression cost is largely fixed per run. CI's baseline for the same pair is 690–970 ms, roughly twenty times that fixed cost, so on CI the signal sits inside the run-to-run variance of launching a background worker on Windows. That is what produced `d1=970/d10=863`.

So no threshold, ratio, or volume rescues it. At ~0.19 ms per thousand rows it would take millions of rows per chunk to clear a variance nobody has measured on the platform that actually fails, and #2160's earlier "4x was not enough, go to 10x" was the same signal being misread as a volume problem.

## What makes it deterministic now — and stronger, not looser

The duration inequality is replaced by an assertion on **work actually done**, which is exact rather than timed: each measured run must have compressed the chunk its own seed created, and that chunk must hold exactly the seeded row count.

```
Assert.Equal(new long[] { 2_000, 50_000 }, work1.RowsPerDay);
Assert.Equal(2, work1.ChunksTotal);
Assert.True(work1.ChunksCompressed == 2, ...);

Assert.Equal(new long[] { 2_000, 50_000, 500_000 }, work10.RowsPerDay);
Assert.Equal(3, work10.ChunksTotal);
Assert.True(work10.ChunksCompressed == 3, ...);
```

With the 1-day chunk interval this hypertable is created at, day groups and chunks are the same thing, so `RowsPerDay` is both the per-chunk row census and the "each seed produced exactly one chunk" check. Row counts are read straight through the hypertable — compressed chunks stay transparently queryable, so `count(*)` is exact and needs none of the pre/post-2.18 columnstore-vs-compression view vocabulary the rest of this file has to hedge on.

**A negative control proves the difference rather than asserting it.** Two sabotage modes, 3 runs each:

| sabotage | old assertion | new assertions |
|---|---|---|
| 10x rows seeded into a **not-yet-eligible** chunk (the hypothesized "both runs did nothing") | FAIL 3/3 | FAIL 3/3, naming `compressed=2` |
| 10x rows seeded into the **1x chunk** (fixture stops producing two chunks at two volumes) | **PASS 3/3, ratio 6–8x** | FAIL 3/3, naming `rows=[2000,550000]`, `total=2` |

The second row is the point: the shipped assertion was not merely flaky, it was **blind to the fixture defect it existed to guard**. It passed with a healthy-looking ratio while the test had stopped measuring two scales at all.

Everything the product actually owns is unchanged and still asserted end to end: a real duration is measured, the V56 `store_metrics` series records both readings in order, and the real `DarlingSelfAlertEvaluator` fires the Critical cadence tier from a real `ReadJobCadenceReadingsAsync` reading. **One gap closed on the way past:** `d10 > 0` was never asserted, and `ReadJobDurationMsAsync` maps a NULL duration to `0`, so a 10x run whose duration was unmeasurable satisfied the telemetry check as `0 == 0` and passed. It now fails, with what the job did attached.

The test is renamed to `ScaleTest_EachRunCompressesItsOwnChunk_TelemetryRecordsBothRuns_AndTheAlertFires_AgainstDevPostgres` so the name stops claiming what it no longer measures, and the summary carries the measurements plus an explicit "do not reintroduce a duration inequality" so the next person has the numbers instead of the temptation.

## Verified / not verified

- **Verified live**: the two new SQL statements were *extracted from this file's shipped source* by the harness and executed against TimescaleDB 2.29 / PG17 — not retyped. 12/12 green on the fix, 6/6 red across the two sabotage modes.
- **Verified**: `Darling.Tests` compiles clean (`EnableWindowsTargeting`, `global.json` restored, not committed); no stacked `/// <summary>` blocks; no other reference to the old test name anywhere in the repo.
- **Not verified locally**: `Darling.Tests` targets `net10.0-windows`, so the xUnit test itself cannot run on macOS. The assertions and their SQL are proven; the xUnit plumbing around them is CI's to confirm. The rig is Linux/aarch64 on TimescaleDB 2.29, while CI is Windows on the bundled `pg-runtime` — that difference is the subject of the finding, not a gap in it, but the absolute timings above are this machine's, not the runner's.


erikdarlingdata:
Reviewed the diff, not just the report — merging.

The replacement is verifiably **stronger**, which is what I was checking for:

- The **negative control** settles it: seeding the 10x rows into the 1x chunk makes the OLD assertion pass with a 6-8x ratio while the fixture has quietly stopped producing two chunks at two volumes, and only the new assertions catch it. A flake fix that still passes the old test is the failure mode I was watching for; this is the opposite.
- Step 2's series assertion is now **exact** (`points[0] == d1`, `points[1] == d10`) where it was "growing". So the product's half of #2136 — record the real durations, in order, feed the cadence alert — is pinned harder than before. What was dropped is only TimescaleDB's compression throughput, which this suite has no business policing.
- `d10 > 0` genuinely was never asserted, and `ReadJobDurationMsAsync` mapping NULL to 0 meant an unmeasurable 10x run satisfied the telemetry check as `0 == 0`. Real hole, closed.
- `ReadCompressionWorkAsync` throwing while `DescribeJobWorkAsync` stays best-effort is the right split: assertion inputs and failure explanations have opposite requirements.

One observation, not a blocker: when `d1 == d10` (the byte-identical case that started this), the two series equality assertions cannot distinguish order. The work assertions carry the weight now, so it is worth knowing rather than fixing.

Item 2 correctly identified as already fixed on dev by #2275/#2281 and left alone rather than re-fixed.

PR #2346, opened 2026-08-19
Bound the per-database watermark read to the clamp horizon (#2344)

### LINKED ISSUE #2344: The watermark read is an unbounded MAX over every chunk in retention — 1-3s per database per cycle, and it scales with store size, not workload
Filed off the #2333 post-install measurement. With the plan-fetch walk gone, the per-database log split shows the phases that remain, and one of them does **not** subside after catch-up: `wm`, running 1–3 seconds per database per cycle on the use1 primaries box.

`wm` is not a query against the monitored SQL Server. It is a read against **our own store**:

```sql
SELECT MAX(last_execution_time) FROM query_store_stats WHERE server_id = $1 AND database_name = $2
```

(`DarlingCollectorRunner.GetLastCollectedTimeForDatabaseAsync`.) `query_store_stats` is a hypertable partitioned on `collection_time`, so a `MAX` over a *different* timestamp column with no time predicate has to touch **every chunk in retention** for that (server, database).

## Measured on the live use1 store (106 GB, 5 chunks, 42 primaries)

Same server/database pair, `EXPLAIN (ANALYZE, BUFFERS)`, current form vs. the same query with a `collection_time` predicate added:

| | current (unbounded) | bounded to 3h |
|---|---|---|
| cold (first touch) | **25,766 buffer reads + 195 written** (temp spill) | 5 chunks excluded, **29 ms** |
| warm (cache primed) | **228 ms** | **29.5 ms** |

Warm is a 7.7x difference. The cold row is the one that matters for the field: ~200 MB of physical I/O per database per cycle, and the cost is a function of **store size and cache residency**, not of anything the monitored server is doing. So it degrades exactly where a customer is weakest — a long-lived store, a busier Query Store, slower disks, a box whose cache is not dominated by this one query. Our dogfood store is 5 chunks; a 90-day-retention store is far more.

## The bound is provably free

Every consumer of this value ends up at `max(stored, now - 1h)`:

- `WatermarkPolicy.MaxCatchup` is **1 hour**, and `ClampCatchup` floors anything older to `now - 1h`.
- A **null** result falls back to the definition's documented first-run window, which is **60 minutes** (`QueryStoreCollector.BuildCutoffParameters`) — the same instant.

So a row whose `last_execution_time` is older than an hour **cannot change the answer**: found or not found, the caller proceeds from `now - 1h`. Adding a time predicate is therefore behaviour-identical, not an approximation — the first thing to check when this is implemented, because "identical" is the whole justification.

## Fix shape

1. Bound the read on the **partitioning** column (`collection_time`), which is what enables chunk exclusion — a predicate on `last_execution_time` alone would not prune. A row's `last_execution_time` cannot exceed its `collection_time` (you cannot collect an execution from the future), so `collection_time > now() - margin` cannot hide a qualifying row. Use a margin comfortably wider than `MaxCatchup` (3h against a 1h horizon) so clock skew between the monitored server and the store cannot bite.
2. **Do not hardcode the horizon in the generic helper.** `GetLastCollectedTimeForDatabaseAsync` serves any definition with a `PerDatabaseWatermarkColumn`; the 1-hour clamp is query_store's policy. Pass the bound in (nullable — no bound = today's behaviour) so a future collector without a clamp keeps correct semantics rather than silently inheriting a horizon it never agreed to.
3. **Check the siblings.** `GetLastCollectedTimeAsync` (server-scoped) and Lite's DuckDB twin have the same unbounded-MAX shape. Lite's store is smaller but the argument is identical, and a fix on one SKU only is the drift this codebase keeps paying for.
4. While here: `idx_query_store_stats_server_db_query_plan_time` exists but is not serving this MAX. Worth checking whether the bounded form wants a supporting index or whether chunk exclusion alone is enough — the measurement above says exclusion alone gets it to 29 ms, so an index is probably unnecessary write tax.

## Why this is worth doing before the release

Erik's framing: "there's no accounting for local factors, e.g. busier query stores, worse hardware." This particular multi-second cost is not the customer's SQL Server at all — it is our store, it scales with how long they have been running, and it is removable by adding a predicate that provably cannot change the answer. That is the cheapest juice left in the collection path.

erikdarlingdata:
Closes #2344. The first of Erik's "further juice to squeeze before releasing" — and it is the phase that survives catch-up, so it is the one that follows an operator home.

## What the split showed

After #2333 removed the plan-fetch catalog walk, the per-database log line reads:

```
query_store [BETA]  => sql:15594ms = wm:238ms + open:1270ms + drain:31ms + plan_fetch:13444ms + text_fetch:611ms
query_store [theta] => sql:7431ms = wm:3134ms + open:566ms + drain:23ms + plan_fetch:2672ms + text_fetch:1036ms
```

`plan_fetch` is catch-up work and drains to near-nothing by design. **`wm` does not.** And `wm` is not a query against the monitored SQL Server — it is a read against our own store.

## Measured, live use1 store (106 GB, 5 chunks, 42 primaries)

Same server/database pair, `EXPLAIN (ANALYZE, BUFFERS)`:

| | unbounded (today) | bounded to 3h |
|---|---|---|
| cold (first touch) | **25,766 buffer reads + 195 written** (temp spill) | 5 chunks excluded, **29 ms** |
| warm (cache primed) | **228 ms** | **29.5 ms** |

Warm is 7.7x. The cold row is the one that matters in the field: the cost is a function of **store size and cache residency**, not of the monitored workload — so it is worst on a long-lived store, a busier Query Store, slower disks, or a box whose cache is not dominated by this one query. Our dogfood store is five chunks; a 90-day-retention store is many more.

## Why the bound is free rather than a tradeoff

Every consumer of this value ends at `max(stored, now - MaxCatchup)`:

- `MaxCatchup` is **1 hour**; `ClampCatchup` floors anything older to `now - 1h`.
- A **null** result falls back to query_store's documented **60-minute** first-run window — the same instant.

So a row older than the horizon cannot change the outcome whether it is found or not, and the unbounded `MAX` was paying to confirm a value the clamp would have produced anyway. `WatermarkPolicyTests` pins both halves: the floor sits **strictly older** than the horizon (the ordering *is* the safety argument), and a buried watermark and a not-found one reach the identical instant.

Two mechanical details that are easy to get wrong:

- The predicate goes on **`collection_time`** (the partitioning column). A predicate on the watermark column alone prunes no chunks. It is safe because a row's watermark value can never exceed its own `collection_time` — an execution cannot be collected before it happens — so nothing qualifying hides behind the bound.
- The bound is **opt-in per caller and name-guarded to query_store**, on both hosts. A ring-buffer collector whose legitimate catch-up spans days (`default_trace_events`) must keep reading its whole history; handing it this floor would silently truncate that. The clamp and the bound travel together — a future definition wanting one has to adopt the other, which the policy remarks say explicitly.
- The server-scoped `GetLastCollectedTimeAsync` is **deliberately untouched**: that path does not clamp, so the equivalence does not hold there.

## Parity and verification

Both SKUs, same shape (Postgres gets chunk exclusion; DuckDB gets min-max pruning and a smaller scan — different mechanism, same predicate and same argument). All four affected projects build.

The policy arithmetic was executed against the built assembly in a throwaway net10.0 harness — which **caught a bug in my own test fixture** (an `InlineData(2)` case asserting a 2-hour-old watermark sits below a 3-hour floor; it does not). That would have been a red CI round. Fixed to 4h before pushing. The Windows test projects build here but cannot run, so the harness is the only pre-CI execution available.

**Not yet measured end-to-end in production**: the before/after above is the query in isolation via EXPLAIN, not the `wm` phase in the running service. That lands when a nightly carrying this installs, and the log split will show it directly.


erikdarlingdata:
Force-pushed a rebase onto clean `origin/dev`. The previous push carried an unrelated commit: three agents were working the same checkout concurrently and one of them briefly landed #2266's `StoreSelfMetricsTests.cs` change on local `dev`, which this branch was then cut from — so the diff showed a test file this PR has nothing to do with, plus that PR's CHANGELOG entry. Both are gone; the diff is now the six files this change actually touches. `origin/dev` was never affected, and #2266's own work is intact on its branch (#2345).

erikdarlingdata:
Reviewed the diff and merging. This is the one of the three I care most about, because it is the only Query Store cost that does **not** subside after catch-up — it is our own store's size, not the monitored server's workload.

Fresh numbers from use1 that make the case, 183 runs over the last 30 minutes post-catch-up:

| | before the activity-driven fetch | now | after this |
|---|---|---|---|
| avg | 46.2s | 7.7s | — |
| median | 40.8s | 6.6s | — |
| p95 | 104s | 17.5s | — |
| max | 313s | 32.5s | — |

The worst live sample splits as `sql:9346ms = wm:3894ms + open:1230ms + drain:124ms + plan_fetch:1952ms + text_fetch:2146ms`. **`wm` is 42% of that server's total**, and the other phases are still shrinking as the backlog drains while `wm` is not — it scales with store size and cache residency. A 106 GB store on modest disk is exactly where an operator is least able to absorb it, which is the release-sensitivity concern.

What I verified rather than took on faith:

- **The floor sits strictly below the clamp horizon.** `now - MaxCatchup - ReadFloorMargin` with a 2h margin against a 1h horizon, so a watermark that could still move the answer can never fall below the bound. The margin exists only for clock disagreement between a monitored server and the store; the correctness argument needs none of it, which is why it is generous rather than tuned.
- **The equivalence is real, not approximate.** Every consumer lands on `max(stored, now - MaxCatchup)`: `ClampCatchup` floors anything older, and NULL falls back to the documented 60-minute first-run window — the same instant. So a row below the floor cannot change the result whether found or not. The `InlineData(4/6/48/2160)` test asserts exactly that: a watermark below the floor clamps to the same instant as finding nothing at all.
- **It bounds the PARTITIONING column.** The predicate is on `collection_time`, not the watermark column, or TimescaleDB prunes nothing. Safe because a row's watermark can never exceed its own `collection_time` — an execution cannot be collected before it happens.
- **It is scoped to exactly one collector.** Guarded on `QueryStoreCollector.Instance.Name`, and the server-scoped `GetLastCollectedTimeAsync` is deliberately untouched. Handing this floor to a ring-buffer collector whose legitimate catch-up spans days would silently truncate it. The bound and the clamp travel together, and a future adopter has to take both.
- **Both stores got it.** Postgres and DuckDB, same shape, or Lite reads a permanently different value.

Prediction I will check on the box after the nightly: `wm` on the worst use1 server drops from 1–3s to ~30ms, and that server's total falls from ~9.3s to ~5.5s.

erikdarlingdata:
Rebased onto `dev` — #2343 landed a `### Fixed` entry in the same spot, so this went CONFLICTING the moment that merged. Resolution keeps both entries in issue order; the code diff is untouched (6 files, same 156/11).

While resolving I also took the cosmetic nit the review bot raised on #2343: the double blank line before the `## [3.5.0]` heading. It was created by that merge and every other version boundary in the file uses a single blank, so it belonged with the conflict resolution rather than in a PR of its own.

Worth flagging for anyone watching CI here: **a CONFLICTING PR silently stops GitHub running `pull_request`-triggered workflows.** The checks do not fail, they simply never queue, and the last green run stays on the page looking authoritative. That trap already cost a round earlier today across three PRs. If checks look stale, compare the run SHA against the head SHA before believing them.

PR #2355, opened 2026-08-19
Stop pretty-printing MCP tool results (#2350)

### LINKED ISSUE #2350: MCP tool results are pretty-printed, and every consumer of them is a model
Every MCP tool result both SKUs return is serialized with `WriteIndented = true`, and the only consumer of an MCP tool result is a language model. We are paying to indent JSON for a reader that does not care.

It is **one line in one shared file**: `PerformanceMonitor.Common/Mcp/McpHelpers.cs:32`

```csharp
public static readonly JsonSerializerOptions JsonOptions = new() { WriteIndented = true };
```

Used by 30 of the 74 Darling MCP files and 24 of the 31 Lite ones. It lives in Common, so one change moves both SKUs together — no parity gap to manage. Two readers (`DarlingAgReader`, `DarlingFleetReader`) declare their own options with the same flag and would need the same treatment.

## Size

Measured on a 15-field record-array result shaped like `get_blocking`:

| rows | indented | compact | indentation's share |
|---|---|---|---|
| 10 | 2,977 | 2,297 | **22.8%** |
| 30 | 8,757 | 6,757 | **22.8%** |
| 100 | 29,082 | 22,462 | **22.8%** |

**These are bytes, not tokens, and the token saving is smaller than the byte saving** — BPE tokenizers pack runs of spaces efficiently, so indentation is cheaper in tokens than it looks in bytes. Anyone quoting this should measure tokens before claiming a number. The direction is not in doubt, though, and the change costs nothing.

## Why bother if it is only whitespace

Because it compounds with the thing we actually care about. Tool results are the bulk of what fills an agent's context when it works a real incident, and the fleet-wide reads are the widest results we return. A result that no longer fits is the difference between an agent reasoning over 100 blocking events and over 40.

## Provenance

This came out of reviewing #2286, which proposed an opt-in third-party wire format and benchmarked it against our **indented** output. Separating the two matters: some part of the win reported there is a serializer flag rather than the format, and the honest baseline for evaluating any compact format is compact JSON. Credit to @blackwell-systems for the measurement work that surfaced it — this half of it carries no new dependency and no new failure mode, so it stands on its own regardless of what happens to that PR.

## Not before 3.5.0

Deliberately not folding this into the release currently staged. It changes the bytes every MCP consumer sees, and the release is already carrying the Aurora fix, the watermark bound, the install-location diagnostic and the peer disclosure. This goes on `dev` after the tag.

## Check before changing

Confirm nothing parses our own tool output positionally or by line. A grep for tests asserting on serialized MCP output with embedded newlines should come first — `CrossAppMcpToolInventoryPinTests` parses the instructions census, not tool results, but it is worth proving rather than assuming.

erikdarlingdata:
Closes #2350.

One property on the shared `McpHelpers.JsonOptions` in Common — 78 call sites across both SKUs' MCP surfaces already route through it — plus the two readers (`DarlingAgReader`, `DarlingFleetReader`) that carry their own options so the `/api/*` endpoints and the MCP tools keep serializing an identical shape.

## The saving, stated honestly

Payload-shaped: **23% of the bytes** on a 15-field record array, **36%** on a narrow one. **The token saving is smaller than the byte saving** — BPE tokenizers pack runs of spaces efficiently, so anyone quoting these numbers as token savings will be wrong. The argument is that it costs nothing and compounds where it matters: tool results are the bulk of what fills an agent's context during a real incident, and the fleet-wide reads are the widest results we return.

## Scope was the whole risk

Flipping `WriteIndented` on something that turns out to write a config file would make `servers.json` a single unreadable line — the kind of damage nobody notices until they open the file by hand at a bad moment. So I checked before changing:

- `McpHelpers.JsonOptions` is **MCP-output only**. Of 78 usage files, exactly one sits outside an `Mcp/` folder (`McpPlanAnalysisFormatter`, still an MCP formatter), and **nothing writes a file with it**.
- `ServerManager`, `ProfileManager` and `ScheduleManager` carry their own `WriteIndented = true` and are untouched.

## Nothing depended on the layout

- No test asserts on serialized MCP output containing newlines or indentation (the two `\n` hits in the suites are an axis tick label and a hand-written JSON literal, both unrelated).
- The 24 test files that read tool JSON use a parser, which is whitespace-insensitive.
- The tests that touch this options object assert field **names** — there is no naming policy on it, so snake_case comes from `[JsonPropertyName]` — not layout.

## Tests

`McpOutputCompactionTests` pins the flag, the observable consequence (no `\n`, no `\r`, no indent run), that compaction changes **layout only and not content** (same parsed values, and genuinely smaller), and the boundary — those three config writers still indent, asserted structurally the way this repo pins invariants it cannot compile.

## Verification

Builds clean, 0 warnings. Suites target `net10.0-windows` so they cannot run on macOS; every assertion in the new test was exercised first in a throwaway `net10.0` harness and passes.

## Provenance

Found while reviewing #2286, which proposed a third-party wire format and benchmarked it against our **indented** output — so part of the win it reported was a serializer flag rather than the format. Credit to @blackwell-systems for the measurement that surfaced this; this half carries no new dependency and stands on its own.


erikdarlingdata:
CI caught a gap in my own pre-change check, and the fix is worth describing because the first instinct would have been wrong.

**What I missed.** I verified "nothing depends on the layout" by grepping for tests asserting on serialized output containing **newlines**. Eighteen assertions depended on the space after the colon instead — `"severity": "Critical"` only exists under `WriteIndented` — so four test classes failed without a single thing they were testing having changed. My check was narrower than my claim.

**Why I did not just retype them compact.** Every one of those assertions reads as a claim about *content*: this field serialized with this value, this enum came out as its string name rather than its ordinal, this null stayed null. They were written as claims about *formatting*. Flipping `": "` to `":"` would have made them pass while leaving exactly the same trap armed for whoever changes the formatting next.

So `JsonAssert.Contains`/`DoesNotContain` normalize both sides — dropping whitespace **between** tokens, preserving whitespace **inside** strings, so `"a": "b c"` and `"a":"b c"` compare equal and the two-space value in `"b c"` survives. Escaping is tracked, because if a `\"` inside a string is read as the end of it the scan falls out, starts stripping real spaces from values, and the assertion silently starts comparing something else. That is verified separately: escaped quote, escaped backslash immediately before a closing quote, inner double-spaces, newline+indent, and empty all pass.

Still substring assertions rather than a full parse, on purpose — they pin one field's serialization without pinning the shape of the envelope around it.

**Four format-coupled assertions are deliberately left alone**: the exported `darling.json`, the network config editor, a stored command result, and a static options constant. Those are config files people read and one literal — none of them MCP output — and CI passed all four, which is the evidence that the scope of the change really is what I claimed.

Issue #2391, opened 2026-08-21
[BUG] #2349's file-growth knobs have no Settings window or MCP surface, so an alert that ships OFF cannot be enabled from any supported surface


gotqn:
### Component

Darling

### Performance Monitor Version

3.5.0

### SQL Server Version

SQL Server 2022

### Windows Version

Windows Server (Darling service host), Windows 11 (Viewer workstation)

### Describe the Bug

[#2349] shipped the **Database File Growth** alert deliberately OFF. Its four knobs exist in storage - V79 adds `file_growth_enabled`, `file_growth_rise_mb`, `file_growth_volume_percent` and `file_growth_lookback_minutes` to `config.config_alert_settings`, and Lite carries the same four in its app settings - but they are absent from every surface an operator would actually use:

- **Darling Viewer → Settings window:** no control. The alert-thresholds section has its usual neighbours (tempdb space usage, volume free space with its CRITICAL grade, monitor self-alerts, store background job cadence, version store (PVS), agent jobs, database state), and the summary line at the bottom of that section reads "Will alert when: Total CPU > 80%, blocking >= 1, deadlocks >= 1, poison waits >= 500ms avg, queries > 30min, tempdb > 80%, disk free < 10% or 5GB, PVS >= 40% of database, jobs > 3x avg, failed jobs (last 60m)" — no mention of file growth.
- **MCP `get_alert_settings` / `update_alert_settings`:** no `file_growth` group. `tempdb`, `low_disk`, `pvs` and `self_alerts` are all present.
- **Lite Settings window:** no control either.

So the only way to switch on a feature that ships OFF is a direct `UPDATE` against `config.config_alert_settings` (Darling) or hand-editing `settings.json` (Lite).

This also departs from the pattern [#2107] established, where every previously-hardcoded threshold was given all three surfaces at once: "six new knobs ride the store control plane, the Viewer's Settings window, and `get_alert_settings`/`update_alert_settings`, clamped on read like their siblings". #2349's knobs have the store plane only.

It looks like an unwired half rather than a deliberate omission, because the rest of the plumbing is complete and consistent with its siblings: `DarlingAlertSettings` already clamps the values on read (`Math.Max(0, …)` on the rise, `[0,100]` on the volume percent, `[5,1440]` on the lookback), and the V79 columns carry sensible NOT NULL defaults.

### Steps to Reproduce

1. Upgrade a Darling install to 3.5.0 and let the store migrate to V79.
[...]

Issue #2386, opened 2026-08-20
Tightening plan_content_retention_days permanently stops retention on query_plan_dim (one-day slice exceeds the 300s delete timeout)


erikdarlingdata:
Setting `plan_content_retention_days` to a value that creates a backlog **turns retention off** on `query_plan_dim` — permanently, silently, and on the largest table in the store. Making the horizon stricter is what breaks it.

Found on the use2 dogfood store by a prediction I registered in advance on #2316 and which did not come true.

## Evidence

The knob went from its default 21 to **14** at 2026-08-19 11:25, moving the cutoff back seven days. Every purge since has failed on the same table:

```
08-19 11:11  58 table(s), 407167 row(s) deleted, 19 chunk(s), 0 failed, 150097ms   <- knob still 21
08-19 17:42  57 table(s),      0 row(s) deleted,  0 chunk(s), 1 FAILED, 301057ms
08-19 18:27  57 table(s),      0 row(s) deleted,  0 chunk(s), 1 FAILED, 301463ms
08-19 20:13  57 table(s),      0 row(s) deleted,  0 chunk(s), 1 FAILED, 600253ms
08-19 20:58  57 table(s),      0 row(s) deleted,  0 chunk(s), 1 FAILED, 301238ms
```

```
Retention purge failed for query_plan_dim: Exception while reading from stream
```

`301057ms` and `301463ms` are not workload variance — that is `DeleteTimeoutSeconds` (300 s) expiring on the command.

The dimension has not moved in a day, and normal ingest has already refilled the one dip:

| day | plan_dim | rows |
|---|---|---|
| 08-18 | 129.0 GB | 12,359,332 |
| 08-19 | 132.6 GB | 11,850,329 |
| 08-20 | 132.6 GB | 12,380,713 |

The 08-19 dip is the 11:11 sweep, which ran while the knob was still 21. Everything after it deleted nothing.

## Mechanism

`TimeSlicedDeleteSql` deletes the oldest **one-day** slice per statement, and that width is the work bound: "one day of arrival volume per statement — exactly what a steady-state daily purge deletes in total".

[...]

Issue #2387, opened 2026-08-20
[BUG] Darling: PLAN_REGRESSION analysis query full-scans/decompresses entire query_store_stats history every cycle (no collection_time bound)


ethosthalsell:
### Component

Darling

### Performance Monitor Version

3.5

### SQL Server Version

SQL Server Version: N/A — this is the Darling PostgreSQL/TimescaleDB store's own internal analysis query, not a monitored-server DMV query. (Monitored fleet is a mix of SQL Server 2014–2022.)

### Windows Version

Windows 11

### Describe the Bug

PgFactCollector.QueryPerf.cs 's  PlanRegressionSql  (feeds the  PLAN_REGRESSION  analysis fact, run every scheduled analysis cycle — default 30 minutes — for every monitored server) filters  collect.query_store_stats  only on  last_execution_time >= $2  (a 14-day window, no upper bound):

FROM v_query_store_stats
WHERE server_id = $1
AND   execution_type_desc = 'Regular'
AND   last_execution_time >= $2

 query_store_stats  is a TimescaleDB hypertable partitioned/compressed on  collection_time  (per  TimescaleSupport.cs ), not  last_execution_time . Since the predicate never touches  collection_time , TimescaleDB can't exclude chunks by time — every cycle has to scan, and for compressed chunks decompress, the entire history of that table for the given server.  PgTableTuning.cs 's covering indexes ( idx_query_store_stats_query_hash ,  idx_query_store_stats_server_db_query_plan_time ) don't cover  last_execution_time  either, so there's no index path around it.

The code comment on  CollectPlanRegressionFactsAsync  even flags this as deliberate: "Unlike other collectors this windows on last_execution_time (14-day comparison window), NOT collection_time." That tradeoff wasn't validated at scale — once the store's  collect  schema passes ~50GB and enough history has moved into compressed chunks (1-day  compress_after  policy), this query becomes a full-history decompress-and-scan, repeated per server, per cycle.

### Steps to Reproduce

1. Run a Darling store monitoring a non-trivial fleet (56 servers here) long enough that  collect.query_store_stats  exceeds ~50GB and most chunks are TimescaleDB-compressed.
2. Let the scheduled analysis job run ( analysis.intervalMinutes , default 30) across all monitored servers.
3. Watch  postgres.exe  CPU/backend count during an analysis pass.

### Expected Behavior

[...]

PR #2395, opened 2026-08-21
3.5.1 release prep: version bump and changelog cut

erikdarlingdata:
Version bump to 3.5.1 across the three live csprojs plus the deprecated Dashboard csproj that the dev→main version check reads, and the changelog cut.

3.5.1 is a patch release for two defects **3.5.0 itself introduced**, both of which are silent:

- **#2388 / #2386** — plan-dimension retention bounded its `DELETE` by a time slice rather than by rows. Slice width is set by data at the old end, not by how far behind retention is, so the statement could exceed the 300s timeout and roll back to **zero rows deleted** — or, worse, succeed on a thin slice and log `0 failed` while the table grew. Measured on the dogfood store: deleting ~6% of what crossed the horizon, falling ~360k rows/day behind, 1.16M expired rows accumulated. Evidence on #2386.
- **#2390** — the PLAN_REGRESSION analysis query never bounded the partitioning column, so chunk exclusion could not engage and every run read the whole `query_store_stats` history.

Plus **#2391** (file-growth knobs reachable on all three surfaces), **#2384** (SHA256SUMS written with LF so `shasum -c` works), and **#2385** (fixture doc comment corrected).

`check-version` is expected RED on this PR — it compares against main and only goes green once dev→main merges. That is the same behavior as the 3.5.0 cut.


erikdarlingdata:
Putting this back to draft — I cut it before validating the thing it exists to fix, which is the same gap that let #2386 and #2387 ship in 3.5.0 in the first place.

#2388's batched drain has never run against real data. Its coverage is a fake-executor loop test plus a string assertion on the SQL; the two have never met a real planner together. #2397 adds that (110 expired rows, cap 25, asserts the total **and** the batch count), and it runs under `DARLING_TEST_PG` in CI.

Gate for taking this out of draft:

1. #2397 green — the drain provably drains rather than peeling one slice.
2. A build carrying #2388 on `prod-sql-use2-monitor-01`, watched through at least one purge cycle, showing the 1.16 M-row backlog actually falling. That store is the exact condition the fix targets and it is sitting right there.

The measurement to watch is `query_plan_dim` row count day over day, not the purge log — the log said `0 failed` throughout the week it was falling ~360k rows/day behind.

erikdarlingdata:
Both gate conditions are met, so this is out of draft.

**1. #2397 green** — the live drain test ran against real PostgreSQL (confirmed executed, not skipped: `DARLING_TEST_PG` was set, 5306 tests, and the payload-dimension skip reason appears zero times in the job log).

**2. Verified on the box in the failure state.** `dev` went onto `prod-sql-use2-monitor-01` and the restart-triggered purge was caught mid-run:

```
pid 2772 | active | 00:00:12.3 | DELETE FROM query_plan_dim WHERE ctid IN (
    SELECT ctid FROM query_plan_dim WHERE last_seen < $1 ORDER BY last_seen LIMIT 50000)

10:46:00  Retention purge: 58 table(s) purged, 451644 row(s) deleted, 23 chunk(s) dropped, 0 failed, 340034ms
```

Against four consecutive runs on 08-19 that hit ~301 s and deleted **zero** rows each. Afterwards: 0 rows past 23 days, oldest surviving `last_seen` exactly 22.0 days old.

## Scope stays as cut, deliberately

`dev` is now 33 commits ahead of this branch and has picked up real features — Lite's MCP alert-settings parity (#2410), the update-check switch (#2415), `settings.sample.json` (#2423), the AG and connection groups (#2421). Folding those in would enlarge the risk surface of a release whose entire purpose is two silent regressions, and would use a patch number for a feature release.

So this ships what it says it ships. Every item in it — #2388, #2390, #2391, #2384, #2385 — is in the build currently soaking on all three monitor boxes, so the contents are validated rather than merely merged. `dev` accumulates toward 3.6.0.

One correction to my own gate note above: I wrote that the 1.16 M figure was the backlog. It was measured against `now() - 21 days`, the knob value, and the purge cuts at knob+1. Measured correctly the eligible set was far smaller, and the "falling 360k rows/day behind" reading was a category error — comparing deletions against total row growth, most of which is rows too young to be eligible. The defect being fixed is unaffected: four runs deleting zero rows because an unbounded DELETE that hits the timeout rolls back everything. Full correction on #2386.

erikdarlingdata:
Closing — superseded.

This was scoped as a five-item 3.5.1 patch. Since then `dev` has moved 300+ commits ahead of `main`: ten viewer surfaces gained endpoints (#2484), fifteen reads learned to distinguish the two kinds of nothing (#2485), every windowed read gained an `as_of` anchor (#2495), the Azure SQL Database read and collection defects were found and fixed (#2511, #2512, #2515), Lite's portable ZIP halved (#2501), and the store schema went 79 → 81.

A patch version cannot carry that, and this branch's changelog cut describes a product that no longer exists. The next release will be **3.6.0**, cut from `dev`, and the work this PR was going to ship is already in it.

Not rushing it out with known validation gaps: no nightly has yet built the self-contained Lite ZIP, nothing has been deployed to a dogfood box to exercise the 79 → 81 migration against a real store, and no Darling instance has actually collected from a live Azure target. Those get closed first.

Issue #2425, opened 2026-08-21
A malformed settings.json silently resets all 88 Lite settings, before logging exists to say so


erikdarlingdata:
Found while making Lite's sample settings file real (#2418). It is the most serious thing that work turned up and it is in neither issue.

## One trailing comma resets all 88 settings, silently

Both settings loaders swallow every exception and fall back to defaults:

```csharp
catch { /* Use default */ }    // App.xaml.cs, LoadDefaultTimeRange
catch { /* Use defaults */ }   // App.xaml.cs, LoadAlertSettings
```

`LoadAlertSettings` wraps the whole `JsonDocument.Parse` plus all 88 `TryGetProperty` calls in that one try. So a single malformed character does not cost one setting — the parse throws before any key is read and **every** setting reverts to its default at once. Alert thresholds, SMTP configuration, webhook enablement, retention, theme, the lot.

## And the app structurally cannot tell you

The loaders run *before* logging exists:

```
LoadDefaultTimeRange();      <- here
LoadAlertSettings();         <- and here
   ... 13 lines ...
AppLogger.Initialize(logDirectory);
```

So there is no log line, no dialog, and no way to add one without reordering startup. The user sees a Performance Monitor that has forgotten its configuration, with nothing anywhere explaining why, and the most likely next move is to reconfigure everything by hand — writing the defaults back over the file that still holds the real settings, at which point the original is gone.

This matters more now, not less. #2418 ships a documented reference precisely so people hand-edit settings.json, and hand-editing JSON is how you get a trailing comma. We are about to make the triggering action more common while the failure stays invisible.

## What to do

The two halves are separable and the first is most of the value.

**Report it.** Capture the exception and surface it once logging exists — a deferred diagnostic the loader records and `AppLogger.Initialize` flushes is enough, and avoids reordering startup. A user who is told "settings.json could not be parsed at line 42, defaults are in use" can fix it in a minute.

**Do not overwrite what you could not read.** This is the part that turns an annoyance into data loss. If the parse failed, `WriteSetting`'s next whole-document rewrite replaces a file whose contents were never understood. At minimum, copy the unreadable file aside (`settings.json.unreadable-<timestamp>`) before the first write, so the original is recoverable.

[...]

Issue #1745, opened 2026-07-27
Stacked <summary> doc comments: six on dev, invisible to every automated check


erikdarlingdata:
Surfaced while fixing two instances of this in #1739/#1744. Filing the class rather than the two, because the two are already fixed and the other six are not.

## The defect class

A member ends up with **two `<summary>` elements stacked**, usually because an edit added a new summary above an old one instead of replacing it:

```csharp
/// <summary>
/// Base service for collecting performance data from remote SQL Servers.
/// Partial class - individual collectors are in separate files.
/// </summary>
/// <summary>
/// Tracks the health state of an individual collector.
/// </summary>
public class CollectorHealthEntry
```

**XML documentation takes the LAST summary.** So IntelliSense, generated docs, and every analyzer render the correct text. The build is clean, the tests pass, nothing warns. The only reader who is misled is a human reading the file top-down — and the first thing they read is the wrong description, attached to the wrong member.

That is what makes this worth a guard rather than a cleanup: it is invisible to every automated check we currently run, and re-reading does not find it. In #1739 the same region was read at least three times by two people over several hours without either of us noticing the duplicate; a regex found it in one pass.

## Verified inventory on dev (6 sites)

Confirmed by walking every `.cs` (excluding `obj`/`bin`) on dev at `f54485ee`:

```
Darling\PerformanceMonitor.Darling.Storage\TimescaleSupport.cs:370
Darling\PerformanceMonitor.Darling.Viewer\MainWindow.xaml.cs:769
Darling\PerformanceMonitor.Darling.Viewer\ViewerDataService.FinOps.Inventory.cs:131
Lite\Services\LocalDataService.QueryStore.cs:419
Lite\Services\RemoteCollectorService.cs:31
deprecated\Dashboard.Tests\ThemeParityTests.cs:195
```

Four projects, all pre-existing, none introduced by #1706.

## Two of them are actively misleading, not benign duplicates

- **`RemoteCollectorService.cs`** — the file's old service-level header ("Base service for collecting performance data from remote SQL Servers") was never removed when a DTO was added at the top of the file, so `CollectorHealthEntry`, a small health-tracking class, now carries the description of the whole service.
[...]

PR #2453, opened 2026-08-21
Make one badly-shaped setting cost one setting (#2444)

### LINKED ISSUE #2444: One badly-shaped value still costs every Lite setting after it, and the message can't say which key
Named by #2428 and again by #2441, unfixed in both because it is its own change rather than a residue of either.

`App.LoadAlertSettings` wraps `JsonDocument.Parse` and all 88 `TryGetProperty` reads in one `try`. #2428 closed the parse half — a malformed document is now reported and quarantined rather than silently resetting everything. What remains is the *value* half: a single key of the wrong shape (a string where an int is expected, a number where a bool is) throws on its own `Get*` call and abandons every read after it.

So the settings that survive depend on where the bad key sits in the file. One wrong value near the top costs almost everything; the same value near the bottom costs almost nothing. Nothing in that behaviour is visible to the user, and the ordering it depends on is an implementation detail of the loader.

This is milder than the parse case, which is why it has been deferred twice and reasonably so: #2428 made it loud, so a user in this state now gets a startup message telling them the file could not be read fully. What they do not get is *which key*, and without that the message sends them to proofread an 88-key document.

## Why it is not just "wrap each read"

88 try blocks would work and would be awful to read and maintain. The shapes worth considering first:

**A read helper that carries the key name.** `TryRead(root, "alert_cpu_threshold", ref CpuThreshold)` swallowing and recording per key gets the naming for free and collapses to one line per setting rather than five. It also creates a place to put the clamp, which today is repeated inline at every call site.

**Report the whole set, not the first.** A user with one hand-edited file probably has one mistake, but a user who pasted a block has several. Failing on the first and stopping means they fix, restart, and discover the next one — several times.

Note the constraint from #2418: `SettingsSampleTests` extracts `TryGetProperty` key literals by regex out of `App.xaml.cs` and requires each to be documented in `settings.sample.json`. A helper that hides the literal behind a method call will make those keys vanish from the extracted set and turn the guard red — #2428 hit exactly this and had to read back through `JsonDocument`. Whatever shape this takes has to keep the key literals greppable, or the extractor has to learn the new shape deliberately rather than by accident.

## The viewer has the same gap, one level up

Split from #2434: the viewer's startup dialog names the file and the parse position but cannot say which settings were lost, because it round-trips a whole object and the deserialize failed before any property existed. Same question, different mechanism — and the same answer would need per-property reads there too.

erikdarlingdata:
Closes #2444.

## The defect

`App.LoadAlertSettings` wrapped all eighty-seven reads in one `try`. #2425 took the parse out of it; the value half was untouched. A key holding a string where an int belongs threw on its own `Get*` call and abandoned every read after it, so which settings survived depended on where the bad key sat in the file — and that ordering is an implementation detail of the method's line order.

Measured against dev's own source rather than described: `alert_cpu_threshold` is read with `v.GetInt32()` **12,389 characters before** `analysis_timeout_seconds`, inside a single **13,902-character** `try` with no inner `try`. `GetInt32()` on a JSON string throws. So `{"alert_cpu_threshold": "ninety"}` costs `analysis_timeout_seconds` and every other key below it.

## The two judgement calls the issue named

**A read helper that carries the key name — yes.** Eighty-seven `try` blocks would work and would be unreadable. `SettingsReader` checks `ValueKind` before it calls a getter and *records* the key it could not read. That is the call `McpSettings.Load` already makes, for the reason it gives on the line: a `catch` around `GetBoolean` is what turned a quoted `"true"` into a silently disabled endpoint, and a hand-edited file is exactly where a quoted boolean comes from.

**Report the whole set — yes, and it is the same fix.** Someone who edited one line has one mistake; someone who pasted a block out of `settings.sample.json` has several, and stopping at the first means they fix, restart, and discover the next one, several times over. The old code could not have done this even in principle — it threw on the first bad value and never reached the rest — so "which key" and "all of them" are not two decisions.

## The extractor trap

**The key literals stay in the shape the extractor already expects, and the extractor is unchanged.** `SettingsReader`'s method is called `TryGetProperty`, so a call site still reads `read.TryGetProperty("alert_cpu_threshold", out v)` and `SettingsSampleTests`'s regex matches it exactly as before. The name is load-bearing, not a stylistic echo, and it is documented as such at the definition, at the call site, and in the guard.

Keeping the shape was the lower-risk half of the choice the issue offered — but it creates a new dependency held together by nothing the compiler checks, so the guard gets the self-test the issue asked for anyway:

- **Both call shapes are seen** — the pre-#2444 `root.TryGetProperty(…)` and the new `read.TryGetProperty(…)` — over a source the test writes.
- **It still catches an undocumented key.** The synthetic sample is missing one key on purpose and the *real* `UndocumentedKeys` comparison names it. A guard that extracts keys but no longer compares them is the same vacuous green as one that extracts nothing, and the only thing standing between that and a green build was a floor of `>= 50`.
- **The trap itself is measured.** The shape #2428 tried — a read helper taking the key as an ordinary argument — is run through the same extraction and yields **nothing**. That is what makes the name load-bearing rather than a claim that it is.
- **The name is asserted directly**, so a rename fails there, with the reason, instead of as an unexplained collapse of the key count elsewhere.
- **The scoping contract** gets its own case: each read belongs to the most recent `*.json` literal above it, which is how `servers.json` and `collection_schedule.json`'s loaders stay out of this set without an exemption — and how a loader moved above the `settings.json` literal would silently leave it.

To do any of that, the extraction and the comparison are split out to run over text. Both are the shipped ones the real tests call, not copies.

The literal is also kept out of the new comments. `LoadDefaultTimeRange` already warns that the extractor cannot tell a comment from code and would read an example as a real key; the first draft of this change put `TryGetProperty("` in two prose lines and would have injected two junk keys into the extracted set.

## Two things that fall out

**A latent overflow, now unreachable.** Two of the inline clamp forms were `(int)Math.Max(0, v.GetInt64())` — floor, then *narrow*. A hand-typed value beyond int range wrapped, so a bigger threshold became a negative one. `Int(fallback, min, max)` clamps before it narrows.

**A second throw site the single `try` hid.** `alert_excluded_databases` called `elem.GetString()` on every element; a number in that array threw, and cost every setting after it. It now filters by element kind, which the fleet-group list beside it already did.

## The message becomes visible, and specific

The value fault logged and stopped there, so a user whose thresholds had silently reverted had no signal unless they went looking. #2425 had already decided, for the document half, that someone looking at an app which has forgotten its configuration should not have to find a log to learn why; the same argument applies to a setting that reverted. What is new is that the message can name the keys *and* say everything else loaded — which is what makes it worth a dialog rather than noise, and is the difference from the message that used to send someone to proofread an eighty-seven-key file.

It reuses the one startup modal #2425 added rather than opening a second. The two cannot both happen on one run: a document that will not parse never reaches a value read.

The `catch` stays with no known cause left in it, so an unforeseen throw cannot take startup down — and its comment now says what it still costs when it fires, because the reads are ordered and anything after a throw is still lost.

## Where the reader lives, and the load/save mental model

`PerformanceMonitor.Common`, beside `SettingsFileGuard`, not in Lite. #2441 rewrote `SaveAlertSettings` so all ten writers mutate one document opened once through `App.SettingsRootForWrite()` and written once; the load side now has the matching shape — one document opened once through `SettingsFileGuard.Read`, one reader over it, one report at the end. Both sides treat a per-key failure as a value to carry rather than an exception to throw, and neither can now lose the rest of the file to one key.

It is in Common rather than Lite because that is the seam the Viewer's dialog needs for the gap #2444 names one level up: it round-trips a whole object, so its deserialize fails before any property exists and it cannot say which settings were lost either. **Deferred, not folded in** — the Viewer's loader is a different mechanism (whole-object deserialize, not per-property reads) and converting it is its own change with its own tests.

## Verification

`Lite.Tests` targets `net10.0-windows` and cannot run on macOS, so everything runnable was compiled into a `net10.0` xUnit-shim harness against the real `PerformanceMonitor.Common` build, with the real `App.xaml.cs` and `settings.sample.json` as fixtures.

- **13 passed / 0 failed** on this branch.
- That includes **all four pre-existing `SettingsSampleTests`**, which is the result that matters most here — they are the guard this refactor was most likely to break, and they are green with two-way symmetry over all **89** documented keys and both exemption sets still empty.
- Against dev the same file runs **6 passed / 0 failed** with the one test that needs the new type removed. The extraction self-tests are guards, not regression tests, and pass on both sides by design — that is what they are for.
- Everything in `SettingsValueReadTests` that exercises `SettingsReader` **cannot compile against dev**, which is the stronger statement.

The two loader-level cases need WPF and can only run on CI, so their premise was measured rather than assumed — the character offsets and the single `try` above, plus `GetInt32()`'s throw, both taken from the shipped dev source and the shipped BCL in the harness. Those are exactly the two facts those cases assert the absence of. Stating the scope precisely: **the reader and the guard ran here; the two end-to-end loader cases did not, and CI is the arbiter for them.**

No CHANGELOG entry, matching #2425, #2428, #2432 and #2442: `[3.5.0]` is a shipped, dated section and there is no `[Unreleased]` on dev.

### DIFF
--- a/Lite.Tests/SettingsSampleTests.cs
+++ b/Lite.Tests/SettingsSampleTests.cs
@@ -12,6 +12,7 @@ using System.IO;
 using System.Linq;
 using System.Text.Json;
 using System.Text.RegularExpressions;
+using PerformanceMonitor.Common;
 using Xunit;
 
 namespace PerformanceMonitorLite.Tests;
@@ -71,10 +72,7 @@ public sealed class SettingsSampleTests
         var read = ReadKeysFromLoaders();
         var documented = SampleKeys();
 
-        var missing = read.Keys
-            .Where(k => !documented.Contains(k) && !LoaderOnlyKeys.Contains(k))
-            .OrderBy(k => k, StringComparer.Ordinal)
-            .ToList();
+        var missing = UndocumentedKeys(read, documented);
 
         Assert.True(
             missing.Count == 0,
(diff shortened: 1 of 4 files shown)


erikdarlingdata:
Addressed the inline finding on `SettingsValueReader.cs:151` — it was right, and it was worse than an odd message: the clamped readers promised "an out-of-range number clamps to the bound" and that promise silently stopped at Int64, so `99999999999999999999` fell into the wrong-shape arm and the **startup dialog** told a user their whole number "holds a JSON number where a whole number belongs".

Fixed by clamping by SIGN rather than by value. A number beyond Int64 in either direction cannot be inside any range a caller can express, so the only question left is which end — read off the raw token rather than through a `double`, because `long.MaxValue` does not survive that round trip and a bound near it would come back wrong in the one case this exists for. The promise is now absolute rather than "up to Int64".

The unclamped `Int(fallback)` is the other side of that boundary and still reports, because with no range there is nothing to put the number into — but it now reports it as *out of range* and names the value, truncated at forty characters since a hand-edited file is where a thousand-digit number comes from and this string reaches a dialog.

Two tests pin both sides (`ANumberBeyondInt64_StillClampsToABound`, `AnUnclampedReadReportsAnOutOfRangeNumber_AsOutOfRangeRatherThanWrongShape`). The earlier nit about the shared log line claiming "every other setting in the file loaded normally" before `LoadAlertSettings` had run is also fixed — it now claims only what its caller can see, and the whole-file statement stays where it was already correct, in the dialog shown after both loaders.

Both branches have been rebased onto `dev` (60c696fc) and re-verified; the harness runs 15 passed / 0 failed.

erikdarlingdata:
Fixed, and thank you — this was a defect I introduced one commit earlier while fixing the *previous* finding in the same place, which is the worse kind.

You were exactly right about the mechanism: `TryGetInt64` returns false for a non-integer token as well as for an overflow, so choosing the bound by sign was right for the vast case and silently turned `analysis_timeout_seconds: 30.0` into 600 with nothing added to `Problems`. Relocating #2444's failure into the fix for it would have been worse than leaving the original alone.

Reading the remainder as a double tells all three cases apart in one step: vast values fall outside any bound and land on one, fractional values truncate into range. The `long` overload compares rather than clamps, because a bound near `long.MaxValue` does not survive a round trip through a double.

On the suggestion to `Reject` a fractional value outright: I kept it clamping for the readers whose caller *declared* a range, because truncation there is the same species of silent adjustment the clamp already is and `30.0` unambiguously means 30. The unclamped `Int(fallback)` has nowhere to put a number that will not fit, so it does the opposite — takes a value that is exact however it was written, and reports the rest naming *which* of the two problems it has ("is not a whole number" vs "is out of range"), since "holds a JSON number where a whole number belongs" is a nonsense sentence about a number.

Two tests, and both were run against the previous commit to confirm they are red there: `AFractionalNumber_ReadsAsItsValue_RatherThanLandingOnTheMaximum` reports 600 where it should report 30 — your example reproduced exactly — and `AnUnclampedRead_TakesAnExactNumber_AndNamesWhyItRejectsTheRest` reports 7 where it should report 30. Harness now runs 16 passed / 0 failed.

erikdarlingdata:
The orphaned-doc inline on `SettingsValueReader.cs` is already fixed on the branch — `DocCommentHygieneTests` caught it first and it is genuinely the funniest way this could have gone, since I was widening that exact rule on #2450 at the time.

Cause and lesson worth writing down: I anchored the insert on `private T Reject<T>(...)`, the *signature* line, which put the new helper between `Reject`'s doc block and `Reject`. XML docs take the last block, so tooling rendered `RejectAsUnusableNumber` correctly and `Reject` silently lost its documentation to a paragraph about a different method. Anchor C# inserts on the previous member's closing brace, not on the next member's signature.

The helper moved below rather than the stray text being deleted — `Reject`'s block is the record of why the message is phrased the way `McpSettings.Load` phrases it, and #1745 found that seven of eight instances of this shape were a displaced block whose real member had been left undocumented, so a blind "remove the extra summary" would have destroyed it here too.

Both branches are rebased onto `dev` (8f692439) and all six checks are green on each.

PR #2459, opened 2026-08-21
Ask sweep_pressure whether one body fits, not only whether the average minute does (#2446)

### LINKED ISSUE #2446: sweep_pressure reports OK while the body overruns: an infrequent 37s collector is invisible to an amortized model
Found soaking `dev` on `prod-sql-use2-monitor-01`. The worker is logging real skipped relaunches while `sweep_pressure` reports the server healthy, and both are correct about what they measure.

## The disagreement

Six overruns in three hours on one server:

```
12:58  shard-A         collection body has not completed after 66s of execution - skipping relaunch
13:29  shard-A         62s
14:01  shard-A         69s
14:23  single-C  62s
15:02  shard-A         60s
15:34  shard-A         61s
```

`get_collection_health` for that server, at the same moment:

```
busy_ms_per_minute: 12250    busy_percent: 20.4    verdict: OK
```

Every collector reads HEALTHY. Nothing in the tool says this server is collecting at a multiple of its configured interval, which is the exact condition #2296 was built to surface.

## Why the model misses it, and why the model is not wrong

`SweepPressureClassifier` amortizes each collector's average duration by its own cadence. Reconciled against the reported number:

| collector | avg duration | cadence | amortized |
|---|---|---|---|
| procedure_stats | 3,205 ms | 1 min | 3,205 ms/min |
| query_store | 13,834 ms | 5 min | 2,767 ms/min |
| query_stats | 2,674 ms | 1 min | 2,674 ms/min |
| plan_correction | 11,246 ms | 5 min | 2,249 ms/min |
| **index_object_stats** | **37,207 ms** | **1440 min** | **26 ms/min** |

The total lands on the reported 12,250 ms/min, so the arithmetic is doing what it says. But `index_object_stats` takes **37 seconds** in a body that has 60, and contributes 26 ms/min to the verdict. On the cycle where it coincides with `query_store` (13.8 s) and `plan_correction` (11.2 s), the serialized body is over 60 s — which is precisely what the log shows, roughly every half hour.

So the metric answers "can this server's total collection demand fit inside its cadence on average", and the operator's question is "does any single sweep overrun". Those diverge exactly when one collector's single-run cost approaches the interval while its amortized cost is negligible — which is the signature of an infrequent heavy collector, not of a saturated server.

**#2296's model choice was right and should not be reverted.** Its own note explains that gap-based detection flags every server at fleet scale because delivered cadence stretches benignly from bounded sweep concurrency. Sustained demand is the correct primary signal. What is missing is a second dimension, not a different one.

## What would close it

The data is already collected — `avg_duration_ms` per collector is right there in the same payload, and the interval is known. Two candidates:

**A burst term in the verdict.** Any collector whose single-run average exceeds some fraction of the fastest cadence is a body-overrun risk regardless of amortized cost. On this server `index_object_stats` at 37 s against a 60 s minute is 62% of the budget in one collector.

**Say it in `heaviest_collectors`.** That list currently ranks by amortized contribution, so the collector actually causing the overruns is not in it — `index_object_stats` is absent while three collectors that fit comfortably are named. Ranking or annotating by single-run cost would put the cause in front of whoever is looking.

Worth deciding whether the verdict should change at all, or whether this is a note rather than an AT_RISK. An operator told "SATURATED" for a once-daily 37-second collector would reasonably ignore the next one; a note saying "index_object_stats takes 37 s of a 60 s budget when it runs" is actionable and does not cry wolf.

## Also worth a look while in here

`query_store` on this server reports `enumeration yielded 0 items - nothing to collect this cycle` on **958 of 1155 runs**, and still costs 13,834 ms average. That is 83% of its executions doing nothing at a cost that dominates its amortized share. Possibly correct (the collector is on a use2 box whose Query Store went quiet by design after the #2296 split) — but if an enumeration that yields nothing still pays 13.8 s, that is worth understanding separately.

---

_Edited: server and database identifiers replaced with role placeholders. This is a public repository and the originals named tenants and customer databases; the measurements, ratios and reasoning are unchanged, because those are what the issue is about._

erikdarlingdata:
Closes #2446.

## The decision

**Reporting changed, verdict did not** — and the issue's own instinct was right, for a reason worth stating rather than assuming.

`sweep_pressure` now answers two questions instead of one, in two fields:

| field | question | values |
|---|---|---|
| `verdict` | does SUSTAINED demand fit the cadence on average? | `OK` / `AT_RISK` / `SATURATED` |
| `peak_cycle_risk` | does ONE scheduled body fit at all? | `FITS` / `BODY_OVERRUN` |

`peak_cycle_ms` is what the body costs on the cycle where every scheduled cadence comes due together — the collectors' averages added *without* being amortized. That cycle is not a hypothetical worst case: the shipped cadences are strictly nested (1 | 5 | 60 | 1440), so all of them fall due simultaneously on a fixed period. On `prod-sql-use2-multi-49` that body is 73,193 ms against a 60,000 ms budget while sustained demand is 20.4%.

`#2296`'s amortized model is untouched. Its note about gap-based detection flagging every server at fleet scale still holds, and the new number is built from execution times for the same reason, so queueing cannot inflate it either.

### Why not fold it into the verdict

A once-daily 37-second collector is not saturation. `SATURATED` recommends capacity or placement; the lever here is the schedule's *shape* — move or split the collector so it stops sharing a cycle. Spending the word on this case would train operators to discount the next one, which costs exactly the signal #2296 was built to give them. Separate field, separate vocabulary (`BODY_OVERRUN` is the watchdog's own wording, so the field you filter on and the log line you grep for say the same thing), and it can never be read as a fourth verdict band.

A *field* rather than a prose note, though, because a fleet scan has to be able to filter on it. A sentence in a nullable string is a footnote, not a dimension.

### `heaviest_collectors`

Still ranked by amortized contribution — that is what explains `busy_percent`, and re-ranking it would break the one thing it does correctly. What was wrong is that anyone reading it to find the collector that overran a body was reading the wrong column with nothing on the row to say so: on multi-49 it named three collectors that fit comfortably while the one taking 37 of the 60 seconds ranked **last of eight** on that key.

- each row now carries `amortized_ms_per_minute` beside `pct_of_sweep_budget_per_run`, so the two costs sit together
- `peak_collector` names the collector that owns the most of one sweep — on every server, not only when it has become a problem
- `peak_cycle_note` explains it, composed in the classifier rather than at each SKU's tool, because it interpolates numbers and two hand-copied format strings would drift

## Measured, not assumed

Three servers off the dogfood fleet, all three reading `OK` today:

| server | `busy_percent` | `verdict` | aligned cycle | `peak_cycle_risk` | peak collector | logged skips? |
|---|---|---|---|---|---|---|
| `prod-sql-use2-multi-49` | 20.3 | OK | 73,193 ms (122%) | **BODY_OVERRUN** | `index_object_stats` 37,207 ms | yes, 6 in 3h |
| `prod-sql-use2-psi-01` | 16.8 | OK | 65,144 ms (109%) | **BODY_OVERRUN** | `index_object_stats` 35,700 ms | yes |
| `prod-sql-use2-alpha-01` | 7.0 | OK | 11,207 ms (19%) | FITS | `index_object_stats` 4,097 ms | no |

The two that overran flag; the quiet one does not. Note that `index_object_stats` is *not* uniformly heavy — 4,097 ms on alpha against 37,207 ms on multi-49 — so the signal discriminates on the thing that actually varies.

## The pins

Both suites carry the identical decision table, as before. The two real profiles are fixtures that reconcile to the `busy_ms_per_minute` the tool actually reported for each server (12,248 against a reported 12,250; 4,209 against 4,210), and one test swaps **only** `index_object_stats` between them — 37,207 ms against 4,097 ms — so the risk flips while the verdict stays 20.4% on both sides. That is the property the change exists for, stated as an assertion: neither dimension is derivable from the other.

Also pinned: the peak-cycle edge is inclusive at exactly the budget and independent of the verdict (60,000 ms once a day is 42 ms/min, so `OK` on both sides while the risk flips); on-load collectors are excluded from the peak cycle for the same reason they are excluded from the amortized sum, and cannot become the peak collector; an empty window yields a null peak collector rather than an invented one; and the note is empty on `FITS`, because a note that fires on the healthy case is how a signal teaches people to skip it.

### Verification

`Lite.Tests` / `Darling.Tests` are `net10.0-windows` and cannot run on macOS, so CI is the arbiter for the suites as a whole. The classifier itself is `net10.0`, so the **shipped test file was compiled verbatim** into a throwaway `net10.0` harness against the real `PerformanceMonitor.Common` build — 11/11 pass. The same fixtures run against `dev`'s classifier return `OK` for both multi-49 and alpha with nothing else in the returned value differing in kind, and `dev`'s `heaviest_collectors` ranking reproduces the live tool's exact output (`procedure_stats, query_store, query_stats`) with `index_object_stats` absent. Both new properties fail on `dev`.

All four touched projects build clean.

## One correction to the issue, and one deferral

**`index_object_stats` is not the cause of the six warnings.** At a 1440-minute cadence it ran 6 times in 7 days, so it explains one guaranteed overrun *per day*, not six in three hours. The half-hourly ones are the 5-minute tier's tail — `query_store` and `plan_correction` average 13.8s and 11.2s and share every fifth cycle — which is variance around an average this model reports honestly at 60% of budget. The new number is still the one that would have put `index_object_stats` in front of whoever was looking, and the aligned cycle it names is real; the mechanism in the issue just does not fit its own timestamps.

**Deferred, noted here rather than filed:** on-load collectors (`FrequencyMinutes == 0`) are excluded from both dimensions, consistently and deliberately — they run on connect, not in a recurring body. But `database_scoped_config` averages **11,913 ms** on multi-49 and 8,290 ms on psi-01, which lands on the connect body and is invisible to both numbers by construction. Worth a look if connect-time overruns ever show up in the log; it is a different body and would need a different question.

**The `query_store` finding from the issue is real, and it is filed as #2460 — but it lands on this change too, so stating it here rather than only there.** The premise inverts: an empty enumeration costs **36 ms**, not 13.8 s (`prod-sql-use2-alpha-01` yields nothing on all 1,551 of its runs and pays 36 ms). Backing that out, multi-49's 197 *productive* `query_store` runs work out to ~80,900 ms each and psi-01's to ~65,500 ms — each one, on its own, larger than the entire 60,000 ms budget.

That is a better candidate for the skipped relaunches than either the issue's mechanism or this one, and it exposes a limit of what this PR adds: `peak_cycle_ms` is built from `avg_duration_ms`, and on a bimodal collector that mean describes neither population. So the new number understates multi-49's worst cycle by roughly 67,000 ms. It is a better question asked of the same blended input. #2460 proposes serving `max_duration_ms` beside the mean, after which the peak cycle should probably be built from the max — "the worst scheduled cycle" computed from averages is a contradiction this change currently lives with, and I would rather say so than have it found later.


erikdarlingdata:
Good catch on the 73,193 / 73,408 mismatch — fixed in 3ac1b752.

Neither number was wrong, which is what made it worth fixing rather than dismissing. **73,408 ms** is what #2446's own verified per-collector table computes and what the fixture pins twice (`PeakCycleMs` and the formatted note). **73,193 ms** is a live `get_collection_health` read of all 41 of that server's collectors taken minutes later — the two differ because the issue's table carries slightly earlier averages (`query_store` 13,834 against a later 13,716, `plan_correction` 11,246 against 11,164). Same measurement twice, ~215 ms apart.

But you are right that a reader verifying the worked example against the tests hits a mismatch with nothing telling them why, and in this file the comments are load-bearing measurements. The doc comment now leads with the reproducible figure and discloses the live read explicitly, so the delta reads as one measurement taken twice rather than two claims; the CHANGELOG takes the fixture's number for the same reason — it is the one anyone can check.

Left `busy_ms_per_minute` reading 12,250 deliberately: that is what `get_collection_health` actually reported and what the issue quotes, and the fixture's 12,248 *is* that number. Writing 12,248 would make the changelog disagree with the tool it describes.

erikdarlingdata:
Took the design note as a change after all, in 0c0e0d4-ish (latest commit) — pushing back gently on "not worth a code change."

You are right that the math holds regardless, and that is exactly why the comment was wrong: I justified `peak_cycle_ms` by the shipped cadences being strictly nested, and `CollectorScheduleEditorWindow` lets a user free-type any integer. So the comment stakes the metric's validity on a property a supported user action breaks, and a reader with a hand-edited schedule would reasonably conclude `peak_cycle_ms` means nothing for their install — when it means exactly what it always did.

The comment now states the general guarantee (any set of positive integer cadences coincides on their LCM, so the aligned body is a periodic certainty for *any* schedule) and demotes `1 | 5 | 60 | 1440` to what it actually is: the reason alignment is *frequent* here, every 1440 minutes, rather than rare. Comment only, no behaviour change.

For the record on the earlier round: all six checks were green on 3ac1b752 before this, and the new pins are confirmed to have executed by count — `Lite.Tests` 2568 → 2573 and `Darling.Tests` 5421 → 5426 against dev, which is exactly the five new facts per suite, 0 failures.

Issue #2455, opened 2026-08-21
Lite's finding INSERT batch takes a read lock to write


erikdarlingdata:
Noticed while threading the analysis token through Lite's store layer for #2443. Not touched there — a token sweep is the wrong PR to change a locking decision in — but it should not go unrecorded.

`Lite/Analysis/FindingStore.cs`, `InsertFindingsAsync`:

```csharp
/* One read lock + one connection for the whole batch, reused for every insert. */
using var readLock = _duckDb.AcquireReadLock();
using var connection = _duckDb.CreateConnection();
await connection.OpenAsync();

foreach (var finding in findings)
    await InsertFindingAsync(connection, finding);
```

`MuteStoryAsync` and `CleanupOldFindingsAsync` do the same — `AcquireReadLock()` around an `INSERT` and a `DELETE` respectively. The comment says "read lock" and means it; the work is a write.

## Whether it is actually wrong

That depends on what the lock is FOR, and the class doc answers it: "Coordinates UI readers with maintenance writers (CHECKPOINT, archive DELETEs, compaction)." The write lock exists so archival/compaction can exclude everyone; the read lock is what everyone else holds to keep archival out while they work. Under that reading, a finding INSERT holding a read lock is **correct on the axis that matters** — it keeps compaction from running underneath it — and DuckDB does its own internal concurrency control for the rest.

So this is probably deliberate and probably fine. What it is not is *stated*. Three things follow from that:

1. **It reads as a bug on inspection**, which costs a reader the same five minutes every time. A comment naming the axis ("read lock because the point is to exclude compaction, not other writers") would end that permanently.
2. **`AcquireReadLock` is not exclusive**, so two of these can interleave. For `analysis_findings` inserts that is harmless — distinct `finding_id`s, no read-modify-write. It would not be harmless for anything that grows into a read-modify-write here later, and nothing marks the boundary.
3. **`_nextId++` in `MuteStoryAsync` is not atomic**, and is reached under a *shared* lock. Two concurrent mutes can collide on `mute_id`. Today only the viewer and the MCP call it, so concurrency is unlikely rather than impossible.

## Suggestion

[...]

Issue #2443, opened 2026-08-21
The analysis token is armed but 167 Darling and ~138 Lite store calls never receive it


erikdarlingdata:
Both SKUs now arm a cancellation token for the analysis pass (#2419 for Lite, #2438 for Darling). Neither store layer receives it. So abandonment happens *between* stages, not inside a read, and a single wedged query still holds its pass for as long as the query runs.

## The counts, and the surprise

| | untokened async store calls |
|---|---|
| **Darling** analysis pipeline | **167 of 209** |
| **Lite** analysis pipeline | ~138 |

Darling's breakdown: `PgFactCollector.*` (93 — the entire fact-collection phase), `PgDrillDownCollector.*` (53), `PgFindingStore` (18), `PgBlockingPairRowQuery` (2), `PgPlanFetcher` (1). Spot-checked independently: every matched call in the first three is the no-argument form.

The surprise is the direction. I assumed in #2430 that Darling would be the cheaper of the two, because Npgsql honours cancellation natively while DuckDB.NET has no async execution at all. That reasoning was about the *driver* and the work is in the *call sites* — Npgsql would honour a token; nothing hands it one. Darling has more of them than Lite, not fewer.

## Why it is worth doing, and why it is not urgent

The reporting added by #2419 and #2438 covers the user-visible symptom: a wedged server is now loudly stuck (minutes stuck, cycles lost, on a doubling backoff) rather than silently skipped forever. That was the actual defect in #2412 and #2430.

What remains is the difference between "we stopped waiting" and "we stopped working". A pass abandoned mid-read keeps a connection and a query alive on the monitored server until it finishes on its own. On a healthy fleet that is invisible; on the pathological case these fixes exist for — a query wedged behind a lock on a monitored instance — the monitoring tool is holding a session it has already given up on.

## Doing it well

This is mechanical but not trivial, and mechanical-but-large is where a sweep goes wrong. Two things worth deciding up front:

**Thread it, do not default it.** Adding `CancellationToken cancellationToken = default` to the signatures and calling it done leaves every existing call site passing `None` while *looking* threaded — the same shape as the exemption lists this repo has spent a day removing. A test that asserts no `Execute*Async()` no-arg call remains in these files would make the sweep verifiable and keep it from regrowing.

[...]

PR #2461, opened 2026-08-21
Make a finding batch commit whole or not at all (#2448)

### LINKED ISSUE #2448: A finding batch that fails mid-insert leaves a truncated set that reads as a complete one
Found while deciding what a *cancelled* persist should mean for #2443. The cancellation path is now closed off before the batch starts, so this is not reachable from a timeout or a stop — but it is still reachable from an ordinary store fault, and the reason it is worth filing is that the damaged state is indistinguishable from a healthy one.

## The mechanism

`PgFindingStore.InsertFindingsAsync` (and its Lite twin `FindingStore.InsertFindingsAsync`, same shape) writes a batch as N separate INSERTs on one connection:

- there is no transaction — autocommit, so row 1..k are durable the moment they run;
- `InsertFindingAsync` catches per row, logs, and the batch **continues** (this is deliberate: "one bad row logs and the batch continues", the Dashboard twin's posture);
- every row in the batch shares one `analysis_time`, stamped once in `FilterMutedFindingsAsync`;
- `GetLatestFindingsSql` selects `analysis_time = (SELECT MAX(analysis_time) ...)`.

So if the store becomes unavailable at row 5 of 40, rows 1–4 are committed under the newest `analysis_time`, rows 5–40 each log an ERROR, and the next reader — viewer, MCP `get_analysis_findings`, the alert routing — sees **four findings and no indication that thirty-six are missing**. It does not look truncated. It looks like an analysis that found four problems, which is to say the server looks *healthier* than the one that just failed to persist its findings.

The 36 ERROR lines are the only evidence, and they are the #2299 shape: one line per row, from a per-row catch, for a single event.

## Why it is not urgent

It needs the store to fail partway through the tail of a pass, which on a local managed Postgres on loopback is rare. The next scheduled pass writes a fresh complete set under a new `analysis_time`, so the damage self-heals within one cycle for the latest-findings read. `GetRecentFindingsAsync` (24h window) keeps the truncated set visible for its retention, mixed in with complete ones.

## Options, and the measurement that should decide

1. **One transaction per batch.** All-or-nothing, which matches what the reader assumes. Costs the per-row isolation: PostgreSQL aborts the whole transaction on any error, so one bad row would take the batch with it unless each row gets a `SAVEPOINT`. Savepoints per row are cheap here (batch sizes are tens) but it is a real behaviour change to a posture that was chosen on purpose.
2. **Stamp completeness.** Persist the intended row count alongside the batch (or a terminal marker row) so a reader can tell a truncated set from a small one. Cheapest, and it fixes the actual defect — the silence — rather than the partiality.
3. **Collapse the per-row ERROR burst** to one line naming the count, whichever of the above lands. Independent of the above and worth doing regardless.

The thing to measure before choosing: how often a batch actually fails partway. `collection_log` will not show it — this is the analysis path — so it wants a counter or a single summary line on the batch, which is option 3 either way. Start there; it turns the question into data.

## Related

- #2443 (the cancellation half — deliberately does not touch this)
- #2299 (the per-metric ERROR-burst shape)

erikdarlingdata:
Closes #2448. Both SKUs, one commit each.

## The option, and why

**Option 1 — one transaction per batch.** Both stores. Option 3 (collapse the ERROR burst) falls out of it rather than being bolted on. Option 2 was rejected and option "do nothing" was not defensible; both reasons are below.

### The measurement the issue asked for

The issue said the thing to measure is how often a batch fails partway, and that `collection_log` cannot show it. That is true, and it was the wrong measurement to wait for — it would have taken a fault landing in a specific tail on the dogfood fleet, and until then the answer would have been "not yet". So the numbers that actually decide are the ones about **cost** and about **whether the fix is buildable identically on both engines**, and those can be taken today.

**Batch size — ~10 rows.** Read off the dogfood store: `prod-sql-use2-multi-37` at `analysis_time 2026-08-21T16:39:03` persisted the contiguous `finding_id` block `639229059608319080`–`089`. Ten rows, on a busy production server with nine co-fired findings. So the transaction spans ten small INSERTs on loopback (Darling) or into an embedded file in-process (Lite). There is nothing to weigh against here.

**DuckDB.NET 1.5.5 and Npgsql 10.0.3, measured on real engines** (DuckDB embedded; PostgreSQL 17.10):

| | DuckDB 1.5.5 | Npgsql / PG 17.10 |
|---|---|---|
| transaction + commit | works | works |
| a failed row inside the transaction | aborts it — every later statement refused (`TransactionContext Error: Current transaction is aborted`) | aborts it — `25P02 current transaction is aborted, commands ignored` |
| `SAVEPOINT` | **not parsed** — `Parser Error: syntax error at or near "SAVEPOINT"` | supported |
| `Commit()` on an already-aborted transaction | **returns normally, commits nothing** | **returns normally, commits nothing** |
| rollback / dispose after the abort | discards the whole batch | discards the whole batch |
| two concurrent append transactions, same table, two connections | both commit | n/a |

Two of those changed the design rather than confirming it, so they are worth reading twice.

**DuckDB has no `SAVEPOINT`.** That is what settles the per-row question. Per-row failure isolation cannot exist inside a transaction on DuckDB at all, so preserving it on Darling with savepoints would have been a fix in one SKU and not the other — exactly the drift this repo has spent a day removing. It also should not survive on its own merits: a batch that silently drops row 5 and commits the other 39 is #2448 at row granularity, a set claiming 39 problems when the analysis found 40.

**`Commit` on an aborted transaction returns normally on BOTH drivers, having written nothing.** So "reaching the commit" is never evidence of a write. A per-row catch inside a transaction would have logged N−k rows' worth of ERRORs, then reached a commit that looked like success and persisted zero rows — strictly worse than what is there now. This is the second, independent reason the row write must not swallow, and it is stated at the code.

**Concurrency was checked before widening the window**, because Lite's batch runs under a *read* lock, which admits concurrent holders (#2455), so two servers' passes can genuinely overlap on this table. Two open append transactions on one DuckDB table from two connections both commit. Nothing to pay.

### Why not option 2 (stamp completeness)

It fixes the silence, but every reader has to learn the new rule — the viewer's Recommendations tab, `get_analysis_findings`, the alert routing, `GetLatestFindings`, `GetRecentFindings` — plus a migration rung on the Darling ladder and a DuckDB schema bump on Lite. Any reader that did not learn it keeps the behaviour being fixed, and a marker written *last* is itself a write that can fail, which flags complete sets as partial. The transaction requires no reader to learn anything, and a rolled-back batch leaves the PREVIOUS pass as the newest complete set — stale, stamped with its own older `analysis_time`, and incapable of misleading anyone. That is a strictly better answer than a fresh set annotated as untrustworthy.

### Why not "do nothing"

The window is not negligible, and the reason is the part I would not have guessed before running it. Darling's per-row catch means the batch **continues** past the failure, so the damaged set is not "the rows before the fault" — it is *every row except the bad ones*. The reverted test persists **four of five rows** and reports failure. A store fault that kills the connection does produce the truncated tail the issue describes, but a data-shape fault produces a near-complete set that is even harder to disbelieve. Neither is rare enough to be worth arguing about once the fix costs ten rows of transaction.

## What changed

Both stores: `InsertFindingsAsync` opens one transaction, enlists every row in it, and commits. The per-row `try`/`catch` in Darling's `InsertFindingAsync` is gone (Lite never had one), so one failure ends the batch and costs **one** log line naming the row and the count — which is option 3, delivered by the same change.

`InsertFindingsAsync` still never throws on Darling (the class's error discipline is unchanged) and still throws on Lite (its shape is unchanged); each SKU's existing posture is preserved exactly. The #2443 abandonment decision is untouched: the connection open — and on Lite the read-lock wait — is still the last cancellation point, and there is still no between-rows token check.

**The issue described the two stores as "same shape" and they were not**, which is worth recording: Darling caught per row and continued, Lite has never caught at all, so a fault at row 5 already ended Lite's batch there. Lite had the quieter half of the same defect — rows 1–4 stayed durable under the new `analysis_time` while the pass reported failure — and gains the fix with no change to error handling at all.

## The tests

Both behavioural, both against a real engine, both verified red on `dev`:

- **Lite** (`FindingStoreTests`) — real DuckDB. Fault induced with a duplicate `finding_id`, which is the reachable per-row fault on that table and not a contrivance: the ids come from `_nextId++` seeded off `DateTime.UtcNow.Ticks`, so two stores in one process can genuinely produce one (#2455). Reverted: **3 rows where 1 was expected**.
- **Darling** (`DarlingAnalysisStoreTests`, `[Collection("live-postgres")]`) — fault induced with a NUL byte in `story_text`, which PostgreSQL rejects as `22021`. That is the realistic per-row fault on a table carrying no primary key, no CHECK and no foreign key whose NOT NULL columns all map to non-nullable properties with defaults. Reverted: **4 of 5 rows persisted**.

Both then commit the same batch *clean* through the same path, so neither assertion can be satisfied by a store that has simply stopped writing.

The #2443 pins in both `AnalysisPassTokenThreadingTests` gain the transaction — begin, commit, and the rows' enlistment are pinned separately, because either half alone is silently useless — plus, on Darling, `Assert.DoesNotContain("catch", rowWrite)`. Their prose said "the insert batch has no transaction"; that claim is now false and is rewritten rather than left to mislead.

## Verification

- Full solution builds clean on macOS (`EnableWindowsTargeting`): **0 errors**, 24 warnings, all pre-existing in untouched files. Each commit builds on its own.
- `Lite.Tests` / `Darling.Tests` are `net10.0-windows` and cannot run on macOS, so **both new tests were compiled into a `net10.0` xUnit shim against the real shipped `FindingStore` / `PgFindingStore`** (Lite's `Lite/Database/**` + `Lite/Analysis/FindingStore.cs` compiled in; one stub for the single legacy symbol `DuckDbInitializer` reaches outside `Lite/Database`, which the finding path never touches) and run on macOS against embedded DuckDB and a local PostgreSQL 17.10. **2/2 pass on this branch, 2/2 fail on `dev`**, with the failure messages quoted above. CI is the arbiter for the rest of the suites.
- Every source pin the Windows-only suites make about these two files was replicated by hand and re-checked after the rebase — including `AnalysisShutdownResidueTests`' `Assert.Equal(2, Count(findingStore, contextFilter))`, which is unchanged: the catch removed was a bare one, not a filtered one.
- Rebased onto `dev` at `8f692439` and everything above re-run after.

## Deferred

**The #2443 cancellation decision is now re-openable, and deliberately not re-opened here.** #2449 withheld the token from the batch because cancelling mid-batch would have *made* the partial set. With the transaction it would roll one back instead, so mid-batch cancellation is now safe and would let a wedged pass drop its store work sooner. That is a different question from this one, the pinned rationale for the absent check is now half-expired, and re-deciding it inside a fault-handling PR is how a settled decision gets changed without anyone noticing. Say the word and I will file it.

### DIFF
--- a/Darling/Darling.Tests/AnalysisPassTokenThreadingTests.cs
+++ b/Darling/Darling.Tests/AnalysisPassTokenThreadingTests.cs
@@ -256,13 +256,19 @@ public sealed class AnalysisPassTokenThreadingTests
 
     /// <summary>
     /// The decision this issue asked to be made explicitly rather than assumed: what a cancelled PERSIST
-    /// means. The insert batch has no transaction (per-row failure isolation is deliberate — one bad row
-    /// logs and the batch continues), every row shares one <c>analysis_time</c>, and the latest-findings
-    /// read keys on <c>MAX(analysis_time)</c>. So a batch cut in half does not read as truncated; it reads
-    /// as a complete analysis that found fewer problems, and the server looks HEALTHIER for having been
-    /// abandoned. That is a third outcome, distinct from the timeout and the shutdown the classifier names,
-    /// and it must not exist. The connection open is therefore the last abandonment point: before the first
-    /// row, or not at all.
+    /// means. Every row shares one <c>analysis_time</c> and the latest-findings read keys on
+    /// <c>MAX(analysis_time)</c>, so a batch cut in half does not read as truncated; it reads as a
+    /// complete analysis that found fewer problems, and the server looks HEALTHIER for having been
+    /// abandoned. That is a third outcome, distinct from the timeout and the shutdown the classifier
+    /// names, and it must not exist. The connection open is therefore the last abandonment point: before
+    /// the first row, or not at all.
+    ///
+    /// <para>#2448 closed the other half of that. #2443 could only reason about the CANCELLATION path,
+    /// and the same truncated set was still reachable from an ordinary store fault mid-batch, where no
+    /// amount of token discipline reaches. The batch is now one transaction, so it is all-or-nothing
+    /// against a fault as well — which is why the between-rows assertion below is still right, but no
+    /// longer the only thing standing between a store fault and a set that understates a server. Both
+    /// halves are pinned here because they are one property.</para>
     /// </summary>
     [Fact]
     public void TheFindingInsertIsAbandonedBeforeItStartsOrNotAtAll()
@@ -278,6 +284,19 @@ public sealed class AnalysisPassTokenThreadingTests
            set rather than prevent it, which is why there is none. */
         var insertBatch = Between(store, "public async Task<List<AnalysisFinding>> InsertFindingsAsync(", "public async Task<List<AnalysisFinding>> SaveFindingsAsync(");
         Assert.DoesNotContain("ThrowIfCancellationRequested", insertBatch, StringComparison.Ordinal);
+
+        /* #2448: the batch is one transaction, and the rows are enlisted in it. Both halves are pinned
+           because either alone is silently useless — a transaction the rows do not join commits nothing
+           of theirs, and enlisting in a transaction nobody commits writes nothing at all. */
+        Assert.Contains("await connection.BeginTransactionAsync();", insertBatch, StringComparison.Ordinal);
+        Assert.Contains("await transaction.CommitAsync();", insertBatch, StringComparison.Ordinal);
+        Assert.Contains("new NpgsqlCommand(InsertFindingSql, connection, transaction)", store, StringComparison.Ordinal);
+
+        /* And the row write must not swallow. A per-row catch inside a transaction is worse than useless:
+           PostgreSQL refuses every later statement once one fails (25P02), so it buys N-k ERROR lines for
+           one event and a CommitAsync that returns normally having written nothing. */
+        var rowWrite = Between(store, "private async Task InsertFindingAsync(", "private static AnalysisFinding ReadFinding(");
+        Assert.DoesNotContain("catch", rowWrite, StringComparison.Ordinal);
 
         /* The row write states the decision where someone changing it will read it. */
         Assert.Contains(ExemptionMarker, Between(store, "/// Inserts one finding on an already-open connection", "private async Task InsertFindingAsync("), StringComparison.Ordinal);
--- a/Darling/Darling.Tests/DarlingAnalysisStoreTests.cs
+++ b/Darling/Darling.Tests/DarlingAnalysisStoreTests.cs
@@ -587,9 +587,12 @@ VALUES ($1, $2, $3, $4, $5, $6)", connection);
             var doomed = await store.FilterMutedFindingsAsync(PartialBatchStories(faultAtRow: 3), context);
             Assert.Equal(5, doomed.Count);
 
-            /* The store never throws to the pass (the class's error discipline); the loss reaches
-               the caller as one logged line, not as an exception. */
-            await store.InsertFindingsAsync(doomed, context);
+            /* It THROWS rather than degrading, which is the one place this store departs from its
+               own no-throw discipline and is deliberate: swallowing a total rollback returns the
+               same list a full success returns, so DarlingAnalysisService would set LastAnalysisTime,
+               fire AnalysisCompleted and log "Analysis complete - 5 finding(s)" over a store holding
+               none of them. That is this very defect moved one layer out. */
+            await Assert.ThrowsAsync<PostgresException>(() => store.InsertFindingsAsync(doomed, context));
 
             /* The assertion #2448 exists for: not "fewer rows", NO rows. A partial set here would
                be indistinguishable from a healthy server. */
--- a/Darling/Darling.Tests/DarlingAnalysisStoreTests.cs
+++ b/Darling/Darling.Tests/DarlingAnalysisStoreTests.cs
@@ -53,6 +53,11 @@ public sealed class DarlingAnalysisStoreTests
     private const int GlobalReaderServerId = -929292;
     private const int GlobalOtherServerId = -939393;
 
+    /* #2448's fixture, kept off every other test's server so the "nothing was persisted" assertion
+       cannot be satisfied by another test's cleanup or broken by its rows. */
+    private const int PartialBatchServerId = -646464;
+    private const string PartialBatchServerName = "analysis-store-partial-batch";
+
     /// <summary>
     /// Every v_&lt;table&gt; view the V4 migration must create: the fourteen the ported fact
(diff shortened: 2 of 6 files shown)


erikdarlingdata:
Good catch, and it is a real defect in what I added rather than a wording nit.

The line exists *because* #2448 asked how often a batch fails partway and observed that `collection_log` cannot answer it — so it is the instrument, and an instrument that silently reclassifies every commit failure as a bad row is worse than no instrument, since the resulting count would look like evidence.

Fixed in 86eb4008: the loop sets `everyRowAccepted` before `CommitAsync`, and a commit failure gets its own branch and its own sentence ("had all N row(s) accepted and then failed to COMMIT them") rather than borrowing the row number, which is identical in both cases and therefore cannot distinguish them. Both branches still lead with the loss rather than the mechanism.

The Lite twin needs no equivalent — it has no logging on this path, throws to `AnalysisService`, and so never claims a row number it could get wrong.

erikdarlingdata:
Both findings were right, and the first is the more serious of the two because this branch made it worse rather than merely leaving it alone. Fixed in 709a0731 (now 4 commits after a rebase onto `c4860e29`).

**The completion announcement.** `PgFindingStore` swallowed and returned the same list a full success returns, and `DarlingAnalysisService` does not read the return value — so it set `LastAnalysisTime`, fired `AnalysisCompleted` and logged "Analysis complete: N finding(s)" over a store holding none of them. You put your finger on why that is not a pre-existing wart I could leave: before the transaction the line was only a *little* wrong, and afterwards it is entirely wrong. #2448's misreading moved one layer out is the last thing this branch should ship.

It now logs the line only it can write — which row, or that the commit failed — and rethrows. The pass's existing catch classifies it, reports Error and returns nothing, so nothing is announced. That costs the class its "no public method throws" discipline for exactly one method, and the class doc now says which and why: *log and degrade* needs something to degrade **to**, and once the batch became atomic there is nothing between all-persisted and none-persisted to land on.

You also called it correctly as a live SKU divergence rather than a Darling-only issue — and in the direction that Lite was already right. Lite has never caught here, so its pass already reported failed. Darling was the odd one out and now matches.

**The Lite diagnostic** is the mirror of that, and is in the same commit: Lite had the correct control flow and none of the measurement, so an operator saw only `AnalysisService`'s generic "Analysis failed for {server}" with no sign a batch was involved. It now logs the same line in the same two shapes (row fault vs commit fault) and then rethrows — control flow unchanged, only the diagnostic added.

Net: both SKUs log one precise store line plus one pass outcome line for one event, and neither announces an analysis it did not persist.

The live test now asserts the throw as well as the empty store; re-verified 2/2 green on this branch and 2/2 red on `dev` against real DuckDB and PostgreSQL 17.

Issue #2422, opened 2026-08-21
[QUESTION] Overview Warnings


ehaar:
### Which component is your question about?

Darling

### Performance Monitor Version

3.5.0

### Is your question about how it works, or the results?

How the tool works

### What's your question?

Watching the Overview tab I can see several Criticals (in red), Warnings (in yellow) and so on. The first server info block in the screenshot says 'Warning' in yellow.
The question: what is it that this text warns me about...? How can I find what it means and how can I resolve this warning?
The warnings in the 'Needs Attention' block point to other servers than the ones displayed in the screenshot.
'+52 More need attention' - Where do I find these warnings?
Thanks in advance.

### Additional Context

<img width="630" height="580" alt="Image" src="https://github.com/user-attachments/assets/c37de563-b943-4b05-b7c3-f865c7383c96" />

erikdarlingdata:
Good questions, and two of the three have answers I am not happy with — the product knows the thing you are asking for and does not show it to you. Details below, and I have opened #2423 to fix both.

## What the "Warning" on a server card means

The band is derived from that card's **own** metrics — nothing hidden, nothing off-screen. `ServerHealthClassifier.ClassifyBand` takes four inputs: is the server online, is it awaiting its first collection, does it have collector errors, and the worst severity across the card's metrics. Those metrics are:

**CPU · Threads · Memory · Blocking · Deadlocks · Collectors**

So the card says Warning because at least one of those rows on that same card is in its warning range, and Critical if any is critical. Offline and "awaiting first collection" are separate states rather than metric-driven ones.

## How to find which one — today, and why it is annoying

Read down the card and find the metric that is yellow or red. That works, but it is a worse answer than it should be, because the code already computes exactly the sentence you want:

```csharp
if (s.CpuSeverity >= HealthSeverity.Warning)     parts.Add($"CPU {s.CpuDisplay}");
if (s.ThreadsSeverity >= HealthSeverity.Warning) parts.Add($"Threads {s.ThreadsDisplay}");
if (s.MemorySeverity >= HealthSeverity.Warning && s.HasMemoryPressure) parts.Add($"Memory: {s.MemoryDetail}");
```

`BuildReason` produces "CPU 94%, Blocking 12" for any server — and it is used **only** for the Needs Attention list, never for the card. The card's status is a bare text block with no tooltip. So the explanation exists, is already computed for the server you are looking at, and is thrown away. That is the first half of #2423.

[...]

Issue #1562, opened 2026-07-18
Darling Web: a browser viewer served by the service (Mac/cross-platform), riding the existing Kestrel + server-side read layer


erikdarlingdata:
The WPF viewer leaves Mac (and Linux) operators with nothing native. Rather than an Avalonia port of a large WPF surface, serve a **web viewer from the service itself** — it already runs Kestrel for MCP, and the decisive asset is that the ~72 MCP data tools read the store through **service-side readers**: the data layer for a web front end already exists, tested and shaped like the viewer's tabs (fleet summary, wait/CPU/memory/file-IO trends, blocking/deadlocks, query performance, alerts, collection health).

**Phase 1 (read-only):** fleet overview cards + per-server core charts + alert history, behind the same token/CIDR exposure model as MCP (designed once with the `--configure-network` wizard). Loopback-only by default, like everything else.
**Phase 2:** grow toward tab parity where it earns its keep; the co-located WPF viewer remains the full-fidelity surface.
**Phase 3:** the custom-views initiative builds on this surface (separate issue).

Crew-scale design when started: auth model, asset pipeline (self-contained, no CDN), charting approach, and which reads are phase-1.

Field driver: Mac users have no path today; a web surface also serves NOC wall-screens and phones for free.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
[...]

PR #2477, opened 2026-08-21
Web server page: twelve sub-tabs over the existing renderPanel seam

### LINKED ISSUE #2475: The web dashboard serves 82 reads and shows 23: the server page is 7 panels where the viewer has 65 tabs
The desktop viewer's per-server surface is a TabControl with **65 `TabItem`s** across 38 `ViewerServerTab.*.cs` partials. The web dashboard's `#/server/{name}` is **one scroll of seven panels**: Overview, CPU, Wait Stats, Active Queries, Memory Trend, File I/O Latency, Collection Health.

That is not a backend gap. The service already dispatches **82 read handlers** at `GET /api/read/{name}` (`DarlingWebEndpoints.BuildReadDispatch`), and `wwwroot/js/panels.js` is already a generic renderer over them — `renderPanel({read, params, viz, span})` with a four-kind viz registry. Of those 82 reads, **the built-in web pages reach 23**. The other 59 are served, documented in `/api/catalog`, reachable by hand-typing a URL, and surfaced nowhere a UAT user will find them.

The specific shape of the problem:

- A user who wants to know why a server is amber has no memory clerk list, no resource semaphore, no plan-cache breakdown, no scheduler pressure, no latch or spinlock stats, no blocking or deadlock detail, no `system_health` parse, no default trace, no configuration snapshot, no configuration *change* history, and no configuration audit — all of which the store holds and the service already serves.
- The single time window is baked into each panel's descriptor (`hours_back: 4` for CPU, `24` for everything else). The desktop viewer has a toolbar range picker; the web page has nothing, so a reader investigating an incident from Tuesday cannot look at Tuesday.
- Two reads are **unreachable from a browser at all** because they require a parameter no UI collects: `get_perfmon_trend` needs a `counter_name`, `get_wait_trend` needs a `wait_type`. The desktop viewer has a picker for each.

### What is genuinely not portable, and should be said rather than imitated

Some of the 65 are desktop things and a web imitation of them would be worse than an honest absence:

- **Plan Viewer**, **Get Actual Plan (re-run)**, **View Cached Plan** — need a plan renderer and a Stage-2 command back to the monitored server. The web seat is read-only over collected rows and has neither.
- **Current Active Queries** — the entire tab is an on-demand `fetch_active_queries` command; there is nothing collected to render.
- **Query Heatmap**, **Block Chain**, the interactive **deadlock graph** — interactive plots/graph views.
- The cross-cutting desktop affordances: chart context menus, Copy Repro Script, CSV export, right-click drill-down to a ±30-minute window.

### Also missing on the read side (this part IS backend work)

Seven desktop tabs have no `/api/read` handler at all, so no amount of front-end work reaches them:

| Desktop tab | Reader with no MCP read |
|---|---|
| Queries > Query Store Regressions | `GetQueryStoreRegressionsAsync` |
| Queries > Query Heatmap | `GetQueryHeatmapAsync` |
| Queries > Performance Trends (3 of its 4 charts) | `GetProcedureDurationTrendAsync`, `GetQueryStoreDurationTrendAsync`, `GetExecutionCountTrendAsync` |
| Blocking > Blocking Stats | `GetBlockingDurationStatsAsync`, `GetDeadlockSeverityStatsAsync` |
| Blocking > Trends (the lock-wait lane) | `GetLockWaitTrendAsync` |
| Blocking > Current Waits | `GetWaitingTaskTrendAsync`, `GetBlockedSessionTrendAsync` |
| Collection Health > Collection Log + Duration Trends | `GetRecentCollectionLogAsync` |

And one that has a read whose SHAPE the web renderer cannot use: `get_database_scoped_config` returns `databases[].settings[]`, a nested array, which `vizTable` cannot render — it needs either a flattened projection or a nested-table viz.

### And one the fleet payload blocks

`DarlingFleetReader.FleetServerCard` carries `engine_edition` — the SQL Server `SERVERPROPERTY('EngineEdition')` — and **no `CollectorTargetEngine` discriminator**. So the browser cannot tell a PostgreSQL target from a SQL Server one, and the eight `get_pg_*` reads have no honest home on a per-server page: a PostgreSQL panel added today would render on every SQL Server in the fleet, permanently empty. Adding the discriminator to the fleet payload is the prerequisite for any PostgreSQL surface in the web dashboard.

erikdarlingdata:
Closes #2475.

The service already dispatches 82 read handlers at `GET /api/read/{name}`, and `panels.js` has been a generic renderer over them since #1562 — `renderPanel({read, params, viz, span})` with a four-kind viz registry whose own header says the seam is stable and nothing downstream of it is built. The built-in pages reached 23 of the 82 reads. `#/server/{name}` was one scroll of seven panels against the desktop viewer's 65 `TabItem`s.

So none of this is backend work. It is descriptors.

## What lands

`pages/server.js` becomes the shell — header, sub-tab bar, time-range picker, panel grid — and the new `pages/server-tabs.js` is the registry: eleven tabs, ~70 panels, reaching **61 of the 82 reads**.

| Tab | What it carries |
|---|---|
| Overview | server summary, properties, CPU, memory, blocking + deadlock trends, file I/O latency, analysis findings, daily summary |
| Wait Stats | wait table with a trend for the wait you pick, waiting tasks, latches, spinlocks |
| CPU | utilization, scheduler pressure, top queries, top procedures |
| Memory | summary, trend, clerks, grants, resource semaphore, pressure events, plan cache + breakdown |
| Blocking | blocking and deadlock trends, blocking events, deadlocks, deadlock graph XML, blocked-process report XML, object contention |
| File I/O | latency per database, per-file stats, tempdb, database sizes, table/index sizes, PVS |
| Queries | active queries, duration trend, top queries, top procedures, Query Store, long-query completions, plan corrections, automatic tuning |
| Configuration | properties, config audit, sp_configure, database config, Query Store health, trace flags |
| Config Changes | server, database and trace-flag change history |
| Activity | perfmon counter picker + trend, perfmon snapshot, sessions, sessions by application, running jobs, index usage |
| System Events | the eight `system_health` parse-on-read families and the default trace |
| Collection Health | sweep pressure, collectors, heaviest collectors |

The tab id rides in the hash — `#/server/{name}/{tab}` — so a section is deep-linkable and survives the 60s refresh, which re-renders the whole route. An unknown or absent id resolves to Overview rather than erroring, and that fallback is what keeps every pre-existing `#/server/{name}` link working. The router splits the hash *before* decoding, so an encoded `/` inside a server name still round-trips.

A page-level range picker (1h / 4h / 12h / 24h / 7d / 30d) is the web twin of `ViewerServerTab.TimeRange.cs`'s toolbar presets. It is module state like the fleet page's sort, so it survives the refresh, and it is deliberately **not** persisted: a page that reopens on a 30-day window is slow for a reason the reader cannot see. Panels whose read takes no time window at all say "latest snapshot" in their own subtitle rather than inheriting a label that would misdescribe them.

Two reads were previously unreachable from a browser because they require a parameter no UI collected. `get_wait_trend` needs a `wait_type` and `get_perfmon_trend` a `counter_name`; both now have a picker. The wait picker is seeded from the rows of the table directly above it rather than from `get_wait_types`, so the reader is choosing from what they can already see — the full distinct set would offer wait types absent from the table and make the two disagree.

## Two things deliberately not done

**No fifth viz kind.** The shape that tempted one is the property grid — a read returning a flat object with a dozen scalars. `stat` handles it (`get_memory_stats`, `get_server_properties`, `get_cpu_scheduler_pressure`), and `table` handles every read that already returns rows. A fifth kind added only to `panels.js`'s `VIZ` would be a page-only special case, and adding it properly means `KnownVizList` (or the composer cannot offer it), `derive.js`'s `deriveVizConfig`, and an editor config arm — composer surface area this change does not need.

**Nothing seeded as stored views.** Curated views as data would need a migration rung, `StorageVersion` bump, four pinned test files and a viewer probe sentinel, for content that is not schema — and a seeded row resurrects itself on the next upgrade after a user deletes it. Code descriptors are covered by the invariant below; a stored view's `read` name is validated once at write time and never again.

## Honesty, in the two places it matters

Every table panel supplies its own empty-state sentence, and `table()` **throws** without one. `renderPanel` already shows a read's `{status,message}` envelope when the read has nothing — better wording than any descriptor could carry — but that does not cover the read returning data whose row array is empty, where `vizTable` falls back to a generic "No rows in this window". On a collector that is off, opt-in, or daily, that reads as a fault. The DOM-shim run builds every tab, so a missing sentence fails there rather than shipping as a blank rectangle.

A tab whose desktop twin does something the browser cannot do says so, in the tab. Plan analysis, the query heatmap, cached-plan retrieval and actual-plan re-execution need a plan renderer and a Stage-2 command back to the monitored server; the block-chain view and the interactive deadlock graph need a graph viewer. Blocking hands over the captured blocked-process-report and deadlock-graph XML verbatim rather than imitating the graph. A reader told to open the desktop viewer is better served than one given a web page that looks like plan analysis and is not.

## Verification

The repository has no JavaScript test runner, so this follows the #2447 / #2451 idiom: xUnit text-scan pins over the shipped module, plus running the shipped JS under a minimal DOM shim.

**The pins are invariants, not spot-checks.** With ~60 read names spread across eleven tabs, a typo is a 400 inside one panel at runtime and completely invisible to inspection. So `ServerPageTabsTests` asserts the category:

- every read name the module mentions exists in `BuildReadDispatch()`;
- every parameter key it sends is one that read actually binds, per `CatalogDescriptors` — this is the half that fails *silently*, since an unknown query key is ignored and `limit` sent to a read that binds `top` quietly returns the default 20 rows;
- every viz named is in `KnownVizList`, and `panels.js`'s registry is that same vocabulary;
- the sub-tab bar is built from the registry and the router carries the tab back;
- no `get_pg_*` read appears — see below.

Both sides of each check come from the shipped artifacts, never a transcribed list.

**The DOM-shim run** exercised the shipped `server-tabs.js` and `server.js` under a stubbed `fetch` across four response shapes — the empty envelope, an error, data with no rows, and data with rows — at three time ranges, plus `renderServer` with an absent, valid and unknown tab id. Zero throws in all cases; 142 distinct requests issued, every one of them checked against `CatalogDescriptors` for parameter validity. Zero problems.

**Negative controls, all three run.** A bogus read name fails the dispatch pin; `limit` swapped for `top` on `get_latch_stats` fails the parameter pin naming both the key and the allowed set; a `get_pg_*` read fails the engine pin. Removing one `emptyText` fails the shim run. Each was restored afterwards.

The xUnit file builds (`Darling.Tests`, `EnableWindowsTargeting`) but cannot execute on macOS. To avoid deferring it to CI, all eight facts were compiled into a throwaway `net10.0` console project against the real `PerformanceMonitor.Darling.Service` build with a minimal xUnit shim, so the assertions that ran are the shipped ones rather than a retyped copy. All eight pass.

## The backend gap this change names but does not close

`DarlingFleetReader.FleetServerCard` carries `engine_edition` — the SQL Server `SERVERPROPERTY` value — and no `CollectorTargetEngine` discriminator. The browser therefore cannot tell a PostgreSQL target from a SQL Server one, so a `get_pg_*` panel added to these tabs would render on every SQL Server in the fleet, permanently empty. `NoPostgresRead_IsOnTheServerPage_UntilTheFleetPayloadCanTellTheEngines` fails until the discriminator exists, rather than letting the panel ship first. The other unreachable readers, and the nested-array shape `get_database_scoped_config` returns, are enumerated on #2475.

---

## Since opening

**Twelve tabs, not eleven.** The prose said eleven while the registry held twelve. The count is now pinned exactly rather than as a floor, so it cannot drift again.

**Two existing pins broke, and fixing only them would have been the scenario-shaped repair.** `DarlingEmptyEnumerationNoteTests` and `ViewerGridPayloadColumnOrderPinTests` scan the web server page *by filename* for `COLLECTOR_COLUMNS` and `ACTIVE_COLUMNS`, and those arrays moved into `server-tabs.js`. Both now scan where the arrays live. But the query-text pin asserted #1949 ordering on **one** array because that was the only web list rendering query text — the sub-tabs added five more, and a rule enforced on one of six is a rule that gets broken on the other five. It is a `[Theory]` over all seven now, and the six new grids were reordered to obey it: Top Queries, Query Store, Blocking and Deadlocks had their text behind their metrics, and Plan Corrections had it dead last.

**The server header's band badge said "Warning" and nothing else.** `Warning` has three unrelated causes — a genuine metric breach, a server awaiting its first collection, a collector error — so that badge was #2422 rebuilt on a new surface. It now carries the fleet's own pre-computed `reason` when the server is in the worst-first ranking, over the fleet card's severity chips, rendered by `fleet.js`'s own `metricBands` (exported, not reimplemented). A server outside the ranking gets the chips and no sentence: inventing one there is the second derivation the fix exists to refuse.

**An empty trend is not a warming-up trend.** `get_blocking_trend` and `get_deadlock_trend` answer an *idle* server with `trend: []` and **no** `{status,message}` envelope — so `renderPanel` had real data in hand and `vizLine` fell through to the chart's "Not enough data points to chart yet". A server with no blocking was being told its blocking chart was still warming up. `vizLine` now renders a descriptor's `emptyText` at **exactly zero** rows; one row still gets the chart's own sentence (there it is the true one), and a descriptor with no `emptyText` still falls through, so every stored view authored before this is unchanged. `line()` requires the sentence the way `table()` does. Verified against the shipped `vizLine` across all three cases, negative control run.

**`docs/uat-onboarding.md` §3.4 is updated here, not deferred.** #2476 listed the wait-type picker, tempdb, the perfmon picker, Running Jobs and the configuration grids as missing; all five ship on this branch, and a UAT tester reading absences that are no longer absent is worse served than by no list. Two counts in it were also wrong and are corrected from the XAML: the viewer has **nineteen** top-level per-server tabs, not fourteen, and **65** `TabItem`s counting the nested ones. What remains missing is now divided by *why*, because the three groups have different answers — reads served but unshown, viewer data with no read endpoint at all (now #2484), and desktop things a web imitation would be worse than. The exclusion bullets above that paragraph are untouched: no plan analysis, no `analyze_server`, no `config_command` enqueue, no alert tuning. Nothing here changes them.


erikdarlingdata:
All three review findings addressed in 2c30e7cc — and the first two were the visible part of a larger set.

**The duplicate fetches: six, not two.** `readTool`/`apiGet` have no cache, so a read feeding several panels on one tab ran several times. Scanning for the category rather than fixing the two named turned up `get_plan_cache_bloat` (memory), `get_session_stats` (activity), `get_health_parser_system_health` (events) and `get_perfmon_stats` (the composite had already fetched exactly what the standalone snapshot table re-fetched) alongside the two you found. New `fanout()` helper takes one read and several ordinary panel descriptors, mapping the same three response kinds `renderPanel` does and carrying the same empty-state rule, so routing a panel through it is never how the sentence gets lost.

Measured under a DOM shim, per tab: **zero duplicate fetches across all twelve**, Collection Health three fetches down to one. `NoTab_FetchesTheSameReadTwice` is the guard — it maps the composites' own reads explicitly and asserts that map against the composites, so it cannot go stale and start passing a tab it no longer describes. Negative control run: reintroducing one duplicate fails it by name.

You were right that `automatic_tuning` ignores `hours_back`/`limit` — with one shared fetch that is moot, and the comment now says so.

**The CHANGELOG headline:** fixed. Worth noting what happened there — the tab-count pin was written precisely to catch eleven-versus-twelve drift, and it did catch it in the code; it just cannot see prose, so the bolded lead-in kept the stale number while the body next to it was correct.

Issue #2468, opened 2026-08-21
query_store's per-database fan-out costs ~10s per database serially, so a multi-tenant box cannot finish a 60s body


erikdarlingdata:
The end of the chain that started with six skipped relaunches on `shard-A` (#2446 → #2459 → #2460). The cause is now established rather than suspected, and this is the issue for the cause itself.

## What the collector does

`DarlingCollectorRunner` returns immediately when the Query Store enumeration yields nothing — that is the 36 ms an empty run costs, measured on `single-A` across all 1,555 of its runs. Otherwise it runs the 55-column `sys.query_store_runtime_stats` payload **once per database, on one connection, in sequence**.

## What that costs on a real server

Verified against the fleet:

```
shard-A    8 user databases
  db-1               630 GB        db-2              583 GB
  db-3             527 GB        db-4   426 GB
  db-5         246 GB        db-6 216 GB
  db-7             174 GB        db-8            21 GB

productive query_store run   ~80,900 ms   ÷ 8 ≈ 10.1 s per database
cadence                       5 minutes
collection body budget       60 seconds
```

So one productive cycle costs more than the whole body it runs in, and it is scheduled twelve times an hour. `single-C` works out to ~65,500 ms on the same arithmetic. `single-A`, whose enumeration yields nothing on every run, logged no skips at all — the discriminator is exactly this.

This is not a slow query. It is eight sequential ones with nothing wrong with any of them individually.

## The lever that is visible from source

`WatermarkPolicy.ClampCatchup` allows a **24-hour catch-up** per database. A database whose watermark has gone stale therefore reads up to a full day of runtime stats in a single cycle — which would make individual runs far heavier than steady state, and would show as `p95 ≪ max` on the columns #2466 just added. That is now a checkable question rather than a hypothesis, and it should be checked before anything is changed: if the ~10 s per database is steady-state cost, the answer is different from if it is catch-up cost.

## Shapes worth weighing

**Parallelise across databases.** The obvious one, and the one to be most careful with — the collection body is what it is because it is serial, and the monitored server is a production instance that did not ask to be asked eight questions at once. Any concurrency here needs a bound and a justification, not just `Task.WhenAll`.

**Give the collector its own cadence per database, or stagger them.** Eight databases at 5 minutes does not have to mean eight databases every 5 minutes. Staggering keeps the per-cycle cost bounded without reducing coverage, at the price of a longer effective interval per database.

[...]

Issue #2460, opened 2026-08-21
avg_duration_ms averages across a bimodal population: query_store's empty runs cost 36 ms and its productive ones ~80 s


erikdarlingdata:
Split out of #2446, which asked why `query_store` on `shard-A` reports `enumeration yielded 0 items - nothing to collect this cycle` on most of its runs and still costs 13,834 ms on average. The premise turns out to be wrong, and what replaces it is more interesting than the original question.

## An empty enumeration costs 36 ms, not 13.8 s

`single-A` is the control, and it is a clean one — same fleet, same box, same collector, same 7-day window:

| server | `query_store` runs | empty-enumeration runs | `avg_duration_ms` |
|---|---|---|---|
| `single-A` | 1,551 | **1,551 (all of them)** | **36** |
| `shard-A` | 1,165 | 968 | 13,716 |
| `single-C` | 1,202 | 982 | 12,013 |

alpha yields nothing on *every single run* and pays 36 ms for it. So the enumeration coming back empty is cheap, exactly as it should be, and the 13.8 s on shard-A is not being paid by the 83% of runs that do nothing.

## Which means the productive runs are enormous

Backing alpha's 36 ms out of the other two (arithmetic only — see the caveat below):

- **shard-A**: 1,165 × 13,716 = 15,979,140 ms total. 968 empty × 36 = 34,848 ms. The remaining **15,944,292 ms across 197 productive runs is ~80,900 ms each.**
- **single-C**: 1,202 × 12,013 = 14,439,626 ms total. 982 empty × 36 = 35,352 ms. The remaining **14,404,274 ms across 220 productive runs is ~65,500 ms each.**

The result is insensitive to the empty-run figure: even at ten times alpha's cost (360 ms per empty run), shard-A's productive runs still come to ~79,300 ms.

**A single `query_store` run on those servers costs more than the entire 60,000 ms sweep budget.** That is a better candidate for the skipped relaunches in #2446 than anything in that issue or in #2459 — it does not need a coincidence of cadences, it overruns on its own.

## The reporting defect underneath it

`get_collection_health` serves one `avg_duration_ms` per collector. When a collector is **bimodal** — and `note_count` vs `total_runs` already says this one is, 968 of 1,165 — that average describes neither population. 13,716 ms is not what a run costs; it is a blend of 36 ms and ~80,000 ms in an 83/17 mix, and it happens to land in a range that reads like a plausible single number. Everything downstream inherits it:

[...]

Issue #2319, opened 2026-08-18
database_config reports query_store as a boolean — collect its health (actual_state, size vs cap, read-only reason) so a thrashing store is visible


erikdarlingdata:
Found during #2312's yardstick. `prod-sql-use1-beta-01`'s single database costs 300+ seconds per query_store collection cycle — 298s for a cycle that shipped ZERO rows — and the leading hypothesis is a huge or capped Query Store whose whole retained history the per-cycle read scans. To check that hypothesis I reached for the product's own surface and found:

```json
"query_store": true
```

That's everything `database_config` knows. `sys.database_query_store_options` carries the fields that would answer it — and that catch the classic silent failure where desired_state says READ_WRITE but actual_state fell to READ_OFF or READ_ONLY (storage cap hit, `readonly_reason`):

- `actual_state` / `desired_state` (and whether they disagree)
- `current_storage_size_mb` / `max_storage_size_mb` (how close to the cap)
- `readonly_reason`
- `size_based_cleanup_mode`, `stale_query_threshold_days`, `max_plans_per_query`

One row per database with QS enabled, tiny and slow-changing — it belongs in the existing `database_config` per-database read (both SKUs, both stores). Alerting can then say "Query Store on X hit its cap and went read-only" instead of nothing, and an investigation like #2312's can rank databases by retained QS size without ad-hoc access to the target.

Not #2312's fix — that issue is chasing the per-cycle scan cost itself — but it's the instrument that investigation was missing.

erikdarlingdata:
PR #2328 implements this — a new `query_store_health` enumerating collector (V76, both SKUs, both stores, hourly), surfaced as a Query Store sub-tab on both apps' Configuration tab.

One deliberate divergence from the framing above, defended in the PR: `database_config` is a single `sys.databases` scan, and these fields need per-database context — so this lands as `database_scoped_config`'s sibling (the same enumeration idiom) rather than columns on `database_config`, preserving that collector's execution model and failure isolation.

Also included beyond the listed fields: `interval_length_minutes` — the runtime-stats grain, which is exactly what an investigation like #2312 needs to interpret per-interval cost on a target like beta.

Remaining after the PR: a `get_query_store_health` MCP read (fleet-rankable QS size without box access), sliced as its own PR the way pvs_stats shipped. Leaving this issue open until that lands.

erikdarlingdata:
Shipped in two parts, both on dev:

[...]

PR #2478, opened 2026-08-21
Record which database a fan-out spent its time in (#2472)

### LINKED ISSUE #2472: collection_log blends a per-database fan-out into one duration, so nothing can say which database cost what
Found while investigating #2468, and it is why that issue cannot be decided yet.

## The mechanism

`collection_log` gets **one row per collector run per server**. `DarlingObservability.LogCollectionAsync` has no database parameter:

```csharp
public static async Task LogCollectionAsync(
    NpgsqlDataSource postgres, ServerRuntime server, string collectorName, string status,
    int rowsCollected, long sqlMs, long storageMs, string? errorMessage, ILogger? logger, CancellationToken cancellationToken)
```

and `duration_ms` is bound from `(int)(sqlMs + storageMs)`, where `CollectorRunResult.SqlMs` / `.StorageMs` are the **sums across every database** in the fan-out (`DarlingCollectorRunner.cs:327-329`, `:992-994`).

So for `query_store` on an 8-database server, one row says 80,900 ms and nothing says which database spent it.

**The per-database numbers already exist.** `EnumeratedCollectorDriver` hands `onItemComplete(item, batchCount, itemSqlMs, itemStorageMs)` per database, and the runner logs them with a full phase breakdown (`DarlingCollectorRunner.cs:931-961`):

```
[{Server}] {Collector} [{Database}] => {Rows} rows
    (sql:{SqlMs}ms = wm:{WatermarkMs}ms + open:{OpenMs}ms + drain:{DrainMs}ms
     + plan_fetch:{PlanFetchMs}ms + text_fetch:{TextFetchMs}ms, pg:{PgMs}ms)
```

They are **log lines only**. Nothing persists them, so no read can reach them. The only per-database state that *is* persisted is the derived watermark, the open-interval stamps, and the backfill hole records — none of which carry cost.

## Why it blocks #2468 rather than being a nicety

`8 × 10.1 s` and `1 × 62 s + 7 × 2.7 s` are the same 80,900 ms and want completely different fixes:

| shape | if the cost is even | if one database dominates |
|---|---|---|
| bounded parallelism | width 3 fits the body | Amdahl eats it; buys almost nothing |
| stagger per database | helps per cycle | helps only if you defer the right one |
| per-database schedule override | little to target | this is the fix — for one database |

Three of #2468's four proposed shapes need this to be **designed** correctly, and two of them need it to be **operated** at all: a per-database override ships a knob with no way to know which database to turn it for. #2466 gave `get_collection_health` `p95_duration_ms` and `max_duration_ms`, which is a real improvement, but both are whole-fan-out figures — they can tell you a collector is bimodal across *runs* and still cannot tell you it is bimodal across *databases*.

Today the only route is grepping `[{Database}]` out of the service log on the box, which is not a route an operator has.

## Shapes worth weighing

**A note on the `collection_log` row.** Tempting — zero schema change, `error_message` already carries the SUCCESS note, and `get_collection_health` already serves `last_note`. **It should not be done.** The health tool's note handling is built on notes *repeating* (`note_summary` reports "N of M runs", and the tool's own contract leans on `note_count == total_runs` meaning "collecting nothing all window"). A note that varies per run makes `note_count` equal `total_runs` for a perfectly healthy collector and destroys that signal. Recording this cheaply by corrupting an existing one is the wrong trade.

**Its own table**, `collect.collector_item_timings` or similar, keyed `(server_id, collector_name, item_name, collection_time)`, carrying `rows`, `sql_ms`, `storage_ms` and the phase split the log line already computes. Costs a migration rung + a Lite DuckDB twin + the literal pins + a read. One row per database per productive run — on `shard-A` that is 8 rows per productive cycle, and the empty cycles write none, so the volume is bounded by the thing being measured. Only the enumerated collectors would write it (query_store plus the two snapshot ones), not all 41.

**A rollup on the existing row** — `slowest_item text`, `slowest_item_ms int` — as two nullable columns on `collection_log`. Much cheaper, answers "is one database dominating and which" without a new hypertable, and that is the *specific* question all three fixes need. Loses the full distribution. **Watch the Lite twin here**: `DuckDbAlertHistoryStore`-style partial-column `INSERT OR REPLACE` is a known trap in this repo, so any column added to a table Lite also writes needs its writer checked, not assumed.

**Do nothing.** Defensible only while #2468 stays unfixed. The moment any of its three active shapes is built, this becomes the measurement that says whether it worked.

## What should decide it

Whether the per-database costs on a real `multi-*` box are near-even or dominated by one database. Ironically that is the question this issue exists to make answerable — but it can be sampled *once*, by hand, off the service log on `the use1 monitor host`, grepping the `[{Database}]` lines for one server over an hour. If it is near-even, the rollup is enough. If one database dominates, the rollup is enough *and* obviously correct. The full table is only worth its cost if the distribution turns out to move around over time.

I would take the rollup on that basis, but the sample should come first and I could not take it in this session (no route to the box).

_Edited: server and database identifiers replaced with role placeholders — this is a public repository. Measurements and reasoning unchanged._

erikdarlingdata:
Closes #2472.

## The decision, and the arithmetic behind it

**Built: the rollup.** Three nullable columns on `collection_log` — `fanout_item_count`, `slowest_item`, `slowest_item_ms` — plus the read that serves them. Not the `collector_item_timings` hypertable, and not "do nothing".

**"Do nothing" was tested first and it fails.** #2460's `p95_duration_ms` / `max_duration_ms` are a real improvement and they cannot close this. Both aggregate over **runs**, and a fan-out run is one blended row:

| shape | per-database | `duration_ms` | avg / p95 / max | **dominance** |
|---|---|---|---|---|
| even | 8 × 10,100 | **80,800** | 80,800 | **1.00** |
| dominated | 61,900 + 7 × 2,700 | **80,800** | 80,800 | **6.13** |

`dominance = slowest_item_ms × fanout_item_count / duration_ms`. That is the whole change: one number that is 1.0 for a perfectly even fan-out and rises with concentration, which is exactly where #2468's remedies fork — near 1.0 the cost is the fan-out's **width** and bounded parallelism is the lever; at 2.0 and above one database dominates and only a per-database override or a stagger reaches it.

(The issue writes these as `8 × 10.1 s` and `1 × 62 s + 7 × 2.7 s`, rounding both to 80,900. They don't actually balance — 8 × 10,100 is 80,800 — and a claim that two runs are indistinguishable can't be shown with two numbers 100 ms apart, so the tests use figures that balance exactly. My harness caught that; it would otherwise have shipped as a test asserting two unequal numbers were equal.)

**The live fleet makes p95 worse here, not better.** On `prod-sql-use2-multi-49`, `query_store`'s 1,196 runs include 999 empty enumerations, so its p95/avg ratio is already saturated by empty-versus-productive — `get_collection_health`'s own description says exactly this. An operator reading that big tail would learn nothing about which database is expensive. The tool description now says so explicitly.

## The storage trade, with real numbers

Measured off the fleet rather than assumed:

- The five fan-out collectors are ~**601** of the ~**31,200** `collection_log` rows a server writes per day — **1.9%**, not the whole table.
- A separate `collector_item_timings` table on a 9-database shard would add ~**3,200 rows/day/server**, roughly a tenth of `collection_log`'s own volume, retained for as long as it is.
- **These three columns add zero rows** and are NULL on ~98% of the rows already being written. On a hypertable compressed and segmented by `server_id`, a ~98%-NULL column is very nearly free.

What is given up is the rest of the distribution. That is the trade, taken deliberately.

## The sample I could and couldn't take

The issue names its own decider: sample the per-database costs off the service log on a `multi-*` box. **I could not take it** — the box needs SSM, which this session's sandbox blocked, same as when the issue was filed. What I could take from the store instead:

- `multi-49` has **9 databases**; the six other shards sampled run **3–9**. (Not the "8–12" the framing assumed.)
- Query Store storage distribution: the worst database holds **2.1×–3.5× an even share** on every one of the seven, with the top three holding 91–100%.

That last one is a **stock proxy for a flow cost** and I'm not claiming more for it — the collector is watermark-driven, so per-run cost tracks new rows, not total size. But it lands *between* the two branches of the issue's own decision rule, and **both branches choose the rollup**. The shape is robust to the sample; only "do nothing" depended on it, and the arithmetic above rules that out.

## What else the measurement turned up

**The fan-out set is five collectors, not three.** The issue and a comment in the runner both say "query_store plus the two snapshot ones"; `query_store_health` (#2319) and `plan_correction` (#1952) joined since.

**`plan_correction` is the one that matters most in steady state.** On all seven servers sampled it is productive on *every* cycle (note_count 0), at 1,043–7,771 ms on a 5-minute cadence — so its blended duration is a fan-out sum every single time, where `query_store`'s is one 16% of the time on a busy shard and never on a quiet one. The issue's framing is the peak case; this is the sustained one.

## Shape of the change

- **V80 / Lite v55.** Nullable, no `DEFAULT` — a catalog-only change in PostgreSQL, instant on the largest table in the store, and not the shape TimescaleDB refuses on a compressed hypertable. No backfill and none possible: a historical row genuinely does not know its fan-out, and NULL says so where 0 would read as "fanned out over nothing".
- **The view refresh.** Postgres freezes a view's `SELECT *` at CREATE, so without `CREATE OR REPLACE VIEW` the passthrough every read goes through would serve eleven columns forever — the write correct, every read blind. V14 exists because that already happened once. The live test reads back **through `v_collection_log`** for exactly that reason.
- **One accumulator, both fan-out shapes, both SKUs.** `query_store` fans out through the enumeration driver on-prem and through the Azure per-database connection loop on Azure SQL DB; two hand-rolled copies would let one collector hold two notions of a slow database.
- **Every completed item is counted, empty batches included.** Their read time is in the blended total, so the ratio needs the same denominator — counting only productive databases would inflate the dominance of whichever one had rows, which is the wrong database to send someone after.
- **`LogCollectionAsync` takes it as a required parameter.** Five collectors fan out and 36 do not; a default would let the five failure paths stand in for a success path that forgot, and a blended number quietly staying the only record is the defect itself. The five say `fanout: null` by name.
- **Lite's `collection_log` writer is a plain full-column INSERT**, not the partial `INSERT OR REPLACE` that resets untouched columns elsewhere in that SKU — checked, not assumed, and left naming every column.

## Verification, and its limits

- Builds on macOS: Common, Collectors, both Darling projects, Lite, `Lite.Tests`, `Darling.Tests`.
- The accumulator and the dominance arithmetic **ran against the real build** in a throwaway `net10.0` console — 13 checks, all passing, including the adversarial tie case and the 0 ms floor.
- The SQL, the ordinals and the rung are CI's: `Darling PostgreSQL tests` runs the migration and the new round-trip against a live store; the Windows job runs the pins.
- Probe cascade **appended, never inserted**: sentinel, reader ordinal 55, map parameter, newest-first arm above V79's. V79's `InvokeMap` arity needed no edit — its fixed leading count doing the job its own comment describes — and its two ladder-top assertions are demoted.

## Deferred

The per-database **distribution over time** (a second-worst database, a per-item p95) is not here. If the rollup shows dominance moving around from cycle to cycle, that is the evidence for the full `collector_item_timings` table, and this is the measurement that would produce it.


erikdarlingdata:
Self-review found the read broken and pushed a fix — worth calling out because the shape of it is the interesting part.

Both health queries aggregate over a subquery that **enumerates its columns** rather than `SELECT *`-ing them. I added three outer aggregates naming `fanout_item_count` / `slowest_item` / `slowest_item_ms` and projected none of them. Valid C#, compiles everywhere, passes every `Assert.Contains` anyone would think to write, fails only when Postgres parses the text. Both SKUs had it.

**Nothing in the suite would have caught it.** `DarlingDataReader.CollectionHealthSql` — the query the MCP and the web dashboard both serve collection health from — had *no live test executing it at all*. The live tests nearby exercise the **viewer's** separate `CollectionHealthSql`; the text-level tests assert about the string. A query that is never run is a query whose dialect nobody has checked.

So the new guard **runs the shipped text** over a fan-out whose answer is known: two `query_store` runs of the same 80,800 ms — one even across eight databases, one dominated by a single one — plus a `wait_stats` run that never fanned out. Known-answer rather than smoke on purpose: a query that parses and returns nulls looks identical to one that works on a store holding no fan-out rows, and "it ran" would have been the same green either way.

Issue #2479, opened 2026-08-21
Six things UAT testers need that the product makes a person answer instead


erikdarlingdata:
Six things a UAT tester needs that the product does not do yet, found while writing the onboarding (#2476). Filed together because they share a cause — each one is a question the product can answer for itself and currently makes a person answer instead — but they are independently shippable and roughly ordered by what a tester hits first.

## 1. `install-darling.ps1` does not check for the .NET runtimes

The service needs **ASP.NET Core 10** unconditionally — `ModelContextProtocol.AspNetCore` pulls the framework reference in even with MCP off — and the viewer needs **Desktop 10**. A stock Windows Server has neither.

The failure is the raw .NET host error with nothing of ours on it, at the very first step, before the tester has any context for debugging. This is the highest-value item here: it is the first thing they hit and the least self-explanatory. A pre-flight check naming both runtimes and the download link costs almost nothing.

## 2. No verb reprints an MCP or web token

`--configure-network` shows the plaintext once. Lose it and the only path is regeneration, which invalidates every client already configured against it — so one mislaid token during UAT means re-onboarding every consumer.

A `--print-mcp-token` (elevated, warning to stderr) would cost nothing and remove a class of unrecoverable mistake. The token is already DPAPI-decryptable in-process on that box; nothing new is exposed that an admin on the host could not already obtain.

## 3. The viewer has no global read-only indicator

You discover your seat one surface at a time, per dialog — which is exactly the #2400 report, and #2403 only improved the wording of the individual refusals rather than answering the question up front.

The viewer already probes `IsReadOnly` once at connect. A status-bar element rendering it would end this class of question instead of re-phrasing it, and it is a strictly smaller change than the four message rewrites already shipped.

**This gets worse in UAT, not better**, because of the next point.

## 4. A remote seat is read-only by default, and nothing says so

`postgres.network.role` defaults to the restricted `viewer` role while a local `connectAs` defaults to `admin` (verified: `PostgresConfig.ConnectAs = "admin"`). So every remote tester connecting from their own desktop — which is what a UAT tester does — gets a read-only seat **by design**, and then meets #2400's message with no indication that this is the expected default rather than a misconfiguration.

The onboarding now states it. The product still does not.

## 5. Rejected HTTP requests are unlogged

[...]

PR #2481, opened 2026-08-21
Check the .NET runtimes before installing, instead of after (#2479)

erikdarlingdata:
Item 1 of #2479, and the one a UAT tester hits first.

## What was wrong

Both shipped binaries are **framework-dependent** publishes, so both name a shared framework in their `runtimeconfig.json` and neither carries one. Verified against the actual build output rather than the summary:

```
PerformanceMonitor.Darling.Service.runtimeconfig.json
  "frameworks": [ Microsoft.NETCore.App 10.0.0, Microsoft.AspNetCore.App 10.0.0 ]
```

and `PerformanceMonitor.Darling.Viewer.csproj` is `net10.0-windows` + `UseWPF=true`, published into the zip's `viewer\` folder by `dotnet publish ... -c Release -o publish/DarlingViewer` — no `--self-contained` — so it takes `Microsoft.WindowsDesktop.App` from the box.

The ASP.NET Core requirement is **unconditional**. `ModelContextProtocol.AspNetCore` brings the framework reference in transitively, so the framework is named whether or not `mcp.enabled` and `web.enabled` are ever turned on. Setting them false does not remove it.

A stock Windows Server image has neither runtime, and what the operator got was worse than nothing: the step-2 pre-flight is the first thing that *runs* the service exe, so it is the pre-flight that surfaces the .NET host's `You must install .NET to run this application` — and the script then reports

> `Pre-flight FAILED (exit N) - the config is invalid or a server is unreachable.`
> `Install anyway? A monitoring service may legitimately be installed while a target is down. [y/N]`

That diagnosis is not merely unhelpful; it points at the wrong file.

## What this does

A new step **1c**, after the install-location guard (a pure string decision that costs nothing, and there is no point telling somebody to install a runtime for a folder about to be refused) and **before** step 2 touches the exe:

- **Missing ASP.NET Core 10 → refusal**, naming the runtime, the version, what was found instead, and <https://dotnet.microsoft.com/download/dotnet/10.0>. The service cannot start without it, so proceeding is never right.
- **Missing Desktop 10 → warning that continues.** A headless collector host nobody opens a window on is a legitimate deployment. But it must not be silent: this script creates Desktop and Start Menu shortcuts a few steps later and a tester will double-click one. Scoped to the co-located `viewer\` folder on purpose — the remote-seat viewer from the Velopack `Setup.exe` is self-contained and needs no Desktop runtime at all.

Detection resolves the dotnet root the way the host does — `DOTNET_ROOT`, then `HKLM:\SOFTWARE\dotnet`'s recorded `InstallLocation`, then Program Files — and enumerates `shared\<framework>\<version>`. `dotnet --list-runtimes` is folded in as a **supplement**, not the primary check: a box with neither runtime installed frequently has no `dotnet.exe` on it at all, which is precisely the box this gate exists for.

It rides the existing `-SkipPreflight` switch rather than adding a second one, and the refusal names it. A check that can be wrong about a private runtime layout needs one documented way past it, not a dead end.

`Darling/README.md`'s `- **.NET 10** to build and run.` is replaced with the two runtimes, which one each binary needs, and the note that ASP.NET Core is not conditional on MCP.

## The boundary, and why it is executed rather than read

`1.10.0` and `110.0.0` both contain the characters `10.` and neither is .NET 10. A `-like '10.*'` gets both wrong in the direction that costs most: the gate reports success, the install proceeds, and the operator gets back the raw host error this gate exists to replace — now with a green line above it claiming the runtime is present. So the test is on the parsed leading integer. Major is the right granularity because it is what the host rolls forward at: any 10.x satisfies a `net10.0` app and an 11.x does not.

## Verification

- **Executed the shipped function**, extracted from `install-darling.ps1` exactly as the test does, over all 16 boundary cases plus the two `Format-FrameworkVersionList` cases. All 18 match. Run in a `mcr.microsoft.com/powershell` container (PowerShell 7 on Linux) because macOS has no PowerShell — the functions use no 7-only syntax, and `DarlingRuntimePreflightTests` re-runs the identical probe under `powershell.exe` (5.1) on the Windows CI job, which is the host that matters.
- `Darling.Tests` builds clean, 0 errors.
- Every new assertion fails against `dev`: none of `Microsoft.AspNetCore.App`, `Microsoft.WindowsDesktop.App`, `$dotnetMajor = 10`, `dotnet/10.0` or `Test-FrameworkMajorPresent` exists in `dev`'s `install-darling.ps1`.
- **Not verified:** the check has not been run against a real Windows Server with a runtime genuinely absent. Manual verification note for whoever does: on a box with only the Desktop runtime, `.\install-darling.ps1` should refuse before probing any SQL Server and name ASP.NET Core; with only ASP.NET Core it should warn about the shortcuts and continue.

## Onboarding

Makes the second bullet of `docs/uat-onboarding.md` §1.1 Prerequisites (#2476) stop needing to say *"`install-darling.ps1` does **not** check for them, and a missing runtime surfaces as the raw .NET host error."* The section should still list both runtimes — installing them first is still faster than being refused — but the workaround sentence goes.

### DIFF
--- a/Darling/Darling.Tests/DarlingRuntimePreflightTests.cs
+++ b/Darling/Darling.Tests/DarlingRuntimePreflightTests.cs
@@ -1,0 +1,343 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using Xunit;
+
+namespace Darling.Tests;
+
+/// <summary>
+/// <c>install-darling.ps1</c>'s .NET runtime gate (#2479, item 1).
+///
+/// <para><b>The gap.</b> Both shipped binaries are framework-dependent publishes.
+/// <c>PerformanceMonitor.Darling.Service.exe</c> names <c>Microsoft.AspNetCore.App</c> in its
+/// runtimeconfig.json — unconditionally, because <c>ModelContextProtocol.AspNetCore</c> brings the
+/// framework reference in transitively, so turning <c>mcp.enabled</c> and <c>web.enabled</c> off does not
+/// remove it — and the co-located <c>viewer\</c> publish names <c>Microsoft.WindowsDesktop.App</c>. A stock
+/// Windows Server image has neither. What the operator saw was the .NET host's own "You must install .NET"
+/// error, emitted by the step-2 pre-flight, which then reported "the config is invalid or a server is
+/// unreachable" — a diagnosis that is not merely unhelpful but points at the wrong file.</para>
+///
+/// <para><b>Why the tests are split the way they are.</b> Everything structural — the ordering, the
+/// fail-versus-warn verdict, which switch governs the gate — is pinned by reading the script, the same way
+/// this project pins every other PowerShell invariant it cannot compile
+/// (<c>DarlingInstallLocationTests</c>, <c>DarlingFirewallCheckTests</c>). The version BOUNDARY is not:
+/// deciding whether <c>1.10.0</c> is a .NET 10 runtime is a computation on a string, its failure mode is a
+/// gate that reports success on a box with no runtime at all — strictly worse than no gate, because it
+/// looks like the check ran — and no source-parsing assertion can see it. So that one is executed against
+/// the function as it ships, under <c>powershell.exe</c> (Windows PowerShell 5.1), which is what Windows
+/// Server hands an operator by default.</para>
+/// </summary>
+public class DarlingRuntimePreflightTests
+{
+    private static string InstallScript => ReadRepoFile(Path.Combine("Darling", "tools", "install-darling.ps1"));
+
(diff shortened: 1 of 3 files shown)


erikdarlingdata:
Good parity catch, and it is sharper than described — filed as #2489 rather than widened into this PR, which is a Darling-installer change.

Verified: Lite's `runtimeconfig.json` names **three** frameworks (`Microsoft.NETCore.App`, `Microsoft.WindowsDesktop.App`, **and** `Microsoft.AspNetCore.App`), so the plain zip needs two runtimes installed, not one. And the documentation has it backwards: `README.md:95` attaches the runtime requirement to `PerformanceMonitorLite-win-Setup.exe`, which is packed by `vpk` from `publish/Lite-velopack` (`build.yml:315`, `--self-contained`) and therefore does **not** need it — while the framework-dependent zip that does need both is undocumented, and `Lite/README.md` has no prerequisites section at all.

So today a tester who reads our one runtime sentence installs a runtime they did not need, and a tester who unzips the plain release gets the bare host error. #2489 has the evidence and the three options.

Not fixable the way this PR fixes Darling, incidentally: the host error happens before any of our code runs, and Lite has no install script to gate — which is exactly why the Darling fix lives in `install-darling.ps1` and not in the service.

Issue #2437, opened 2026-08-21
Lite's card and the web dashboard have the two Overview defects #2429 just fixed in the Viewer


erikdarlingdata:
#2429 fixed both Overview defects in the Darling Viewer. Two other surfaces have the same ones, verified in source.

## Lite's card has the identical unexplained status

`Lite/MainWindow.xaml:442`:

```xml
<TextBlock Text="{Binding StatusDisplay}" Foreground="{Binding StatusBrush}"
           FontSize="11" Margin="0,0,0,10"/>
```

Same bare binding, same absent `ToolTip`. A Lite user reads "Warning" and scans the metric rows for the yellow one, exactly as @ehaar had to on the Viewer in #2422.

**This is a port rather than a wiring change**, which is the whole difficulty: Lite has no `FleetRollup` and no `BuildReason`, so the reason has to be built rather than bound. The Viewer's version is worth copying carefully — `BuildReason` is deliberately assembled from the card's OWN metric displays so it can never drift from the six rows underneath, and a Lite implementation that recomputed severity independently would lose that property and be worse than no tooltip.

One difference in Lite's favour: it has no `AwaitingFirstCollection` flag, so the status/tooltip desync family that #2429 spent four review rounds on cannot occur there. Only the unexplained-word half applies.

## The web dashboard's "+N more" is the same dead end

`Darling/PerformanceMonitor.Darling.Service/wwwroot/js/pages/fleet.js:168`:

```js
nodes.push(el("div", { class: "muted", style: "margin:0.4rem 0 0.2rem",
                       text: "+ " + d.additional_problem_count + " more need attention" }));
```

A plain muted `div`, no handler, no navigation — the same "here is a number you cannot act on" the Viewer had.

This one should be **easier** than the WPF version was, because the seam already exists: `fleetFilter` is a module-level variable at line 33, `redrawCards()` at 183 already filters `lastCards` through `cardMatches(c, fleetFilter)` before sorting, and the count itself arrives on the payload as `additional_problem_count`. So option (a) from #2424 — make it a link that filters the grid to the servers behind the number — ports nearly as-is.

Keep the two properties #2429 settled on, since they are what make the filter safe rather than confusing: filter on the *same* predicate the count is computed from, so the destination cannot disagree with the label; and make the active state visible (`showing 12 of 57`) so a filtered grid can never be mistaken for an unfiltered one.

## Worth doing together

[...]

PR #2482, opened 2026-08-21
Five ready-made dashboards for the empty Custom Views page

### LINKED ISSUE #2480: A new user's Custom Views page is empty, and its three 'popular starting points' are inert text
No custom view ships seeded, so the first thing a new user sees at `#/views` is an empty page. The first-run hero (#1563 M4) sells the feature well and then hands over two buttons onto a blank canvas over an 82-read catalog and a composed-measure vocabulary — with no worked example of either.

The hero already knows what would help. It ends with this:

```js
const suggestions = ["Top waits by server", "CPU trend over time", "Slowest procedures by database"];
...
el("div", { class: "hero-suggest-chips" }, suggestions.map((s) => el("span", { class: "hero-chip", text: s })))
```

Three chips, no handler, no href. They read as things you can click on the one page a new user arrives at with nothing — which is the same defect shape as #2437's "+52 more need attention": a promise rendered as a caption. Notebooks already solved this for themselves (`NOTEBOOK_TEMPLATES` + `#/notebook/new/{template}`, #1563 D7); dashboards never got the equivalent, so the "New from template" menu is notebooks-only and the dashboard half of the feature has no starting point at all.

The fix should be **templates, not seeded rows**. Seeding curated views as data would mean a migration rung, a `StorageVersion` bump, four pinned test files and a viewer probe sentinel — for content that is not schema — and a seeded row resurrects itself on the next upgrade after the user deletes it, with no reset path if a `can_edit` seat edits one. A template is created only when someone asks for it, and what they get is an ordinary view they own.

One wrinkle worth stating: a v1 read panel's `server` param is **static** — `renderView` only threads variables and range into *composed* panels — so a dashboard template has to be created against a chosen server, unlike a notebook template which can link straight to the composer and re-scope live. That argues for a server picker beside the dashboard templates rather than a bare link.

erikdarlingdata:
Closes #2480.

No custom view ships seeded, so the first thing a new user sees at `#/views` is an empty page. The hero (#1563 M4) sells the feature well and then hands over two buttons onto a blank canvas over an 82-read catalog. It also ended with three chips reading "Top waits by server", "CPU trend over time" and "Slowest procedures by database" — `<span>`s, no handler, no href. That is #2437's defect shape (a promise rendered as a caption) on the one page nobody arrives at with context.

## What lands

Five starter dashboards in a new `wwwroot/js/view-templates.js`:

| Template | Panels |
|---|---|
| Server health at a glance | summary, CPU, memory trend, wait stats, collection health |
| CPU investigation | utilization, scheduler pressure, top queries, top procedures |
| Blocking and deadlocks | both trends, blocking chains, deadlocks |
| Memory pressure | memory split, trend, clerks, grants, plan-cache bloat |
| Configuration review | audit, sp_configure, database options, trace flags, week of changes |

The hero's chips are now those templates, rendered *from* the registry rather than from a second hardcoded list, and they also join the existing "New from template" menu beside the notebook seeds — one affordance, two groups, not a new one bolted on.

## Templates, not seeded rows

Seeding curated views as data would need a migration rung, a `StorageVersion` bump, four pinned test files and a viewer probe sentinel — for content that is not schema. Worse, a seeded row **resurrects itself** on the next upgrade after the user deletes it, and there is no reset path if an editing seat breaks one. A template is created only when someone asks for it, and what lands is an ordinary view the user owns, edits, and can delete for good.

The trade-off is worth stating: a stored definition is validated once at write time and never again, so a definition naming a later-renamed read becomes a permanently broken panel that nothing greps. These templates being **code** is exactly what makes that fixable — see the pins below.

## One wrinkle that shaped the design

A v1 read panel's `server` param is **static**: `renderView` threads variables and range into *composed* panels only. So a dashboard template cannot re-scope itself the way a notebook template can, and the menu carries a server picker rather than a bare link. The chosen server also goes into the view's **name**, so two servers' copies of one template do not collide on the unique-name constraint and turn the second click into a 409.

## Degrading out loud

- A fleet with no servers gets a sentence saying the dashboards are scoped to a server and appear once the fleet has one — not an empty `<select>`.
- The fleet read is allowed to fail: the notebook templates and every other affordance keep working, and the dashboard group says why it cannot offer itself.
- A refused create surfaces the backend's message **verbatim** — it re-validates as the authority, and flattening its reply to "could not create" would hide the one line naming the bad panel. A 409 gets its own sentence, since a second click of the same template is the likely failure.

## Verification

No JS test runner here, so this follows the #2447 / #2451 idiom.

`ViewTemplatesTests` carries the same invariant the built-in pages do — every read named exists in `BuildReadDispatch()`, every parameter key is one that read actually binds per `CatalogDescriptors`, every viz is in `KnownVizList` — plus the structural rules `ValidateDefinition` enforces (no raw `path`, span ∈ {1,2}, no non-hex series colour), unique keys, and that the hero renders from the registry rather than a second list. Both sides come from the shipped artifacts.

Behaviour was verified two ways. Every template was rendered through the shipped `renderPanel` under a minimal DOM shim across two response shapes (empty envelope, error) — zero throws, 21 distinct requests, every one checked against `CatalogDescriptors`. And every definition was dumped from the shipped module and fed to the service's own `DarlingWebEndpoints.ValidateDefinition` — the authority that would refuse the POST. All five come back storable.

Three negative controls, all run and each restored afterwards: a bogus read name fails the dispatch pin; `limit` swapped for `top` on `get_top_queries_by_cpu` fails the parameter pin, naming both the key and the allowed set; an unknown viz fails the storability pin.

`Darling.Tests` builds (`EnableWindowsTargeting`) but cannot execute on macOS, so the six facts were compiled into a throwaway `net10.0` console project against the real service build with a minimal xUnit shim — the assertions that ran are the shipped ones, not a retyped copy. All six pass.

---

## Since opening: honest on day one

A starter dashboard is the first screen a UAT tester opens, on the store with the least data it will ever have — a fresh install has collected nothing for its first cycles. A wall of unexplained blank rectangles there is a worse first impression than the feature not existing.

All 19 data panels (12 tables, 7 charts) now carry their own empty-state sentence, pinned as an **exact count** rather than a spot-check: `emptyText` appears nowhere else in the module, so one dropped sentence is one panel that renders blank in front of a new user. Negative control run — removing one fails the pin at 18 of 19.

The chart case is the one that bites without looking like it. `get_blocking_trend` and `get_deadlock_trend` answer an *idle* server with `trend: []` and **no** `{status,message}` envelope, so the panel has real data in hand and falls through to the chart's "Not enough data points to chart yet" — a warming-up message on a server that simply never blocked. Those panels now say an empty trend means none happened. (The renderer change that makes `emptyText` win at exactly zero rows is in #2477; these sentences are inert without it and correct with it, so there is no ordering dependency between the two PRs.)

Two reads are **deliberately absent** rather than merely unused, for the same reason: analysis findings (the pass writes nothing for 24 hours) and Query Store (a target with it switched off has nothing, ever). Both are on the built-in server page, where the reader arrived looking for that specific thing rather than for a first impression.


erikdarlingdata:
Correct as read, and it names a merge-ordering hazard I created by splitting these into two PRs — so let me state the safe order explicitly rather than leave it in the PR body.

**`vizLine` reading `desc.emptyText` is #2477.** On this branch alone the field is inert, exactly as you describe. The two PRs are independent by *file* (this one never touches `panels.js`) but not by *effect*, and I should have said so on the PR itself rather than only in the body.

**Safe merge order: #2477 first, then this.** The dependency is one-directional and the failure mode is benign — if this merges first, the eight line-panel sentences sit dead and the charts render today's `"Not enough data points to chart yet."`, which is the current behaviour, not a regression. #2477 then activates them with no further change here. There is no ordering in which anything breaks; the only cost of the wrong order is that the improvement is late.

**On dropping the field until it is wired:** I would rather not. It would have to be re-added verbatim by #2477, and a template authored without the sentence is how the sentence never comes back — the panels that need it most (`get_blocking_trend`, `get_deadlock_trend`, the two that answer an idle server with `trend: []` and no envelope at all) are the ones a later pass is least likely to revisit. Inert-and-correct beats absent-and-forgotten, given the fix is already written and reviewed on the other PR.

**On the test only pinning the source string — that is a fair hit and worth being precise about.** `EveryTemplateDataPanel_ExplainsItsOwnEmptyState` is a *content* pin: it asserts every data panel carries a sentence, as an exact count so one dropped sentence fails. It deliberately does not exercise the render path, because on this branch there is no render path to exercise. That proof lives in #2477, where the renderer change is, and it is a real one rather than another text scan: the shipped `vizLine` was driven directly under a DOM shim across all three cases — zero points with an `emptyText` renders the descriptor's sentence, one point still renders the chart's own (there it is the true statement), and zero points *without* an `emptyText` still falls through, so every stored view authored before it is byte-identical.

PR #2486, opened 2026-08-21
Say which gate turned a request away, without letting a scanner fill the log (#2479)

erikdarlingdata:
Item **5** of #2479.

## The gap

All five refusal sites across the two listeners were a bare `StatusCode = …; return;`:

| Host | Gate | Status | Logged before |
|---|---|---|---|
| MCP | Host header allowlist (DNS-rebinding guard) | 400 | nothing |
| MCP | bearer token | 401 | nothing |
| MCP | source CIDR (`mcp.network.allowFrom`) | 403 | nothing |
| Web | Host header allowlist | 400 | nothing |
| Web | source CIDR (`web.network.allowFrom`) | 403 | nothing |

So *"is my token wrong, or my CIDR wrong?"* was answerable only from the client — which sees one opaque status code and cannot tell a Host-allowlist 400 from a malformed request, or a CIDR 403 from anything else. The person who can fix either is on the server, and the server said nothing.

Each gate now emits a WARN naming itself and the config key that governs it.

**The web host's `ShowLogin` path is included too**, even though it answers **200** rather than 401 — but only when a token was actually *presented and did not match*. That is precisely the state an operator asks about when a `?token=` they pasted did not work, and the status code hides it. A first visit with no token is the normal route to the login page, so logging that would make every bookmark a warning.

Out of scope on purpose: `DarlingWebEndpoints`' validation 400s and the CSRF 415 already return a body that explains itself to the client. This is about the gates that refuse *silently, before any handler*.

## The rate limit, and why it is shaped this way

This is the load-bearing half, so it is measured rather than asserted about.

- **Key is (gate, source address), and a first sighting always logs immediately.** A limiter that makes the tester wait out a window before their own answer appears has not shipped anything. Only *repeats* fold — 10 minutes — which costs nothing, because a second identical refusal from the same address through the same gate carries no information the first did not. Fix the token and now fail the CIDR check, and that is a different key, which logs at once.
- **Per source, not per gate alone.** Per-gate is a smaller key set and it is wrong: one scanner's line would win the window and hide the tester's refusal completely — this defect reintroduced by the fix for it.
- **Which makes the map unbounded, so there is a cap.** Past 16 distinct sources in a window, further sources fold into one per-gate aggregate that **says so** ("speaking for several sources — the per-source log budget is full, which usually means this port is being scanned"). Bounds memory and volume to `cap + gate count` entries and one line per gate per window.
- **The accepted cost, stated rather than discovered:** under an active scan the per-source slots may be taken and a tester arriving mid-scan lands in the aggregate. That is visible rather than silent, and if the port is being scanned, that is the more urgent thing for the operator to know.
- **Eviction is keyed on last SEEN, not last logged.** Keyed on last logged, a source still being refused every second would be dropped at the two-window horizon and re-log as a fresh first sighting — turning the rate limit into a rate *multiplier*, which is worse than no limit because it looks like one is working.
- **The budget belongs to the started server, not the process**, so a rebind starts clean rather than inheriting the previous listener's scan and swallowing the new one's first refusals.

**Measured ceiling: 108,000 refused requests over an hour (30/s across three gates from a few hundred sources) produce 100 log lines and 19 tracked entries.**

## What is never logged

The token, any prefix of it, and its length. A line may say whether a credential was **presented** — an absent `Authorization` header is a client that was never configured; a wrong one is a token that does not match `mcp.network.encryptedToken`, and each points at a different next step — and nothing about its value. Pinned per call site rather than trusted.

The Host header is the only request-derived value echoed, and it goes through a sanitizer first. A log file is a text file: a header carrying `\r\n` would otherwise forge whole log entries.

## Verification

- **Ran the shipped `DarlingHttpRefusalLog.cs`** — `<Compile Include>`'d into a throwaway `net10.0` console harness, since `Darling.Tests` is `net10.0-windows` and cannot run on macOS. **26/26 assertions pass**, including the 108,000-request scan ceiling, the first-sighting-is-immediate guarantee, the noisy-neighbour case, the cap announcement, the last-seen eviction invariant, and the log-injection cases.
- **Simulated both structural pins against the real host files**: MCP has 3 of 3 4xx sites reported, Web 2 of 2, 6 `Report` calls total, no forbidden token identifier in any argument list.
- **Re-checked the existing `HostHeaderGuardTests` assertions against the edited files** — `FirstMiddlewareAfterBuild` still finds `IsAllowedHost(context.Request.Host.Host` and `Status400BadRequest` in the first middleware, and the guard still precedes `if (networkMode)` and `MapMcp`. Adding lines inside that lambda was the one thing that could have broken them.
- Service and `Darling.Tests` both build clean, 0 errors.
- Against `dev` the whole test class fails to **compile** — `DarlingHttpRefusalLog` does not exist there.
- **Not verified:** no request has been pushed through a live Kestrel host, so the exact rendered line and the interaction with the file/Event Log sinks are unexercised. Manual check on a box: with `mcp.network` exposed, `curl -H "Authorization: Bearer wrong" http://<ip>:5152/` should produce one WARN naming the access-token gate, and a second identical curl within ten minutes should produce none.

## Onboarding

Makes `docs/uat-onboarding.md` §4.4 *"When it says enabled and still will not connect"* substantially unnecessary — the section exists because the server-side answer had to be reconstructed from the client's status code. The tell it teaches (401 = token, 403 = CIDR, 400 = Host) is now stated in the service log at the moment it happens, naming the config key. §5's *"the things that will actually bite you"* entry on the same subject can go the same way. The gate list is still worth keeping as a reference table; the "you can only tell from the client" framing is what stops being true.

### DIFF
--- a/Darling/Darling.Tests/DarlingHttpRefusalLogTests.cs
+++ b/Darling/Darling.Tests/DarlingHttpRefusalLogTests.cs
@@ -346,6 +346,23 @@ public class DarlingHttpRefusalLogTests
     }
 
     /// <summary>
+    /// The verb agrees with the status code.
+    ///
+    /// <para>Review catch on #2479: not every gate rejection ends in a 4xx. The web dashboard answers an
+    /// in-CIDR request with a WRONG <c>?token=</c> using a 200 and the login page, deliberately — and
+    /// "refused a request … with 200" is self-contradictory on its face, misleading anyone reading the log
+    /// cold or filtering for denials on status.</para>
+    /// </summary>
+    [Theory]
+    [InlineData(400, "refused")]
+    [InlineData(401, "refused")]
+    [InlineData(403, "refused")]
+    [InlineData(200, "did not authorize")]
+    [InlineData(302, "did not authorize")]
+    public void TheOutcomeVerb_AgreesWithTheStatusCode(int statusCode, string expected) =>
+        Assert.Equal(expected, DarlingHttpRefusalLog.DescribeOutcome(statusCode));
+
+    /// <summary>
     /// Every refusal site on both hosts reports. This is the half that decays: a gate added later is a
     /// gate that refuses silently again, and no behavioural test can see one that does not exist yet.
     /// </summary>
--- a/Darling/Darling.Tests/DarlingHttpRefusalLogTests.cs
+++ b/Darling/Darling.Tests/DarlingHttpRefusalLogTests.cs
@@ -1,0 +1,385 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Runtime.CompilerServices;
+using PerformanceMonitor.Darling.Service.Hosting;
+using Xunit;
+
+namespace Darling.Tests;
+
+/// <summary>
+/// Rejected HTTP requests get a rate-limited WARN naming the gate that refused them (#2479, item 5).
+///
+/// <para><b>The gap.</b> All five refusal sites across the MCP and web listeners were a bare
+/// <c>StatusCode = …; return;</c>. So "is my token wrong, or my CIDR wrong?" was answerable only from the
+/// client, which sees one opaque status code and cannot tell a Host-allowlist 400 from a malformed
+/// request, or a CIDR 403 from anything else — while the person who can fix either is on the server, and
+/// the server said nothing.</para>
+///
+/// <para><b>Why the policy is tested this hard.</b> The rate limit is not a nicety here. These ports are
+/// LAN-exposed on purpose, so an unthrottled line per refusal makes the service log a denial-of-service
+/// target: a scanner fills the file an operator debugs from, and the log stops being readable exactly when
+/// it matters. Equally, a limiter tuned the other way is worthless — if a tester's own first refusal waits
+/// out a window, the feature has not shipped. Both halves are measured below rather than asserted about.</para>
+/// </summary>
+public class DarlingHttpRefusalLogTests
+{
+    private static readonly DateTime T0 = new(2026, 8, 21, 12, 0, 0, DateTimeKind.Utc);
+    private static readonly TimeSpan Window = TimeSpan.FromMinutes(10);
+
+    /// <summary>
+    /// The tester's own first refusal is never delayed. A limiter that makes the operator wait out a
+    /// window before their answer appears has not fixed anything.
+    /// </summary>
+    [Fact]
+    public void FirstSighting_LogsImmediately_ForEveryGate()
+    {
+        var log = new DarlingHttpRefusalLog(Window, 16);
+
+        foreach (DarlingRefusalGate gate in Enum.GetValues<DarlingRefusalGate>())
+        {
+            Assert.True(log.Observe(gate, "10.0.0.5", T0).Log, $"the first {gate} refusal must be logged at once");
+        }
+    }
+
+    /// <summary>
+    /// Repeats fold, and the count rides out on the next line that does emit — so the volume is visible
+    /// without being written. And the count resets once reported, or every later line over-reports.
+    /// </summary>
+    [Fact]
+    public void RepeatsFold_AndTheSuppressedCountRidesTheNextLine()
+    {
+        var log = new DarlingHttpRefusalLog(Window, 16);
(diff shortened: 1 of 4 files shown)


erikdarlingdata:
Good catch — fixed in 2a857dc1 by hoisting the header to a local, so `StringValues.ToString()` runs once and both readers share it.

I did not take the fuller version (having `IsBearerTokenAuthorized` return what it parsed): its signature is pinned by `DarlingMcpHostTests.IsBearerTokenAuthorized_MatchesOnlyExactBearerToken` and `DarlingHostBindingTests`, and threading a result type through it to save one parse on an already-refused request is not a trade worth making. The comment now says that, so the next reader does not re-open the question.

erikdarlingdata:
Right, and the analysis is exact — the aggregate bucket folds *many different addresses* through one gate, so "from this source" is false there, and false in the direction that costs most: it sends the operator looking for one busy client when what is happening is a broad scan, which is the very thing the aggregate bucket exists to tell them.

Fixed in 84086e43. The clause now branches on `Decision.Aggregated`:

- aggregate: `399 further refusal(s) through this gate, from other sources, were not logged in the last 10 minute(s).`
- per-source: `7 further refusal(s) from this source were not logged in the last 10 minute(s).`

Your point about the coverage gap was the useful half — `PastTheCap_TheLineSaysItSpeaksForSeveralSources` only exercises the *first* overflow, before any folding, so the clause that ships with the folded count was never rendered by a test. The new `TheAggregateSuppressionClause_DoesNotClaimTheyCameFromOneSource` fills the budget, pushes 400 distinct overflow addresses through one gate, crosses the window, and asserts both wordings — including that a per-source entry still says "from this source", so the fix cannot flatten both into one hedge. Rendered output verified against the shipped class.

erikdarlingdata:
Correct, and it is the same defect this PR exists to fix, sitting inside the fix. `ExtractBearerToken` returns null for *both* "no header" and "a header that is not a well-formed Bearer", so the line told an operator nothing was presented while their client was sending `Authorization: Basic …` every second.

Fixed in 4bae26dc. Three states now, because each is a different next step:

| header | message |
|---|---|
| absent / blank | `no 'Authorization: Bearer <token>' header was presented` |
| `Basic …`, a bare token, `Bearer ` with an empty or blank token | `an Authorization header WAS presented but is not a 'Bearer <token>' (wrong scheme, or an empty token after 'Bearer')` |
| well-formed Bearer that misses | `the presented bearer token does not match mcp.network.encryptedToken` |

Pulled out into a pure `DescribeBearerRefusal` so all three are pinned by a table rather than by reading a lambda — and so the "never the value" property becomes structural: it reads the header's *shape* and has nothing else to leak. The table covers the wrong scheme, a bare token with no scheme, `Bearer ` with empty and blank tokens, and the case-insensitive scheme forms that must still reach the third state. Verified against the shipped method, 14/14.

erikdarlingdata:
Agreed — and it is worse than a nit, because a log line that says "refused … with 200" argues with itself and would mislead anyone filtering for denials on status. The reasoning for including the 200 case was in the PR description, which is not where the person reading the log is standing.

Fixed in df057b6b. The verb is derived from the status and the code is stated as `HTTP N`:

```
MCP refused a request from 10.0.0.5 (HTTP 401): the access token gate rejected it — ...
Web dashboard did not authorize a request from 10.0.0.5 (HTTP 200): the access token gate rejected it — the presented ?token= does not match web.network.encryptedToken, so the login page was served instead.
```

Kept as a pure `DescribeOutcome` so the boundary is pinned rather than asserted about, and so the structured template keeps its named placeholders instead of being concatenated into one opaque string.

erikdarlingdata:
Both fixed in 3041b739, and both were narrower statements than they sounded.

**The sentence.** You are right that `DescribeOutcome` was only half the fix — the middle clause still said the gate "rejected it", so the line contradicted itself two clauses later. The gate is now stated as the *reason* rather than as a second verdict, which is accurate at every status:

```
Web dashboard did not authorize a request from 10.0.0.5 (HTTP 200), because of the
access token gate — the presented ?token= does not match web.network.encryptedToken,
so the login page was served instead.
```

Your observation that nothing tested the *full* message is the part that mattered — that is exactly how the contradiction survived the first fix. `TheRenderedLine_NeverContradictsItsOwnStatusCode` now renders through a real `ILogger` and asserts no clause anywhere in a sub-400 line calls it a rejection or a refusal.

**The budget.** Also right, and the doc overstated it by ~3x. Renamed to `DefaultMaxTrackedKeys` and sized at `16 × gate count`, so the budget an operator actually cares about — distinct *addresses* — is the 16 the design intends rather than the ~5 a per-key cap of 16 delivered. `SixteenDistinctAddresses_AllGetTheirOwnLine_EvenTrippingEveryGate` drives 16 addresses through all three gates and asserts not one falls into the aggregate.

Re-measured the ceiling with the larger cap: **108,000 refused requests over an hour now produce 258 lines and 51 tracked keys**, against 100 and 19 before. Still a hard ceiling.

Issue #2414, opened 2026-08-21
The firewall verbs name the rule from darling.json's port while the endpoint binds the store's


erikdarlingdata:
Found while working #2389, which fixed the *diagnostic* for the same split. This one is a functional consequence and deserves its own lane.

`mcp.port` / `web.port` are control-plane-authoritative once the store is seeded — the supervisor resolves `config_service.mcp_port` and binds Kestrel there. But every firewall surface derives the rule name and the port from **darling.json**:

- `DarlingCliCommands.ReconcileEndpointFirewallAsync` (the firewall half of `--enable-mcp` / `--disable-mcp` / `--enable-web` / `--disable-web`) — `McpFirewallRuleName(config.Mcp.Port)`, `config.Mcp.Port`
- `--configure-firewall`'s plan builder — same two, for MCP and web alike

The runtime side does not. `DarlingMcpHostService` calls `CheckMcpFirewallAsync(effectivePort, …)`, where `effectivePort` is the store's port.

## The failure

Toggle the MCP port in the Viewer's Settings (5152 → 5199) on a box with an `mcp.network` block:

- the endpoint rebinds to 5199 within one poll interval, correctly, no restart
- the service then looks for `PerformanceMonitor Darling MCP (port 5199)` and reports it missing
- an admin runs `--configure-firewall` as instructed, and it creates/keeps `PerformanceMonitor Darling MCP (port 5152)` for port 5152 — the wrong rule, from the stale file value, and it will keep creating that one however many times it is run
- the exposed endpoint has no firewall path; LAN clients are blocked, loopback still works, so it reads as a client or CIDR problem

The stale rule for 5152 also stays open on a port nothing is listening on, which is the opposite of what the scoping is for.

Same shape for the web dashboard.

## Why it survived

Nothing agrees or disagrees until someone changes the port in the store, which is rare — and when it happens the symptom (LAN clients can't connect, loopback can) points at the CIDR or the token, not at a rule name. #2389 now warns when the two ports disagree, so the precondition is at least visible, but the elevated verb still writes the wrong rule.

## Options

[...]

PR #2487, opened 2026-08-21
Reprint a lost endpoint token, and say when a network block can change (#2479)

erikdarlingdata:
Items **2 and 6** of #2479. Both come down to the service knowing something and not saying it.

---

## Item 2 — `--print-mcp-token` / `--print-web-token`

`--configure-network` shows each generated token's plaintext exactly once. Lose it and the only path was regeneration, which invalidates every client already configured against it — so one mislaid token during UAT meant re-onboarding every consumer of that endpoint. An unrecoverable mistake made out of a recoverable one.

### The security reasoning, checked — and it is stronger than the issue states

The issue's claim was *"the token is already DPAPI-decryptable in-process on that box, so an admin on the host gains nothing they could not already get."* That holds, and the real position is broader:

- The token is a `DarlingSecrets` blob: **DPAPI at `DataProtectionScope.LocalMachine`**, entropy `"PerformanceMonitor.Darling.v1"` — a constant compiled into an **open-source** binary (`DarlingSecrets.cs:27`). LocalMachine scope means *any process on the box* can `Unprotect` it, and the entropy is not a secret because it is published in this repository. `DarlingCliCommands.cs:2191-2194` already says as much: read access to `darling.json` **is** the secret.
- **`darling.json` grants `INTERACTIVE` read deliberately** — `DarlingWorker.cs:737` and `DarlingFileSecurity.HardenFile(path, allowInteractiveRead: true)`, mirrored by `install-darling.ps1` — because the Viewer and these very CLI verbs are run by the interactive operator. `Darling/README.md`'s principal model states it outright: *"`INTERACTIVE` == the operator, so the admin/viewer credentials are readable by the Viewer with zero configuration."*

So an ordinary logged-on **non-administrator** can already read and decrypt the blob in about four lines of PowerShell. Shipping the verb discloses nothing new to anyone.

**Which means the elevation gate is not a confidentiality boundary, and the code says so rather than implying one.** It is still worth having for what it actually does — consistency with the other verbs carrying `(run elevated)`, and making a reprint a deliberate act rather than something a script or a shared shell picks up in passing — and the refusal message spends five lines saying that the token's real protection is the file's ACL, with a pointer at `--harden-files`. An operator who believes elevation is protecting the token stops guarding the thing that is.

Worth flagging for the record: `--print-viewer-connection` prints a live **database password** to stdout and is *not* elevation-gated. That asymmetry is now real. I have not changed it — it would break existing runbooks and is outside this issue — but it is the obvious next question.

### Both verbs, not one

The issue names *"an MCP or web token"*. They have the identical loss problem, the identical DPAPI shape and adjacent config blocks; shipping one would leave half a fix behind.

### Emission

Follows `--print-viewer-connection`'s posture exactly (D8): **stdout is the token and nothing else**, so `> token.txt` and `| clip` both yield a pasteable value, and every warning is on **stderr ahead of the payload** so a redirect cannot swallow it. The warning also separates the two cases that look alike: reprinting recovers a **mislaid** token and is *not* the answer to a **leaked** one, where `--configure-network` generates a new one and every configured client must be updated.

Elevation is passed **in** rather than measured inside the shared body, the way the pure alert gates take `nowUtc`. That makes the refusal path drivable on any runner — a refusal only exercised when the CI agent happens to be unprivileged is a refusal nobody has checked.

---

## Item 6 — saying when a network block can change

`mcp.network` / `web.network` are loaded once and held for the process lifetime **by design**: exposure should require touching the host, and a LocalMachine-scoped DPAPI token could not live in the store anyway. **No hot reload here** — that design is right and is not changing.

But it produces "I enabled it and it is still loopback-bound" in another costume, because nothing said the value froze the moment the process started. #2389/#2411 fixed the seed-versus-store confusion by making the split **visible** rather than by removing it, and #2414 did the same one layer down for the firewall port. This is that treatment applied to the third face of the same object.

**The specific gap:** the network-mode start line already carried `listen/allowFrom/token from darling.json mcp.network (file-only, restart-only)`. The **loopback** line carried nothing about the block at all — and loopback-when-you-expected-LAN is precisely the state an operator is diagnosing. So the new line is emitted in **both** modes.

Three states, three consequences, only the true one emitted:

> `MCP network exposure: mcp.network was read from darling.json at start-up and is FIXED for the lifetime of this process - LAN-bound on 10.197.53.214 for 10.197.0.0/16, behind the configured token, until the service is RESTARTED. Editing darling.json now changes nothing until then. The control plane still owns whether this endpoint runs and on which port (config.config_service.mcp_enabled / mcp_port, moved by --enable-mcp / --disable-mcp or the Viewer's Settings, and applied without a restart) - it simply cannot move where this binds, who may reach it, or what token it requires.`

> `MCP network exposure: darling.json DEFINES mcp.network, but this endpoint is bound loopback-only. That block was read at start-up and is FIXED for the lifetime of this process, so editing darling.json now changes nothing until the service is RESTARTED. If you just edited it to expose this endpoint, restart the service; if you already restarted, the mcp.network line logged above says why the exposure was refused. […]`

> `MCP network exposure: darling.json has no mcp.network block, so this endpoint is loopback-only. Adding one is a FILE edit that takes effect on the next service RESTART; there is no store setting and no Viewer control that can expose it. […]`

Only the trap state gets a **call to action**. An endpoint already serving the LAN exactly as configured is the likeliest state, and telling that operator to restart is where a start-up line spends its credibility — the same reasoning that made `DescribeToggleOverride`'s consequence clause state-specific. And all three name the half the control plane **does** own without a restart, or "cannot change until restart" reads as "this endpoint is immovable" and sends someone to restart the service for something `--enable-mcp` already does live.

---

## Verification

- **Ran both new pure surfaces against the real service assembly** — a throwaway `net10.0` console with a `ProjectReference` to `PerformanceMonitor.Darling.Service`, borrowing the friend-assembly name so it reaches the same internals `Darling.Tests` does. **28/28 checks pass**, covering all five `PrintEndpointToken` branches (not elevated / no block / empty token / decrypt failure / success) and all three lifetime states. The critical invariants: **stdout is empty on every refusal path**, and the token appears on stderr in none of them.
- Service and `Darling.Tests` both build clean, 0 errors.
- **Checked the pins the new verbs could break:** `IsKnownVerb_ReachesEveryVerbClassifierOnTheClass` derives `--print-mcp-token` / `--print-web-token` correctly from the classifier names via `VerbLiteralFor`, and both new help lines are pure ASCII for `UsageText_ListsTheKeyVerbs_AndIsAscii`.
- Against `dev` the whole test file fails to **compile** — none of these members exist there.
- **Not verified:** nothing has run on a real Windows host, so the actual elevated round-trip through a DPAPI blob and the rendered start-up log lines are unexercised. Manual checks: `--print-mcp-token` from a non-elevated shell should refuse with empty stdout; from an elevated one, `... --print-mcp-token > t.txt` should leave exactly the token in `t.txt` with the warning on the terminal. Restarting the service should show one `MCP network exposure:` line whichever mode it comes up in.
- **Deliberately not done:** `postgres.network` is restart-only in the same way and gets no equivalent line — the issue names `mcp.network` and `web.network`, and the store already logs `Fix postgres.network and restart to expose the store` on a degrade. Worth a follow-up if the store's silent-when-healthy case turns out to bite the same way.

## Onboarding

- **Item 2** removes the need for `docs/uat-onboarding.md` §4.2's warning that the MCP token is shown once and is otherwise unrecoverable, and the matching line in §3.2 for the web token. The "save it now" advice is still good; the "or else re-onboard every client" consequence stops being true for a token that is merely mislaid.
- **Item 6** makes §4.4's *"When it says enabled and still will not connect"* and §3.5's restart caveat unnecessary as *instructions* — the service now says it at every start, in the mode where it matters. §0's "the file is a first-run seed / the network block is the opposite" framing is still worth keeping; it is the conceptual model, and the log line is the runtime reminder of it.

### DIFF
--- a/Darling/Darling.Tests/DarlingPrintEndpointTokenTests.cs
+++ b/Darling/Darling.Tests/DarlingPrintEndpointTokenTests.cs
@@ -239,8 +239,8 @@ public class DarlingPrintEndpointTokenTests
     /// <summary>Both verbs are dispatched from Program.cs behind the Windows guard the DPAPI material
     /// requires — a verb that is classified but never dispatched is #1581 the other way round.</summary>
     [Theory]
-    [InlineData("IsPrintMcpTokenVerb", "PrintMcpTokenAsync", "--print-mcp-token requires Windows")]
-    [InlineData("IsPrintWebTokenVerb", "PrintWebTokenAsync", "--print-web-token requires Windows")]
+    [InlineData("IsPrintMcpTokenVerb", "PrintMcpToken", "--print-mcp-token requires Windows")]
+    [InlineData("IsPrintWebTokenVerb", "PrintWebToken", "--print-web-token requires Windows")]
     public void ProgramDispatchesBothVerbs_BehindTheWindowsGuard(string classifier, string entryPoint, string guard)
     {
         var source = ReadSource(Path.Combine("Darling", "PerformanceMonitor.Darling.Service", "Program.cs"));
--- a/Darling/Darling.Tests/DarlingPrintEndpointTokenTests.cs
+++ b/Darling/Darling.Tests/DarlingPrintEndpointTokenTests.cs
@@ -1,0 +1,355 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.IO;
+using System.Runtime.CompilerServices;
+using PerformanceMonitor.Darling.Service;
+using PerformanceMonitor.Darling.Service.Hosting;
+using Xunit;
+
+namespace Darling.Tests;
+
+/// <summary>
+/// Reprinting an endpoint's access token (#2479, item 2).
+///
+/// <para><b>The gap.</b> <c>--configure-network</c> shows each generated token's plaintext exactly once.
+/// Lose it and the only path was regeneration, which invalidates every client already configured against
+/// it — so one mislaid token during UAT meant re-onboarding every consumer of that endpoint. An
+/// unrecoverable mistake made out of a recoverable one.</para>
+///
+/// <para><b>What it discloses, and what the elevation gate is not.</b> Nothing new. The token is a
+/// <c>DarlingSecrets</c> blob: DPAPI at <c>LocalMachine</c> scope, with an entropy constant compiled into
+/// an open-source binary — so any process on the box can decrypt it, and the entropy is not a secret
+/// because it is published. And <c>darling.json</c> deliberately grants INTERACTIVE read, because the
+/// Viewer and these very CLI verbs are run by the interactive operator. So an ordinary logged-on
+/// NON-administrator can already read and decrypt the blob in four lines of PowerShell. The gate is worth
+/// having for what it actually does — consistency with the other "(run elevated)" endpoint verbs, and
+/// making a reprint a deliberate act rather than something a script picks up in passing — and the refusal
+/// message says so rather than implying a boundary that is not there. The token's protection is the file's
+/// ACL, and it always was.</para>
+/// </summary>
+public class DarlingPrintEndpointTokenTests
+{
+    [Theory]
+    [InlineData("--print-mcp-token", true)]
+    [InlineData("--PRINT-MCP-TOKEN", true)]
+    [InlineData("--print-web-token", false)]
+    [InlineData("--print-viewer-connection", false)]
+    [InlineData("--nonsense", false)]
+    public void IsPrintMcpTokenVerb_RecognizesOnlyItsOwnVerb_CaseInsensitively(string arg, bool expected) =>
+        Assert.Equal(expected, DarlingCliCommands.IsPrintMcpTokenVerb(arg));
+
+    [Theory]
+    [InlineData("--print-web-token", true)]
+    [InlineData("--PRINT-WEB-TOKEN", true)]
+    [InlineData("--print-mcp-token", false)]
+    [InlineData("--nonsense", false)]
+    public void IsPrintWebTokenVerb_RecognizesOnlyItsOwnVerb_CaseInsensitively(string arg, bool expected) =>
+        Assert.Equal(expected, DarlingCliCommands.IsPrintWebTokenVerb(arg));
+
+    /// <summary>The #1581 trap: a verb with a dispatch block and help text that <c>IsKnownVerb</c> never
+    /// learned classifies as an unknown option, and its dispatch is unreachable.</summary>
+    [Fact]
+    public void IsKnownVerb_ReachesBothTokenVerbs()
+    {
+        Assert.True(DarlingCliCommands.IsKnownVerb("--print-mcp-token"));
+        Assert.True(DarlingCliCommands.IsKnownVerb("--print-web-token"));
+    }
+
+    /// <summary>Both are listed, and both carry the elevated marker the other endpoint verbs use — an
+    /// operator should not have to discover the requirement by being refused.</summary>
+    [Fact]
+    public void UsageText_ListsBothVerbs_AndMarksThemElevated()
+    {
+        var usage = DarlingCliCommands.UsageText();
+
+        foreach (var verb in new[] { "--print-mcp-token", "--print-web-token" })
+        {
+            var at = usage.IndexOf(verb, StringComparison.Ordinal);
+            Assert.True(at >= 0, $"{verb} is not in the help text");
+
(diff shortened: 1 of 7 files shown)


erikdarlingdata:
Both correct, and this is a blocking bug rather than a nit — thank you. Fixed in 59f61f9a.

`config.Mcp.Network` / `config.Web.Network` are nullable and are `null` on the **default secure config**, which is exactly the state the new line exists to describe. The dereferences at 313/317 are safe because they sit inside `if (networkMode)`, where the bind resolution has already proven a block exists; mine runs unconditionally. So it would have thrown on every start of an un-exposed server, been swallowed by the catch below, and retry-failed forever because `config` never changes between attempts — the feature that explains *why* an endpoint is loopback-only would have been the thing keeping it from starting.

Two things worth recording:

- It is the shape this repo keeps finding — a call site that never learned a concept the code around it already knew. The CLI half of *this same PR* guards correctly (`network is not null && network.IsConfigured`), so it was written right once and wrong once, three files apart.
- **The compiler said so and I did not read it.** The buggy form emits `CS8602`; the build filter I was using only looked for `error`. Verified by rebuilding the pre-fix text — `CS8602` on line 765, gone after the fix.

Pinned per host with `TheLifetimeReport_ReadsTheNetworkBlockNullSafely`, since no unit test can reach a Kestrel start and the next unconditional read of a nullable config block should fail by name rather than in the field.

erikdarlingdata:
Right on both counts, including that the suite only exercised the `CryptographicException` branch — which is how the other one shipped unexamined. Fixed in 6267081e.

The DPAPI advice is now gated on the exception that earns it. The general arm reports only what it knows:

```
Could not read the MCP bearer token: mcp.network.token references environment
variable 'DARLING_MCP_TOKEN', which is not set (or empty).
That is what mcp.network resolved to. Fix the source it names (an env: reference
needs the variable set in THIS shell; a file: reference needs a path this account
can read), then re-run - the token itself is fine and does not need regenerating.
```

The "destructive fix" framing in your comment is the part that made this worth doing properly rather than just softening the wording: `--configure-network` invalidates every client configured against the token, and the token was never the problem. `Elevated_WhenAnEnvOrFileReferenceFails_DoesNotBlameDpapi_OrSuggestRegenerating` now asserts the real cause survives while `DPAPI` and `--configure-network` appear nowhere, and the DPAPI arm still says both. Both arms still print nothing to stdout.

erikdarlingdata:
Correct — renamed to `PrintMcpToken` / `PrintWebToken` in the follow-up commit. They read one file and decrypt one blob, with nothing to yield on, so they follow `HardenFiles` rather than `PrintViewerConnectionAsync` directly above them.

Added a line saying which of the two neighbours they follow and why, so the next reader does not restore the suffix for symmetry with the method above it.

erikdarlingdata:
Both fixed in 57776357, and the first one is this issue's own defect committed by the fix for it — thank you for catching it.

**The promised line.** Exactly as you describe: `IsConfigured` is true if *any* of listen/allowFrom/token is set, but a block whose `listen` is absent or loopback resolves to `LoopbackByDefault`, whose severity maps to null, so nothing is logged. Setting `allowFrom` and a token and leaving `listen` unset is the partial edit this feature exists to catch, so the message was wrong in the case it was built for.

I took something close to your option (b): the hosts pass `DarlingNetwork.IsExposedListenAddress(config.Mcp.Network?.Listen)` — the fact the branch actually needs, already available at the call site — rather than plumbing `BindReason` through. Four states now:

| state | what it says |
|---|---|
| no block | loopback-only; adding one is a file edit that takes effect on the next restart |
| block, no LAN `listen` | *"no exposure was requested and none was refused - there is no further log line about it. Set mcp.network.listen to this host's LAN address ... and RESTART"* |
| block, exposure refused | *"...so it was REFUSED ... the mcp.network line logged above says why"* |
| exposed | LAN-bound on `<listen>` for `<cidr>`, fixed until restart |

Pinned per host with `EachHost_TellsTheReportWhetherExposureWasEvenAskedFor`, so a future call site cannot drop the argument, and `ABlockWithNoLanListen_NamesTheMissingField_InsteadOfPromisingALogLine` asserts the no-listen state contains neither "logged above" nor "why the exposure was refused".

**The corrupted blob.** Also right — `Convert.FromBase64String` runs before `ProtectedData.Unprotect`, so a hand-edited `encryptedToken` raises `FormatException` and fell into the arm that says *"the token itself is fine and does not need regenerating"*, which is the opposite of true. Both now share one arm and one verdict — this token cannot be recovered, regenerate it, and that invalidates every client configured against it — while the `env:`/`file:` arm keeps saying the token is fine, because for those it is. `Elevated_WhenTheBlobIsCorrupted_SaysItCannotBeRecovered_NotThatItIsFine` covers it.

PR #2491, opened 2026-08-21
Scrub fleet identifiers from main

erikdarlingdata:
Closed — `check-branches` refuses it by design ("pull requests to main are only allowed from dev"), and that gate is worth keeping. The branch is retained in case the rename is easier to graft than to re-derive.

Two things worth keeping from the attempt, both fixed elsewhere:

- A bare-token rename touched 50 files, because `apex` is domain vocabulary here — the root of a blocking chain. Scoped to whole instance names it is 9 files.
- A word boundary failed where a hostname really was: in a C# literal following an escape, the `n` of `\n` is a word character. The #2490 guard had the same flaw and no longer does.


erikdarlingdata:
Closing this — `check-branches` refuses it by design:

> Pull requests to main are only allowed from dev. Please target the dev branch instead.

That gate is release discipline and worth keeping, so this is the wrong instrument. The scrub is already on `dev`; `main` gets it the moment `dev` merges to `main`, and the sanctioned path there is a release.

So this is not a patch decision, it is a release-timing one. Tracking in #2492 rather than leaving a permanently red PR open. The branch is kept in case the content is useful as a cherry-pick during the release.

Worth keeping from the attempt, since both were found here and are fixed elsewhere:

- **A bare-token rename corrupted 50 files.** `apex` is domain vocabulary — the root of a blocking chain. Scoped to the instance-name shape it is 9 files, not 58.
- **A word boundary failed where a hostname really was**: `"servers:\nprod-..."`, where the `n` of the escape is a word character. The #2490 guard had the same flaw and no longer does.

Issue #2485, opened 2026-08-21
Six MCP reads answer 'nothing happened' and 'nothing collected' with identical bytes, and three of them disagree between the SKUs


erikdarlingdata:
The MCP half of a defect just fixed on the web half in #2477.

`get_blocking_trend` serializes `{ server, hours_back, trend }` **unconditionally**. On a server that had no blocking, `trend` is `[]` — the same bytes a server with nothing collected yet produces. There is no `{status, message}` envelope, no count, nothing.

That was visible on the web dashboard because `renderPanel` had real data in hand and fell through to the chart's "Not enough data points to chart yet" — a warming-up message on a server that simply never blocked. #2477 fixes the browser at the renderer.

**It is worse for an agent than it was for a person.** A human sees an empty chart in a page whose other panels are populated and infers the server is fine. An MCP client has only the JSON. An LLM asked "is this server blocking?" can reasonably answer *"there is no data"* when the true answer is *"no, and that is good news"* — and on a monitoring tool that is close to the worst available failure, because it converts a clean bill of health into an apparent gap in coverage.

## Scope of the pattern

Scanned every `[McpServerTool]` in `Darling/PerformanceMonitor.Darling.Service/Mcp` and `Lite/Mcp` for a top-level collection with no early-return on empty.

**Darling — no distinction between "none happened" and "none collected":**

| Tool | Empty shape | Web read? |
|---|---|---|
| `get_blocking_trend` | `trend: []` | yes |
| `get_deadlock_trend` | `trend: []` | yes |
| `get_memory_clerks` | `clerks: []` | yes |
| `get_wait_types` | `wait_types: []` | yes |
| `get_mute_rules` | bare list | yes |
| `compare_analysis` | bare comparison | yes |

Neighbours that already do this correctly, which is what makes the six read as an omission rather than a decision: `get_wait_stats`, `get_memory_stats`, `get_file_io_trend`, `get_memory_trend`, `get_query_duration_trend` and ~50 others all return `McpHelpers.Status("unavailable", "…")`; the three `*_changes` tools go further and return a purpose-built `NoChanges(...)` carrying `hints: { server, snapshot_count }` — which is precisely the "queried successfully, zero rows, here is the denominator" shape the trend tools want; and `list_servers` guards its empty registry with prose. So the vocabulary exists and these did not adopt it.

**Lite — the same six, plus three where the two SKUs disagree:**

[...]

Issue #2484, opened 2026-08-21
Ten viewer surfaces read data that has no /api/read endpoint, so neither the web dashboard nor the MCP can reach it


erikdarlingdata:
Split out of #2475, which named it but is not the place to fix it.

The web dashboard's per-server page now carries twelve sub-tabs reaching **61 of the service's 82 `/api/read/*` handlers** (#2477). Three groups of viewer data remain unreachable from a browser, and only one of them is web work:

- **(a) served but unshown** — the eight `get_pg_*` reads (blocked by the fleet payload carrying no target-engine discriminator), `get_database_scoped_config` (a nested `databases[].settings[]` the table renderer cannot draw), `get_store_metrics` (store-wide, no per-server home). Front-end and payload work, tracked on #2475.
- **(c) desktop-only by design** — the plan viewer, query heatmap rendering, block-chain reconstruction, the interactive deadlock graph, and the grids' filter/export/repro/drill-down affordances. A standing product boundary, now stated in `docs/uat-onboarding.md` §3.4.
- **(b) — this issue.** Data the WPF viewer reads that **has no `/api/read` endpoint at all**. No amount of front-end work reaches it, and it is invisible to the MCP surface too, so an agent asking about it gets nothing.

## The list

Every one of these is a projection over data the store **already holds** — the viewer opens no connection to a monitored server and reads the same Postgres store the MCP tools do, so these are **endpoints, not collectors**. None needs new collection.

| Viewer tab | Reader | Note |
|---|---|---|
| Queries › Query Store Regressions | `GetQueryStoreRegressionsAsync` (`ViewerDataService.QueryStoreRegressions.cs`) | Regression detection over `query_store_stats`. The only #2477 tab that is entirely unreachable rather than reduced. |
| Queries › Query Heatmap | `GetQueryHeatmapAsync` (`ViewerDataService.QueryHeatmap.cs`) | The *read* is portable even though the interactive plot is not — a bucketed table is a real answer. Needs a bucket-size decision. |
| Queries › Performance Trends (3 of 4 charts) | `GetProcedureDurationTrendAsync`, `GetQueryStoreDurationTrendAsync`, `GetExecutionCountTrendAsync` (`ViewerDataService.QueryTrends.cs`) | `get_query_duration_trend` exists and ships; these three are its missing siblings, same shape. |
[...]

PR #2494, opened 2026-08-21
Serve the raw collection log (#2484)

erikdarlingdata:
First item on #2484's list, and the one it ranks first: the collection log is what people reach for when the rollup says HEALTHY and collection still looks wrong, and it covers two viewer tabs (Collection Log and Duration Trends).

## The gap

`/api/read/*` is machine-derived from the MCP read catalog, so a read with no tool is absent from **both** surfaces. `get_collection_health` serves the seven-day per-collector rollup; the raw per-run log had nothing. No projection of a rollup gives back the individual runs, so the WPF viewer on a Windows desktop was the only way to this data — invisible to the web dashboard and to any agent.

**No new collection.** The store already holds these rows. This reads `v_collection_log`, the same view the viewer uses, so the three surfaces cannot drift.

## Three decisions worth stating

**The window is bounded on both sides.** Most neighbouring reads take a single now-relative `hours_back`. Someone asking about a past incident wants the runs from *then*, and an hours-back-from-now span cannot express that. Matches how the viewer's own tab windows this read.

**Zero rows says which kind of nothing it is.** This is the substance of the change:

| State | Status | Told to |
|---|---|---|
| Collected before, quiet window | `empty` | widen `hours_back` |
| Never collected at all | `unavailable` | go find out why collection is not running |

Both are literally "no runs in the last 24 hours". The responses are opposites — telling the second caller to widen the window sends them off filling one that will never fill. That is exactly the defect #2485 catalogues across six other reads, so this one is not allowed to add a seventh. It costs one `LIMIT 1` probe, taken only on the empty path.

**Duration is split three ways** — total, time on the monitored server, time writing to the store. A collector slow because the target is slow needs a different fix from one slow because the store is, and the total alone cannot tell them apart. That is the first question anyone asks of a slow collector, and it is why the raw log earns its place beside the rollup.

## Verification

Live store test (`DarlingCollectionLogReadTests`, runs in the PostgreSQL job) covers all four paths:

1. Registered but never collected → `unavailable`, message contains "EVER", and asserts it does **not** say "widen".
2. Rows seeded 48h out, asked for 1h → `empty`, asserts it does **not** say "EVER", and does say "widen". Same row count as case 1 — zero — and it must not reach for the same word.
3. Rows in window → newest-first ordering, and the `sql`/`store` split values.
4. `limit: 1` → `truncated: true`, so the cap announces itself instead of being inferred.

Service and test projects build clean, 0 errors. JS parses clean.

The existing parity pin (every `[McpServerTool]` needs a catalog entry) self-enforces the wiring; the 12-tab pin and the no-duplicate-read-per-tab pin both still hold, since this adds a panel to an existing tab rather than a tab.

## Remaining on #2484

Blocking Stats, Current Waits, Significant Waits, the three Performance Trends siblings, Query Store Regressions, and the heatmap — in the order the issue suggests.


erikdarlingdata:
All findings addressed. The first one was the important one — thank you for catching it.

**1. The panel never rendered and never fetched.** Correct, and it is the whole web half of this change. A raw object literal where every sibling goes through `table()`; `appendChildren` handles arrays and `Node`s and stringifies everything else, so it would have shipped as the literal text `[object Object]` and `get_collection_log` would never have been requested. Now goes through `table(...)`.

Worth naming what this says about the pins: the change built, the JS parsed, and every page-level pin passed. They check that a named read exists, that a named viz is in the vocabulary, that no tab fetches the same read twice — all true of a descriptor the mounter cannot consume. **A panel that renders nothing looks identical to one nobody looked at.**

**2. `truncated` off-by-one.** Fixed by over-fetching one row and reporting what was observed rather than comparing the count to the cap. A window holding exactly `limit` runs is full, not truncated, and this is the one field whose entire premise is that the cap should not have to be inferred. Pinned on both SKUs: exactly-the-cap asserts `false`, under-the-cap asserts `true`.

**3. The row cap.** `Math.Clamp(limit, 1, 5000)` accepted five times the shared ceiling and silently rewrote out-of-range values. Now `McpHelpers.ValidateTop`, which refuses — a caller asking for 5000 is told no rather than handed 1000 while believing it got 5000.

**On the parity note:** Lite got `get_collection_log` in the same change rather than a line on the divergence ratchet — that ratchet is documented as only ever shrinking, and #2485 is already open about SKU divergence being a defect. Lite now has a tool-level test too, and writing it surfaced something: **Lite does not store a server id, it derives one from the storage name.** Seeding under a hardcoded id would have written rows the tool looks straight past, and the never-collected assertion would have passed for exactly the wrong reason.

**One thing I overclaimed and am correcting rather than quietly fixing:** the description said the two-sided window lets a caller ask about a past incident. The reader is bounded on both sides and the viewer uses that, but the MCP tool exposes only `hours_back`, always anchored to now — so as shipped, no caller can express "last Tuesday". The two-sided bound is real and the web panel now follows the page's time picker rather than a hardcoded 24 hours, but an arbitrary window is not reachable through the tool. Filed as follow-up rather than grown into this PR.

PR #2497, opened 2026-08-22
Serve blocking severity, and share the deadlock aggregation (#2484)

erikdarlingdata:
Third item on #2484, and the one that needed its own PR: the deadlock aggregation was `internal` to the Darling viewer, so the headless service could either duplicate it or move it.

## The move is the substance

There were already **two** copies computing victim counts and total wait from parsed graphs — the viewer's and Lite's. A third would be how three surfaces end up quietly disagreeing about one deadlock.

- `DeadlockSeverityStatsPoint` and the aggregation now live in `PerformanceMonitor.Common`.
- The viewer's `AggregateDeadlockSeverity` stays as a delegate, so the **nine existing pins** written against it keep working — and because they now exercise the moved code, they *are* the evidence the move changed nothing.
- `TruncateToMinute` was copied **verbatim** rather than rewritten to an equivalent, so not a single bucket boundary can shift.
- Lite keeps its own aggregation *body*: it parses into `DeadlockProcessDetail` where Darling uses `DeadlockProcessNode`. Only the record is shared. Unifying the parsers is a real question and not this change's.

## What the read adds

`get_blocking_trend` and `get_deadlock_trend` count incidents. **Ten one-second blocks and one ten-minute block are the same count and are not the same problem.** This is severity — total, max, and process-weighted average — the half that existed only in the desktop viewer.

## The empty answer checks both capture paths

Either can be off on a given server: the blocked-process report needs its XE session running, the DMV snapshot needs its collector enabled.

- Probing **one** would report "never captured" for a server capturing fine through the other.
- Probing **neither** would let a silent capture gap read as a clean bill of health.

That second failure is the worst outcome here, because *"no blocking"* is exactly what an operator wants to hear. So `unavailable` says outright this is **NOT a clean bill of health** and names both prerequisites.

## Verification

- `DeadlockSeverityAggregatorTests` — **run locally**, 2/2 passing, against the real `Common` assembly via a `net10.0` shim. Pins that total wait sums **every** process (4500, not 1500) rather than only victims, that the average is process-weighted, that buckets come back in time order when fed out of order, and that an unparseable graph contributes **nothing** rather than a zero bucket — an empty bucket would read as "a deadlock happened and cost nothing", and absent evidence is not evidence of calm.
- First draft of that fixture marked victims with an attribute and failed; rewritten against the shape the viewer's proven suite uses.
- All six projects build clean. JS parses clean.
- Web panels go through `fanout()` — one read, two charts. Counts are deliberately **not** charted beside durations: a count and a millisecond on one y-domain is the two-units mistake the CPU chart's dropped idle series exists to avoid.

Census 103 → 104 tools, 78 → 79 shared, Darling-unique unchanged.


erikdarlingdata:
Good catch, and a real one — fixed in both SKUs.

The verdict fires only when the blocking series **and** the deadlock series are both empty, but the probe behind it checked only the two *blocking* sources. A server whose deadlock capture had never run, with blocking capture working fine, would have been told the window was "genuinely clear" on the strength of a source that says nothing about deadlocks.

That is exactly the failure the probe was written to prevent. The doc comment even argued the case for checking both paths rather than one — it was counting the wrong "both". Now all three.

**There was no live test for this tool**, only for the pure aggregator, which is why nothing caught it. `DarlingBlockingStatsReadTests` now pins it, deliberately from the direction that would otherwise flip: a deadlock *outside* the window with no blocking rows at all must read as a clear **window**, not as never-captured. Telling an operator to go fix collection that is working is the same defect pointing the other way, and a probe fixed carelessly would trade one for the other.

Issue #2495, opened 2026-08-21
MCP reads window only backwards from now, so no caller can ask about a past incident


erikdarlingdata:
Split out of #2494 rather than grown into it.

## What

`get_collection_log`'s reader is bounded on **both** sides — it takes a window start and a window end — and the viewer's Collection Log tab uses that to honour a custom From/To range. The MCP tool exposes only `hours_back`, always anchored to `end = now`.

So the capability exists in the read and is unreachable through the tool. An agent asked *"what were the collectors doing during the incident last Tuesday?"* can only say "the last N hours" and widen until Tuesday falls inside — dragging along everything since.

I claimed the past-incident case worked in #2494's description. It does not, and this is the correction.

## Scope, because it is not one tool

`get_collection_log` is where it was noticed, not where it starts. The same `hours_back`-from-now shape is the norm across the read surface — `get_wait_stats`, `get_query_store_top`, `get_blocking`, `get_alert_history` and others all anchor to now. Any read whose stored data outlives its default window has the same gap.

That makes this a **surface-wide parameter convention question**, not a bug in one tool:

1. **Add optional `from`/`to` to reads that already window**, keeping `hours_back` as the default path. Backward compatible; the widest change.
2. **Add an `as_of` anchor** — keep `hours_back` as the span and let a caller move the endpoint off "now". Smaller and covers the incident case, which is nearly always "N hours around a known time".
3. **Leave it.** Callers widen and filter client-side. Cheapest, and gets worse the more history the store keeps — which is the direction it is going.

Option 2 is the smaller change and covers the actual use case, but this wants deciding once for the surface rather than per tool, which is why it is an issue and not a patch.

## Not a workaround

Widening `hours_back` until the window reaches back far enough is not equivalent: it changes what a top-N returns, what an average is taken over, and how much a capped read truncates. For an aggregate read, a wider window is a **different answer**, not the same answer with more rows.

erikdarlingdata:
**Done in #2504.** `as_of` — an optional ISO-8601 UTC anchor that moves the window's END off "now" while `hours_back` stays its LENGTH. Applied to **57 Darling / 50 Lite** reads with identical name, semantics and description on both, plus all 57 web catalog descriptors and dispatch lambdas, so it is reachable from `/api/read/*` and not only from MCP.

[...]

PR #2509, opened 2026-08-22
Serve the Query Heatmap on the viewer's own 5-minute bins (#2484)

erikdarlingdata:
Closes the last item on #2484. The other nine shipped in #2494, #2496, #2497, #2498, #2503, #2507 and #2508.

The interactive plot stays desktop-only by design (#2484 group (c)) — there is no `heatmap` viz in the web page's vocabulary and inventing a fifth one is not what the issue asked for. The **read** behind it is portable, and a bucketed table is the same answer: one row per (time bin x magnitude bucket) cell, with the most-executed query in each.

What it adds over every other query read is the **time axis**. `get_top_queries_by_cpu` ranks a whole window and cannot show that the window had a quiet half and a bad half, which is the first thing anyone asks about an incident that has already ended.

## What the viewer's bucketing turned out to be

**A constant, not a derivation.** `ViewerDataService.QueryHeatmap.cs` bins with a literal `INTERVAL '5 minutes'` whatever range is on screen — it is Lite's `time_bucket(INTERVAL '5 minutes', collection_time)` ported to `date_bin`, and neither one looks at the window length. So there was no derivation to reproduce, only a default to honor:

- **Time bins: 5 minutes**, exposed as `bucket_minutes` and defaulting to exactly that. A browser, an agent and a desktop pointed at the same server over the same window draw the same grid unless the caller asks for something else. The result echoes `bucket_minutes_matches_desktop_viewer` so a consumer can tell at a glance whether it is looking at the viewer's picture.
- **Magnitude buckets: the viewer's seven**, `< 1 / 1-10 / 10-100 / 100-1K / 1K-10K / 10K-100K / > 100K`, labelled per metric family (milliseconds for duration and CPU, plain counts for the other three). The labels ship with every result — a bare `bucket_index` is unreadable.
- **The metric expressions are byte-identical** with `ViewerDataService.HeatmapMetricExpr` and Lite's `GetMetricColumn`, all five of them.

The SQL is the viewer's verbatim apart from two things a desktop chart does not need: the bin width is a bound parameter, and the tail carries `ORDER BY time_bin DESC` + `LIMIT` so a capped call keeps the recent end of the window.

## The divergence the parameter exposed

Making the width a parameter surfaced a cross-SKU bug the default was hiding.

Postgres `date_bin` **requires** an origin and the viewer passes the Unix epoch. DuckDB's `time_bucket` defaults to a **different** origin — 2000-01-03. The two are **15,780,960 minutes apart**, and 5 divides that exactly, so the default agrees *by luck*. 7 does not:

| stride | Postgres `date_bin` | DuckDB `time_bucket` (default origin) |
|---|---|---|
| 5 min | `10:05:00` | `10:05:00` |
| **7 min** | `10:02:00` | **`10:01:00`** |

The first caller to pass an odd `bucket_minutes` would have had Lite and Darling bin the same row a minute apart. Lite's read now passes the epoch explicitly. Verified identical across `{1, 5, 7, 13, 60, 90, 360, 1440}`-minute strides against PostgreSQL 17 and DuckDB 1.5.5 — 32 combinations, zero mismatches. Lite's WPF chart keeps its origin-less form deliberately: it only ever asks for 5 minutes, where the two agree.

## Which kind of table this is, and why the empty branches are what they are

**`query_stats` is a PERIODIC table, not an edge table.** The collector writes rows every cycle for whatever is in the plan cache, whether or not anything interesting happened. So an existence probe **on the data** is the right denominator here — the exact opposite of what #2508 corrected for blocking and deadlocks, where zero rows is the healthy answer and a data probe sends someone to fix collection that works. The probe reads `v_query_stats`, the same relation the read itself uses, so it cannot disagree with the read about which rows exist.

Zero cells is **three** states:

1. **`unavailable`** — no rows for this server at all. Nobody looked; go find out why collection is not running.
2. **`empty`** — rows exist outside the window, none in it. Widen `hours_back`, or check `get_collection_health`; a collector that stopped looks exactly like this.
3. **`empty`** — captures exist *in* the window and every one recorded a **zero execution delta**. This is the branch only this read has: collection is healthy and the server is idle. Telling that caller to widen the window would be advice pointed at the wrong problem. The same sentence covers a `database_name` filter that matched nothing collected, and the two-cycle warm-up that delta collection needs.

All three are pinned on both SKUs, asserted to be three different sentences, and the three shared sentences are added to `McpMissMessageParityPinTests` so they cannot drift a dash apart between the trees.

## The cap, and the partial-column problem

`limit` caps **cells** (default 500) and is validated by `McpHelpers.ValidateTop` — out of range is refused, never clamped. Over-fetch by one, so truncation is observed rather than inferred.

Rows come back newest-bin-first so the cap cuts the **oldest** end, which is what anyone looking at an incident wants. That creates a hazard nothing else on this surface has: the cut can land in the middle of a bin and hand back a **column missing its low buckets**, which reads as "nothing fast ran then" rather than "we stopped looking" — a quiet wrong answer that a grid makes very easy to believe. So the partial column is dropped, and `first_time_bin` / `last_time_bin` say which slice of the window actually came back. (It is kept only when it is the *only* column, i.e. a cap below one bin's seven cells, where there is nothing to fall back to.)

`bucket_minutes` is the lever to reach for before the cap: a 7-day window at 5-minute bins is 2,016 columns, and widening the bin covers the window in fewer cells.

An unknown `metric` is **refused**, not silently turned into duration — a caller who asked for CPU and got elapsed time would read the wrong grid with nothing to tell them so.

## Also in here

`docs/uat-onboarding.md` §3.4 listed "data with no read endpoint at all", named the query heatmap in it, and said no amount of web work reaches it. That list is now empty, so it says so. Its counts were seven tools stale (the whole #2484 wave landed without them moving); they are **re-derived from the source rather than incremented** — 90 read endpoints out of 109 tools, 69 of them reached by the per-server page. The "desktop things a web imitation would be worse than" paragraph keeps the heatmap, narrowed to the interactive **plot**.

Both SKUs, not the divergence ratchet. Census recounted by scanning the `McpServerTool(Name = ...)` attributes: **108 → 109 tools, 83 → 84 shared, 25 Darling-unique unchanged.**

## The window anchor (#2495 / #2504)

`as_of` landed across 57 reads while this was open, and the heatmap now takes it too — on the shared `McpHelpers.AsOfDescription` constant, not a second wording of the same idea.

It earns the parameter more than most reads do. A heatmap **is** a time axis, so "the four hours ending Tuesday 03:00" is the shape of every question anyone brings to it. And widening `hours_back` until an old incident falls inside is not the same question here even in the weak sense the general argument makes: on this read the extra hours arrive as extra **columns**, which push the incident's own columns past the cell cap.

`#2495`'s own failure mode is a tool that advertises the anchor, validates it, refuses a bad one correctly — and then queries `DateTime.UtcNow` anyway, so the validation succeeding is exactly what makes the caller believe the window moved. Eight of those shipped past a green suite during #2504. This is therefore proved by **content** on both SKUs: rows are seeded outside every default window, the anchored call returns them, and the same-**length** window at the default anchor does not. Neither tool names `DateTime.UtcNow`, and Lite threads the resolved instant into both `GetQueryHeatmapCellsAsync` and `GetQueryHeatmapCoverageAsync` rather than letting either compute its own.

That also subsumes the window-skew fix from the review below: one resolved instant now decides the query window, the coverage probe's window, and the window reported back.

## Merge conflicts, and how they were taken

`dev` moved under this branch (#2504, #2505, #2507, #2508). Three files conflicted:

- **`DarlingWebEndpoints.cs`** catalog + dispatch — dev's `as_of` on the neighbouring reads, plus `PAsOf()` / `as_of: AsOf(c)` on `get_query_heatmap`.
- **Both instructions tables** — dev's rows win (they carry `as_of`); the heatmap row is re-inserted in place with `as_of` appended.

The census sentence did **not** conflict, and was recounted anyway rather than trusted: scanning the `McpServerTool(Name = ...)` attributes gives **109 / 84 / 25**, which is what it says. The `uat-onboarding` counts were re-derived the same way and are unchanged at 90 read endpoints, 109 tools, 69 reached by the per-server page.

## How this was verified

- **The shipped Postgres SQL was executed against real PostgreSQL 17** (a local timescaledb container) over a synthetic `v_query_stats`, and the text in `DarlingQueryHeatmapReader` is asserted equal to the text that ran, modulo whitespace. Five-minute and sixty-minute bins, the database filter, and the `LIMIT` path all produce what the tests expect.
- **The Lite SQL was executed against real DuckDB 1.5.5** — the exact version pinned in `Directory.Packages.props` — and produces cell-for-cell identical output to the Postgres run on the same rows.
- The Lite test's whole scenario (bin counts, cell counts, the top-query-per-cell tie-break, the truncation trim, the idle-probe branch) was simulated end to end against that DuckDB before the C# was written.
- Full solution builds clean on macOS with the `EnableWindowsTargeting` ritual; `global.json` is untouched in every commit.
- The `AsOfWindowAnchorTests` source-scan rules (anchor reaches the query, no `DateTime.UtcNow` in an anchored body, no anchorable service call left un-anchored) were replayed locally over both trees before pushing: 58 anchored tools scanned on Darling, 51 on Lite, zero offenders.
- `node --check` on `server-tabs.js`. The panel is built with `table(...)`, not an object literal.
- Locally re-derived and checked: line endings (CRLF, zero bare LF across all 13 files), stacked-`<summary>` hygiene, the census against a fresh attribute scan, the shared-sentence parity in both trees, and that every param key the panel sends is one the catalog binds.

## What I could NOT verify

- **The test suites themselves.** Both projects are `net10.0-windows`, so they build here and run in CI. The live Darling test is gated on `DARLING_TEST_PG`.
- **A real monitored server.** Everything above is synthetic rows; nothing was run against a production store, so the shape of a real busy server's grid (and whether 500 cells is the right default in practice) is untested by anything but arithmetic.
- **The rendered panel.** The descriptor is validated by `node --check` and by the page-level pins, but no browser has drawn it.

One judgement worth flagging for review: the metric list is the viewer's five, **including** `logical_reads` and `logical_writes`. Dropping them would have made the two surfaces disagree about what the heatmap can show, which is the thing this PR exists to prevent.

## CI

All six checks green. One red run along the way was `Lite.Tests.LiteServerTagsStoreTests` timing out on the DuckDB **write** lock — the documented writer-starvation shape of that suite's one process-wide `ReaderWriterLockSlim` (`SharedDuckDbFixture`'s own doc comment describes it; #2208 is the same shape). Unrelated to this change and green on re-run, but worth naming rather than quietly re-running.


erikdarlingdata:
Good catch on the window skew — fixed in bfef9cb6.

`window_start`/`window_end` were taken from a fresh `DateTime.UtcNow` *after* the DuckDB call returned, so the reported window drifted by however long the read took. It is now hoisted before the call, the same move `GetTopQueriesByCpu` in that file already makes and for the same reason. Darling's twin never had it (it computes start/end once and passes them into the query).

Pinned on **both** SKUs rather than only the one that was wrong: the reported window must span exactly `hours_back` and bracket every bin returned. Darling was correct by construction and now has the assertion too, so that cannot quietly regress.

Issue #2515, opened 2026-08-22
The tempdb Space alert's percentage has no size floor, and on Azure SQL Database that makes it fire on one temp table


erikdarlingdata:
Found while removing the Azure SQL Database gate on `TempDbStatsCollector` (#2512). The collection is right and ships; this is the consequence downstream, and it is a policy call rather than a bug fix, so it is filed rather than folded in.

## The mechanism

`TempDbSpaceInfo.UsedPercent` is

```csharp
TotalReservedMb / (TotalReservedMb + UnallocatedMb) * 100
```

and `AlertEngine.CheckTempDbSpaceAsync` fires when that clears `TempDbSpaceThresholdPercent` (default **80**, in both SKUs).

Both halves of that denominator come from `tempdb.sys.dm_db_file_space_usage`, which reports the tempdb files **as currently allocated**. On a box with a pre-sized tempdb that denominator is effectively the ceiling, so the ratio is real headroom and the alert means what it says.

On Azure SQL Database it is not. The platform creates tempdb small and autogrows it toward the tier cap: **4 files, ~62 MB total** on both 2-vCore tiers measured (`GP_S_Gen5_2` and `HS_S_Gen5_2`, EngineEdition 5). So the ratio measures distance to the next autogrow, not distance to the ceiling.

## The measurement

From #2512, on `GP_S_Gen5_2`, allocating ~57 MB into one `#temp` table:

| | `total_reserved_mb` | `unallocated_mb` | `UsedPercent` |
|---|---|---|---|
| idle | 1.88 | 60.69 | **3.0%** |
| after one `#temp` | 59.75 | 2.69 | **95.7%** |

One ordinary temp table, and the server is over the 80% default with 60 MB at stake. That is not a tempdb running out of room; that is a tempdb that has not grown yet. Every Azure SQL Database target doing real work would page someone on its first busy minute.

## The second threshold, failing the other way

`FactScorer.ScoreTempDbVersionStore` scores `max_version_store_mb` at absolute bars — **>5000 MB CRITICAL, >2000 HIGH, >1000 MEDIUM** — inherited from `install/47_create_reporting_views.sql`. Those are unreachable while tempdb is at its initial ~62 MB, so the arm scores 0.

[...]

Issue #2511, opened 2026-08-22
On Azure SQL DB, twelve reads tell you to go start a capture that cannot exist there


erikdarlingdata:
Found by actually connecting to a live Azure SQL Database rather than reasoning about one.

## The evidence

Live probe, General Purpose serverless (`GP_S_Gen5_2`, `EngineEdition` 5, `12.0.2000.8`):

```
sys.dm_xe_sessions          ERROR: Invalid object name 'sys.dm_xe_sessions'
sys.dm_xe_database_sessions 2
sys.dm_os_wait_stats        1819
sys.dm_os_waiting_tasks     59
sys.dm_os_performance_counters 4216
```

`sys.dm_xe_sessions` does not exist on Azure SQL Database, so the **`system_health` ring buffer can never be read there**. The collector knows this — `SystemHealthEventsCollector.AppliesTo(target) => !target.IsAzureSqlDb`. It is a deliberate, correct, permanent gate.

## The defect

**The reads do not know.** On an Azure SQL DB target, all nine `get_health_parser_*` tools answer:

> No wait_info events have EVER been captured for `<server>`, so this is NOT an all-clear — there is nothing here to be clear about. This read is served from the collected system_health ring buffer: **check that collection is running for this server and that its system_health session is started** before concluding nothing was waiting.

Every clause is wrong for this target. Collection *is* running. The session cannot be started, because the DMV that exposes it does not exist on this engine. The operator is sent to fix something that is not broken and cannot be changed.

This is the same failure #2485 was about and the same one #2508 fixed for edge tables, on a third axis: **engine capability** rather than event rarity. And the miss vocabulary already has the right word — `not_collected`, *"the input names something this server does not collect"* — which is exactly true here and is currently used only for per-counter and per-wait-type misses.

## Scope: not one family

Twelve collectors gate themselves off on Azure SQL DB, and every one of them feeds reads that will now confidently mis-explain their permanent emptiness:

[...]

PR #2516, opened 2026-08-22
Collect tempdb on Azure SQL Database: the gate's stated reason was false (#2512)

### LINKED ISSUE #2512: Azure SQL DB gets no tempdb monitoring, and the reason recorded for that is not true
Found by running the collector's own SQL against live Azure SQL Database, both service tiers.

## The gate

```csharp
// TempDbStatsCollector.cs:70
public override bool AppliesTo(CollectorTargetInfo target) => !target.IsAzureSqlDb;
```

Its doc comment explains why:

> The second result set (`sys.dm_db_session_space_usage`, two-part and in-database) WOULD work on Azure SQL DB. Recovering that half needs an Azure-specific query variant rather than a gate … the immediate defect is a collector that can only ever fail

The premise is that the FIRST result set — the three-part `tempdb.sys.dm_db_file_space_usage` — cannot bind on Azure SQL Database.

## It binds, and returns real data

The collector's SQL, verbatim, both result sets, run over an Entra token (2026-08-22):

**General Purpose serverless** (`GP_S_Gen5_2`, `EngineEdition` 5, `12.0.2000.8`)

| result set | returned |
|---|---|
| 1 — `tempdb.sys.dm_db_file_space_usage` | user 5.44 MB, internal 1.81 MB, version store 0.00 MB, total 7.25 MB, unallocated 54.19 MB |
| 2 — `sys.dm_db_session_space_usage` | session 74, 0.13 MB, 1 session over threshold |

**Hyperscale serverless** (`HS_S_Gen5_2`, `EngineEdition` 5)

| result set | returned |
|---|---|
| 1 | user 1.88 MB, internal 0.00 MB, version store 0.00 MB, total 1.88 MB, unallocated 60.69 MB |
| 2 | session 1, 0.00 MB, 0 sessions over threshold |

`SELECT COUNT(*) FROM tempdb.sys.dm_db_file_space_usage` returns **4** on both tiers.

So the collector does not "only ever fail" on Azure SQL Database. It works, on both tiers, with no query variant needed.

## What this costs today

Every Azure SQL Database target — the whole tier, both GP and Hyperscale — collects **no tempdb data at all**. No `get_tempdb_trend`, no tempdb space alerting, nothing in the viewer's tempdb surfaces. On a platform where tempdb contention is a common and consequential problem, and where the customer cannot simply look at the box themselves.

Worse, after #2485 the reads will now explain that absence confidently and wrongly — see #2511.

## What to check before flipping the gate

I have verified the SQL binds and returns plausible values. I have **not**:

- run it through the actual collector pipeline against an Azure target (payload composition, column typing, the `RunsPerDatabase` interaction);
- checked whether the numbers are *meaningful* on Azure SQL DB, as opposed to merely returned — `tempdb` on Azure SQL DB is per-replica and its size is tier-governed, so "unallocated_mb" may not mean what it means on a box. **This matters more than the binding**: a number that returns but describes something different is worse than a gap, and is precisely the trap this collector's own comment was trying to avoid;
- confirmed behaviour on Managed Instance is unchanged (the gate is `IsAzureSqlDb` specifically, so MI already collects).

So: flip the gate only after deciding what these figures MEAN on the tier, and say so in the doc comment. If they turn out to be misleading rather than merely available, the right fix is to keep the gate and **correct the stated reason**, because the current one is checkably false and will invite exactly this issue again.

## Related

- #2511 — the reads mis-explain permanent Azure gaps; this issue is a gap that should not exist in the first place.

erikdarlingdata:
Closes #2512.

## What the gate cost

```csharp
public override bool AppliesTo(CollectorTargetInfo target) => !target.IsAzureSqlDb;
```

Every Azure SQL Database target — General Purpose **and** Hyperscale — collected **no tempdb data at all**. No `get_tempdb_trend`, no tempdb space alerting, nothing in the viewer's tempdb surfaces, nothing for the analysis pass's `TEMPDB_USAGE` fact to score. On a platform where tempdb contention is common and the customer cannot go and look at the box themselves.

Its doc comment said the first result set's three-part `tempdb.sys.dm_db_file_space_usage` reference could not be served there, so the collector "could only ever fail". That premise was checkable.

## The live evidence

The collector's SQL, verbatim, both result sets, over an Entra token (2026-08-22):

| tier | result set 1 | result set 2 |
|---|---|---|
| `GP_S_Gen5_2` (EngineEdition 5, 12.0.2000.8) | user 5.44 MB, internal 1.81 MB, version store 0.00 MB, total 7.25 MB, unallocated 54.19 MB | session 74, 0.13 MB, 1 session over threshold |
| `HS_S_Gen5_2` | user 1.88 MB, internal 0.00, version store 0.00, total 1.88 MB, unallocated 60.69 MB | session 1, 0.00 MB, 0 over threshold |

`SELECT COUNT(*) FROM tempdb.sys.dm_db_file_space_usage` returns **4** on both. No Azure-specific query variant is needed, and **both** result sets bind — so the one row this collector promises is a whole row on Azure SQL DB, which is what the old comment's "half a row is a different contract" objection was about.

And the figures *describe* something rather than merely returning, which was the open question. Allocating ~57 MB into a `#temp` table on `GP_S_Gen5_2`:

| step | `user_mb` | `unallocated_mb` |
|---|---|---|
| before | 1.88 | 60.69 |
| after ~57 MB | **59.75** (+57.87) | **2.69** (−58.00) |

with `sys.dm_db_session_space_usage` attributing 59.25 MB to the session that did it. The counters track actual allocation; the session view attributes it correctly.

**Worth more here than on a box.** The tempdb ceiling on Azure SQL Database is governed by the **service tier**: you cannot add files and cannot grow past the cap. On a box "tempdb is filling" means "go look at the disk". Here it means "you are approaching a limit you cannot raise without changing service objective".

## What each red test was really pinning

**`CollectorGateSurfacePinTests.TempDbStats_AppliesTo_SkipsOnlyAzureSqlDb`** — the only test that goes red, and it was pinning **two different claims in one `Assert.False`**:

1. *That the DMV cannot be served on Azure SQL DB.* False, per above. Flipped.
2. *That a login might not be able to READ it.* True — but that is a property of the **login**, not of the tier, so it cannot be decided by a gate. It moved to the fault classifier (below).

The other half of that test — **`Assert.True(...AppliesTo(AzureMi))`** — never changed, and it is why the test survives rather than being deleted. Its original comment says outright that asserting both directions is what stops an "anything Azure" gate creeping in. That risk now runs the other way: this must not be re-narrowed later on the strength of the stale comment, so both directions are still asserted, and the test now also asserts the **composed** `CollectorCatalog.AppliesTo` and the by-name surface Lite dispatches through — neither of which `AppliesTo` alone can see.

Nothing else was pinning the gate. `AzureSweepScopeTests`, `CollectorRunnerConnectionEngineTests` and `DarlingWorkerTests` don't reference this collector at all; `DarlingCliCommandsTests.ToTargetInfo_RoundTripsTheFactsTheGateReads` asserts *all* SQL Server collectors are off at a **PostgreSQL** target, which is the engine half of the composed gate and is unaffected — verified by asserting the same thing directly against the real build.

## Coordination with #2511

The composer's capability map **is derived, not transcribed** — `DarlingWebEndpoints.BuildAppliesToNode` calls `collector.AppliesTo(s_azureSqlDbTarget)` per measure — so the three `tempdb_*` measures flip to `azureSqlDb: true` with no edit here, and `DarlingComposeTests.Catalog_Measures_CarryAppliesTo_ReflectingTheCollectorGate` doesn't name tempdb. If #2511's `not_collected` mapping is derived the same way it will flow through automatically; if it goes red on this branch, that is the signal the map was transcribed.

## What I verified about payload composition

The issue listed what it had **not** checked. All of it, executed against the real build rather than reasoned about:

- **The composed gate**, both halves: `CollectorCatalog.AppliesTo(def, azureSqlDb)` → true, `(def, azureMi)` → true, `(def, postgres)` → **false** (the engine half still holds). Plus `CollectorCatalog.AppliesTo("tempdb_stats", azure)`, the by-name surface Lite dispatches through.
- **`RunsPerDatabase(azure)` → false**, and `BuildEnumerationQuery` → null. So on Azure this takes the plain single-connection path, not the per-database loop `file_io_stats` and `index_object_stats` take. Per #2220 a registration naming a database is scoped to that database, so this is one tempdb snapshot per registration, not N.
- **Query text is byte-identical** on Azure and on-prem — no variant, as the measurement says none is needed.
- **Column typing.** `sys.dm_db_session_space_usage.session_id` is `smallint`, so the driver hands back a `short`; the definition reads it through `Convert.ToInt32(GetValue(0))` precisely for that. Fed a real `short`, `ReadAsync` widens it and `WritePayload` emits `int` for `top_session_id INTEGER`. The existing parity pin feeds an `int` and can never see this. `total_sessions` is `COUNT_BIG` → `long` → `bigint`.
- **Payload order and arity**: 8 values, `PayloadColumns` order, CLR types `decimal ×5, long, int, decimal`, matching the generated DDL (`numeric(18,2)` ×5, `bigint`, `integer`, `numeric(18,2)`).

**It composes cleanly.** Nothing had to be forced.

## The permissions half — what actually makes this safe

The #2150 field report was 11x consecutive with **error 262**, "VIEW DATABASE PERFORMANCE STATE permission denied in database 'tempdb'". That is real, and removing the gate without handling it would have regressed a fixed bug — 262 was classified **`Unclassified`, which means ERROR, every cycle, forever**, which is exactly the collection-health pollution the gate was created to stop.

262 is now a permission denial in both SKUs, so such a login degrades to a non-fatal PERMISSIONS skip with an explanatory Azure sentence, the same treatment error 300 already gets (#1631) — while every properly-permissioned Azure target collects.

Those numbers lived in **three** hand-maintained places and **had already drifted**: 916 was in `SqlServerTargetProvider.Classify` and in neither runner's catch, so the engine seam whose stated purpose is that "the runner and the provider cannot disagree about what an error means" disagreed. All three now read one shared `SqlServerPermissionErrors.IsPermissionDenied`. That is also the only way to *pin* them — `SqlException` cannot be constructed in a test, so a shared predicate is the difference between a tested rule and three unverifiable literals. **Note the behaviour change**: Lite and Darling's runners gain 916 → PERMISSIONS. That is the correct classification per `CollectorTargetFault`'s own definition, and it is called out here rather than smuggled in.

## Thresholds that need attention — filed as #2515

**`unallocated_mb` is free space in the tempdb files AS CURRENTLY ALLOCATED**, not headroom to the tier ceiling. Azure creates them small — 4 files, **~62 MB total** on both 2-vCore tiers measured — and autogrows toward the cap. So `total_reserved / (total_reserved + unallocated)`, which is exactly what `TempDbSpaceInfo.UsedPercent` computes and what the `tempdb Space` alert fires on at a default of **80%**, measures distance to the **next autogrow**, not to the ceiling.

Run the alert's own arithmetic on the measured numbers:

| | `UsedPercent` |
|---|---|
| idle | **3.0%** |
| after one ~57 MB `#temp` table | **95.7%** |

One ordinary temp table puts an Azure SQL DB target over the default with 60 MB at stake. **That would fire on the first busy minute of every Azure target this PR enables**, which is the "worse than the gap" case. The alert needs a size floor — the shape `ScorePlanCacheBloat` already uses ("a high % on a trivially small footprint is noise").

I did **not** retune it here. The floor value is a policy call that suppresses alerts existing on-prem and RDS targets get *today*, and I cannot measure what those customers rely on from here — note that `AlertEngineTests.TempDb_FiresAtThreshold_AndResolutionCarriesTheCurrentPercent` uses a fixture totalling exactly 1000 MB, so a 1 GB floor flips an existing pin. **#2515** carries the mechanism, the three options, and the measurement that should decide (the distribution of total tempdb size at the moment `UsedPercent >= 80` across collected history). If you want the floor in this PR instead, say so and I will add it.

Second threshold, failing the other way and **needing no change**: `FactScorer.ScoreTempDbVersionStore` bars at 1000/2000/5000 MB are unreachable while tempdb sits at ~62 MB, so that arm scores 0. It is self-consistent — a >1 GB version store implies a >1 GB tempdb — and produces no false alerts. Worth knowing because **RCSI is on by default on Azure SQL Database**, which makes the version store the tempdb consumer most likely to matter there; the space-fraction arm is the only cover until tempdb grows past a gigabyte. Noted in #2515, not acted on.

`get_tempdb_trend` (both the MCP read and the viewer's), `TempDbSpaceSql`, and the analysis facts needed **no change**: they are column reads with no absolute-MB assumptions, and the `numeric(18,2)` → `double precision` / `bigint` / `integer` typing is already what an Azure row produces.

## Testing

- `CollectorGateSurfacePinTests.TempDbStats_AppliesTo_EveryTarget_IncludingAzureSqlDb` — flipped Azure, kept MI, added the composed and by-name surfaces.
- `TempDbStatsCollectorDefinitionTests.AzureSqlDb_MeasuredValues_ComposeThroughToThePayload` — the composition run above, using the values `GP_S_Gen5_2` actually returned and a real `short` session id.
- `SqlServerPermissionErrorsTests` — the six permission numbers, six non-permission numbers that must stay loud, and the 262 hint on/off Azure.

Test projects are `net10.0-windows` and cannot run on macOS, so every assertion above was additionally executed against the real build through a throwaway `net10.0` harness first. CI is the arbiter for the suites themselves.


erikdarlingdata:
## Review finding fixed — `c4c3197`

`claude[bot]` caught a real one on `AzureDmvPermissionHint`, and it was the exact thing I hedged on in wording instead of solving in code: the hint dispatched on the **error number alone**, so any 262 on an Azure SQL DB target got the tempdb sentence appended — including a denial that named someone's **user** database.

That is worse than appending nothing. "Reach tempdb's space DMVs through server-level state access" sends the reader after a `##MS_ServerStateReader##` membership that would not have helped, for a denial where the raw error already names the database and the grant genuinely *is* the fix.

That difference is now stated where the two arms are, because it is the reason they are shaped differently:

- **300 is server-scoped.** On Azure SQL Database it can only ever be the service objective, so the number settles it.
- **262 names a DATABASE**, and the advice inverts on which one — in tempdb there is no grant to issue, in a user database there is. The number cannot settle it, so it reads the message.

Both call sites now pass `ex.Message`, and the arm matches the **quoted** identifier that error 262's format string interpolates (`permission denied in database '%.*ls'`). Quoted rather than bare, because `tempdb` is a reserved name and no user database can collide with `'tempdb'` in that position — while a bare match would read a database called `reporting_tempdb_stage` as tempdb. Case-insensitive, because the quoted name follows the server's collation and not this code's.

A null or absent message **stays silent rather than guessing**. That direction is deliberate: a call site that forgets the message loses a helpful sentence instead of gaining a wrong one.

Four new pins cover it — the tempdb message, a user-database message, the `reporting_tempdb_stage` look-alike, null/empty, and the three casings.

## First commit's CI

All six checks passed on `6c3867a` (`mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`), including the Windows `build` job where the flipped pin and the two new test classes run. This push re-runs them.

## #2511 / PR #2517 — composes, one assertion to reconcile

Checked rather than assumed: their `CollectorEngineCapability.IsCollectedOnEngineEdition` sweeps `TargetsWithEngineEdition` through `CollectorCatalog.AppliesTo`. **The mapping is genuinely derived — there is no enumerated gap list** — so this gate change flows through it automatically and I edited nothing of theirs. `tempdb_stats` simply drops out of the derived Azure set, taking their count 13 → 12.

The one place that needs reconciling is a pin, not the map: `Darling.Tests/EngineCapabilityMissTests.cs:480-481` and `Lite.Tests/EngineCapabilityMissTests.cs:103` assert `get_tempdb_trend` answers `not_collected` on an Azure server. After both land the correct answer is `unavailable` again, because the collector really does run and the absence really is worth looking into. Raised on #2517 with the options and left as their call rather than edited from here — landing a change that reds their open PR *and* editing their file to match would be forcing the outcome twice over.

Worth noting for its own sake: that `not_collected` branch was about to tell Azure callers tempdb is a permanent capability gap, which the measurements in #2512 show is false. Their own argument — a confident permanent-gap message is worse than a vague one when the gap is not permanent — lands on their own PR here.

erikdarlingdata:
## Second review finding fixed — `3743fa0`

Fair catch, and the sharper of the two: the doc comment claimed the drift was closed while three more hand-written copies were still sitting there, which made the claim *less* true than before rather than more.

I took the census properly this time instead of fixing the instances the first pass happened to touch:

| site | 229 | 262 | 297 | 300 | 916 | 8189 |
|---|:-:|:-:|:-:|:-:|:-:|:-:|
| `SqlServerTargetProvider.Classify` | ✓ | | ✓ | ✓ | ✓ | ✓ |
| `DarlingWorker` collector catch | ✓ | | ✓ | ✓ | | ✓ |
| Lite `RunCollectorAsync` collector catch | ✓ | | ✓ | ✓ | | ✓ |
| `DarlingWorker` failed-jobs msdb read | ✓ | | ✓ | ✓ | ✓ | |
| Lite failed-jobs msdb read | ✓ | | ✓ | ✓ | ✓ | |
| Lite XE-session-ensure catch | ✓ | | ✓ | ✓ | | |

**Six lists, four distinct sets, one question.** Not one of them had 262. All six now ask `SqlServerPermissionErrors.IsPermissionDenied`.

One set is right for all of them because they all answer the same question — *did the login get told no?* — even though they act on it differently: the collector catches record `PERMISSIONS` instead of `ERROR`, the failed-jobs reads return an empty list instead of failing the alert cycle, the XE catch marks the capture unavailable. What each one *does* with the answer is its own business; what counts as a denial is not. That is now stated on the type.

Widening is **correct** at each site, not merely tolerable, and I checked rather than assumed:

- **failed-jobs msdb reads** gain 262 and 8189. A 262 naming msdb is precisely the least-privilege case those catches exist for, and it used to fall through to the warning arm. 8189 is `sys.traces` and cannot arise from an msdb read at all.
- **the XE-session catch** gains 262, 916 and 8189. 8189 is inert there for the same reason; 262 and 916 both mean the login was denied where it asked, which is exactly the `PERMISSIONS` that arm already records for 229/297/300.

This is the second instance of one category in this PR — a rule stated in one place and transcribed into others — so it was worth taking the whole census rather than fixing the three the review named and leaving a seventh to be found later.

erikdarlingdata:
Reviewed and ready — **deliberately not merging yet.**

This PR is correct and I want it. The hold is the consequence it documents: enabling Azure tempdb collection while `UsedPercent` still divides by the *currently allocated* file size arms the `tempdb Space` alert on every Azure target, and it fires on the first busy minute. Shipping collection that immediately pages people is worse than the gap it closes, which is exactly the judgment this PR made in refusing to retune the threshold itself.

Since then the question got settled rather than deferred. Probing the same `GP_S_Gen5_2` live: tempdb is 4 ROWS files of 16 MB, each with `max_size` = 2,097,152 pages = **16 GB**, so the real ceiling is **65,536 MB**. The 57 MB `#temp` that reads 95.7% full against the allocation is **0.09% full** against the cap. So the alert is not mistuned — it is measuring distance to the next autogrow, and it is nearly right on a box only because a pre-sized tempdb has already grown to its ceiling.

#2515 now has that evidence and is being fixed as a denominator correction (`SUM(max_size)` over ROWS files where it is a real value; today's behaviour where growth is unlimited). **This merges immediately after it.**

Two things in here I want to note separately, because they are worth more than the gate:

- **Error 262 was classified `Unclassified`, so a permission denial logged an ERROR every cycle, forever.** That is precisely the log pollution the gate was supposed to prevent — the gate was hiding a bug rather than avoiding one.
- **That error-number set lived in six hand-written places that had drifted three ways.** Now one shared predicate, which is also the only way to pin it, since `SqlException` cannot be constructed in a test.

And thank you for checking #2511's map rather than assuming: confirming it is genuinely derived from `CollectorCatalog.AppliesTo` is what makes this gate change flow through automatically instead of needing a hand-edit on their branch.

erikdarlingdata:
#2517 merged first (`f5a365eb`), so this is now `CONFLICTING` and needs a `dev` merge. When you take it, **two lines here will go red** — flagging them with the exact patch so it costs you nothing.

`EngineCapabilityReadWiringTests.EveryWiredRead_NamesARealCollectorThatIsGatedOffOnAzureSqlDb` asserts that every read which asks the engine-capability question names a collector that is *actually* gated off on Azure SQL Database. Once you remove `TempDbStatsCollector.AppliesTo`'s `!IsAzureSqlDb`, `get_tempdb_trend`'s question becomes dead code and the pin says so. That is the pin working — it exists to stop a stale permanent-gap claim outliving its gate — but it means the un-wiring has to land with the gate change.

**`Darling/PerformanceMonitor.Darling.Service/Mcp/DarlingMcpDataTools.cs`** (~line 396):

```diff
             if (rows.Count == 0)
-                return await DarlingEngineCapability.NotCollectedStatusAsync(postgres, resolved.ServerId, resolved.ServerName, "tempdb_stats")
-                    ?? McpHelpers.Status("unavailable", "No TempDB data available.");
+                return McpHelpers.Status("unavailable", "No TempDB data available.");
```

**`Lite/Mcp/McpTempDbTools.cs`** (~line 31):

```diff
             {
-                return await McpEngineCapability.NotCollectedStatusAsync(dataService, resolved.ServerId, resolved.ServerName, "tempdb_stats")
-                    ?? McpHelpers.Status("unavailable", "No TempDB data available.");
+                return McpHelpers.Status("unavailable", "No TempDB data available.");
             }
```

Nothing else of #2511's needs touching: the derivation is by-sweep, so `tempdb_stats` simply drops out of the Azure gap set on its own, and `CapturePathByCollector["tempdb_stats"]`… **is** the one other thing — `EveryCapturePathEntry_NamesARealCollectorThatIsActuallyGatedSomewhere` asserts every prose entry describes a collector that is still gated. So also drop this line from `PerformanceMonitor.Collectors/CollectorEngineCapability.cs`:

```diff
-            ["tempdb_stats"] = "the tempdb space-usage snapshot",
```

Three lines total. My test files no longer reference `get_tempdb_trend` at all (I swapped them to `get_default_trace_events` in 5f83a19d precisely so they would not be in your way), so nothing else in #2511's suite moves.

**Why in your PR rather than a follow-up from me**, having offered the other way round earlier: mine merged first, so splitting it would leave `dev` red between your merge and my follow-up, and a red `dev` is worse than either of us carrying three lines. If you would rather not, say so and I will land them the moment yours goes in — but the coherent change is one PR that opens the gate and retires the claim that depended on it.

And to close the loop on my earlier note: you were right that the gate's stated reason was false, and #2511 was about to convert that into a message telling Azure callers a working DMV "is not collected for this server, and never will". Good catch, and it is a better illustration of #2511's own thesis than anything in my PR body.

erikdarlingdata:
#2515 is implemented in #2521, so the caveat this PR carries is now resolved rather than deferred: **the ratio did not need a size floor, it needed a different denominator.** `tempdb_stats` collects `max_size_mb` = `SUM(max_size)` over tempdb's ROWS files, and `TempDbSpaceInfo.UsedPercent` divides by that where there is a real cap. On `GP_S_Gen5_2` the `#temp` measurement recorded here — 59.75 MB reserved, 2.69 MB unallocated — goes from **95.7%** against the allocation to **0.09%** against the 65,536 MB cap, so it no longer clears the 80% default. Where any data file is unlimited (`-1`) there is no ceiling and the allocation stays the denominator, which is today's behaviour, so nothing an unlimited-growth on-prem or RDS target reports moves.

**This PR is safe to ship, and it should go after #2521** — or at least not before it, since until #2521 lands, enabling Azure collection arms the alert this PR's own doc comment warns about.

**Three things to expect when the two meet**, from a trial merge of this branch into `fix/2515-tempdb-ceiling-denominator`. Only the first is a git conflict; the other two merge clean and are wrong afterwards:

1. **`TempDbStatsCollector.cs` class summary** — a genuine conflict, both sides rewrote it. Keep this PR's "applies to every SQL Server target" sentence *and* #2521's ceiling paragraph.
2. **`AppliesTo`'s doc comment merges cleanly and is then out of date.** The `unallocated_mb` paragraph says the ratio "needs a size floor before it can safely carry the `tempdb Space` alert — filed as #2515". It got a denominator instead, and that paragraph should say so.
3. **`Lite.Tests/TempDbStatsCollectorDefinitionTests.AzureSqlDb_MeasuredValues_ComposeThroughToThePayload` merges cleanly and then does not compile.** It builds a five-column first result set and an eight-argument `Row`; #2521 appends `max_size_mb` to both, so it needs a sixth column and a ninth `Row`/writer value. Worth knowing before the merge rather than from a red build after it.

(Plus the usual CHANGELOG link-definition block — both branches add `[#2512]:` and `[#2515]:`; keep one copy of each.)

PR #2524, opened 2026-08-22
A fresh store never seeds, so the control plane silently overrides darling.json

erikdarlingdata:
Found by running the shipped nightly container against a live Azure SQL Database — not by reading code. This is the first-run path of every container deployment.

## What happens

```
Could not seed the config store from darling.json — running on the file config;
the store will seed on a later start: 42601: INSERT has more target columns than expressions
```

`config_service` names **sixteen** columns and supplies **fifteen** values. `$12` is bound on the command and never written into the `VALUES` list, so everything after it shifts one column left — `'seed'` lands on `updated_at`, and `updated_by` gets nothing:

| column | got |
|---|---|
| `compose_statement_timeout_seconds` | `0` |
| `config_version` | `$6` |
| `updated_at` | `'seed'` |
| `updated_by` | *nothing* |

## Why it has been shipping unnoticed

The failure is **caught and logged**, and the service carries on with the file config. That looks survivable. It isn't, because of what runs next: the control plane is authoritative after first contact, and it was answering from a row that had never been written — where every toggle is `false`.

So on a fresh store:

```
Web dashboard configuration disagrees across the two planes and the CONTROL PLANE WINS:
  enabled is true in darling.json (web.enabled) but false in config.config_service.web_enabled
MCP configuration disagrees ... enabled is true in darling.json (mcp.enabled) but false ...
```

Both endpoints then never listen. Verified: `curl` to `:5153` and `:5152` returns nothing at all. **A container user who sets `enabled: true` gets no web dashboard and no MCP, and the only clue is one logged line that says the config will "seed on a later start" — it will not, because the statement is malformed, not racing.**

The control-plane-wins behaviour is correct and deliberate. The bug is that it was deciding from a row the seed failed to write.

## The fix

`$12` restored to the `VALUES` list. Sixteen columns, sixteen values, and the mapping becomes coherent — `compose_statement_timeout_seconds ← $12`, `config_version ← 0`, `updated_at ← $6`, `updated_by ← 'seed'`.

## The pin, and why it reads text

No test drives the seed against an empty store, because the seed methods are `private` and reaching them needs a live one — which is exactly how this went unpinned. Arity is a property of the text, so the text is what gets read:

1. **Every seed INSERT supplies as many values as it names columns.**
2. **Every seed INSERT references every positional parameter it binds.** This catches the precise shape here: a hole in `$1..$max` silently drops a computed value and shifts everything after it onto the wrong column.

Covers all four seed statements, and asserts it parsed at least four — a regex that stops matching fails rather than passing for free.

## Verification

- Pin **green** on the fix, **red** on the bug, naming `updated_by` as the first unmatched column.
- Independently, both statements prepared against a live Postgres: the old one reproduces `42601`, the fixed one prepares.
- Service and test projects build clean.

### DIFF
--- a/Darling/Darling.Tests/ConfigSeedStatementArityTests.cs
+++ b/Darling/Darling.Tests/ConfigSeedStatementArityTests.cs
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) 2026 Erik Darling, Darling Data LLC
+ *
+ * This file is part of the SQL Server Performance Monitor.
+ *
+ * Licensed under the MIT License. See LICENSE file in the project root for full license information.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
+using Xunit;
+
+namespace Darling.Tests;
+
+/// <summary>
+/// The config-store seed statements agree with themselves: as many values as columns, and every
+/// parameter the command supplies is actually referenced by the SQL.
+///
+/// <para>Written after a live first-run failed with <c>42601: INSERT has more target columns than
+/// expressions</c>. <c>config_service</c> named sixteen columns and supplied fifteen values, because
+/// <c>$12</c> was bound on the command and never written into the VALUES list. Every subsequent value
+/// shifted one column left, so <c>'seed'</c> landed on <c>updated_at</c> and <c>updated_by</c> had
+/// nothing at all.</para>
+///
+/// <para><b>What that cost, which is why this is worth a pin.</b> The seed is how a FRESH store learns
+/// the file's settings. It failed, the service logged it and carried on with the file config — and then
+/// the control plane, which is authoritative after first contact, answered from an unseeded row where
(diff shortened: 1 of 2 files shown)


erikdarlingdata:
**Verified end to end against a fresh store, on the shipped nightly image built from `dev` after this merged.** Same compose stack, same config, `docker compose down -v` first so the store really is new.

| | before | after |
|---|---|---|
| seed | `42601: INSERT has more target columns than expressions` | no error |
| "control plane wins" disagreements | 2 (web + MCP) | **0** |
| web `:5153` | HTTP 000 — not listening | **HTTP 200** |
| MCP `:5152` | HTTP 000 — not listening | **HTTP 405** (correct for a GET; listening and authenticating) |

And the row itself:

```
web_enabled=true  mcp_enabled=true  updated_by=seed
```

`updated_by=seed` is the definitive bit — `'seed'` is the literal that previously landed on `updated_at` and left `updated_by` with nothing. It is in the right column now, and both toggles carry the values from `darling.json` rather than the false defaults of a row that was never written.

Issue #2532, opened 2026-08-22
Engine-kind capability: the PostgreSQL AppliesTo axis, and the reads that still cannot answer


erikdarlingdata:
Two deferrals from #2531 (the #2530 foundation), both about the same mechanism, both filed rather than smuggled in because each needs a guard that does not exist yet.

## 1. The PostgreSQL `AppliesTo` axis, and the guard it needs first

#2531 taught `CollectorEngineCapability` the engine-KIND axis, but only the **engine half** of the dispatch gate: "a SQL Server collector is never sent at a PostgreSQL target, and vice versa". The PostgreSQL collectors' own `AppliesTo` gates still answer `unavailable`, exactly as msdb access does on the SQL Server side.

That leaves a real claim on the table. Now that the store records `aurora-postgres` distinctly from `postgres`, Aurora-ness is **fixed by the engine kind** the way `IsAzureSqlDb` / `IsAzureManagedInstance` are fixed by the engine edition. So `pg_wait_stats` — which reads `aurora_stat_system_waits()`, a surface core PostgreSQL has in **no** version — is a genuine permanent gap on a stock PostgreSQL target, and today its read tells the operator to go check that collection is running.

The mechanism is the exact mirror of the edition axis:

- a `TargetsWithEngineKind(engineKind)` sweep, with `IsAurora` **fixed by the kind** and `PostgresMajorVersion` / `PostgresVersionNum` / `IsInRecovery` **varied** (all three are fixable — an upgrade moves a floor, a writer connection moves recovery state — so a permanence claim over them would be the #2511 over-claim one engine over);
- `IsCollectedOnEngineKind` asking `CollectorCatalog.AppliesTo` over that sweep instead of `EngineMatches` alone;
- capture-path prose entries for the PostgreSQL collectors that are genuinely gated, so the message is specific rather than "the data this read is served from".

[...]

Issue #2530, opened 2026-08-22
PostgreSQL targets have no UI on either SKU, and both UIs render SQL Server tabs at them


erikdarlingdata:
The stated goal is as much PostgreSQL parity as we can muster. It is currently an afterthought in both the WPF viewer and the web dashboard, and this issue is the parent for closing that.

## What exists, which is the expensive half

Eight collectors — roughly 1,900 lines — already collect and store: `PgAutovacuumStats`, `PgBlocking`, `PgIoStats`, `PgReplicationSlots`, `PgStatementStats`, `PgWaitStats`, `PgWraparoundStats`, `PgXminHorizon`. Eight MCP reads already serve it: `get_pg_autovacuum_health`, `get_pg_blocking`, `get_pg_io_stats`, `get_pg_replication_slots`, `get_pg_top_queries`, `get_pg_wait_stats`, `get_pg_wraparound_risk`, `get_pg_xmin_horizon`.

**The data is in the store right now.** This is a presentation gap, not a collection gap — the same shape as #2484, one layer up: missing UI rather than missing endpoint.

## What a PostgreSQL target actually looks like today

Verified, not assumed:

- **Neither UI has any concept of engine kind.** `git grep` for `IsPostgres` / `PostgresMajorVersion` / `IsAurora` returns **zero** files in the viewer, and the web's `server-tabs.js` has no engine branch at all. `SERVER_TABS` is a flat list applied to every server.
- So an Aurora target renders **twelve SQL Server tabs** — Wait Stats, CPU, Memory, Blocking, File I/O, Queries, Configuration, Config Changes, Activity, System Events, Collection Health — nearly all permanently empty, several meaningless for the engine (tempdb, Trace Flags, Query Store).
- The `servers` table records `sql_engine_edition` and `sql_major_version` and **nothing that says a target is PostgreSQL**. A pg target lands with edition `0`.
- #2511's capability machinery treats `0` as `UnknownEngineEdition` and deliberately **makes no claim** — correct, and it means pg reads fall through to `unavailable` ("never collected, check that collection is running"). True, and pointing at the wrong cause: those collectors do not apply to this engine at all.
- The eight `get_pg_*` reads are reachable **only through MCP**. An agent can ask about Aurora; a human gets no screen.

## The work, in dependency order

[...]

Issue #2518, opened 2026-08-22
The engine-capability derivation is never proven to FOLLOW a gate that moves


erikdarlingdata:
#2511 made the engine→capability answer **derived**: `CollectorEngineCapability.IsCollectedOnEngineEdition` sweeps `TargetsWithEngineEdition` through `CollectorCatalog.AppliesTo` rather than consulting a list. Every test that exists for it asserts a *property* of that derivation against the collectors as they are shipped today.

**What none of them proves is that the answer MOVES when a gate moves.** Every current assertion would pass just as well against a hard-coded set that happened to match today's gates — which is precisely the failure mode the derivation exists to prevent. The two guards that come closest both stop short:

- `EveryWiredRead_NamesARealCollectorThatIsGatedOffOnAzureSqlDb` catches a read whose collector stopped being gated, but only for the collectors reads are wired to, and only in the "gate opened" direction.
- `CollectorEngineCapabilityTests` pins properties (a fixable gate is not an engine gap, a version floor is not an engine gap, box editions have zero gaps) against real definitions, all of whose gates are fixed at test time.

This nearly cost something already. #2512/#2516 moved `TempDbStatsCollector`'s gate while #2511 was open, and the obvious move was to use that as the proof — but pinning the derivation with the one collector whose gate is in flight is pinning the wrong thing, and #2511 deliberately swapped it out for a gate that is not moving. So the gap is real, it was noticed by two lanes independently, and it is still open.

## The shape

A **fake** `ICollectorSchemaInfo` whose `AppliesTo` the test controls, run through the same sweep — assert it reports a gap, flip the gate, assert the same call now reports none. Nothing real moves, and it is the only version of this that is a genuine test rather than a restatement.

That needs one new entry point, because the sweep currently goes through `CollectorCatalog` by NAME and the catalog only knows real definitions:

```csharp
public static bool IsCollectedOnEngineEdition(ICollectorSchemaInfo definition, int engineEdition)
```

with the existing by-name overload delegating to it. Today that overload would have exactly one caller (the test), which is why it was not added speculatively in #2511 — but it is a small, honest seam, and it is the same by-name / by-definition pair `CollectorCatalog.AppliesTo` already carries.

## Worth checking while in there

[...]

PR #2536, opened 2026-08-22
Sweep the PostgreSQL AppliesTo gates too, and ship the IL guard first (#2532)

erikdarlingdata:
Half of #2532: the PostgreSQL `AppliesTo` axis, with the guard #2532 named as its prerequisite. The other half — wiring the remaining SQL Server reads on both SKUs — is a separate PR against `dev`, because the read wiring is the user-visible half and should not wait on this.

## What was wrong

#2530 taught the miss vocabulary the engine-KIND axis, but only the **engine half** of the dispatch gate: "a SQL Server collector is never sent at a PostgreSQL target, and vice versa". The PostgreSQL collectors' own `AppliesTo` gates were invisible to it, and **none of the eight `get_pg_*` reads asked the capability question at all**.

Five of those eight do not even use the `unavailable` vocabulary — they return a bespoke all-clear. So a SQL Server target asked `get_pg_xmin_horizon` got:

> `no_holder` — "Nothing is holding back the xmin horizon in this window. Vacuum is free to reclaim dead rows, so bloat growth has a different cause…"

and `get_pg_autovacuum_health` reported `no_pending_maintenance` — "No table on this server has dead tuples". That is not a weak answer about a mechanism the engine does not have; it is a fabricated clean bill of health, and it is exactly what an agent asking by name over MCP gets, because an agent has no tabs.

## The sweep

`CollectorEngineCapability.TargetsWithEngineKind(engineKind)` is the engine-kind twin of `TargetsWithEngineEdition`, and it splits the facts the same way:

| fact | role | why |
|---|---|---|
| `IsAurora` | **fixed by the kind** | nothing turns a stock PostgreSQL server into an Aurora one — the same standing `IsAzureSqlDb` has on the edition axis |
| `PostgresMajorVersion` | varied | an upgrade crosses a floor |
| `PostgresVersionNum` | varied | its own dimension, because the gates that will read it are MINOR-version gates (`aurora_stat_resource_usage()` needs 16.9+/17.5+ and is absent on 17.4) |
| `IsInRecovery` | varied | a writer connection leaves recovery |

`IsCollectedOnEngineKind` now asks `CollectorCatalog.AppliesTo` over that sweep instead of `EngineMatches` alone. The SQL Server arm of the sweep **varies** the two Azure flags where the edition sweep fixes them, which is what keeps the axes from answering each other's question: a collector gated off on Azure SQL Database still runs on some SQL Server, so it is not a kind gap, and the edition axis is left to say so with the edition named.

Measured over the shipped catalog (42 SQL Server / 8 PostgreSQL definitions):

- `pg_wait_stats` and `pg_statement_stats` — `aurora_stat_system_waits()` and `aurora_stat_statements()`, surfaces core PostgreSQL has in **no** version — are permanent gaps on stock PostgreSQL, and on nothing else;
- `pg_stat_io`'s PG16 floor and the writer-only autovacuum read are **not**, and keep the `unavailable` vocabulary that correctly sends an operator to look;
- the dialect split is unmoved in both directions, and an unknown/absent/unrecognised token still claims nothing for anyone.

**The cross product deliberately includes incoherent pairs** (major 16 alongside `server_version_num` 170005). A superset of the real target shapes can only find MORE shapes that pass a gate, so its error direction is *under*-claiming — the read keeps `unavailable`, which is the status quo. The opposite error is the one this whole mechanism exists to prevent, and a superset cannot make it.

## The message names the real reason

There are two shapes of kind gap and they are not the same thing:

> **not_collected** — aurora-01 runs Aurora PostgreSQL. The `system_health_events` collector is written against SQL Server and **the dispatch gate's engine half never sends it at another engine**, so this server does not collect the system_health extended-events ring buffer, and never will. …

> **not_collected** — pg-01 runs PostgreSQL. The `pg_wait_stats` collector does not run on that engine — **its own AppliesTo gate excludes it** — so this server does not collect the aurora_stat_system_waits() cumulative wait counters, and never will. …

The second borrows the edition axis's wording verbatim past the engine name, because they are the same finding. One sentence for both would tell a PostgreSQL operator that a PostgreSQL collector is not written for PostgreSQL, which is the sort of wrong that costs a reader their trust in the rest of the message.

## The guard, which is why this could ship

#2532 said explicitly this must not ship without the PostgreSQL twin of #2518's IL-scanning sweep-dimension guard, and that machinery generalised exactly as it predicted — `FactsReadBySqlServerGates` became `FactsReadByGatesOf(engine)`, `RoleOf` became parameterised over the shape families, and the two axes now share one decoder.

`EveryFactAPostgresGateReads_IsVariedBySweepOrFixedByKind` decodes the IL of every PostgreSQL definition's `AppliesTo`, records which `CollectorTargetInfo` getters it calls, and fails if any of them names a fact the kind sweep leaves at its CLR default. Both halves are read out of shipped artifacts — the facts from the gates' IL, the roles from running the sweep — so neither can be brought into line by editing the test.

Alongside it: `ThePostgresGateScan_ReadsTheGatesTheSourceActuallyCarries` (including `pg_blocking`, which reads nothing, so the decoder is shown to discriminate — and an assertion that no SQL Server fact leaks in, because a filter that stopped filtering would look exactly like a guard that was working), `TheKindSweep_IsTheFullCrossProductOfTheDimensionsItVaries`, `EveryTargetFact_LandsInExactlyOneDerivedKindRole`, and `AnUnrecognisedKind_SweepsNothing_AndStillClaimsNothing` — the last because an empty sweep read as a gap set is the one failure here that would be catastrophic rather than merely wrong.

## Proven red

The suites are `net10.0-windows` and cannot run on macOS, so CI is the arbiter — but every assertion above was run locally against the real built `PerformanceMonitor.Collectors` in a throwaway `net10.0` harness, over the shipped catalog and the shipped MCP source. **89 checks green.**

Then, red on demand:

- **the guard**: fixing `IsInRecovery` at `false` in the sweep turns `EveryFactAPostgresGateReads_IsVariedBySweepOrFixedByKind` red naming `IsInRecovery — read by pg_autovacuum_stats`, and `IsInRecovery` slides from `VariedBySweep` to `ConstantEverywhere`. Restoring it turns both green. That is the exact failure #2518 exists for, reproduced one engine over: without it, `pg_autovacuum_stats` would have been announced as a permanent gap on both PostgreSQL tokens.
- **the derivation**: a synthetic definition whose gate the harness owns is moved through `IsAurora`, `!IsAurora`, `PostgresMajorVersion >= 16`, `PostgresVersionNum >= 170005`, `!IsInRecovery`, `_ => false` and `_ => true`, and the answer follows every move in both directions. Only the two Aurora-shaped gates produce a permanent gap, and each produces it on exactly one token.

## Reads wired here

All eight `get_pg_*` reads, Darling only — Lite has no PostgreSQL tools and no PostgreSQL target seam, and `McpEngineCapability` still passes `engineKind: null` explicitly, which I re-checked rather than assumed. Every call sits on the **miss path**, after the read came back empty, so a server whose registry row disagrees with its collected rows still gets its data. `get_pg_blocking` asks only when `captures_total == 0`: a window with captures is a window the collector ran in, so the engine cannot be in question there.

`CapturePathByCollector` gains the eight PostgreSQL noun phrases, so the messages name a surface rather than "the data this read is served from".

## Two tests changed meaning, deliberately

- `AShutAppliesToGate_IsNotAnEngineKindGap` → `AShutAppliesToGate_IsAnEngineKindGap_ButAFixableOneIsNot`. The narrowing it protected — never report a fixable gate as permanent — is unchanged and is now enforced by the sweep on its own merits rather than by the axis refusing to look; the second half of the test is exactly that.
- `EveryWiredRead_NamesARealCollectorThatIsGatedOffOnAzureSqlDb` → `…NamesARealCollectorWhoseCapabilityBranchCanFire`. **This is weaker and the doc comment says so.** The Azure-specific claim was right while the edition axis was the only one; the kind axis makes every SQL Server collector a permanent gap on a PostgreSQL target, so demanding an Azure gap would refuse the wiring this issue is about. What survives is the property that matters at a call site — the branch is reachable — and the sharper claims stay in the derivation tests, over the gates rather than over the reads.
- `EveryCapturePathEntry_NamesARealCollectorThatIsActuallyGatedSomewhere` keeps its teeth by excluding the dialect half: an entry must name a collector its **own** gate shuts out somewhere, **or** one a shipped read asks about. A check phrased as "is this a gap somewhere" would now pass for any name anyone typed.

## One thing in #2532 that I could not confirm

#2532's own gate on this work was: *"the claim to check before starting is whether any monitored PostgreSQL target is not Aurora — if none is and none is planned, this is theory."* I could not re-measure the fleet from this session. On the stated fleet the stock-PostgreSQL claim buys nothing **today**; what it buys immediately is the other seven-eighths of this PR — the eight `get_pg_*` reads no longer answering a SQL Server target with a fabricated all-clear — and the guard, which is what makes the sweep safe to have shipped at all.

Closes nothing on its own; #2530 stays open for the UI waves, and #2532 stays open for the read wiring.


erikdarlingdata:
Good catch, and it is the same discipline this PR is about, so it should not have shipped with the disjuncts left in. Fixed in 006540cb.

All three fallbacks now name only the states that can still reach them:

- **`get_pg_wait_stats`** — an Aurora target with no rows in the window, or a row whose `engine_kind` is NULL. The "if it is PostgreSQL but not Aurora" half is exactly what the capability answer now says, permanently, so repeating it here was offering a live-looking alternative to a settled question.
- **`get_pg_top_queries`** — the one genuinely diagnosable Aurora cause (`pg_stat_statements` missing from the connected database), plus the NULL-kind case.
- **`get_pg_wraparound_risk`** — this collector has no `AppliesTo` gate at all, so the capability answer catches *every* engine the store can classify and "means the server is SQL Server" was unreachable whenever `engine_kind` is known.

Each now points at `list_servers` for the NULL-kind case, which is the read that actually answers it.

PR #2537, opened 2026-08-22
Wire the engine-capability question into every read that can answer it (#2532)

erikdarlingdata:
The user-visible half of #2532. The other half — the PostgreSQL `AppliesTo` axis and the IL guard it needed first — is #2536, also against `dev`. Neither depends on the other; see **Merge order** at the bottom.

## What was wrong

#2511 wired the capability helper into the reads whose collectors are gated off on Azure SQL Database — twelve collectors' worth. That was the right scope for the **edition** axis, because a collector that runs on every SQL Server had nothing to say.

The engine-**kind** axis changes that arithmetic completely: on a PostgreSQL target *every* SQL Server read is a permanent gap, not just twelve. So `get_database_config`, `get_wait_stats`, `get_top_queries_by_cpu` and the rest still answered

> **unavailable** — No database configuration data available. The config collector may not have run yet.

…about a collector that does not run on that engine at all. True, and pointing at the wrong cause. `EngineCapabilityMissLivePostgresTests.APostgresTarget_AnswersNotCollectedNamingTheEngine_WhileAnUnprobedServerKeepsItsMiss` asserts that behaviour explicitly today, which is how the gap stayed visible rather than becoming folklore.

Once #2530's UI waves land a human pointed at a PostgreSQL target sees PostgreSQL tabs and never asks a SQL Server read. But an **agent** over MCP has no tabs — it asks by name — and agents are half the product's surface.

## What it does now

> **not_collected** — aurora-01 runs Aurora PostgreSQL. The `wait_stats` collector is written against SQL Server and the dispatch gate's engine half never sends it at another engine, so this server does not collect the sys.dm_os_wait_stats cumulative totals, and never will. This is a permanent engine capability gap, not a collection outage: checking collection health, enabling a collector or starting a capture cannot change it.

**72 Darling reads and 71 Lite reads** ask the question, up from 17 on each. Every call site sits **after** the read came back empty — never before it — so a server whose registry row says one engine while its collected rows say another still gets its DATA rather than a confident explanation of why it cannot have any. That is the property `DarlingEngineCapability`'s "called only on the miss path" note exists to protect, and it is why #2532's Option 3 (answer at the resolver) was not taken: it is cheaper and it is the one shape that can hand a mis-registered server an explanation instead of its rows.

`CapturePathByCollector` gains **28** noun phrases, so the sentence names a surface — "the sys.dm_os_wait_stats cumulative totals", "the sys.dm_io_virtual_file_stats per-file I/O totals" — instead of falling back to "the data this read is served from".

## What I deliberately did NOT wire, and why

#2532 warned against blanket-wiring 83 reads without thinking, and four groups genuinely have a better answer than `not_collected` or no answer to give:

| Not wired | Why |
|---|---|
| `analyze_server`, `get_analysis_facts`, `compare_analysis`, `get_daily_summary`, `get_daily_summary_range`, `get_server_summary` | each rolls up a dozen collectors and can name none of them. Naming one would be arbitrary and the message would be specific about the wrong thing. The honest answer is a rollup-level one — "every input to this read is a SQL Server collector" — which needs a second message shape this PR does not build. **This is the residue, and it is the one worth filing if it is not closed here.** |
| `get_collection_health`, `get_collection_log` | served by `collection_log`, which a PostgreSQL target **genuinely populates**. Their existing messages are about collection, and on an Aurora target they are correct. Wiring these would have been the same mistake in the other direction. |
| `list_servers`, `get_fleet_overview`, `get_store_metrics`, `add_servers`, `remove_server`, the six alert tools, the eight custom-view tools | resolve no single server, so there is no engine to speak for. `get_alert_history` takes an optional server and is fleet-wide by default; `get_ag_health` already handles that shape correctly and is the model if these ever need it. |
| `analyze_plan_xml` | parses XML the caller supplies. No store read, no server. |
| `get_wait_trend` / `get_perfmon_trend`'s "not a wait type / counter this server collects" branches | only reached when the collector HAS produced rows, so the engine cannot be in question there. Their sibling `unavailable` branches — "nothing collected at all" — **are** wired. |
| the eight `get_pg_*` reads | wired in #2536, where the axis they need lives. |

`get_blocked_process_reports` (Lite-only) and `get_blocking` (Darling-only) are each wired on their own SKU; they are not compared by the parity pin because the pin only looks at tools both surfaces expose.

## One signature change

Lite's `analyze_query_store_plan` fetches the plan **live** from the monitored instance rather than from the store, so it was the one plan tool with no `LocalDataService` to ask. Its three siblings all take one. I added it — DI-injected, no user-facing parameter — because the capability question is the same question and `BothSkus_WireTheSameReadsToTheSameCollectors` is right to insist the two SKUs answer it identically. On Lite it is inert today (`engineKind` is still passed as `null`, which I re-checked rather than assumed), and it cannot become wrong.

## Two pins changed meaning, both still derived

- `EveryWiredRead_NamesARealCollectorThatIsGatedOffOnAzureSqlDb` → `…NamesARealCollectorWhoseCapabilityBranchCanFire`. **This is weaker and the doc comment says so.** The Azure-specific claim was right while the edition axis was the only one; the kind axis makes every SQL Server collector a permanent gap on a PostgreSQL target, so demanding an Azure gap would *refuse the wiring this issue is about*. What survives is the property that matters at a call site — the branch is reachable on some (kind, edition) pair the store can hold — and it is derived by asking `NotCollectedMessage`, not by a list.
- `EveryCapturePathEntry_NamesARealCollectorThatIsActuallyGatedSomewhere` now requires an entry to be gated by its **own** `AppliesTo` somewhere **or** named by a wired read. Two derived arms, no hand list, and it catches dead prose in both directions: an entry for a collector nothing gates and nothing reads now fails.

`AllNineHealthParserReads_AskTheQuestion` and `BothSkus_WireTheSameReadsToTheSameCollectors` are untouched and both still hold.

## Verification

The suites are `net10.0-windows` and cannot run on macOS, so CI is the arbiter — but the wiring is a source-scan property, so all of it was run locally in a throwaway `net10.0` harness against the **shipped MCP trees and the real built `PerformanceMonitor.Collectors`**, using the same regexes the pins use:

- 72 Darling / 71 Lite wired reads found — the scan parsed something, which is the failure mode a source guard has;
- every wired collector name resolves in `CollectorCatalog`, has a `CapturePathByCollector` entry, and has a reachable capability branch;
- **zero drift** across the 55 tools both SKUs expose;
- both health-parser families still nine;
- every capture-path entry is gated-by-its-own-gate or wired;
- and the answers that must NOT change did not: `wait_stats` on a SQL Server box is still silent, and so is any collector when the kind is unknown (a pre-#2530 row, or Lite).

**Proven red:** dropping one collector's `CapturePathByCollector` entry fails the wiring pin naming the tool and the collector; pointing a wired read at a name the catalog does not carry fails the first assertion. Both were run.

Both SKUs and both test projects build clean on macOS with `EnableWindowsTargeting`.

## Merge order

#2536 and this PR both edit `CollectorEngineCapability.CapturePathByCollector` (adjacent entries) and `EngineCapabilityMissTests.cs`. **The shared hunks in the test file are byte-identical in the two branches on purpose** — the wiring-pin rewrite and the capture-path rewrite are the same edit on both sides, so they merge without a conflict whichever lands first. The only region this PR touches alone is the live-Postgres test, which #2536 leaves untouched. Just the dictionary needs the second PR to carry both sets of entries, which is a one-hunk rebase. Neither PR depends on the other's behaviour.

## The pin that caught the one real defect

`EngineCapabilityMissLivePostgresTests` deliberately pinned the LIMIT this PR removes — that `get_database_config` still answered `unavailable` on a PostgreSQL target — so the gap could not later be mistaken for a regression. It is the one check that went red here, which is the pin doing exactly what it was written for. It now asserts the closed shape, plus `get_wait_stats` beside it: a read from a family with nothing to do with configuration, whose collector is gated off on no engine edition at all, so nothing but the kind axis could ever make it speak.

Closes nothing on its own. #2530 stays open for the UI waves; #2532 stays open for the rollup residue named above.


erikdarlingdata:
Right, and the part of that worth recording is the cost argument rather than the query count: this is not a rare-path waste that amortises. On an engine with a permanent gap the branch is taken on **every** call, forever, so the extra query repeats indefinitely for that server/tool pair — and the probe could never have told a gated engine anything, because there is nothing for it to find.

Fixed in 151434ae across six reads on both SKUs:

- **`get_blocking_trend` / `get_deadlock_trend`** — the `collection_log` capture-count denominator, which is only consumed by `EmptyTrend(...)`.
- **`get_wait_trend` / `get_perfmon_trend`** — the distinct-values probe. Those needed a slightly bigger edit than a reorder: the gate had been folded into the `??` on the `collected.Count == 0` branch, so it is now a hoisted `gated` block above the probe and the fallback goes back to its original single form.

Both queries are still on the miss path, so the ordering does not touch the property that matters — the DATA read has already come back empty before either runs. The ungated case (every SQL Server target, which is all of them today) is byte-identical.

Checked the rest of the list and the other five are already in the right order: `get_memory_trend`, `get_file_io_trend`, `get_query_duration_trend`, `get_procedure_duration_trend` and `get_query_store_duration_trend` pass their existence probe to `EmptyTrendAsync` as an argument, and the `gated` block sits above that whole statement — so the probe task is never created when the gate fires. Same for `get_current_waits_trend` and `get_blocking_stats` on both SKUs.

Issue #2539, opened 2026-08-22
Collect pg_stat_database: temp-file spills, cache hit ratio, deadlocks and commit ratio in one read


erikdarlingdata:
Part of #2530. The cheapest meaningful PostgreSQL win available — one view, four questions we currently cannot answer.

## What is missing

Nothing collects `pg_stat_database`. Confirmed by grep across all eight `Pg*Collector` files. That single view carries:

- **`temp_files` / `temp_bytes`** — work_mem spills to disk. This is the PostgreSQL answer to *"why is this query slow"*, and it is the gap that matters most: we already collect `pg_stat_statements`, so we can tell a user a query is expensive and **cannot** tell them it is spilling. Direct analogue of the tempdb-spill story on the SQL Server side, which we do tell.
- **`blks_hit` / `blks_read`** — cache hit ratio. The conventional target is >99%, and it is one of the first numbers any PostgreSQL DBA looks at.
- **`deadlocks`** — a server-level counter. We collect blocking pairs but have no deadlock count for pg at all.
- **`xact_commit` / `xact_rollback`** — commit versus rollback ratio, which is how a rollback storm shows up.

Four independent needs, one cheap read, no extension required and nothing to configure on the server.

## Why this ranks above the more obvious ones

Bloat and index usage are the answers people *expect* to want. Spills are the one they need first, because a spilling query looks identical to a slow query in `pg_stat_statements` — same calls, same total time, no indication that the cause is memory rather than plan shape. Without it the statement stats we already collect are missing their most common explanation.

## Notes for whoever picks it up

- These are **cumulative counters since stats reset**, like most `pg_stat_*` surfaces, so they need the same delta treatment the existing pg collectors already apply. `PgIoStatsCollector` is the closest model.
- `stats_reset` is on the view and is worth capturing: a reset silently zeroes every delta, and a counter that went backwards should be visible as a reset rather than read as a negative rate.
- Per-database rows, so decide deliberately whether to store per-database or aggregate — the existing collectors vary and the choice should be stated rather than inherited.

Sources: [PostgreSQL Wiki — Monitoring](https://wiki.postgresql.org/wiki/Monitoring), [Datadog — Collecting metrics with PostgreSQL monitoring tools](https://www.datadoghq.com/blog/postgresql-monitoring-tools/).

erikdarlingdata:
[...]

PR #2547, opened 2026-08-22
Give a PostgreSQL target PostgreSQL tabs on the web (#2530)

erikdarlingdata:
Closes the web half of #2530. The desktop viewer is untouched and still has no PostgreSQL screens — that stays open.

## What ships

`server-tabs.js` now holds **two** registries, and `serverTabsFor(card)` is the only thing that chooses between them, from the fleet card's server-derived `is_postgres` (#2531).

| Tab | Reads |
|---|---|
| **Overview** | `get_pg_wraparound_risk`, `get_pg_xmin_horizon`, `get_pg_autovacuum_health`, `get_pg_replication_slots` (vitals as tiles) + `get_analysis_findings`, `get_collection_health`, `get_collection_log` |
| **Activity** | `get_pg_blocking` (fanned out to sampling denominator + chains + cycles), `get_pg_top_queries`, `get_pg_database_stats` |
| **Vacuum** | `get_pg_xmin_horizon`, `get_pg_autovacuum_health`, `get_pg_wraparound_risk` |
| **Waits** | `get_pg_wait_stats` |
| **I/O** | `get_pg_io_stats` |
| **Replication** | `get_pg_replication_slots` |

All **nine** `get_pg_*` reads land on exactly one tab — nine, not eight, because #2539's `get_pg_database_stats` landed on `dev` while this was in review and the pin below refused the merge until Activity showed it. Six against SQL Server's twelve is the design, not a shortfall: parity was explicitly not the constraint, and the tabs that are missing (tempdb, Query Store, trace flags, plan cache, `system_health`) have no PostgreSQL analogue to fill.

Three SQL-Server-named reads are borrowed, and only three: `get_collection_health`, `get_collection_log`, `get_analysis_findings` read the collection log and the findings store, not a SQL Server collector's output, so they answer a PostgreSQL target honestly. `get_server_summary` and `get_daily_summary` **look** engine-neutral and are not — they roll up SQL Server wait types, memory pressure and deadlocks, and would answer `unavailable` with a sentence about a collector that will never run. A pin names the three explicitly for that reason.

**Vacuum is one tab.** xmin holds the horizon, the horizon starves vacuum, vacuum falling behind ends in wraparound. The three reads sit in that causal order — horizon, backlog, headroom — under a `note` that says why, because read separately each looks survivable.

## The Waits decision: SHOWN on stock PostgreSQL, not hidden

I went the other way from the issue's inclination, and the reason is a fact about the code rather than a preference.

`get_pg_wait_stats` reads `aurora_stat_system_waits()`, so on `engine_kind = 'postgres'` it can never have content. But since #2532 it does not answer that state with a blank: `DarlingEngineCapability.NotCollectedStatusAsync` returns `not_collected` naming the server, the engine, the collector and the exact Aurora surface, and saying **"and never will"**. `classifyResponse` maps that envelope to `kind: "empty"`, `renderPanel` renders the read's own sentence, and the tab reads as an explanation rather than a blank rectangle.

That flips the argument:

1. **The defect in #2530 is unexplained emptiness, not emptiness.** Twelve SQL Server tabs at an Aurora target say nothing about why they are blank. One Waits tab that says *"…runs PostgreSQL. The pg_wait_stats collector does not run on that engine … so this server does not collect the `aurora_stat_system_waits()` cumulative wait counters, and never will"* is the opposite of that.
2. **A tab set that changes shape between two PostgreSQL servers in one fleet is its own confusion** — as the issue says. Aurora and stock would disagree about what tabs exist, and nothing on either page would explain the difference.
3. **Hiding makes the one Aurora-specific capability we have invisible from a stock instance.** Someone evaluating on RDS PostgreSQL would never learn we do Aurora wait sampling.
4. **Hiding is browser-side re-derivation of a capability the server already decides, in prose.** This repo's R1 rule is that the browser does not re-derive; `is_aurora → hide the tab` is exactly that, and it would have to be maintained beside `CollectorEngineCapability`'s gate.

The tab carries a `note` stating the Aurora dependency plainly, so a reader who never clicks the panel still learns it. §3.4 of the UAT doc calls it out too.

## NULL `engine_kind`

Only a **positive** claim switches:

```js
return card && card.is_postgres === true ? POSTGRES_TABS : SERVER_TABS;
```

`is_postgres` is derived server-side through `MonitoredEngineKind`, and is false for a NULL kind, an unrecognised token, and a card that never arrived — none of which is evidence for either engine. Falling back to the SQL Server registry there is the pre-#2530 behaviour, unchanged. The unclaimed population is servers that have not connected since the V82 rung landed, and guessing PostgreSQL from an absence would break every one of them. The `=== true` is load-bearing and pinned — a truthy non-boolean must not pass.

The header's new engine badge follows the same asymmetry: a card with no `engine_kind` gets **no badge**, rather than one reading "SQL Server". The tabs it gets are a default, not a finding.

## Evidence the reads are genuinely fetched

`node --check` on all three modules, plus a DOM shim run (scratch, not committed) that loads the **shipped** modules and builds every tab of both registries against a stubbed `fetch`, at three windows, across four response shapes — envelope, error, data-with-no-rows, and the awkward data-carrying-a-`finding`-and-no-summary-keys body several pg reads answer their healthy case with:

```
registries: SERVER_TABS=12 tabs, POSTGRES_TABS=6 tabs
SERVER_TABS  fetched 72 distinct reads
POSTGRES_TABS fetched 12 distinct reads: get_analysis_findings, get_collection_health,
  get_collection_log, get_pg_autovacuum_health, get_pg_blocking, get_pg_database_stats,
  get_pg_io_stats, get_pg_replication_slots, get_pg_top_queries, get_pg_wait_stats,
  get_pg_wraparound_risk, get_pg_xmin_horizon
  overview     -> get_pg_wraparound_risk, get_pg_xmin_horizon, get_pg_autovacuum_health,
                  get_pg_replication_slots, get_analysis_findings, get_collection_health,
                  get_collection_log
  activity     -> get_pg_blocking, get_pg_top_queries, get_pg_database_stats
  vacuum       -> get_pg_xmin_horizon, get_pg_autovacuum_health, get_pg_wraparound_risk
  waits        -> get_pg_wait_stats
  io           -> get_pg_io_stats
  replication  -> get_pg_replication_slots
PASS
```

Those are recorded **requests**, not descriptor names. The harness also asserts no node renders as the literal `[object Object]`, no tab fetches a read twice, every request carries a server and the canonical `hours` key, and every `serverTabsFor` fallback case lands on `SERVER_TABS`.

**Proved it fails without the fix**: swapping one `table(...)` call for a bare object literal produces

```
FAIL (6):
  - POSTGRES_TABS/waits @24h: build() returned a NON-NODE ({"title":"Wait Events", …})
    — a bare descriptor object renders as [object Object] and never fetches
  - POSTGRES_TABS never FETCHED get_pg_wait_stats
```

The twelve pins in `ServerPageTabsTests` were also **executed** on macOS against the compiled `Darling.Tests.dll` (a throwaway net10.0 host that loads the assembly by type name — enumerating types drags in PresentationFramework, which is why it is by name), plus `MonitoredEngineKindStoreTests`, `FleetPageAttentionFilterTests` and `ViewTemplatesTests`. All pass. Each new pin was proved red first:

| Mutation | Pin that went red |
|---|---|
| drop `get_pg_wait_stats` from `POSTGRES_TABS` | `EveryPostgresRead_IsReachableFromThePostgresRegistry_AndFromNoOtherTab` — *"the service serves these PostgreSQL reads and no PostgreSQL tab shows them…"* |
| `is_postgres === true` → `is_postgres` | `TheTabSet_SwitchesOnlyOnAPositivePostgresClaim` |
| remove the `vizStat` empty guard | `EveryDataPanel_ExplainsItsOwnEmptyState` |

## What happened to the existing pins

I read each one for what it was asserting before changing it; none was relaxed to pass.

- **`NoPostgresRead_IsOnTheServerPage_UntilTheTabRegistryBranchesOnTheEngine`** → **`EveryPostgresRead_IsReachableFromThePostgresRegistry_AndFromNoOtherTab`**. The old pin refused *any* pg read while the registry was flat — correct then. The placement half survives verbatim (a pg read in `SERVER_TABS` is the same defect it always was), and a **second direction is new and derived from `BuildReadDispatch()`**: add a ninth `get_pg_*` read and this goes red naming it. That is the mechanism that stops the graphical surface silently falling behind the reads a second time, which is exactly how the first eight spent three releases MCP-only — **and it fired three hours after it was written**, on #2539's `get_pg_database_stats`. See the merge note below.
- **`TheSubTabBar_IsBuiltFromTheRegistry_AndTheRouterCarriesTheTab`**: the `Assert.Equal(12, ids.Length)` exact count was written because prose drifted from the registry at "eleven against twelve". It is now **two** exact counts, one per registry region, so the same protection covers the second number the prose gained. Ids are unique *within* a registry; `overview` / `activity` / `waits` / `io` collide *across* them deliberately, so those deep links survive a server turning out to be the other engine.
- **`NoTab_FetchesTheSameReadTwice`**: now scans per registry. Its blocks are delimited by the next tab's `id:`, so a single whole-file sweep would have run the SQL Server registry's last tab straight through the PostgreSQL registry's header comment and attributed its reads to a tab that does not fetch them.
- **`EveryDataPanel_ExplainsItsOwnEmptyState`**: gained the `vizStat` guard alongside the `vizTable` / `vizLine` ones it already had.

## Two supporting changes

**`vizStat` gained `vizLine`'s `emptyText` guard.** `get_pg_xmin_horizon`, `get_pg_autovacuum_health`, `get_pg_io_stats`, `get_pg_replication_slots` and `get_pg_blocking` answer their *healthy* case with `{server, hours_back, status, finding}` — prose under `finding`, none of the summary keys, and **not** the `{status, message}` envelope, so `classifyResponse` calls it data, it reaches a viz, and a tile set renders as a row of em-dashes saying nothing. The guard fires at exactly all-null and only for descriptors that opt in, so every stored view and every SQL Server tile is unchanged. `stat()` grew an optional 7th parameter to carry it (optional, not required like `table()`'s, because ~30 SQL Server tiles over reads that always return their object cannot reach the case).

**`FleetServerCard.engine_description`.** Derived on the DTO through `MonitoredEngineKind.DescribeEngineKind`, null for an absent kind. The header badge needs the engine's *name*, and my first cut mapped the three tokens to three words in JavaScript — a second description table in a language the first one cannot be shared with, which is what `DescribeEngineKind`'s own comment says drifts. Derived rather than settable so no future card builder can forget it. A pin asserts the browser contains no `card.engine_kind`, no `"aurora-postgres"`, and no `card.is_postgres`.

## Docs

`docs/uat-onboarding.md` §3.4 rewritten. Every count **re-derived from source**, not incremented: 93 dispatch entries (`BuildReadDispatch`) and 112 `McpServerTool` declarations, 72 reads fetched by `SERVER_TABS` and 9 more by `POSTGRES_TABS` → **81 on the per-server page, 82 across all five pages**. The derivation reproduces the old 72/73 exactly on the pre-change tree, which is why I trust it. §3.4's *"92 read endpoints out of 111 tools"* was left stale by #2539 and is corrected here too.

## What could not be verified without a browser

- **No pixels.** Column widths, the two-column grid at six tabs, the tab strip's horizontal scroll at six labels rather than twelve, and the new `.badge.engine` against both themes are all unexercised. The shim asserts structure and fetches, never layout.
- **No live PostgreSQL target.** Every response in the harness is synthetic. The column *keys* were read out of the nine tool sources rather than transcribed from a sample, and the reads' own tests cover their shapes, but a real the PostgreSQL monitoring host render against `looker-poc` is the check that has not happened. Specifically unverified against real data: the `-1` sentinel actually appearing in `root_xact_duration_ms`, `writes` rendering blank (not `0`) on Aurora, and `threshold_ratio` being null on a never-analyzed table.
- **The loading-strip window on a FIRST visit.** Only the first render of a given server waits for `/api/fleet` (see the review round below — repeat renders paint from the remembered card and never blank). How long that one strip is visible on a large fleet is still a real-browser question.
- **Windows test suites** did not run locally (net10.0-windows). CI is the arbiter; the twelve pins were executed via the reflection host described above, which is not the xUnit runner.

## Filed separately

`queryid` is a PostgreSQL `int8` and routinely exceeds 2^53, so `JSON.parse` rounds it — the Query ID column can render a value that will not match `pg_stat_statements`. Pre-existing in the MCP contract (any JSON parser using doubles has the same problem), not introduced here, so it is filed rather than fixed in this PR.

## Review rounds

Three findings from `claude[bot]`, all real, all fixed with a harness proved red first. Two of the three were regressions this PR introduced rather than defects it exposed, and both are the same shape: a caller that never learned the thing it calls had changed what it writes.

| Finding | Fix |
|---|---|
| **Stale-render race.** Moving the tab-set decision into the async callback made it write *module* state (`current`, and the grid `redrawPanels()` reads) where it had only ever touched nodes captured in its own closure. Two renders in flight → the last response wins regardless of what is on screen. Reproduced: **7 reads for `server-A` fired into `server-B`'s page**, PostgreSQL reads under a SQL Server header. | Generation counter (628e4bf). Not an `AbortController` — the losing render must not cancel the shared `/api/fleet` fetch out from under the winner. |
| **The 60s poll blanked the page.** `route()` re-renders on every sub-tab click and every poll; gating both on a fetch the page needs to make *once* blanked the bar and grid once a minute and made a tab click wait on a read it does not need. Reproduced, with a symptom review did not name: **every poll was refetching the tab twice**. | `lastCard` remembers the card by name; a repeat render paints synchronously, and the fresh card rebuilds the bar/grid only if it chooses a *different* registry. A card that does not arrive leaves a painted page alone (f0183cb). |
| **Badge register.** `DescribeEngineKind`'s `"an unrecognised engine"` is worded to sit mid-sentence in the capability messages; as a badge beside "SQL Server" it is the wrong string, not the wrong case. | `engine_description` has three answers: null (absent), the describer's words (recognised), **the raw token** (unrecognised) — the string an operator would search their store for (f0183cb). |

Two more scratch harnesses came out of it, both driving the shipped `server.js`: one holds two `/api/fleet` responses and releases them out of order, one renders a server twice and inspects the page before the second response lands. Each was proved red by reverting only the guard under test.

## Merge note: the pin caught a real one, same day

`dev` moved while this was in review and the PR went `CONFLICTING`, which is worth flagging on its own: a conflicting PR **dispatches no workflows**, so `gh pr checks` showed a lone passing `check-branches` and looked exactly like a settled green. Only `mergeable`/`mergeStateStatus` told the truth.

Merged `dev` in (not rebased — a force-push orphans the inline review anchors above). One text conflict, both lanes adding a first entry under `[Unreleased]` / `Added`; kept both.

The interesting part: #2539 added a **ninth** `get_pg_*` read, `get_pg_database_stats`, and `EveryPostgresRead_IsReachableFromThePostgresRegistry_AndFromNoOtherTab` went red naming it — the pin doing exactly its job, hours after being written. Relaxing it to an enumerated eight would have been the anti-pattern it exists to prevent, so the read got a panel:

- **On Activity, directly under Top Query Shapes**, because that is the question it answers. A statement whose time makes no sense from its row count usually spilled, and `pg_stat_database`'s temp counters are the only evidence of that we collect — the statement stats cannot see it. Deadlocks and the commit/rollback split ride along from the same free read.
- **Not** the raw `blks_hit` / `blks_read` columns: `cache_hit_pct` and the read's own `cache_finding` say what they are for without two more columns of block counts to be misread as a workload measure.
- The summary tile keeps the read's own honesty about its `LIMIT` — `cache_hit_pct_of_returned` is labelled "of returned" and `limit_reached` is a tile rather than fine print — and `statistics_were_reset_in_window` sits on the tile because every total beside it is a lower bound when it is true.

`PgDatabaseStatsCollector.AppliesTo` is an unconditional `true`, so the Aurora-only note pin correctly does **not** demand a note for it.


erikdarlingdata:
Review's stale-render finding is right, and it is a defect this PR **introduced** rather than one it exposed — worth stating plainly, because the mechanism is the same shape as the seam lessons in this repo: the code each side was individually correct, and the caller never learned that the thing it calls had changed what it writes.

`enrichServerHead` used to touch only header nodes captured in its own closure, which a newer render had already detached from the document, so a late `/api/fleet` response wrote to garbage and nobody saw it. Choosing the tab set moved `current =` and `redrawPanels()` into that same callback, and both read **module** state — so the last response to land won regardless of which render was on screen.

Reproduced against the shipped module under the DOM shim: render `server-A` (Aurora), render `server-B` (SQL Server) before A's fleet fetch answers, then resolve B's and A's in that order.

```
FAIL (2):
  - the stale render issued 7 read(s) for server-A after server-B was on screen,
    e.g. /api/read/get_pg_wraparound_risk?server=server-A&hours=24
  - a PostgreSQL read fired at the SQL Server page
```

Seven reads for the wrong server, painted into the right server's grid, and PostgreSQL panels under a SQL Server header — exactly the two-servers case the review called out.

Fixed in 628e4bf with a generation counter, **not** an `AbortController`: the losing render must not cancel the shared `/api/fleet` fetch out from under the winning one. The fetch is fine; only its result is stale.

Pinned in `ServerPageTabsTests.TheTabSet_SwitchesOnlyOnAPositivePostgresClaim`, because the guard is invisible — the page looks correct without it until two renders overlap, which is precisely why it was missing.

erikdarlingdata:
Both round-2 findings are right; fixed in f0183cb.

**The 60s poll blanking the page.** This is the sharper of the two, and review has the mechanism exactly: `route()` re-renders on every sub-tab click *and* on the poll, and I gated both on a fetch the page only genuinely needs to make **once**. The engine is a property of the server, not of the render.

`lastCard` now remembers the card by name and a repeat render paints synchronously, exactly as this page behaved before it was engine-aware. The fresh card always refreshes the header, and rebuilds the bar and grid only if it chooses a *different* registry — the two registries are module constants, so that comparison is exact and, in practice, never true. A card that does **not** arrive leaves a painted page alone rather than repainting it as SQL Server: a failed fleet read is not evidence a server stopped being PostgreSQL.

Reproduced against the shipped module before fixing:

```
first render: 7 reads. repeat render: 0 before the card, 3 after.
FAIL (5):
  - the repeat render blanked the page behind a loading strip while waiting on /api/fleet
  - the repeat render painted no tab bar before its fleet response landed
  - the repeat render started NO panel reads before /api/fleet answered
  - the repeat render did not paint the PostgreSQL Vacuum tab it was asked for
  - the fresh card repainted an unchanged registry: 3 duplicate panel read(s)
```

That last line is a symptom the review did not name and I had not noticed: because the async callback painted unconditionally, **every poll was refetching the tab twice**. The same guard removes it. Only the first render of a given server shows the loading strip, which is honest — at that point the engine genuinely is unknown.

**The badge register.** Also right, and the fix is not a capitalisation change. `DescribeEngineKind`'s `"an unrecognised engine"` is worded to sit mid-sentence in the capability messages, so it is the wrong string here rather than the wrong case. `engine_description` now has three answers: null for an absent kind, the describer's words for a recognised token, and **the raw token** for one this build has never heard of — which reads as a label and is the more useful of the two anyway, being the string an operator would search their own store for. It deliberately is not mapped onto a default, for the same reason the describer refuses to guess.

Pinned: `TheTabSet_SwitchesOnlyOnAPositivePostgresClaim` covers the remembered-card paint and the no-repaint-on-an-unchanged-registry rule, and `TheFleetCardSerializesTheDiscriminator_RawAndDerived` now asserts all three description answers including the unknown-token one. Both harnesses (the stale-render race and this one) were proved red first by reverting only the guard under test.

erikdarlingdata:
Right, and verified before fixing: `PgStatementStatsCollector.AppliesTo => target.IsAurora` (line 129), exactly like `PgWaitStatsCollector`'s, and `CapturePathByCollector` already lists the two together as *"the first two [that] read an Aurora-only surface"*. `PgBlockingCollector.AppliesTo => true`, so the rest of the tab is fine.

This is the second instance of one **category**, not a second bug — I wrote 25 lines arguing the Waits case and left the tab with the identical gate silent. And it matters *more* on Activity than on Waits: the rest of that tab fills, so one empty grid among three reads as a fault in a way a wholly-empty tab does not.

Fixed in 514f1ce:

- **The Activity tab has a note**, and it states the positive half too — the blocking panels are collected at every PostgreSQL target *including standbys*, where a recovery conflict is blocking that happens nowhere else.
- **The doc was wrong, not just incomplete.** "One PostgreSQL tab can be permanently empty" is now "two of the eight capture paths are Aurora-only", naming both, plus the distinction review's wording implies: **Waits is the only tab that is *wholly* Aurora-only.**
- **The category is pinned, not the instance.** `EveryAuroraOnlyPostgresRead_SitsOnATabThatSaysSo` **derives** which reads are Aurora-only from `CollectorEngineCapability.IsCollectedOnEngineKind` over the shipped `AppliesTo` gates — move a gate and the pin moves with it, and a ninth PostgreSQL read landing on a note-less tab goes red rather than shipping an unexplained blank. The read→collector map is explicit because that half is a naming fact with no table anywhere (each read passes its collector's name as a string literal), and the map is asserted to cover every served `get_pg_*` read so it cannot go stale quietly.

Proved red first, by removing only the new note:

```
FAIL  ServerPageTabsTests.EveryAuroraOnlyPostgresRead_SitsOnATabThatSaysSo
        Assert.Contains() Failure: Sub-string not found
        String:    "    id: "activity",\n    label: "Activity",\n\n    /*"···
        Not found: "note:"
```

Issue #2252, opened 2026-08-14
[QUESTION]  Add new server to monitoring


000al000:
### Which component is your question about?

Darling

### Performance Monitor Version

3.4.0 (server) / 3.3.0 (viewer)

### Is your question about how it works, or the results?

How the tool works

### What's your question?

Configuration: 
Darling headless (on windows server) 
Viewer on remote PC (downloaded as version 3.4, but in fact 3.3)

One server was added and start to darling.json and working fine
Try to add second server to darling.json, --test-connection seen 2 servers, result ok

Try restarting service and connect with viewer - but second server not seeing in viewer

How tell server to "reread" servers config from json?

000al000:
If i run PerformanceMonitor.Darling.Service.exe i cant see in console output second server

config
  "servers": [
    {
      "name": "plvs-itebd",
      "host": "plvs-itebd",
      "auth": "integrated",
      "trustServerCertificate": true,
      "monthlyCostUsd": 0,
      "alertDeliveryModeOverride": null,
      "excludedDatabases": []
    },
    {
      "name": "dcvs-bd01",
      "host": "dcvs-bd01",
      "auth": "sql",
      "username": "PerfomanceMonitor",
      "encryptedPassword": "hided_from_all",
      "monthlyCostUsd": 0,
      "trustServerCertificate": true,
      "alertDeliveryModeOverride": null,
      "excludedDatabases": []
    }
  ],

000al000:
PS C:\PerformanceMonitorDarling>  .\PerformanceMonitor.Darling.Service.exe --test-connection
Validating connectivity to 2 server(s)...
  [PASS] plvs-itebd: SQL major version 13, Enterprise, msdb access: yes
  [PASS] dcvs-bd01: SQL major version 14, Standard, msdb access: yes
All servers reachable.

000al000:
Error, if i try to add from remote viewer

warn: PerformanceMonitor.Darling.Service.DarlingWorker[0]
      [dcvs-bd01] Connect failed, retrying in 60s: Key not valid for use in specified state

erikdarlingdata:
Your config is fine and the tool is behaving as built — but the behaviour is wrong, and you've found two separate faults. Nothing you can put in `darling.json` will fix the first one.

## Why the second server never appears

**`darling.json` seeds the server list exactly once, and after that the store is authoritative.** From `StoreConfigProvider`:

```csharp
[...]

Issue #1236, opened 2026-06-25
[FEATURE] Per-server override for alert notification mode (per-event vs summary)


erikdarlingdata:
Follow-up split out from #1141.

The per-event vs batched-summary notification mode shipped as a **global** setting in #1141 (`AlertNotificationMode` + `PerEventNotification`, Settings UI in both apps, default Summary, per-cycle cap). That closes the core request.

This issue tracks the remaining **optional per-server override** the original issue mentioned ("global and/or per-server") — e.g. per-event for a noisy prod box, summary everywhere else.

**Blocked on infrastructure:** the Dashboard has no per-server settings store yet (its settings are global); Lite has per-server config. A per-server delivery-mode override needs that store to exist first.

Low priority — the global setting covers the primary use case.

PR #2556, opened 2026-08-23
Report when darling.json disagrees with a REGISTERED server's registry row (#2552)

### LINKED ISSUE #2552: Editing a registered server's settings in darling.json is silently ignored — only ADDING a server is warned about
Found while standing up a live PostgreSQL target against a container build of `dev`.

## What happens

`WarnAboutFileOnlyServersAsync` reads `SELECT server_id, name FROM config_monitored_servers` and compares **names and ids only**. If every server in `darling.json` is already registered, `fileOnly.Count == 0` and it returns immediately. Nothing anywhere compares the file's per-server *settings* against the store's — `grep` for a property-drift comparison in `StoreConfigProvider.cs` finds none.

So for a server that is already registered, **every setting in `darling.json` is dead text**: `trustServerCertificate`, `username`, `password`, `port`, `database`, `encryptMode`, `excludedDatabases`, `capturePlans`, `displayName`. The file is read, parsed, validated, logged as `Loaded configuration from /config/darling.json: 1 server(s)` — and then the store's row is used instead, with no indication that the two disagree.

## Why this is worse than it sounds

#2252/#2254 already warn — well — about the *adjacent* case:

> darling.json lists N server(s) that are NOT monitored and never have been … **The store is authoritative after the first seed, so adding a server to the file does not register it** and a restart cannot change that

An operator who reads that learns "adding via the file does not work." The natural inference is that the file *does* still drive the servers it already knows about. It does not, and nothing corrects that inference.

The `web.network` / `mcp.network` blocks each log a full paragraph explaining they are file-only and restart-only and that the control plane owns the rest. The **server list** — far more likely to be edited than either — says nothing.

## How I hit it, which is the realistic path

1. Added a PostgreSQL target. First connect failed: `[pgtarget] Connect failed, retrying in 60s: SSL connection requested. No SSL enabled connection from this host is configured.` Good message — names the target and the cause.
2. Enabled TLS on the target. Now: `Exception while performing SSL handshake` (self-signed cert, so `SslMode.VerifyFull` correctly refuses).
3. Applied the documented fix — set `"trustServerCertificate": true` in `darling.json` — and restarted.
4. **Byte-identical error.** No new information, no hint the edit was not read.

The store row explained it:

```
name     | trust_server_certificate |         created_at         |        modified_at
pgtarget | f                        | 2026-08-22 22:27:15.069708 | 2026-08-22 22:27:15.069708
```

`created_at == modified_at`, still `f`. The edit never left the file.

That is a bad loop to be stuck in: a connection failure is *exactly* the kind of problem an operator fixes by editing config and restarting, and this is the one class of edit that produces an unchanged error with no explanation.

## Worth deciding, not assumed

I am **not** proposing that the file start overriding the store — store-authoritative is a deliberate design and #2254 is explicit about it. The gap is that the disagreement is **invisible**. The minimum fix is to extend the existing warning to cover servers present in *both*, comparing the settings that actually affect behaviour and naming the ones that differ, in the same voice as the existing message: what the file says, what the store says, and that the store wins and how to change it (Viewer's Add/Edit Server, or MCP `add_servers`).

Two things to settle first:

- **Which fields count.** Comparing everything invites noise from cosmetic normalization — `encrypt_mode` is already normalized on seed (`StoreConfigProvider.cs:350` mirrors the normalizer precisely so a casing choice cannot fail the seed), so a naive comparison would warn on a difference that does not exist. The comparison has to run against the same normalizer.
- **Passwords must not be compared or printed.** The file may carry a `file:`/`env:` reference while the store carries a DPAPI blob; those are not comparable and the answer is almost certainly to exclude the credential from the drift check entirely rather than to try.

A `--test-connection` caveat probably belongs here too: per the #2252 message it already reads `darling.json` and reports PASS for servers that collect nothing. The same is true for settings — it would test the *file's* `trustServerCertificate`, not the store's, so it can report PASS for a configuration the service will never use.

Related: #2252, #2254, #2258.

erikdarlingdata:
Closes #2552 — with two corrections to the issue's framing, at the bottom.

## What was wrong

`StoreConfigProvider.WarnAboutFileOnlyServersAsync` read `SELECT server_id, name FROM config_monitored_servers` and compared names and ids only, then returned early once `fileOnly.Count == 0`. So for a server the store already had, **every per-server setting in `darling.json` was dead text** — `host`, `database`, `auth`, `username`, `encryptMode`, `trustServerCertificate`, `excludedDatabases`, `port`, `engine`, the display name. Read, parsed, validated, logged as `Loaded configuration … 1 server(s)`, and then not used.

The reproduction in the issue is the loop that makes it expensive: a PostgreSQL target refused a self-signed certificate, the operator applied the documented fix (`"trustServerCertificate": true`), restarted, and got a **byte-identical error** — the store row still said `f`, with `created_at == modified_at`. A connection failure is exactly the class of problem an operator fixes by editing config and restarting, and this was the one class of edit that produced an unchanged error with no explanation.

#2252/#2254 already warn — well — about the *adjacent* case, and that warning made this one worse: it teaches "adding a server to the file does not register it", from which the natural inference is that the file still drives the servers the store already knows about.

**The store still wins.** Store-authoritative is deliberate (#2254) and none of the ordering changed. The defect was that the disagreement was invisible.

## What ships

One new WARNING, once per start, naming every registered server whose file entry disagrees with its registry row, every field, and both values:

```
darling.json disagrees with the registry about 1 monitored server(s), and the registry is what the
service uses: pgtarget: trustServerCertificate (file=true, store=false). The store is authoritative
after the first seed, so editing a registered server's settings in the file changes nothing and a
restart cannot change that — edit them in the Viewer's Manage Servers window (the MCP add_servers
tool cannot: an already-registered server is skipped as a duplicate). Note --test-connection reads
darling.json, so it probes the FILE's settings and can report PASS for a connection the service will
never make.
```

`WarnAboutFileOnlyServersAsync` is now `WarnAboutFileVersusStoreAsync`, one store read serving both causes, with the file-only lines split into their own method so the drift pass is no longer gated on there being anything to say about cause A.

### Which fields count, and the fold each one runs through

Every fold below is a real one somewhere else in the service, not a convenience. A check that warns about differences that do not exist is the same silence with extra steps.

| field | fold | why |
|---|---|---|
| `name` | blank → host, then case-**sensitive** | the store's spelling is what the Viewer and every alert render |
| `host` | trim, case-insensitive | matches `DarlingWorker.ServerDefinitionEquals` |
| `database` | blank → the engine's implicit default (`master` / `postgres`); **case-sensitive on PostgreSQL**, case-insensitive on SQL Server | what `MonitoredServerConnection` substitutes; and PostgreSQL matches `pg_database.datname` byte for byte, so `ReportingDB` and `reportingdb` are two databases there |
| `auth` | trim, case-insensitive | validation already restricts it to integrated/sql |
| `username` | **only when the store row uses SQL auth** | the only case it reaches the connection string |
| `encryptMode` | the connect path's fail-closed fold (trim, upper, unknown → Mandatory), **plus `Strict` ≡ `Mandatory` on PostgreSQL** | **the hazard the issue named** — `strict` vs `Strict` is not drift, nor a typo against `Mandatory`; and the Npgsql builder branches on `OPTIONAL` alone, so the other two are one connection there |
| `engine` | `MonitoredServer.TargetEngine` | `aurora` and `postgres` are one engine |
| `port` | **PostgreSQL only**, `0` → `5432` | SQL Server carries its port in the host as `host,1433` |
| `readOnlyIntent`, `multiSubnetFailover` | **SQL Server only** | SqlClient concepts the Npgsql builder never sees |
| `trustServerCertificate` | everywhere **except** PostgreSQL + a stored `Optional` | that branch picks `SslMode.Prefer` without reading the flag; SQL Server's `Optional` still validates the certificate |
| `excludedDatabases` | trim, drop blanks, de-dupe, **order**; case follows the engine | spliced as a `NOT IN` set (`DatabaseExclusionFilter`) — live on **both** engines, since `PostgresTargetProvider.BuildDatabaseListPlan` uses the same filter against `datname` to pick the per-database fan-out |
| `monthlyCostUsd` | numeric | `decimal` carries its scale, so `1200` and `1200.00` are one figure |
| `alertDeliveryModeOverride` | null = `(inherit)` | #1236 |
| `password` / `encryptedPassword` | **never read, never compared** | see below |
| `capture_plans` | not read | no per-server file counterpart to disagree with |
| `is_enabled` | read, never compared | no file counterpart either; it decides which of the two report lines a server lands on — see below |

The engine gates read the **store's** engine (and its stored `encryptMode`/`auth`), because the store is what the service connects with; a disagreement about the engine itself is reported on its own line.

Four of those folds are **engine-asymmetric**, all four found in review by asking what the message *claims* rather than what the code does. The first two would have been false **positives** (`encryptMode`, `trustServerCertificate`); the last two false **negatives** (`database`, `excludedDatabases`) — the silent direction, which is what #2552 is about, so those were the more expensive half. For `encryptMode` the displayed value and the compared key are deliberately **split**: folding the display too would print `store=Mandatory` for a row that literally says `Strict`, i.e. a message inventing a value neither side holds.

### The credential

`encrypted_password` is **not in the SELECT list**. That is structural rather than editorial — there is no blob in memory for a later edit to this file to leak into a log line. It must not be compared even if it were read: a file entry legitimately carries an `env:`/`file:` reference (`DarlingSecretSource`) or a dev plaintext password while the store row carries a DPAPI blob, and `BuildServerFromRow` backfills exactly that pairing at read time. That is the **supported** shape, so comparing them would report a working configuration as drift on every start.

Proven by reverting: adding an `encryptedPassword` comparison puts `env:ALPHA_MONITOR_PASSWORD` and a DPAPI blob straight into the rendered line, and the guard goes red.

### A paused server gets its own sentence (review finding)

A control-plane-**disabled** server is still registered, so it still pairs and its drift is still real — but *"the registry is what the service uses"* is not a true sentence about a server nothing is connecting to. It is reported on its own line, at **Information**, which is the same two-line shape cause A already uses for "never monitored" (warning) versus "deliberately removed" (information). That line drops the claim, the remedy and the `--test-connection` caveat.

Filtering the store read to `is_enabled = TRUE` was the obvious fix and is the one option that **must not** be taken: that read also feeds `ServersOnlyInFile`, whose question is whether a file entry is *registered* — and a paused server is — so filtering there would report it as never-monitored and advise re-adding it, which is the #2158 defect. Dropping it from the drift pass silently is rejected more narrowly, since #2552 is a defect about silence being expensive and the drift is exactly what the operator walks back into on re-enable.

### The invariant — the category, not the instance

#2552 is not "trustServerCertificate was not compared"; it is "a per-server darling.json setting that nothing compares is silently dead text". So the covered set is **derived from both ends** rather than listed: `MonitoredServer`'s `[JsonPropertyName]` keys on one side, and **`CompareServerSettings`' own output** on the other — driven with two entries that differ in every field, twice, because two fields are engine-gated and one pass can only see one side of that gate. Field labels are the darling.json keys spelled exactly, so there is no mapping table in between to drift.

Today: **16 keys, 14 compared, 2 excluded** (the credential, asserted as such). It fails in both directions — a key nothing compares, and a field label that is not a real key. Proved red by dropping `monthlyCostUsd` from the comparison, and by adding a new per-server key nothing compares (`uncovered=[connectTimeoutSeconds]`) — the future defect, caught the day it lands.

### Pairing

Same either-or key as `ServersOnlyInFile` (stored `server_id`, then the display name case-folded), with **one rule on both arms**: the match must be unambiguous in both directions or the entry is skipped rather than guessed at. Nothing enforces display-name uniqueness, and two file entries can derive one `server_id` (identical addresses, where the seed's `ON CONFLICT DO NOTHING` left a single row) — a guess would print two entries' settings as one server's drift and could contradict itself on the same line. That is `BuildServerFromRow`'s "exactly one match" discipline, for the same reason.

### The `--test-connection` caveat — yes, it belongs, but conditionally

Appended **only when a connection-relevant field drifted**. That verb probes the file's connection settings, so it can report PASS for a connection the service will never make — which is precisely the reproduction. But it does not exercise the display name, the exclusion list, the cost figure or the delivery override, and a warning must not claim more than it can support.

### Docs

`darling.sample.json`'s `servers` block now carries the seed-once paragraph the `web.network` / `mcp.network` blocks have always had. The issue is right that the server list — far more likely to be edited than either — said nothing.

## Two corrections to the issue

1. **`capturePlans` is not a per-server setting.** `MonitoredServer` has no such property; `capturePlans` is a top-level service setting. The `config_monitored_servers.capture_plans` column exists but the seed writes it `NULL`, so there is nothing to compare. Nothing else in the issue's field list is wrong.
2. **The remedy cannot include `add_servers`.** The issue (and the obvious reading of the #2252 message) suggests "the Viewer's Add/Edit Server dialog, or MCP `add_servers`". `add_servers` **cannot change a registered server**: `PartitionDuplicates` skips an already-monitored server as `status:"duplicate"` before it validates anything. Pointing an operator there would send them somewhere that silently does nothing — the exact failure this warning exists to end. The message names the Viewer's **Manage Servers** window and rules `add_servers` out explicitly. (There is no MCP edit verb at all; `remove_server` + `add_servers` is the only headless path, and it is a heavier hammer than this warning should recommend.)

One thing worth recording separately, not fixed here: the issue cites `StoreConfigProvider.cs:350` as the `encrypt_mode` seed normalizer. That line is `NormalizePlanXmlCompression` — `plan_xml_compression`, whose V62 CHECK is case-sensitive. `encrypt_mode` has no CHECK and the seed writes it **raw** (line 512). The hazard is real but lives at the *connect* path, not the seed, which is why the comparison folds through `MonitoredServerConnection`'s own fail-closed switch rather than through a seed-side normalizer.

## Verification

- **Builds on macOS**: service, viewer and `Darling.Tests` all clean, no new warnings.
- **Pure logic RUN locally**, not deferred to CI. `PerformanceMonitor.Darling.Service` is `net10.0`, so a throwaway console harness referencing the real project executed the same 57 assertions against the shipped build: **all pass**.
- **Every guard proven RED first**, reverting only the thing under test, then restored green. Eighteen reverts, each failing exactly the checks that name it and no others:

  | reverted | went red |
  |---|---|
  | the whole per-field comparison (#2552 as filed) | 16 checks, including the field report |
  | the encryptMode fail-closed fold | 6 |
  | the blank-database engine default | 4 |
  | the port-0-means-5432 fold | 1 |
  | the engine-alias fold | 5 |
  | the excludedDatabases set fold | 1 |
  | the credential exclusion | 2 — with the blob and the `env:` reference visible in the rendered line |
  | the engine gate on `readOnlyIntent`/`multiSubnetFailover` | 2 |
  | the username-only-under-SQL-auth gate | 1 |
  | the ambiguous-display-name skip | 2 |
  | the id-arm ambiguity guard | 1 — two contradictory lines for one server |
  | the display-budget truncation | 1 |
  | the enablement carried onto the drift entry | 1 |
  | `monthlyCostUsd` dropped from the comparison | 2, including the invariant naming it |
  | a new per-server key added and never compared | 1, the invariant naming `connectTimeoutSeconds` |
  | the PostgreSQL `Strict`≡`Mandatory` collapse | 1 |
  | the `trustServerCertificate` inert gate | 1 |
  | the engine-sensitive name comparison | 2 (`database` and `excludedDatabases`) |
  | the engine-sensitive exclusion-set comparer | 1 — **and this one had to be rebuilt to bite** |

  That last row is the one worth reading. Reverting the exclusion set's *inner* comparer alone left everything green, because the outer comparison catches a case difference either way — a test that passes with and without the fix is worse than none. What only the comparer decides is what the message **says**: a PostgreSQL store excluding both `Scratch` and `scratch` renders as `store=Scratch, scratch` with the fix and `store=Scratch` without, which is wrong information about the server in the one sentence the operator gets. Asserting the rendered value is what makes it fail.

- **`Darling.Tests` cannot RUN on macOS** (`net10.0-windows`); the 32 facts + 3 theories (43 cases with the `InlineData` rows) in `RegisteredServerSettingDriftTests` build here and execute in CI. The harness above covers the same assertions against the same code, so the logic is verified — what is not verified locally is that xUnit discovers and runs them, which the `build` check is the arbiter of.
- **Not verified at all**: the live path. Nothing here ran against a real seeded store, so the `SELECT` shape, the `MonitoredServer` hydration from the reader and the log line as it actually renders through the service's logger are unexercised outside CI's live-Postgres suite. The SQL selects only columns that exist in the V17 table plus V68's `engine`/`port` and V18's `alert_delivery_mode_override`, all of which `ReadMonitoredServersAsync` already reads at the same ordinals.

No schema change, no migration rung, no store write. Darling only — Lite has no control-plane store to disagree with.


erikdarlingdata:
Addressed, though not the way it was suggested — the suggestion would have traded this defect for #2158's.

**`WHERE is_enabled = TRUE` on that read is the one option that must not be taken.** `ReadRegisteredServersForComparisonAsync` is a *shared* read: it also builds the `storeIds` / `storeNames` sets that feed `ServersOnlyInFile`, whose question is "is this file entry **registered** at all" — and a paused server is. Filtering there would report a server the operator deliberately disabled as "NOT monitored and never have been", and advise re-adding it. That is precisely the #2158 failure this method was already fixed once to avoid, on every start, about a server the operator had just paused on purpose.

**Suppressing it from the drift pass only** was the other candidate, and it is rejected more narrowly: #2552 is a defect about silence being expensive, so answering it by adding a new silence is the wrong direction — and the drift is exactly what the operator walks back into the moment they re-enable.

What was genuinely wrong is the **claim**, which is the sharp half of the finding: *"the registry is what the service uses"* is not a true sentence about a server nothing is connecting to. So the row now carries its enablement (a new `RegisteredServer` record) and the report splits into two lines — the same two-line shape cause A already uses for "never monitored" (warning) versus "deliberately removed" (information):

```
[Information] darling.json also disagrees with the registry about 1 server(s) that are registered but
DISABLED: alpha: trustServerCertificate (file=true, store=false). Nothing is connecting to them, so
neither value is in force today — but the registry's is the one that would take effect if they were
re-enabled from the Viewer, not the file's.
```

The paused line drops the claim, drops the remedy, and drops the `--test-connection` caveat. `is_enabled` is read but never *compared* — it has no darling.json counterpart; it only decides which sentence a server belongs under. A store mid-migration with a NULL there reads as **enabled**, which is the direction that keeps the drift visible.

Pinned by `ADisabledServerStillPairs_ButIsReportedSeparately`, and proved red by hardcoding the flag to `true` at the point it is carried onto the drift entry.

---

**And, since the review found this by asking what the message CLAIMS rather than what the code does, I took the same question one level up.** #2552 is not "trustServerCertificate was not compared" — it is "a per-server darling.json setting that nothing compares is silently dead text". Fixing the reported field leaves the next one to be found the same way, in the field, by an operator staring at an unchanged error.

So the covered set is now **derived from both ends** rather than listed. Field labels are the darling.json keys spelled exactly (`displayName` → `name`), which lets `EveryPerServerDarlingJsonKeyIsEitherComparedOrDeliberatelyExcluded` take `MonitoredServer`'s `[JsonPropertyName]` properties on one side and **`CompareServerSettings`' own output** on the other — driven with two entries differing in every field, twice, because two fields are engine-gated and one pass can only see one side of that gate. No mapping table in between, and the shipped function is the evidence rather than a reimplementation that could keep agreeing while the code drifts.

Today: **16 keys, 14 compared, 2 excluded**, and the 2 are the credential, asserted as such. It fails in both directions — a key nothing compares, and a field label that is not a real key (which would send an operator to edit something that is not there).

Proved red twice: dropping `monthlyCostUsd` from the comparison (`uncovered=[monthlyCostUsd]`), and adding a new per-server key nothing compares (`uncovered=[connectTimeoutSeconds]`) — which is the future defect, caught the day it lands instead of in the field.

erikdarlingdata:
Both correct, both fixed — and the second one generalizes to a field the comment did not name.

**`encryptMode` on PostgreSQL.** Confirmed against `BuildPostgresConnectionString`: it branches on `== "OPTIONAL"` and everything else takes the same `SslMode` arm, so `Strict` and `Mandatory` are one connection there. Reporting them as drift with `AffectsConnection = true` — pulling in the `--test-connection` caveat — was reporting a connection difference the builder cannot express, against this PR's own stated principle. The comparison now collapses them on a PostgreSQL target.

One detail worth flagging, because "collapse `Strict`→`Mandatory`" has a trap in it: display and comparison are now **split for this one field**. Folding the printed value too would have shown `store=Mandatory` for a store row that literally says `Strict` — a message inventing a value neither side holds. So the key is collapsed and the words are not:

```
encryptMode (file=Optional, store=Strict)      ← on a PG target, compared as Optional vs Mandatory
```

**`trustServerCertificate` — the parenthetical.** Also right, and gated on the **store's** mode, since the store's is the one in force: on PostgreSQL with a stored `Optional`, `SslMode.Prefer` is chosen without ever reading the flag. It stays live on SQL Server's `Optional`, where SqlClient still validates the certificate if the server negotiates encryption — so this is narrower than "Optional means inert".

**`database` case — and `excludedDatabases` with it.** Right, and not as low-confidence as the comment hedged. The generalization is the part worth having: **`excludedDatabases` is live on PostgreSQL too.** `PostgresTargetProvider.BuildDatabaseListPlan` splices the *same* `DatabaseExclusionFilter` against `pg_database.datname` to choose the per-database fan-out, and `NOT IN` is case-sensitive there — so folding case would treat two genuinely different exclusions as one. Both fields are now engine-gated: `Ordinal` on PostgreSQL, `OrdinalIgnoreCase` on SQL Server (matching `ServerDefinitionEquals`, with the case-sensitive-server-collation limit stated in the doc rather than left implicit).

Worth naming the direction, since it flips: the first two findings were false **positives**. The last two were false **negatives** — the silent direction, which is the one #2552 is about, so those were the more expensive half.

---

**One of the four guards nearly shipped without biting**, which is the part I want on the record. Reverting the exclusion set's *inner* comparer alone left everything green: the outer comparison catches `"Scratch"` vs `"scratch"` whether or not the set was de-duplicated case-sensitively, so the first test passed with and without the fix — worse than no test, because it converts an open question into false confidence.

What only the inner comparer decides is what the message **says**. A PostgreSQL store excluding both `Scratch` and `scratch` renders as `store=Scratch, scratch` with it and `store=Scratch` without — wrong information about the server, in the one sentence the operator gets. Asserting the rendered value is what makes it bite:

```
======== REVERTED: the engine-sensitive exclusion-set comparer ========
FAIL  both case variants survive the postgres exclusion set
      -> excludedDatabases(file=scratch,store=Scratch,conn=False)
```

All four reverted individually and watched go red, then restored: the `Strict`/`Mandatory` collapse (1), the `trustServerCertificate` gate (1), the engine-sensitive name comparison (2 — `database` and `excludedDatabases` together), and the set comparer (1). Local harness against the shipped build is now **57 assertions, all passing**; the xUnit file is 32 facts + 3 theories.

PR #2560, opened 2026-08-23
Collect PostgreSQL index usage and table bloat, with the facts that stop bad advice (#2541, #2542)

### LINKED ISSUE #2541: No index usage collection for PostgreSQL, where an unused index costs more than on SQL Server
Part of #2530.

## The gap

Nothing collects `pg_stat_user_indexes` (or `pg_stat_all_indexes`). SQL Server users already get `get_index_usage`; PostgreSQL users get nothing equivalent.

`idx_scan`, `idx_tup_read` and `idx_tup_fetch` are the whole story for "is this index earning its keep", and `pg_stat_user_indexes` joined to `pg_class` gives the size to weigh it against.

## Why it matters more here than on SQL Server

On SQL Server an unused nonclustered index costs write throughput and space. On PostgreSQL it costs the same **plus** it participates in every `VACUUM` — index cleanup is a large share of vacuum work, so dead indexes directly slow the maintenance that #2530's Vacuum surface exists to monitor. An unused index on a hot table is a vacuum problem as much as a write problem.

It is also the cheapest tuning advice a monitoring tool can give: dropping an index nobody scans has no downside to model.

## Notes

- `idx_scan` is cumulative since stats reset. `stats_reset` on `pg_stat_database` (#2539) is the way to tell "never scanned" from "reset yesterday" — worth doing after that one so the distinction is available, because reporting a freshly-reset index as unused is exactly the wrong advice.
- Index *size* wants `pg_relation_size`, and on a large database that is not free — check what `PgAutovacuumStatsCollector` already does about relation sizing rather than inventing a second approach.
- Partial and expression indexes read as unused when the planner cannot match them; the read should carry enough (`indexdef`) that a human can tell "unused" from "unusable".

### LINKED ISSUE #2542: No bloat measurement for PostgreSQL: we collect the whole cause chain and none of the damage
Part of #2530.

## The gap

We collect autovacuum lag (`PgAutovacuumStatsCollector`), wraparound risk (`PgWraparoundStatsCollector`) and the xmin horizon (`PgXminHorizonCollector`) — the entire **cause** chain for bloat — and nothing that measures the **damage**. A user can watch vacuum fall behind and has no way to see what it has cost them, which is the question they ask immediately afterwards.

## The measurement decision, which is the substance

- **`pgstattuple`** is the canonical answer and gives exact live/dead tuple counts and free space. It performs a **full relation scan**, so on a large table it is far too expensive to run on a collection cadence. It is also an extension, so availability on managed PostgreSQL has to be checked rather than assumed.
- **Estimate queries** (the well-known `pg_class.reltuples` / `relpages` versus expected-size calculations) are cheap enough to run regularly and are approximate — they can be badly wrong on tables with unusual column widths or after a schema change.
- **`pgstattuple_approx`** sits between the two.

Picking one is the work. My inclination is the estimate on cadence with `pgstattuple` available on demand for a single table, so the routine path is cheap and the precise answer is reachable when someone is actually investigating — but that should be decided against a measurement on a real table, not on reasoning.

## Note

Whatever ships must not present an estimate as though it were exact. A bloat number that is quietly ±40% will be used to justify a `VACUUM FULL` on a production table, and the difference between "estimated" and "measured" needs to survive into the read's payload rather than living only in a doc.

Source: [PostgreSQL Wiki — Monitoring](https://wiki.postgresql.org/wiki/Monitoring) (names `pgstattuple` as the bloat tool).

erikdarlingdata:
Closes #2541. Closes #2542. Part of #2530.

Two collectors, two MCP reads, two schema rungs (**v84**, **v85**) and a **Storage** tab on both front ends. They ship together because both add rungs and would otherwise collide on `StorageVersion.SchemaVersion`.

## #2541 — index usage: the scan count is the easy half

`pg_stat_user_indexes` was uncollected, so SQL Server users had `get_index_usage` and PostgreSQL users had nothing — and an unused index costs **more** here: the same write throughput and space, **plus** participation in every `VACUUM`, where index cleanup is a large share of the work. A dead index on a hot table slows the maintenance #2530's vacuum surface exists to watch.

The scan counts are trivial to collect. What makes this shippable is that **"unused" is not "droppable"**, and the tool refuses to say otherwise. Every row carries whether the index backs a primary key, a unique or exclusion constraint, or the table's replica identity; whether it is partial, an expression index, or INVALID; and its full `pg_get_indexdef` text. That is not metadata around the finding, it is the finding — **the single most common zero-scan index on any schema is a unique index enforcing a constraint**, because enforcing uniqueness is not a scan of the kind `idx_scan` counts. Advice derived from the counter alone tells somebody to drop their primary key.

Verified live on the rig, not argued:

| index | scans | the fact that stops a DROP |
|---|---|---|
| `gadget_sku_status_uq` | 0 | `supports_constraint = t` |
| `gadget_sku_uq` | 0 | `is_replica_identity = t` |
| `gadget_partial_big_idx` | 0 | `is_partial = t` — may be unusable, not unused |
| `gadget_expr_idx` | 0 | `is_expression = t` |
| `gadget_broken_idx` | 0 | `is_valid = f` — the one case where zero scans really does mean drop it |
| `churny_v_idx` | 0 | 852,057 block hits: maintained by writes, read by nobody |

It also refuses a disuse claim on fewer than two samples. PostgreSQL records no index creation time anywhere, so how long **we** have been watching is the only evidence an index is old enough to judge.

### The scan count only a monitoring store can produce

`idx_scan` is cumulative since the last statistics reset, which is all a live query can give: an index with nine million lifetime scans and none in ninety days reads as heavily used. The read reports **both** — the lifetime counter, so it cannot appear to disagree with psql, and `scans_in_window` differenced across stored samples.

Two guards on that difference, both **proven RED against a live store** by reverting only the thing under test:

| guard | shipped | reverted |
|---|---|---|
| `GREATEST(scans - first, 0)` on a 9000 → 50 → 100 reset | `scans_in_window = 0` | **`-8900`** — a negative scan count at the top of a least-used-first list |
| `IS DISTINCT FROM` on a **first-ever** reset (`stats_reset` NULL → timestamp) | reset reported | **NULL** — completely invisible, in the one case the counter arithmetic also cannot see |

The `quietreset` fixture is why both signals travel: 1000 → 3000 across a reset leaves every difference positive, so the arithmetic sees an ordinary busy day and only the timestamp knows.

Retention is **90 days** because the retention window *is* the evidence. 30 days cannot clear a monthly report. It still cannot clear an annual one, which the read says rather than glossing.

## #2542 — bloat: the measurement decision, settled by measurement

We collected the entire **cause** chain for bloat (`pg_autovacuum_stats`, `pg_wraparound_stats`, `pg_xmin_horizon`) and nothing that measured the **damage**. The issue asked for the option to be picked against a real table rather than by reasoning, so it was:

| option | cost, measured | verdict |
|---|---|---|
| `pgstattuple` | **11.8 ms** on a 54 MB fully-cached table (~4.6 GB/s); **860 ms** over 2,001 tables | reads the whole relation — a 1 TB table costs minutes per cycle even entirely in RAM. Also an extension |
| `pgstattuple_approx` | 0.58 ms on the same table | that is its **best** case and unrepresentative: the pages it must still read are the recently-written ones, so on the churning table you care about it degrades toward the full scan. Same availability problem, no cost fix |
| **statistics estimate** | **44 ms** over the same 2,001 tables / 854 MB | ships. Reads `pg_class` + `pg_stats` only, so it costs by table **count**, not table **size** |

The 19× is a **floor** on the ratio — those tables were tiny and fully cached.

**On the on-demand half of the issue's proposal:** "`pgstattuple` available on demand for a single table" is **not implementable as a tool** — the MCP surface is strictly read-only over collected rows and has no ad-hoc path back to a monitored server. Each row instead records whether `pgstattuple` is installed on that database and ships the exact command that would settle it, which is the reachable form of the same idea. (Noted on #2542.)

### How wrong it can be — measured, and the good case is not the interesting one

With current statistics the estimate was within **~2 percentage points** of `pgstattuple` on every table tested: churn-bloated 75.14 vs 74.82, delete-bloated 60.73 vs 59.52, clean 0.58 vs 0.50.

Then the failure modes, each reproduced:

| case | estimate | truth | error |
|---|---|---|---|
| **stale width statistics** | **92.64%** | 10.93% | **+81.7 pp** |
| control: same table, ANALYZEd | 11.01% | 10.93% | +0.08 pp |
| never analyzed | 92.68% | 48.86% | +43.8 pp (caught by the flag) |
| fillfactor 70 | 3.79% | 32.52% | **not an error** — `pgstattuple` counts fillfactor-reserved space as free; the estimate is the more useful reading, and `fillfactor` travels so a read can say which question it answered |

The stale pair is the headline: **two byte-identical 8,998-page tables with identical true bloat, 81 points apart**, differing only in whether `ANALYZE` had run after a widening UPDATE — with nothing in the arithmetic to show it. `mods_since_analyze` is the signal (300,000 against 150,000 live rows vs 0), and it is a *sound* proxy rather than a coincidence: width statistics can only go stale **through** modifications.

### The permissions trap, which is the real finding

`pg_stats` is filtered by `has_column_privilege(..., 'select')`, and **`pg_monitor` — the one grant the PostgreSQL runbook says Darling needs — confers no SELECT on user tables.** Measured against a `pg_monitor`-only role on the live PG16 target: **zero** `pg_stats` rows visible, and *the estimator did not fail*:

| table | reported | true |
|---|---|---|
| `gadget` | **88.59%** | 0.50% |
| `churny` | 95.03% | 74.82% |
| `narrow` | 22.57% | 0.46% |

Every row an argument for rewriting a production table. Adding `pg_read_all_data` (PostgreSQL **14+**; the role does not exist on 13, where explicit `GRANT SELECT` is the only route — both verified) restored **byte-identical** agreement with the superuser's numbers.

So `estimate_unavailable` is **load-bearing, not advisory**: the read suppresses the number outright rather than captioning it, because a footnote under a large red percentage is not a safeguard. What survives is the **dead-tuple fraction** — the server's own counters, no width model, no SELECT needed — so a permissions gap *degrades* the answer instead of removing it. The runbook now carries the grant and the reasoning.

`VACUUM FULL` is never the lead remedy, and a pin asserts `pgstattuple` is named before it.

## Version floors: measured on six live majors, not read from docs

Every column checked by listing the real catalogs on **13, 14, 15, 16, 17 and 18**, then both shipped queries executed on all six. **Exactly one floor exists:**

- **`last_idx_scan` is PG16+** — proven RED by running the gated-on form against 13/14/15: `ERROR: column i.last_idx_scan does not exist`, which fails the *whole* collection every cycle. Substituted with a typed NULL so the row shape does not change across a mixed fleet; NULL is honest, because a PG15 server genuinely does not know.

Everything else confirmed present on all six and stated as checked: `pg_stat_user_indexes`, `pg_statio_user_indexes`, `pg_index`'s droppability columns, `pg_constraint.conindid`, `pg_stats`, `pg_class.reltuples`/`relpages`. **The bloat query carries no gate at all** — byte-identical text and identical results on every major (3.8–15.9 ms).

MAXALIGN detection names `aarch64|arm64` explicitly (Graviton RDS and Aurora are aarch64) and **stores the value it used**, so a platform where detection is wrong shows in the data rather than skewing every row silently.

## Fan-out cost, budgeted rather than discovered

Both are per-database fan-outs — the second and third after `pg_autovacuum_stats`, whose approach they follow — and on PostgreSQL that means one **connection** per database per cycle.

- **Index usage: daily (1440/90)**, inherited from `index_object_stats` — the SQL Server collector answering the same question — not from the PostgreSQL sibling. "Has anything scanned this index" is structural; hourly would re-record the same catalog facts 24× a day at 24× the connections.
- **Bloat: hourly (60/90)**, matching `pg_autovacuum_stats` deliberately: this measures the damage whose cause that one measures, and *"vacuum fell behind at 14:00 and bloat grew"* is only a sentence the data can support at a common grain.
- **Size floors as the cost control** — 64 KB per index, 1 MB per table, below which there is nothing to model (the whole argument for dropping an index is write, space and vacuum cost, all of which scale with size). On a 2,001-table / 6,005-index database the floors removed 2,000 tables and 2,001 indexes. An **invalid** index is kept regardless of size, and that escape hatch was verified live: a 16 KB invalid index came through a 64 KB floor.

## Both UIs

A new **Storage** tab in the web dashboard and the WPF viewer — same id, same grouping. Both reads answer one question: where the space went and whether it is earning its keep. Deliberately **not** on Vacuum despite bloat being what vacuum lag costs — that tab is the cause chain in causal order, and dropping the damage into the middle breaks the sequence that makes those three panels one story; the bloat panel points back at it instead.

Both ratchets that would have caught an MCP-only ship were exercised: the web pin derives its set from `BuildReadDispatch()`, the viewer pin from `CollectorCatalog`. The viewer grid renders a suppressed estimate as a **dash**, never a number, and never paints it as high-bloat — that colour would assert exactly what the suppression denies.

**The suppression rule is one function, shared.** The first draft had the MCP tool and the WPF projection each carrying their own copy of the churn threshold — matching on the day and free to drift at the next edit, silently and in the direction where one surface publishes a figure the other had already withheld. It now lives in `DarlingPgTableBloatReader.EstimateIsUnpublishable`, in `Darling.Storage`, which both consumers already reference; the tool keeps only the prose. The agreement test that guards it **would have passed before that change**, because both literals were `0.2` — which is exactly why passing is not evidence — so the local harness also simulates the drift: a copy at `0.3` is **detected** by the same comparison, and a copy at the real ratio still agrees as the control.

Derived, not enumerated: catalog **53**, PostgreSQL **11**, per-shape gate counts Aurora writer 11/11, Aurora reader 8/11, self-managed 16 writer 9/11, self-managed 15 reader 5/11. CI TimescaleDB sizing re-derived in both workflows (hypertables 52→**54**, `max_background_workers` 54→**56**, `max_worker_processes` 65→**67**; 5-line diff per file, no renormalization). Tool census **recounted by scanning** `McpServerTool(Name = ...)`: 112 → **114**.

## What I ran, and against what

- **Live PostgreSQL 16.15 target** with real churn (~225k-row `widget` plus purpose-built `gadget`/`bloaty`/`churny`/`widey`/`stale2`/`fresh2`/`ffy`/`toasty` fixtures): both shipped queries, dumped from `BuildQuery(ctx).Text` rather than retyped.
- **PostgreSQL 13.23 / 14.24 / 15.19 / 17.11 / 18.6** in Docker: catalog-column census, both shipped queries, and the gate proven red.
- **`pgstattuple` ground truth** on the same tables for every accuracy figure above.
- **A `pg_monitor`-only role** on the live target for the permissions finding, and again after `GRANT pg_read_all_data`.
- **The shipped reader SQL against a live store**, extracted from the C# raw string literals with `$n` rewritten in textual order, over an adversarial fixture (two databases with a same-named index, a second server's rows, four reset shapes) — and both guards reverted to prove them red.
- **A `net10.0` harness driving `ReadAsync` over a real Npgsql connection to real PostgreSQL 17** — every ordinal, every `GetInt32`/`GetInt64`/`GetDecimal`/`GetDouble`/`GetBoolean`/`GetDateTime` decode, and `WritePayload` arity (24/24 and 19/19) against real rows. This is the half a text assertion on the query string can never see.
- Ladder-generator diff for both rungs, ladder ordering/density, catalog uniqueness, `pg_catalog` shadow check against six live catalogs, and the suppression/droppability classifiers as pure functions.

## What I could not verify

- **The test suites do not execute on macOS** (`net10.0-windows`). Everything builds; CI is the arbiter for the xUnit assertions. What ran locally is listed above.
- **No live Aurora run.** Everything was measured on stock PostgreSQL 13–18. Both queries are core PostgreSQL and Aurora removes no columns from these catalogs, but that is reasoning, not measurement. Worth a probe from the PostgreSQL monitoring host.
- **No end-to-end service run.** The collectors were driven through `BuildQuery` → `ReadAsync` → `WritePayload` against live PostgreSQL, but not through a running `DarlingWorker` sweep into a store.
- **The bloat accuracy figures are from one instance's fixtures**, not a fleet. The direction (estimate slightly over) held on every table tested; the magnitude is not calibrated across workloads.
- **Index bloat is not covered** and is a real gap — `REINDEX CONCURRENTLY` is a much safer remedy than `VACUUM FULL`, so index bloat is arguably the more actionable half. It needs its own estimator over `pg_stats` on the indexed columns. Filing separately rather than widening this PR.
- **The `pgstattuple` escalation is a printed command, not a tool.** Making it a real on-demand read would need an ad-hoc query path to a monitored server, which the product deliberately does not have.


erikdarlingdata:
All six findings addressed in `6b7be5af`. Two of them were real bugs and the first is the sharpest thing anyone has said about this PR — thank you.

**1. The `LiveTuples == 0` gap — correct, and it is the worst kind of miss this feature can have.** `live_tuples = 0` is a real state distinct from the `-1` sentinel, the `> 0` guard short-circuited the whole staleness clause there, and the shape that reaches it — a table analyzed while empty and then heavily inserted into — is exactly what the "autovacuum has fallen behind" scenarios this collector exists for produce. Now `>= 0`, which at zero reduces to "any modification suppresses": correct, because with no live rows recorded any write makes the row count stale by definition.

Proven red rather than just fixed. The harness reproduces the defect with the old predicate and confirms the new one catches it, plus a control that a genuinely empty *unmodified* table still publishes — without that control the fix could have been satisfied by suppressing every zero-row table, which is a different wrong answer:

| case | old `> 0` guard | shipped `>= 0` guard |
|---|---|---|
| analyzed empty, then 400k modifications | **publishes** (the defect) | suppressed |
| empty, no modifications (control) | publishes | publishes |

**2. The sort key covered one of three suppression reasons — correct.** A row suppressed for stale statistics carries an arbitrarily large untrustworthy estimate (the measured case was 94.28%), so it could take the top of a `LIMIT` that reads as a work queue. All three conditions are in the `ORDER BY` now. You are right that this is a duplicated rule; an `ORDER BY` cannot call into C#, so it is a forced twin of `EstimateIsUnpublishable` and I have pinned it as one — `TheSortKeyCoversEverySuppressionReason` asserts each condition is present, and `TheSortKeyUsesTheSharedChurnRatio` asserts the inlined `0.2` equals the shared constant, so tuning the constant and missing the query goes red instead of quietly re-ranking.

**3. The `estimated_heap_pages IS NULL` path — correct, and worth fixing despite being low-probability.** It is now a fourth `estimate_unavailable` condition, OR-ed in at the outermost select because that is where the value first exists. You are right that the 1 MB floor probably makes it unreachable; "probably unreachable" is not the standard a flag documented as *"not advisory... a footnote is not a safeguard"* gets to be held to.

**4. The cadence doc contradiction — mine, and exactly the kind that gets "corrected" in the wrong direction.** Both comments now say daily and why (inherited from `index_object_stats`, not from the autovacuum sibling), and the harness pins that no comment in the file claims hourly while the schedule says 1440.

**5. The web/desktop severity parity gap — correct, and the fix had to go server-side.** `panels.js` deliberately never re-derives a band, so the browser could not have coloured these rows no matter what the columns declared. Both reads now emit a computed `severity` and the columns consume it through `sevKey`. The ranking mirrors the XAML triggers exactly, including the two cases that are easy to get wrong:

- a **suppressed** estimate bands as `Unknown`, not a severity — giving it one would assert precisely what the suppression denies, and it matches the neutral grey the desktop paints;
- an index we have **not watched for two samples** bands `Healthy`, not `Warning` — too-early-to-say must not look like a finding, same reason the droppability sentence refuses a disuse claim there.

You are right that the previous "never paints a suppressed estimate as high-bloat" claim was true on the web only vacuously. It is true structurally now.

**6. The runbook section split a list — correct.** Moved after the full list. The list's intro also said "Two results" over four bullets, which was stale before this PR touched it, so that is corrected too.

Re-verified after all six: the bloat query returns byte-identical percentages against the live PostgreSQL 16 target (churny 75.14, widget 27.29, widey 5.26, gadget 0.58, narrow 0.45, `namey` still suppressed), and both queries parse and return rows on PostgreSQL 13 and 18.

PR #2563, opened 2026-08-23
Surface the sessions behind a pinned xmin horizon (#2540)

### LINKED ISSUE #2540: Nothing surfaces idle-in-transaction sessions, which is what pins the xmin horizon we already collect
Part of #2530.

## The gap

`PgXminHorizonCollector` already collects the xmin horizon, and `PgAutovacuumStatsCollector` collects vacuum lag. So the product can already show that vacuum is falling behind and that something is holding the horizon back — and gives the user **no way to see what**.

The usual culprit is a session sitting `idle in transaction`. `pg_stat_activity` carries `state`, `state_change`, `xact_start`, `backend_xmin` and `wait_event`, and it is **already queried** by both `PgBlockingCollector` and `PgXminHorizonCollector` — so the source is wired and the connection is made. Nothing surfaces session state as a read.

That is an odd place to stop: we show the symptom and withhold the cause, when the cause is one column away in a view we already read.

## What it should answer

- Which sessions are `idle in transaction`, and for how long (`now() - state_change`).
- Which session holds the oldest `backend_xmin` — the one actually pinning the horizon.
- Long-running transactions generally (`xact_start`), since a slow transaction pins the horizon just as effectively as an idle one.

## Why it matters beyond tidiness

`idle in transaction` is the classic PostgreSQL foot-gun: an application that opens a transaction and waits on something external holds the horizon, which starves vacuum, which grows bloat, which ends in wraparound risk. We collect the last three links of that chain (#2530 groups them into a Vacuum surface) and not the first, so the chain reads as "your database is degrading" with no actionable end.

## Note

`pg_stat_activity` shows query text, which is subject to the same handling as any other captured statement — worth checking what the existing pg collectors already do with `query` before adding another copy of that decision.

erikdarlingdata:
Closes #2540.

## What was missing

`PgXminHorizonCollector` already names the CLASS of thing holding the horizon back. When that class is `session` it emits one row — a pid and a formatted detail string — which is enough to know a session is the cause and not enough to do anything about it. `PgAutovacuumStatsCollector` and `PgTableBloatStatsCollector` show the consequences. Nothing showed the session.

## The correction the issue needs

The issue frames `idle in transaction` as the culprit. Measured on the live PostgreSQL 16.15 rig, **two of the four idle-in-transaction shapes pin nothing at all**:

| shape | `backend_xmin` | `backend_xid` | pins the horizon |
|---|---|---|---|
| `BEGIN; SELECT …` READ COMMITTED | NULL | NULL | **no** — snapshot released at statement end |
| `BEGIN; UPDATE …` matching **0** rows | NULL | NULL | **no** — no xid was ever assigned |
| `BEGIN; UPDATE …` matching a row | NULL | set | yes, via its xid |
| `BEGIN ISOLATION LEVEL REPEATABLE READ; SELECT …` | set | NULL | yes, via its snapshot |

So duration in `idle in transaction` is **not** evidence that vacuum is being starved, and a read that reasons from the state string and the clock will tell someone to kill a session that costs them nothing. `horizon_age` is stamped from the GREATER of the two ages, and is `-1` rather than `0` when the session holds neither — 0 would read as *holding the newest possible xid*, the opposite finding. The MCP tool's finding text refuses the vacuum argument on those rows explicitly.

## The permission trap, measured

Same check #2560 made for `pg_stats`. Without `pg_monitor`, PostgreSQL does **not** refuse the read — it returns every row with `state`, `state_change`, `xact_start`, `backend_start` and `query_id` NULL and the query text replaced, **while leaving `backend_xmin` and `backend_xid` visible**. The horizon still reads as pinned and nothing can say by what. On one live capture over the same nine backends the privileged role saw four idle-in-transaction sessions and the unprivileged one saw **zero**. `state_is_redacted` is stamped from the privilege literal — not from `state IS NULL`, because background workers report NULL state under full privilege too (measured: `checkpointer`, `walwriter`, `autovacuum launcher`).

## Query text and PII

**No raw query text is stored**, departing from `pg_blocking` deliberately. `pg_stat_activity.query` carries literal parameter values inline (verified — a probe session's row came back with its literal argument verbatim). `pg_blocking` stores it because blocking is exceptional and the text is the finding; this fills on a duration floor an ordinary application crosses. Stored instead: `query_id` — PostgreSQL's own normalised fingerprint, joinable to `pg_statement_stats` whose text is already `$1`-normalised — and `command_tag`, the leading SQL keyword matched against a closed whitelist (a substring would leak, since ORMs prepend comment blocks).

## Version floors, measured not assumed

`pg_attribute` read for the view on a live **13.23** instance: 21 columns, including `leader_pid`, `backend_type`, `backend_xid` and `backend_xmin`, and **no `query_id`**. That one column is version-substituted with a typed NULL below 14 so the row shape is constant across a mixed fleet; the shipped PG13 text was run against that instance.

## Verification

Everything below ran the **shipped** string — dumped from `BuildQuery(ctx).Text`, never retyped.

- Shipped PG16 SQL against the live target with four real held-open transactions: correct rows, correct `is_horizon_holder`, `command_tag` = `UPDATE`/`SELECT`, no literals anywhere.
- Shipped PG13 SQL against a live PostgreSQL 13.23: parses and runs, identical column shape.
- Same shipped SQL as a least-privileged role: the redaction path above.
- **A real defect this caught:** the first draft's comment contained a literal SQL comment opener. PostgreSQL block comments **nest**, so the whole collection failed to parse. Text assertions could never see it.
- Ladder-generator diff across all twelve PostgreSQL rungs, **proven RED first** by changing V86's `pid` column type.

## Schema

V86 / `StorageVersion.SchemaVersion = 86`. Worker sizing re-derived in both workflows: 54-collector catalog + `collection_log` = 55 hypertables → 57 / 68.

Also fixes a drift artifact in `ViewerDataService`: three comment blocks had stacked above the top arm, each rung inserting its own above the last, so #2530's prose was explaining an arm three rungs away.


erikdarlingdata:
All four review findings addressed. Two were real, and one of them changed what the code says to a user at runtime.

### 1 & 2 — the panel that did not exist (`ViewerDataService.Postgres.cs:309`, `docs/uat-onboarding.md:510`)

Correct, and already fixed in `ef3ecbcf` — the review was reading the previous head. The Vacuum tab now carries the sessions panel as its **first** panel on both surfaces. First, not appended: that tab's own note calls itself a causal chain ("something holds the xmin horizon, the horizon starves vacuum, and vacuum falling behind ends in wraparound"), and this is the link upstream of panel 1 — it names *which* session holds the horizon. Four panels now, note rewritten, `POSTGRES_TABS` still seven tabs.

The three ratchets that were red are green on that commit: `EveryPostgresCollector_IsShownOnExactlyOnePostgresTab`, `EveryPostgresRead_IsReachableFromThePostgresRegistry_AndFromNoOtherTab`, `EveryAuroraOnlyPostgresRead_SitsOnATabThatSaysSo`.

### 3 — `active_sessions` under redaction (`PgSessionStatesCollector.cs:308`)

Right, and the correction matters. Redaction is **per-backend and ownership-based**, not per-role-wide — I re-measured it, and an unprivileged role does see its OWN backend with a live `state` and a real `query` while every other row is redacted. So those two counts do not go to zero on principle; they count only the backends the login owns, which after this collector's self-exclusion is usually zero but is not guaranteed to be. They **under-report** rather than report nothing, which is the more dangerous shape of the two. Comment rewritten to say that instead of "therefore 0 there".

### 4 — the five disagreeing restatements (`PgSessionStatesCollector.cs:77`)

The best catch of the four, because the discrepancy was baked into the finding string a user reads at runtime. Rather than pick whichever of the two lists looked right, I measured **every column** of `pg_stat_activity` as an unprivileged role against a backend it does not own, on live PostgreSQL 16.15:

| | |
|---|---|
| **survives** | `pid`, `application_name`, `datname`, `usename`, `backend_xid`, `backend_xmin` |
| **NULL** | `state`, `state_change`, `xact_start`, `query_start`, `backend_start`, `wait_event_type`, `wait_event`, `backend_type`, `client_addr`, `leader_pid`, `query_id` |
| **replaced** | `query` → `<insufficient privilege>` |

Neither of the two competing lists was complete. All five call sites now state it as **what survives**, which is shorter, exactly checkable, and sharper — the last two survivors are precisely the two that make the trap what it is.

Two things fell out of the measurement that no restatement had:

- **`leader_pid` is redacted.** The parallel-worker guard is `leader_pid IS NULL OR leader_pid = pid`, so without `pg_monitor` it stops excluding anything and a parallel worker comes through as its own row. That errs toward *including* rather than dropping — the safe direction for a guard whose only job is avoiding double-counting, and the row is flagged untrustworthy anyway — but it is now written down rather than accidental.
- **`wait_event_type` and `wait_event` are redacted too**, which nothing had mentioned.

### One more, found while wiring the UI

`SessionSeverity` returned a private lower-case vocabulary (`critical`/`warning`/`info`/`none`/`unknown`) where the house set is `Healthy`/`Warning`/`Critical`/`Unknown` and the shared `sev-*` CSS defines **only those four**. `sev-critical`, `sev-info` and `sev-none` match no rule, so the badge would have rendered **unstyled rather than failing** — the kind of defect that ships. Fixed in the read rather than worked around in the UI, so the web grid now uses `sevKey` like its siblings. A passing sighting collapses into `Healthy` rather than getting a band of its own: every write transaction is briefly the oldest xmin holder, so painting that would report ordinary traffic as a cause.

### What was verified, and how

Everything below ran the **shipped** string, dumped from `BuildQuery(ctx).Text` / the reader's `const`, never retyped.

- Shipped PG16 SQL against the live target with four real held-open transactions covering all four idle-in-transaction shapes; re-run after **every** subsequent comment edit, because a comment inside this query has already broken its parse once (PostgreSQL block comments nest).
- Shipped PG13 SQL against a live PostgreSQL 13.23.
- Same shipped SQL as a `pg_monitor`-only role and as a no-grant role.
- Shipped **reader** SQL against the live store over an adversarial fixture: the culprit named `zzz-worker` and an equal-duration decoy named `aaa-app` (so any alphabetical tiebreak picks wrong), pid 100 reused by two backends, three unrelated situations sharing one capture, and one capture truncated. The holder sorted first, the reused pid stayed two rows, and the decoy kept `peak_horizon_age = -1`.
- Ladder-generator diff across all twelve PostgreSQL rungs, **proven red first** by changing V86's `pid` column type.
- Classifier boundaries against the real build (the xUnit suites are `net10.0-windows` and cannot run on macOS): 4/10, 5/10 and 9/10 against the half-share threshold, exactly 5 minutes on the attention floor, and redaction outranking a would-be critical. 19/19.

**Not verified by me:** the xUnit suites themselves, and the rendered appearance of either UI. CI is the arbiter for the first; the second is compile-and-pin-verified only.

erikdarlingdata:
Both new findings addressed in `24ebd39b`, plus the red `build` — which was my own test, not the code.

### The red build: my guard couldn't tell what it forbids from what it requires

`Lite.Tests…NoRawQueryTextIsEverProjectedOrStored` failed on `Assert.DoesNotContain("AS query", sql)` — because `AS query_id` **contains** `AS query`, and `query_id` is the fingerprint column this collector deliberately emits. The guard failed against correct code.

Now a regex with a trailing word boundary: `\b` after `query` requires a non-word character next and `_` is a word character, so `AS query_id` no longer matches while `AS query,` still does. Verified both directions against the shipped string.

I did **not** tighten it to `(?!_id)` — that form also fires on `AS query_duration_ms`, which this collector emits too. The `PayloadColumns` assertion below it already catches a `query_text` column at the storage layer, which is the layer that actually matters.

Wrong in the safe direction (it cried wolf rather than staying quiet), but a guard that cannot distinguish the thing it forbids from the thing it requires is not yet a guard.

### `SessionSeverity`'s Warning branch — real bug, opposite fix

The disagreement you found is real and worse than described: that row got **three different answers** — `severity: "Warning"` from the MCP, a `finding` saying "not a finding", and **no highlight at all** in WPF, all for the same session.

But the suggested fix (`&& r.PeakHorizonAge < 0`) is the wrong direction. **`is_horizon_holder` means OLDEST ON THE INSTANCE, not HOLDS.** On a busy instance several genuinely long idle transactions each take turns being oldest — so none of them clears the sustained threshold, and that gate would paint **every one of them Healthy**. That is the dangerous direction, and it is exactly the population someone needs to see.

So the Warning band stays "long idle in transaction, full stop" — whether it pinned is an orthogonal fact the horizon columns already carry — and the **finding text and the viewer flag were changed to agree with it**. `IsIdleWithoutHorizon` is renamed `IsLongIdleTransaction` because that is now what it means.

You were also right that the parity test couldn't have caught it: no `InlineData` row combined `HolderSamples > 0` with `PeakHorizonAge >= 0` **and** a long idle. `[InlineData(10, 3, 10, 600_000L, 40_000L, false, "Warning")]` added, with the reasoning above in the comment.

### `backend_duration_ms` — fixed as a category, not an instance

Correct, and it is the "captured data that no read reports" category, so the repair is an invariant rather than one more remembered column:

```
EveryStoredColumn_IsReadBackBySomeRead()
```

It derives both sides from the **shipped** artifacts — `PayloadColumns` and the reader's own SQL constants — so it cannot keep passing while the two drift underneath it. Any future column that is stored and never selected fails by name.

The column is now surfaced on the MCP payload and both grids as **backend age**, which earns its place: a connection created ten minutes ago that has been idle in transaction for all ten is a pool handing out a session nobody finished with; a three-day-old worker holding one for ten minutes is a code path that forgot to commit. Same transaction duration, different bug.

Reader arity re-verified after the insert — 30 projection columns, 30 record fields, 30 contiguous `MapRow` ordinals.

### Verification on this round

Viewer, Service, `Darling.Tests` and `Lite.Tests` all build clean on macOS; `node --check` passes; the classifier harness re-run against the real build is 19/19. The xUnit suites still cannot execute on macOS — CI is the arbiter, and this is the first push where the Windows `build` job has a chance to run to completion rather than being cancelled by the next push.

PR #2600, opened 2026-08-25
Attribute per-database PostgreSQL rows to their database (#2599)

### LINKED ISSUE #2599: PostgreSQL: two collectors report per-database facts with no database_name (pg_extension_availability, pg_column_stats)
Two collectors shipped in v89 and v91 report **per-database facts without recording which database they came from**. Found on a live Aurora PostgreSQL target, where two of our own collectors read the same catalog on the same cluster and returned contradictory answers.

## The evidence

`pg_table_bloat_stats` (per-database) and `pg_extension_availability` (single connection) both read `pg_catalog.pg_extension` for `pgstattuple` on the same cluster in the same collection cycle:

| collector | scope | what it reported for `pgstattuple` |
|---|---|---|
| `pg_table_bloat_stats.pgstattuple_available` | per-database | **true** — a `pg_extension` row exists |
| `pg_extension_availability.state` | one connection | **`available`** — i.e. `installed_version IS NULL`, no `pg_extension` row |

Both readings are correct. `pg_extension` is per-database, so they were simply looking at different databases — and neither row says so.

## Defect 1 — `pg_extension_availability` reports a per-database state from one connection

`PgExtensionAvailabilityCollector` never overrides `RunsPerDatabase`, so it runs once per target against the configured database. But `installed` and `outdated` are derived from `pg_extension`, which describes **only the connected database**. `available` and `absent` come from `pg_available_extensions` and are genuinely cluster-wide.

So the table mixes two scopes in one `state` column, and the answer flips depending on which database the target happens to be configured for — with nothing in the output to say which.

The collector's own doc comment already says this is wrong (`PgExtensionAvailabilityCollector.cs:41-45`):

> `pg_available_extensions` is cluster-wide (what the binaries offer), while `pg_extension` is PER-DATABASE (what has been created here). Measured: `pgstattuple` reported installed in one database and not in another on the same cluster [...] So `installed` and `outdated` are claims about THIS DATABASE ONLY, and the column names and the read both have to say so — reporting "not installed" from the maintenance database about an [extension installed elsewhere is misleading].

The documentation is right; the collector does the thing it warns against.

## Defect 2 — `pg_column_stats` is per-database with no `database_name`

`PgColumnStatsCollector` sets `RunsPerDatabase => true` but does not declare a `database_name` payload column, so the store table is:

```
collection_id, collection_time, server_id, server_name,
schema_name, table_name, column_name, n_distinct, null_frac,
avg_width, correlation, top_value_frequency, common_value_count
```

Rows from different databases collide on `(server_id, schema_name, table_name, column_name)`. Any cluster with the same schema in two databases — the normal multi-tenant shape — produces two indistinguishable rows per column. `pg_table_bloat_stats` and `pg_index_bloat` both declare `database_name`; this one was missed.

Currently masked: the collector returns 0 rows against a `pg_monitor`-only role, because `pg_stats` is filtered by `has_column_privilege` (documented, and the runbook already prescribes `pg_read_all_data`). The collision appears as soon as that grant is added.

## Fix

Schema **v95**:

1. Add `database_name` to `pg_column_stats` and to `pg_extension_availability`.
2. `PgExtensionAvailabilityCollector.RunsPerDatabase => true`.
3. Have both collectors emit `current_database()`.
4. Rename or caption the state so the two scopes are legible: `installed`/`outdated` are per-database, `available`/`absent` are cluster-wide. A read that asks "is pgstattuple usable here" must be answered per database.
5. Pin it — a test asserting every `RunsPerDatabase` collector declares a `database_name` column. That assertion would have caught both of these at build time, and catches the next one.

Item 5 is the one that matters beyond this issue: this is the fifth time a PostgreSQL catalog's scope has not matched its name in this codebase.

erikdarlingdata:
Fixes #2599.

Two of our own collectors contradicted each other on a live Aurora target, which is how this surfaced:

| collector | scope | `pgstattuple` |
|---|---|---|
| `pg_table_bloat_stats` | per database | **installed** — a `pg_extension` row exists |
| `pg_extension_availability` | one connection | **`available`** — no `pg_extension` row |

Both reads were correct. `pg_extension` is per-database, so they were looking at different databases — and neither row said which.

## What changed

**Schema v95** adds `database_name` to `collect.pg_column_stats`, `collect.pg_index_bloat` and `collect.pg_extension_availability`, and `PgExtensionAvailabilityCollector` now returns `RunsPerDatabase => true`.

`pg_extension_availability` was reporting `installed`/`outdated` — per-database facts from `pg_extension` — alongside `available`/`absent`, which are cluster-wide facts from `pg_available_extensions`, all through a single connection. The collector's own doc comment (lines 41–45) already explained why that is wrong. The documentation was right; the code did the thing it warned against.

**The readers were the user-visible half.** All three used `DISTINCT ON` over a key that did not include the database:

```sql
SELECT DISTINCT ON (schema_name, table_name, column_name)   -- before
SELECT DISTINCT ON (database_name, schema_name, table_name, column_name)   -- after
```

On any cluster carrying the same schema in two databases — the ordinary multi-tenant shape — the rows collapsed to one and the newest `collection_time` silently picked which database the grid was describing. Each of the three grids now shows a Database column.

## The pin

`PerDatabaseCollectorAttributionTests` walks `CollectorCatalog.All` and fails the build for any collector that returns true from `RunsPerDatabase` without declaring a `database_name` payload column.

It was **proven red before the fix**, and it immediately found a third offender this investigation had missed — `pg_index_bloat`, merged the same day:

```
ok        pg_table_bloat_stats
OFFENDER  pg_column_stats
OFFENDER  pg_index_bloat
-- per-database collectors examined: 14
```

and after:

```
ok        pg_extension_availability
ok        pg_column_stats
ok        pg_index_bloat
-- per-database collectors examined: 15
```

`RunsPerDatabase` is declared on the generic `ICollectorDefinition<TRow>` while the catalog is typed as the row-agnostic `ICollectorSchemaInfo`, so the test reaches it by reflection — there is no non-generic surface to call it through.

## Migration shape

The column is added to the V89/V91/V94 `CREATE TABLE` statements **and** as `ADD COLUMN IF NOT EXISTS` in V95. Both are needed and they converge:

- a store new enough to create the tables gets the column from the CREATE, and V95 no-ops;
- a store that already ran V89/V91/V94 gets it from V95.

The CREATE edits are required because `EveryPostgresRung_IsIdenticalToTheGeneratedSchema` holds those hand-written rungs identical to the generator's output, and the generator now emits the column. Verified locally that all three still match, that the rung ladder tops out at 95, and that `StorageVersion.SchemaVersion` is 95.

Nullable with no DEFAULT and no backfill — a catalog-only change in PostgreSQL, so it stays instant on a compressed hypertable, and rows collected before this rung genuinely do not know which database they came from. NULL is honest there; guessing the target's default database would be wrong for every row the sweep collected elsewhere.

## Notes

- No new tables, so the hypertable count and the CI worker sizing are unchanged.
- Probe arity checked: 71 select expressions, 71 map parameters, ordinals 0–70.
- This is the fifth time a PostgreSQL catalog's scope has not matched its name in this codebase. The test is the part meant to outlast the fix.


erikdarlingdata:
**Cost of making `pg_extension_availability` per-database, measured rather than assumed.**

The cluster-wide half of that collector (`available`/`absent`, from `pg_available_extensions`) is now duplicated once per database. On the live Aurora target:

- 102 rows per run today
- **5 databases** on the cluster (only 1 carries user indexes)

So ~510 rows per target per day, against a 1440-minute cadence and 365-day retention — roughly 186k rows per target per year. That is the honest price of making the per-database half attributable, and it is not a number worth optimising around.

**Also fixed here:** `StoreSchemaProbe_ColumnCount_MatchesTheMapArity` failed on the first run, and the cause was not the arity. It counts top-level commas by paren depth, and a comma inside a `/* */` comment sits at depth zero — indistinguishable from a select-item separator. The explanatory comment beside the new V95 sentinel took the count from 71 to 73.

Comments are now stripped before counting; the arity assertion itself is unchanged. The sentinels are the part most worth commenting, so the test should not tax anyone for doing it. Verified locally with the test's exact algorithm: 71 columns, 71 map parameters.

Issue #2594, opened 2026-08-25
[BUG] Lite - Table ".main.*" could not be found


dphugo:
### Component

Lite

### Performance Monitor Version

3.5.0

### SQL Server Version

SQL Server 2022 (RTM-CU16)

### Windows Version

Windows 11 23H2

### Describe the Bug

The Lite dashboard would be running fine and seem to sporadically enter a state where it can't find any of the tables in DuckDB.
All the graphs goes blank and it shows an error at the bottom in Red.
Unfortunately I don't have a screenshot of it.

### Steps to Reproduce

It is sporadic as I've done normal navigation and couldn't find a consistent pattern to reproduce.

### Expected Behavior

It should not error. 😊

### Actual Behavior

The Lite dashboard would be running fine and seem to sporadically enter a state where it can't find any of the tables in DuckDB.
All the graphs goes blank and it shows an error at the bottom in Red.
Unfortunately I don't have a screenshot of it.

### Error Messages / Log Output

```text
2026-08-25 08:48:45.664 [ERROR] [Collector] COLLECTION LOGGING FAILED (1x): DuckDBException: Catalog Error: Table with name collection_log does not exist!
Did you mean "pg_collation"?
2026-08-25 08:48:45.664 [ERROR] [Collector] Failed to log collection for index_object_stats (failure #1) | DuckDBException: Catalog Error: Table with name collection_log does not exist!
Did you mean "pg_collation"?
2026-08-25 08:48:45.688 [ERROR] [Collector] Stack:    at DuckDB.NET.Data.PreparedStatement.PreparedStatement.PrepareMultiple(DuckDBNativeConnection connection, String query, DuckDBParameterCollection parameters, Boolean useStreamingMode)+MoveNext()
   at DuckDB.NET.Data.DuckDBCommand.ExecuteNonQuery()
   at System.Data.Common.DbCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location ---
   at PerformanceMonitorLite.Services.RemoteCollectorService.LogCollectionAsync(Int32 serverId, String serverName, String collectorName, DateTime startTime, String status, String errorMessage, Int32 rowsCollected, Int64 sqlMs, Int64 duckDbMs) in C:\Users\daniel.hugo\source\repos\erikdarlingdata\PerformanceMonitor\Lite\Services\RemoteCollectorService.cs:line 749
[...]

PR #2601, opened 2026-08-25
Status-bar size read must not hold an unlocked DuckDB connection (#2594)

erikdarlingdata:
Addresses one concrete gap found while investigating #2594. **It does not close that issue** — the diagnosis there still depends on the reporter's `Database files deleted, reinitializing` timestamp — but this is worth fixing whatever that answer turns out to be.

## The gap

`DuckDbInitializer.GetUsedDataSizeMb` opens a DuckDB connection **without taking the read lock**:

```csharp
using var connection = CreateConnection();
connection.Open();
using var cmd = connection.CreateCommand();
cmd.CommandText = "SELECT (used_blocks * block_size)::BIGINT FROM pragma_database_size()";
```

Every other read in that class takes the lock first. This one is also the only **periodic** unlocked site: it is called from `UpdateStatusBar()`, driven by a `DispatcherTimer` with a 30-second interval (`MainWindow.xaml.cs:125`), to render the "Database: X / Y MB" figure. So while the dashboard is open, a live handle lands on the database file every 30 seconds — including at moments the archival path may be deleting and recreating it.

A bare `catch { }` returning null means it reports nothing when it fails, so this has been invisible.

## Why the obvious fix is wrong

Taking `AcquireReadLock()` like every other read would run on the **dispatcher thread**. A long archival holds the write lock, and the window would freeze while waiting to draw a number nobody is reading.

So the contract is a **bounded** attempt — a new `TryAcquireReadLock(TimeSpan)` that returns null instead of waiting. The status bar gives up after 100 ms and the caller renders the file size alone, which it already does when this returns null. The degraded answer is a smaller status bar, not a stalled window.

`TryAcquireReadLock` takes a timeout rather than a cancellation token, unlike `AcquireReadLock(CancellationToken)`. That asymmetry is deliberate and documented: the token exists there because the right budget is the caller's remaining budget, and this caller genuinely has a number to pick — what a UI thread may spend.

## The red run is on purpose

The first commit on this branch is **the test alone**, so the pin is demonstrated failing rather than asserted to fail. Lite's test project is `net10.0-windows` and cannot run on this host, so CI is the only place that demonstration can happen.

Expect the first run to fail `GetUsedDataSizeMb_WhenTheWriteLockIsHeld_GivesUpInsteadOfBlocking`: the current code never consults the lock, so it opens its connection and returns a size where the test expects null. The fix commit follows.

The test asserts **both** halves — that the read gives up, and that it gives up quickly. Asserting only "it takes a lock" would pass a fix that hangs the UI.

## Still open, deliberately not touched here

`CreateArchiveViewsAsync` is a second unlocked connection site, and unlike this one it is unlocked **on purpose** — `ArchiveService.cs:193` says "Refresh archive views outside write lock — view creation is fast and safe". That is an assertion rather than a proof, and it is a bigger call than this PR. Flagging it rather than widening the change.

### DIFF
--- a/Lite.Tests/StatusBarSizeReadLockTests.cs
+++ b/Lite.Tests/StatusBarSizeReadLockTests.cs
@@ -35,7 +35,7 @@ public class StatusBarSizeReadLockTests
     private static readonly TimeSpan WriteLockHold = TimeSpan.FromSeconds(10);
 
     [Fact]
-    public void GetUsedDataSizeMb_WhenTheWriteLockIsHeld_GivesUpInsteadOfBlocking()
+    public async Task GetUsedDataSizeMb_WhenTheWriteLockIsHeld_GivesUpInsteadOfBlocking()
     {
         var tempDir = Path.Combine(Path.GetTempPath(), "pmlite-statusbar-" + Guid.NewGuid().ToString("N"));
         Directory.CreateDirectory(tempDir);
@@ -62,7 +62,7 @@ public class StatusBarSizeReadLockTests
             stopwatch.Stop();
 
             release.Set();
-            holder.Wait(TimeSpan.FromSeconds(15));
+            await holder.WaitAsync(TimeSpan.FromSeconds(15));
 
             /* Null, not a number: the caller renders the file size alone in this state, which is the
                degraded answer this is choosing on purpose. */
--- a/Lite.Tests/StatusBarSizeReadLockTests.cs
+++ b/Lite.Tests/StatusBarSizeReadLockTests.cs
@@ -1,0 +1,88 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using PerformanceMonitorLite.Database;
+using Xunit;
+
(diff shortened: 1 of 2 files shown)


erikdarlingdata:
**The pin is demonstrated red.** First run on this branch, test-only commit `d64ac32`:

```
PerformanceMonitorLite.Tests.StatusBarSizeReadLockTests.
  GetUsedDataSizeMb_WhenTheWriteLockIsHeld_GivesUpInsteadOfBlocking [FAIL]
    Assert.Null() Failure: Value of type 'Nullable<double>' has a value
    Expected: null
    Actual:   0

Lite.Tests  Total: 3018, Errors: 0, Failed: 1, Skipped: 0
```

The write lock was held for the whole call and the read returned a number anyway — because it never consulted the lock. One failure out of 3018, and it is the intended one, which is the part worth checking: a red run for an unrelated reason would prove nothing.

The fix commit `56305f9` follows. It also makes the test `async` so its task wait is awaited rather than blocking, which the first run flagged as `xUnit1031`.

Issue #2629, opened 2026-08-26
Nine PostgreSQL collectors are readable only from the Windows Viewer — the MCP and web surfaces cannot see them


erikdarlingdata:
Found while exercising every read tool against the self-hosted rig.

## The gap

Nine PostgreSQL collectors have a WPF Viewer panel and **nothing else**. No MCP tool, no web read endpoint:

| collector | rows on the rig right now | reachable from |
|---|---|---|
| `pg_kernel_stats` | 253 | Viewer only |
| `pg_column_stats` | 240 | Viewer only |
| `pg_predicate_stats` | 240 | Viewer only |
| `pg_buffer_usage` | 189 | Viewer only |
| `pg_index_bloat` | 160 | Viewer only |
| `pg_extension_availability` | 100 | Viewer only |
| `pg_wait_sampling` | 63 | Viewer only |
| `pg_lock_stats` | 0 (idle rig) | Viewer only |
| `pg_write_stats` | 1 | Viewer only |

Thirteen `get_pg_*` reads are served. These nine are not among them, and the web `POSTGRES_TABS` registry carries the same thirteen.

## Why it matters more than a missing endpoint

**The Viewer is Windows-only WPF.** Every one of these is collected on a Linux or macOS host and then unreadable from that host — and unreadable by an agent anywhere, which is the surface this product is differentiated on. We are paying the collection cost on every cycle and shipping the answer to one GUI.

It is sharpest for `pg_wait_sampling`. #2625 just shipped a message that tells a stock-PostgreSQL operator, correctly, that `pg_wait_stats` is Aurora-only and **`pg_wait_sampling` answers the same question**. It points at a panel the MCP cannot read. On a non-Aurora target "why is this slow" now has an answer that is only visible if you are sitting at a Windows machine.

`pg_index_bloat` is the same shape: #2617 and #2622 were both spent getting it to return rows, and it returned 160 of them measuring 89.6% average leaf density — into a table nothing but the Viewer queries.

## Suggested order

By what question each one closes, not by table size:

1. **`pg_wait_sampling`** — waits by queryid. The one #2625 explicitly promises, and the primary "why is it slow" answer on every non-Aurora target.
2. **`pg_kernel_stats`** — OS CPU and disk per queryid. Pairs with the above: elapsed vs. on-CPU is the first split any tuning question needs.
[...]

Issue #2625, opened 2026-08-26
Self-hosted PostgreSQL has no top-queries answer at all: pg_statement_stats is Aurora-only


erikdarlingdata:
Surfaced by the self-hosted rig — the first non-Aurora PostgreSQL target this product has ever run against.

## The gap

`PgStatementStatsCollector` reads `aurora_stat_statements(false)` and gates itself `AppliesTo => target.IsAurora`. That is correct for what it reads. The consequence is that on a self-hosted PostgreSQL target, **the single most-asked question a database monitor exists to answer — which queries cost the most — has no answer at all.**

```
GET /api/read/get_pg_top_queries
{"status":"not_collected","message":"… runs PostgreSQL. The pg_statement_stats collector does not run
on that engine — its own AppliesTo gate excludes it — so this server does not …"}
```

The rig has `pg_stat_statements` loaded and populated. `pg_kernel_stats` is collecting 251 rows of OS CPU and disk **keyed by the very queryids** that would identify those statements, and `pg_predicate_stats` is collecting 240 rows of predicate selectivity for them. We collect everything *around* the statement and not the statement.

The collector's own doc comment already says it: "pg_stat_statements would serve those — but this definition reads the Aurora-extended function."

## What to build

A sibling `pg_statement_stats` capture path over vanilla `pg_stat_statements`, selected at dispatch by flavor the way `pg_plan_capture` already picks between the RDS log API and `pg_read_file`. Not a new collector name and not a new table if the column set can be shared — the Aurora-extended columns are a superset, so the vanilla path fills the shared columns and leaves the Aurora-only ones null, which is exactly what the miss vocabulary is for.

Notes for whoever picks it up:

- **The delta key must stay `(queryid, dbid, userid, toplevel)`** — the full `pg_stat_statements` identity. Aurora's function returns the same identity columns, so this is unchanged, but it is the thing that breaks silently if the vanilla query selects a narrower key.
- **`toplevel` needs a version guard.** It arrived in `pg_stat_statements` 1.9 / PG 14. The rig is 17; a 13 target has no such column and the query must not simply fail there.
- **Do not read logical or physical block counts.** `pg_stat_statements` exposes them; they are not what this product reports.
[...]

Issue #2612, opened 2026-08-25
HypoPG: test an index without building it — needs a decision about the product acting, not just reading


erikdarlingdata:
Split out of #2545, which is otherwise complete.

## Why this is not another collector

Every other extension on #2545 was a passive data source: something accumulates numbers, the collector reads them on a schedule. **HypoPG is not that.** It works by having you create a hypothetical index — visible only to your own session, never written to disk — and then running `EXPLAIN` to see whether the planner would have used it.

So there is nothing to collect. It is an **experiment**, and the product would be running it against a customer's database on their behalf.

## What it would be worth

It closes the loop on the collector that just shipped. `pg_predicate_stats` (#2611) names columns that are filtered on heavily with nothing supporting them — the index candidates. HypoPG is what turns "this column looks like a candidate" into "the planner would switch to it, and the estimated cost drops by this much." Without it every recommendation is an inference; with it there is an answer.

That pairing is the reason to do it at all, and it is why this is worth more now than it was when #2545 listed it as "further out."

## What has to be decided first, because it is a change in product posture

Darling reads. This would make it **act** — running `EXPLAIN` on a customer's statements with planner state the customer did not ask for.

The mitigations are real: a hypothetical index costs no disk, exists only inside one session, and disappears when that session ends. `EXPLAIN` without `ANALYZE` does not execute the statement. So the blast radius is genuinely small.

But small is not none, and the questions are not technical:

- **Whose statements?** Replaying a captured statement means re-planning something a customer ran, at a time nobody chose.
- **On which server?** A primary under load is the worst place to do speculative planning; a replica has different statistics.
- **Triggered by whom?** A scheduled sweep that quietly tests indexes is a very different product from a button someone presses while looking at a recommendation.

My read: **on demand only, never scheduled**, driven from a specific predicate row a human is already looking at. That keeps it an answer to a question somebody asked rather than a background activity nobody authorised. But that is a call about what this product is, not an implementation detail, so it belongs to Erik rather than to whoever picks the issue up.

## Groundwork already done

- `hypopg` 1.4.3 verified installable and working locally, and it is **`available` on Aurora** — one of the few on #2545's list that is.
- `pg_predicate_stats` supplies the candidates, with the selectivity and the planner's estimate error already separated (they are different problems and need different remedies).

erikdarlingdata:
**Erik's call: on demand only, never scheduled.**

[...]

PR #2652, opened 2026-08-26
Read statement text from the vanilla view too, and guard the class

### LINKED ISSUE #2651: pg_statement_text reads aurora_stat_statements only, so self-hosted PostgreSQL has no query text — and test_hypothetical_index cannot work there
Found by running `test_hypothetical_index` (#2612) end to end against the verification rig. It could not resolve a statement, and the reason is one line.

## The defect

`PgStatementText.FetchSql` reads Aurora's function and has no vanilla path:

```sql
FROM aurora_stat_statements(true) AS s
```

`aurora_stat_statements()` does not exist on a self-hosted PostgreSQL server, so `collect.pg_statement_text` is **never populated there**. Measured on the rig after a full collection cycle:

```
pg_statement_stats  486 rows      <- collecting fine since #2625
pg_predicate_stats  160 rows
pg_statement_text     0 rows      <- and it never will be
```

This is #2625 exactly, in the sibling table, and I missed it while fixing that one.

## What it breaks

**`get_pg_top_queries` returns `query_text: null` for every row** on every self-hosted target, permanently. I saw that in a read sweep and misread it as "text has not been captured for this queryid yet" — which is what the field's own documentation says it means, and which is true on Aurora. Here it means *never*.

**`test_hypothetical_index` (#2612) cannot work at all.** It resolves the statement from this table by queryid, so on the only platform it can currently be tested against it always answers:

> no statement text is stored for queryid N on 'X'

A feature shipped yesterday, dead on arrival on self-hosted PostgreSQL, and the message blames a refresh cadence rather than a missing source.

## Fix

The same shape #2625 used for `pg_statement_stats`: branch the source on `IsAurora`, and read `query` and `queryid` from the vanilla `pg_stat_statements` view otherwise. The columns needed here are the two that both sources have, so the vanilla arm is strictly simpler than the Aurora one — no NULL-filling, no ordinal juggling.

Two details worth carrying over:

- **`toplevel` is filtered, not grouped**, and it arrived in `pg_stat_statements` 1.9 (PostgreSQL 14). The same version guard `PgStatementStatsCollector` already applies belongs here, or the query fails on a 13 target rather than degrading.
- **`ORDER BY total_exec_time DESC LIMIT $1`** must stay: the point is the text for the statements that matter, not every statement the server has ever parsed.

## The wider miss

`pg_statement_text` is not a collector, so it was not in the catalog sweep that found #2629's nine Viewer-only collectors, nor in the `AppliesTo` audit that found #2625. It is refreshed directly by the worker. Worth a guard that catches Aurora-only SQL anywhere in the PostgreSQL path, not only inside `ICollectorDefinition` implementations — `git grep aurora_stat_` returns three sites and only two of them were ever examined.

erikdarlingdata:
Fixes #2651. Found by running `test_hypothetical_index` (#2612) **end to end** against the verification rig, rather than by reading code — the experiment class had been tested directly and passed, and the path from a queryid to a statement had not.

## The defect

`PgStatementText.FetchSql` read `FROM aurora_stat_statements(true)` with no vanilla path, so off Aurora `collect.pg_statement_text` was **never populated**:

```
pg_statement_stats  486 rows      <- collecting fine since #2625
pg_predicate_stats  160 rows
pg_statement_text     0 rows      <- and never would be
```

Two silent failures on every self-hosted target:

- **`get_pg_top_queries` returned `query_text: null` on every row, forever** — while that field's own documentation says null means "not captured *yet*". True on Aurora. A lie here. I saw exactly that in a read sweep and accepted the documented meaning.
- **`test_hypothetical_index` could never work at all.** It resolves its statement from that table, so it always answered "no statement text is stored" — a feature shipped yesterday, dead on arrival, with a message pointing away from the cause.

This is #2625 again in the sibling table. `pg_statement_text` is not a collector — the worker refreshes it directly — so it was in neither the catalog sweep behind #2629 nor the `AppliesTo` audit behind #2625.

## Fixing it exposed a second defect immediately

```
[pg-selfhosted] pg_statement_text refresh failed: 21000: ON CONFLICT DO UPDATE command cannot affect row a second time
```

`pg_stat_statements` keys on `(queryid, userid, dbid, toplevel)`, so **one queryid comes back once per user and database that ran it**. The upsert keys on `(server_id, queryid)` and met those duplicates by abandoning the whole batch — every cycle, quietly, because the caller deliberately treats a text-refresh failure as non-fatal.

`DISTINCT ON` fixes it, **inside the subquery** so the cap is still spent on the costliest statements rather than on duplicates of one. There is a test pinning that ordering, because reversing it would look correct and quietly waste the cap.

`toplevel` is guarded for PostgreSQL 13 and older, where nested tracking did not exist and every row *is* top level — the same guard `PgStatementStatsCollector` already applies.

## Verified end to end on a real self-hosted PostgreSQL

47 statement texts stored where there were zero. Then the command, through the queue, the worker and the store lookup:

```json
{ "candidate": "CREATE INDEX ON public.t7 (amount)",
  "planner_would_use_it": true,
  "estimated_cost_before": 1059.34,
  "estimated_cost_after": 438.71,
  "hypothetical_index_name": "<13630>btree_t7_amount" }
```

> The planner WOULD use this index: estimated cost falls from 1,059.34 to 438.71, a 58.6% reduction.

Session left clean afterwards — `hypopg_list_indexes` returns 0.

## Worth following up

`git grep aurora_stat_` returns three SQL sites. Two had been examined; this was the third. A guard that catches Aurora-only SQL anywhere in the PostgreSQL path — not only inside `ICollectorDefinition` implementations — would have caught it.


erikdarlingdata:
Added the guard I flagged in the issue rather than leaving it as a note.

`AuroraOnlySqlIsGatedTests` walks the **source**, not `ICollectorDefinition` implementations — which is exactly why this defect survived the audit that found #2625. `pg_statement_text` is not a collector; the worker refreshes it directly, so it was invisible to both the catalog sweep behind #2629 and the `AppliesTo` audit behind #2625.

Every file naming an Aurora-only surface must be accounted for as one of:

- **GATED** — `AppliesTo` really does exclude non-Aurora, because the capability is absent there (`PgWaitStatsCollector`)
- **PAIRED** — a vanilla path chosen by flavor, because the *source* differs and the capability does not (`PgStatementStatsCollector`, `PgStatementText`)
- **PROSE** — it only names the surface in an operator-facing message (`CollectorEngineCapability`)

Four files today, all three kinds represented. The allow-list is the deliverable rather than the scan: each entry is a claim that somebody looked and decided, and the failure message asks a newcomer for that decision rather than for a suppression. A second test fails on a **stale** entry so the list cannot outlive what it describes, and the two PAIRED entries are checked against the code — an entry claiming a vanilla alternative that does not exist would record a decision nobody implemented.

Proven red first: removing `PgStatementText.cs` from the list names exactly that file, the one that shipped the bug.

Issue #2711, opened 2026-08-31
Postgres fleet (pgmon) has no per-target operational alerting — only tool-health and 3 outage-predictor alerts fire


erikdarlingdata:
## Summary

Part of a broader alerting-integration effort scoped 2026-08-31 (headless notification-config exposure and Datadog-parity payload structure are two sibling issues being filed in parallel).

Tonight, SQL Server per-target alerts (Deadlocks Detected/Cleared, Blocking Detected/Cleared, High CPU/CPU Resolved, Poison Wait, Long-Running Query/Cleared, Failed Agent Job) fired repeatedly and correctly on the use1/use2 fleets. The Postgres fleet (pgmon) never produced a single one of these — every pgmon alert observed was tool-health (Collection Stopped, Collector Cost Regression, Store Job Over Cadence). This is not because Postgres alerting is unwired — it's a real but narrower and differently-shaped gap than "Postgres has nothing."

## Confirmed mechanism (traced, not guessed)

`EvaluateAlertsAsync` (`Darling/PerformanceMonitor.Darling.Service/DarlingWorker.cs:2724`, called unconditionally from `ProcessServerSweepAsync` at `:1718` for every connected server every 30s, **no engine check** at either call site) does two things:

1. `await engine.EvaluateServerAsync(snapshot, cancellationToken)` — the shared `AlertEngine`, which evaluates Deadlocks/Blocking/CPU/Poison-Wait/Long-Running-Query/Failed-Agent-Job for **every server regardless of engine**. Its read adapter (`DarlingAlertReadAdapter.cs`) queries exclusively SQL-Server-only-populated tables:
   - Deadlocks → `FROM deadlocks` (`:247`)
   - Blocking → `FROM dmv_blocking_snapshots` / `FROM blocked_process_reports` (`:106`, `:83`, `:201`)
   - Poison Wait → `FROM wait_stats` (`:296`)
   - Long-Running Query → `FROM query_snapshots` (`:356`)
   - Failed Agent Job → `FROM running_jobs` (`:710`)
   - High CPU → a separate read, `ReadLatestCpuAsync` (`DarlingWorker.cs:2865`), `FROM cpu_utilization_stats` selecting `sqlserver_cpu_utilization`/`other_process_cpu_utilization` — column names are SQL-Server-specific.

[...]

Issue #2716, opened 2026-08-31
Postgres alert watermarks (5 alerts) don't survive a service restart, unlike AlertEngine's


erikdarlingdata:
Follow-up from review on #2713 (Postgres Deadlocks/Blocking alerts, part of #2711).

## The finding

\`AlertEngine\`'s SQL Server watermarks (\`_lastAlertedDeadlockCount\`/\`_lastAlertedBlockingCount\`) are seeded from \`IAlertStateStore\` on startup via \`EnsureWatermarksSeededAsync\`, so a host restart doesn't re-fire an event still sitting in the rolling window.

The Postgres side has no equivalent for ANY of its five alerts:
- The three existing Tier 0 predictors (\`_lastPostgresAlert\` — wraparound/xmin/replication slot)
- The two added in #2713 (\`_lastPgDeadlockAlert\`/\`_activePgDeadlockAlert\`/\`_lastAlertedPgDeadlockCount\` and the Blocking equivalents)

All of these are purely in-process \`ConcurrentDictionary\`s with no persistence or seeding. A Darling service restart (routine deployment, crash-and-recover, etc.) resets every watermark to 0. If a deadlock/blocking situation — or a wraparound/xmin/slot condition — is still inside its relevant window at that point, the very next sweep sees the condition still present with a zeroed watermark and fires again for something already reported before the restart. This is the #1091/#2704/#2708 re-fire class, re-opened at the restart boundary instead of the cooldown boundary.

## Why this wasn't fixed in #2713

This isn't specific to the two alerts #2713 added — it's a property Postgres alerting has had since the three Tier 0 predictors shipped, and #2713 just matched that existing precedent rather than fixing it unilaterally for two of five alerts while leaving the other three inconsistent. Fixing the category (all five) properly means porting \`AlertEngine\`'s \`IAlertStateStore\`-seeding pattern to the Postgres side, which is a real, separate piece of work — not something to bolt onto a PR scoped to "add two alerts."

## Scope, if picked up

Whatever the fix, it should cover all five Postgres alerts uniformly (the three existing predictors AND the two from #2713), not just a subset — this is exactly the "fix the category, not the instance" situation. Options, not yet decided:
- Port \`AlertEngine\`'s \`EnsureWatermarksSeededAsync\`/\`IAlertStateStore\` seeding pattern to the Postgres alert path.
[...]

PR #2728, opened 2026-08-31
Add Postgres Poison Wait analogue alert (#2711)

erikdarlingdata:
The last open piece of #2711. This is a **concrete proposal, not a settled answer** — the issue's research comment deliberately left the definition open, so everything below (the candidate, the threshold shape, the numbers) is framed to be rejected or re-tuned cheaply. Every choice traces to that comment's fleet data.

## Which candidate, and why

**Candidate 2 from the research comment: the `IPC:BtreePage` + `IPC:BufferIo` pair.** The comment's own words call it "the stronger candidate for something SQL-Server's-Poison-Wait-shaped that doesn't already exist elsewhere in the alert catalog," and the data backs it:

- **Near-zero baseline** — exactly zero occurrences on 4 of 5 fleet servers over 24h, the defining trait of the SQL Server poison set (THREADPOOL / RESOURCE_SEMAPHORE / RESOURCE_SEMAPHORE_QUERY_COMPILE): any sustained accumulation is inherently abnormal.
- **Structural, not human** — sustained index-page or buffer-I/O pile-up between backends, not a lock a query is holding. No overlap with any shipped alert.
- **Candidate 1 (`Lock/Relation`) rejected** on the comment's own caveat: a relation-level lock wait IS blocking, and the #2713 Blocking alert already owns that ground — a second alert firing on the same incident is how alerts earn mute rules.

## Threshold shape and defaults

The research comment is explicit that SQL Server's avg-ms-per-wait bar does not port: on the one server where these events appeared, they averaged **1-2 ms per wait at six-figure volumes** — a shape a per-wait average can never see. So the trigger is the comment's suggested alternative: **total accumulated wait time crossing a bar**, normalized to the window so the number reads as "backends continuously stuck, on average."

- **Window**: 10 minutes (matching the SQL Server poison read's own recency window). Summed per-interval deltas from the already-collected `pg_wait_stats` table — no new collector, no new schema. Partial window coverage undercounts and therefore under-fires, the correct failure direction.
- **Warning**: 600s of accumulated wait per 10 minutes (avg **1.0 backend continuously stuck**). Calibrated against the comment's data: the worst fleet server averaged ~0.006 concurrent BtreePage waiters over 24h, so this sits **~160x above the worst healthy baseline observed** — nothing measured to date would have fired it. A test pins that: even packing that server's entire day of BtreePage wait into one hour stays silent.
- **Critical**: 6,000s per 10 minutes (avg **10 backends stuck**) — an active contention collapse.
- **Per event, not summed** — the two events are different incidents with different remedies (hot index insert point vs. backends stacked behind in-flight reads), and each finding carries its own remedy text.

**Settings**: gated on the existing shared `PoisonWaitEnabled` switch (the exact #2722 precedent — one preference, both engines). `PoisonWaitThresholdMs` is deliberately NOT reused (wrong semantics per the research); the two Postgres bars are documented constants for this first cut, the same reasoning as `PgDeadlockCountThreshold` — a knob (new `config_alert_settings` column + migration + Settings-window work) the moment anyone wants a different number.

## Mechanics

- **Path**: the engine-gated sibling path — `PostgresAlertEvaluator.EvaluatePoisonWaits` (pure), `DarlingPostgresAlertReadAdapter.GetPoisonWaitPressureAsync`, delivered from a failure-isolated `EvaluatePgPoisonWaitAsync` in `DarlingWorker`. No engine branching in the shared `AlertEngine`; Lite never sees any of it.
- **Metric name**: the exact SQL Server string `"Poison Wait"` + `"Poison Waits Cleared"` (Detected/Cleared pair) — #2713's parity reasoning for mute rules/history/dashboards. Mute dimension is `WaitType`, matching the SQL Server twin.
- **Per-subject cooldown** keyed `server|metric|subject` per the documented #1140 guidance — `IPC:BtreePage` firing must not consume `IPC:BufferIo`'s cooldown.
- **#2704 guard**: re-fires only on a store row newer than the one last fired on — the collector's delivered cadence and the alert cooldown are independent clocks.
- **#2716/#2723 restart survival**: cooldown seeded once per key from `GetLastAlertTimeAsync`'s per-fingerprint filter (the subject IS the dedup fingerprint), and the same seed floors the #2704 guard — no new persisted state, no schema.
- **Restart/reset safe by construction**: `CollectorDeltaCalculator` writes 0 for first sightings, counter resets and gap re-baselines, so the accumulation can undercount after a disruption but never spike.
- **Aurora-only source, honestly**: the cumulative wait counters are Aurora-only; on other targets the read returns no rows and no rows is silence. Extending the definition to self-hosted targets via `pg_wait_sampling` needs its own calibration (sampled counts, not accumulated time) and its own fleet evidence first — deliberately not faked here.

## Verification

- `dotnet build` clean on macOS (full solution, 0 errors, no new warnings).
- 12 new tests in `PostgresAlertEvaluatorTests.cs` (boundaries, the fleet-quiet pin, per-event independence, ordering, parity metric name, remedies, subject/guard lockstep). The Windows-only suites can't run locally; the same 21 assertions were executed on macOS against the real `PerformanceMonitor.Alerting` build via a throwaway net10.0 harness — all pass.
- The **shipped** `PoisonWaitSql` string (extracted from the source, not retyped) was executed against a live PostgreSQL 17 with the V63 rung's exact DDL and an adversarial fixture: sums and casts correct, and all five exclusions held (out-of-window, other server, `Lock/relation`, non-poison IPC event, NULL event name), with a NULL-delta first-sighting row neither breaking the sum nor hiding the freshness marker.

Closes #2711.


erikdarlingdata:
Took the casing nit — it's narrow, but it under-fires at exactly the moment a major upgrade makes contention most likely, which is the wrong failure to be quiet about. 1124f73f groups on lower() with a MAX() representative casing for the subject; verified live against PG 17 with a mixed-casing fixture: the shipped string now sums both casings into one 690,000 ms row (over the 600,000 ms bar), where the old grouping split it 340,000/350,000 — each silently under. The one-off dedup-fingerprint flip at upgrade time is the accepted cost, per the new doc comment.

Issue #2719, opened 2026-08-31
Postgres High CPU alert isn't a light port — there's no instance-level CPU collector to alert on


erikdarlingdata:
## Summary

Part of #2711 (Postgres alert-type parity). #2711's own table classified High CPU as "needs light design" — this issue corrects that: it's not a design question, it's a missing data source. There is currently **no collector anywhere in this codebase that captures instance-level CPU utilization for a Postgres/Aurora target.** Building a real "High CPU"/"CPU Resolved" pair requires new collector infrastructure, not just an alert-wiring change alongside `EvaluatePostgresAlertsAsync`.

## What SQL Server's High CPU actually reads

`AlertEngine.CheckCpuAsync` (`PerformanceMonitor.Alerting/AlertEngine.cs:334-388`) thresholds on `AlertServerSnapshot.SqlCpuPercent`/`TotalCpuPercent`, populated for SQL Server by `ReadLatestCpuAsync` (`Darling/PerformanceMonitor.Darling.Service/DarlingWorker.cs:3246`):

```sql
SELECT sqlserver_cpu_utilization, other_process_cpu_utilization
FROM cpu_utilization_stats
WHERE server_id = $1
ORDER BY sample_time DESC
LIMIT 1
```

This is an **instance-level percentage-of-capacity gauge**, sourced from a SQL Server collector that reads `sys.dm_os_ring_buffers`' RING_BUFFER_SCHEDULER_MONITOR entries (SQL Server tracks its own and the OS's CPU% internally). The fleet overview's `cpu_percent`/`total_cpu_percent` fields (`DarlingFleetReader.FleetCpuSql`) read the exact same table — confirming this is a SQL-Server-only signal end to end, not just in the alert path. **The fleet overview shows no CPU number for any pgmon server today either** — this isn't a gap unique to alerting, it's a gap in what's collected at all.

## What Postgres actually has

[...]

Issue #2710, opened 2026-08-31
Alert deliveries lack Datadog-parity structure: no structured context tags, no linked triage artifact


erikdarlingdata:
## Context

Part of a broader alerting-integration effort scoped 2026-08-31 (two sibling issues cover: exposing Slack/Teams/PagerDuty/webhook configuration on a headless Darling box, and auditing SQL-Server-vs-Postgres alert-type parity). This one is about the *shape* of what actually gets delivered.

A real Datadog alert Erik uses today looks like this:

```
Warn: High level of timeouts in greendiamond over the last 5 minutes (webapi) on env:greendiamond,resource_name:<the actual SQL query text, parameterized>
* Datadog notebook: <link to an auto-generated triage/investigation page>
* env: greendiamond
* resource_name: <query text>
@slack-eng-alrt-llx-db
```

The bar: (1) structured context tags (an `env`/tenant equivalent, a `resource_name`/object-identity equivalent), not just a text blob, and (2) a link to an auto-generated, pre-populated triage artifact bundling diagnostic context — not just the bare alert.

## Current state (confirmed by reading the shipped code)

**Delivery is already multi-channel and fully implemented** (`PerformanceMonitor.Notifications/WebhookAlertService.cs`, ~1350 lines): Slack (Block Kit, `BuildSlackPayload`), Teams (MessageCard, `BuildTeamsPayload`), PagerDuty (Events API v2, `BuildPagerDutyPayload` + `BuildPagerDutyCustomDetails`), and a fully generic, operator-templated webhook (`BuildGenericPayload`). None of the four builders emit anything resembling Datadog's structured `env:`/`resource_name:` tags today, and none link to any external artifact.

**The underlying data model is richer than the current payloads use.** `PerformanceMonitor.Notifications/AlertContext.cs` already carries, per alert firing:
- `AlertContext.Details: List<AlertDetailItem>` — heading + label/value `Fields`, or free-text `Body`, or an `IsCodeBlock` T-SQL block (webhooks currently render code blocks as a "see email/in-app dialog" hint, never inline — `TsqlWebhookHint`, `WebhookAlertService.cs` Teams section).
[...]

Issue #2677, opened 2026-08-28
[BUG] 3.5.0->3.6.0 Deadlocks


ZedZipDev:
### Component

Lite

### Performance Monitor Version

3.6.0

### SQL Server Version

Azure SQL DB

### Windows Version

W11

### Describe the Bug

After 3.5.0->3.6.0 Update I opened the Blocking->Deadlocks page and see only 2 deadlocks  records and in fact it is 1 deadlock
2026-08-21 Victim
2026-08-21

But there are more deadlocks , I retrieve them using query.
I have "Last 7 days" so it looks like PMLite did not detect any deadlocks after 21.08 and shows me only these 2 rows.
But it (3.5.0) worked 24hrs/7. 
By the way we discussed it in other issue. It seems the PMLite does not detect deadlocks anymore.

Let me know what additional details  can I provide.
Thanx.

### Steps to Reproduce

Run PMLIte, go to Blocking->Deadlocks

### Expected Behavior

Deadlocks detected

### Actual Behavior

No deadlocks

### Error Messages / Log Output

```text

```

### Screenshots

<img width="809" height="309" alt="Image" src="https://github.com/user-attachments/assets/045a66d7-7b92-448b-bfc3-e63091979866" />

<img width="820" height="329" alt="Image" src="https://github.com/user-attachments/assets/f9fbacd1-7aeb-48fe-a3de-51247ec466fa" />

<img width="802" height="317" alt="Image" src="https://github.com/user-attachments/assets/0f9397da-ab16-4724-b19d-dffd9c5956db" />

erikdarlingdata:
Found it, and thank you for the persistence — this is a real bug, and it is *not* a 3.5.0→3.6.0 regression (the deadlock code barely changed between them). The timing was a coincidence.

**What actually happened.** On Azure SQL DB our deadlock capture is a database-scoped Extended Events session. After an Azure **failover or scale operation**, that session can end up listed in `sys.database_event_sessions` but missing from `sys.dm_xe_database_sessions` — it exists, but the ring-buffer reader can't see it. That state is *transient* (it clears once the failover settles). Our recovery code recreated the session once and, if that didn't restore visibility on the spot, set an internal 'gave up' flag — and that flag **was never cleared until the app restarted**. So one transient Azure blip on ~Aug 21 stopped capture, and because PMLite runs 24/7 for you, it never got the restart that would have cleared it. Blocked-process capture rides the same code path, so it went quiet too.

[...]

PR #2731, opened 2026-08-31
Add auto force-plan bot phase 2: the write path (#2138)

erikdarlingdata:
> **Carved. #2745 is PR A and is open for review; this PR is now PR B — the write path, and nothing else.**
>
> [#2745](https://github.com/erikdarlingdata/PerformanceMonitor/pull/2745) — *Add force-plan bot phase 1: detection, evidence journal, dry-run* — took detection, the evidence journal, the V107 rung, the policy and self-review state machines, the config gates and the dry-run ledger. It ships `IPlanForceExecutor` as a declared seam with **no implementation**, and pins that three ways (byte-scan of the compiled service assembly for `sp_query_store_force_plan` / `sp_query_store_unforce_plan` / `FREEPROCCACHE`; no concrete implementer in any shipped assembly; no executor-typed member on `PlanForceBot`). So PR A reviews as a pure detection/audit feature that cannot reach a monitored server.
>
> **This branch is not rebased yet and its diff below still shows both halves.** Once #2745 lands I will rebase onto dev, at which point what remains here is the write path alone:
>
> - `SqlServerPlanForceExecutor` — `sp_query_store_force_plan` / `_unforce_plan` in the #1914-verified named two-argument form, the read-only verify, and the targeted `DBCC FREEPROCCACHE(plan_handle)` evict lever
> - the bot's force arm (intent row → executor → completion row linked by `related_action_id`)
> - the bot's review-unforce arm, including the write-gate RE-CHECK that journals a verdict as withheld when a gate closed after the force, and the review half deliberately outliving the `Enabled` switch so disarming cannot orphan an outstanding force
> - the consumer of `PgPlanForceActionStore.GetPendingReviewsAsync` (the read itself, its orphaned-intent grace window and its live-PostgreSQL spec already landed with PR A's journal)
> - `PlanForceExecutorTests` and the four write-path `PlanForceBotTests` scenarios
>
> **The three no-write pins in PR A go red when this lands, by design.** They assert "the write path is not in this build" — a claim that is supposed to stop being true here, in the diff whose whole subject is making it stop being true, so relaxing them is a reviewable act rather than a silent side effect. Do not loosen them for any other reason.
>
> Small enough after the rebase to review character by character, which is what the first write path to production servers deserves.

---

**Draft, per your call that this is too much for a single pass.** The honest headline first: **the live write path IS in this diff** — `sp_query_store_force_plan` / `_unforce_plan` / targeted `DBCC FREEPROCCACHE(plan_handle)` all exist in `PlanForceExecutor.cs`, complete and pinned, reachable only behind two gates that both ship closed.

## How to review this (the risk lives in 5 files, in this order)

1. **`PerformanceMonitor.Analysis/ForcePlanBotPolicy.cs`** (256 LOC) — the decision table. *(Landed in #2745.)* Invariants pinned by `ForcePlanBotPolicyTests`: shipped defaults are OFF + DRY RUN; a target with any `ForcePlanBlockers` output (PSP, secondary-replica evidence) can never evaluate `Force` or `WouldForce` whatever the gates say — the #2140 never-auto-force rule as a testable contract, same function agents read in `structured_remediation` (#2146); a live `Force` requires dry-run off AND per-server opt-in, each closed gate named in the journaled reasons; failure memory is a sliding window with no reset API (#2677).
2. **`Darling/.../PlanForceExecutor.cs`** (286 LOC) — **the whole subject of this PR after the rebase.** The ONLY statements that can ever reach a monitored server. `PlanForceExecutorTests` pins them character-exact: the #1914-verified named two-argument force form (no `@replica_group_id`, no `@disable_optimized_plan_forcing`), no `USE`/`CONVERT` anywhere, evict is the bounded handle form — the argumentless cache-wide `FREEPROCCACHE` is pinned ABSENT, and a garbage plan-hash refuses to parse rather than evicting by accident.
3. **`Darling/.../PlanForceBot.cs`** — the orchestration, where a write would escape. The evaluate-and-journal half landed in #2745; the **force and review-unforce arms** stay here. `PlanForceBotTests`' write-path scenarios pin: intent-row-then-completion-row journaling linked by `related_action_id`; and the write-gate RE-CHECK on unforce — a self-review verdict journals as withheld, executor untouched, when ANY gate (including the global Enabled) closed after the force. The review half deliberately outlives the Enabled switch: disarming the bot cannot orphan a force it placed while armed.
4. **`Darling/.../PgPlanForceActionStore.cs`** *(landed in #2745)* — the windowed history reads ARE the cooldowns. `PlanForceActionStoreTests` (live PG, CI-gated) pins: an intent+completion force pair spends ONE budget slot; failure memory heals purely by the window sliding; a pending review is closed only by a related review/unforce row; an ORPHANED INTENT (a force that landed but lost its completion journal write) surfaces for review after a 10-minute grace, so no live force can escape review through a store blip; the fleet-wide null-server read works.
5. **`PerformanceMonitor.Analysis/ForcePlanSelfReview.cs`** *(landed in #2745)* — the review verdict table; `ForcePlanSelfReviewTests` is its complete spec (not-net-benefit unforces at the FIRST due checkpoint; exactly-at-the-bar keeps; a quiet query keeps its force with the reason saying the keep rests on absence of evidence; "no longer forced" outranks "failing to force").

## Phased plan (what's here vs. what follows)

| Phase | Content | Where |
|---|---|---|
| 1a | Detection consumption, evidence journal, dry-run/shadow mode, cooldowns + budgets | **#2745** |
| 1b | The write path + self-review execution, armed only by both gates | **this PR** |
| 2 | Evict-first orchestration (evict → observe 25 execs/30 min → force only if the bad plan returns) — the executor lever + tests already exist | follow-up |
| 2 | MCP/web read surface `get_plan_force_actions` (web-parity pin touches `DarlingWebEndpoints.cs`, owned by the #2710 lane right now) | follow-up |
| 2 | Viewer/settings surface for the opt-in + control-plane knobs; shadow scoring of aged would_force rows (self-healed / still regressing / human forced); alert on live bot action | follow-up |
| 3 | Lite twin, if an interactive SKU should ever act autonomously at all | discussion |

## Numbers (pre-carve)

21 files, 2,866 insertions / 12 deletions. Production code ~1,600 LOC across 8 files; tests ~1,430 LOC across 6. Verified: full ladder V1→V107 + journal round-trip green on live PostgreSQL 17; all 50 policy/review/orchestrator scenarios executed green on macOS against the real assemblies (the Windows suites are CI's). Renumbered V106→V107 mid-flight after #2719 merged claiming V106; #2745 re-derived the same 107 as `max(dev) + 1` and keeps it. CI rounds 1-2 caught five registration ratchets (probe reader ordinal, round-trip catalog, role ACL + its provision-script mirror, the #1902 cleanup pattern) and the review bot five real findings across three rounds (stale V107 comment; evict no-op misclassified as failure; missing orchestrator spec; disarming the bot orphaned outstanding forces from self-review; a force that lost its completion journal write became un-reviewable) - all addressed in follow-up commits, no force-pushes.


erikdarlingdata:

**Closing: divergent second implementation of a phase that already landed, 672 commits behind `dev`.**

Phase 1 reached `dev` via `d8f22c40c` ("Wire the force-plan bot into the service, with no write path at all") plus `e4f40a0dd`. This branch does **not** contain `d8f22c40c`, so the two are independent implementations rather than a stack. It conflicts in `PlanForceBot.cs`, `PgPlanForceActionStore.cs`, `PlanForceBotTests.cs`, `DarlingConfig.cs`, `DarlingWorker.cs`, `StoreConfigProvider.cs` and `darling.sample.json` — its own subject matter. A merge is a rewrite, and it would be a rewrite against a phase-1 that is not the one this branch assumed.

**#2138 stays open, because the work this branch aimed at is genuinely outstanding and deliberately fenced.** `dev` carries `PlanForceNoWritePathTests`, which asserts the **absence** of `sp_query_store_force_plan`, `sp_query_store_unforce_plan` and `FREEPROCCACHE` — the bot detects and journals, and the write path is held closed on purpose. `IPlanForceExecutor.ForcePlanAsync` is the seam left for it.

So phase 2 should be rewritten against `dev`'s seam rather than resurrected from here. What is worth reading off this branch before it goes cold, as design input rather than code: `2323d28d3` (policy gate and self-review state machine), `5df09b862` (the bot's decision table, review spec, write-path SQL and journal pins), `9f5748d6e` ("a live force can never escape its review"), and `26a610c6e` (mirroring the `plan_force_bot_enabled` grant into `provision-roles.sql` — an easy thing to forget twice).

Closing by hand; closing keywords do not fire here.

PR #2745, opened 2026-09-01
Add force-plan bot phase 1: detection, evidence journal, dry-run (#2138)

erikdarlingdata:
**This PR cannot write to any monitored server.** It is the detection, evidence and dry-run half of #2138, carved out of #2731 per the split that PR's own body offered. #2731 stays open as the draft parent and now tracks **only the write path**.

## What "cannot write" means here, and how it is proven

`IPlanForceExecutor` ships as a declared seam with **no implementation**. `PlanForceBot` holds no executor, no connection factory, and no member that names the seam. Three tests in `PlanForceNoWritePathTests` assert it independently, because each one alone leaves a way through:

1. **The shipped artifact carries no write statement.** The compiled `PerformanceMonitor.Darling.Service` assembly is read as bytes and searched, at both UTF-16 alignments, for `sp_query_store_force_plan`, `sp_query_store_unforce_plan` and `FREEPROCCACHE`. Reading the DLL rather than the source means it cannot be satisfied by a copy that drifts, and it sees SQL built inline in a command constructor — which a scan of `const` fields would miss.
2. **The seam has no implementation.** Reflection over the three shipped assemblies finds no concrete `IPlanForceExecutor`. Test fakes don't count; they live in `Darling.Tests` and never ship.
3. **The orchestrator has no route to one.** `PlanForceBot`'s fields, properties, constructor parameters and method signatures are walked (through generic arguments, so `Func<string, IPlanForceExecutor>` is caught) and none mentions the seam — so there is no door to open from outside the assembly either.

**Proven red, not just green.** I dropped `SqlServerPlanForceExecutor` back into the service project and re-ran: 1 and 2 both fail (3 correctly stays green — the bot itself still has no executor member, which is exactly the separation the three prongs are for). Removed it, all green again. When #2731 lands the write path these go red on purpose: relaxing them becomes the reviewable act in the diff whose whole subject is making the claim stop being true, rather than a silent side effect.

Scope stated so it cannot be overread: the search is the **service** assembly, the process that holds connections to monitored servers. `PerformanceMonitor.Analysis.FactRemediation` deliberately *renders* those statements as text — the copy-and-run remediation script an operator reads, unchanged and long-standing. Rendering advice and executing it are different acts, and only the second is what "cannot write" is about.

## What's here

| Piece | File |
|---|---|
| The decision table — the shared gate, the three switches, the cooldowns | `PerformanceMonitor.Analysis/ForcePlanBotPolicy.cs` |
| The review verdict table (pure, complete, unwired) | `PerformanceMonitor.Analysis/ForcePlanSelfReview.cs` |
| V107 rung: the journal + the per-server opt-in column | `Darling/.../PgMigrations.cs` |
| The journal store and its windowed reads | `Darling/.../PgPlanForceActionStore.cs` |
| Evaluate-and-journal, hooked to the scheduled analysis pass | `Darling/.../PlanForceBot.cs` |
| The unimplemented seam | `Darling/.../IPlanForceExecutor.cs` |

**The shared gate.** The bot never re-derives whether a target is forceable — it passes `FactRemediation.ForcePlanBlockers`' output straight through, so the gate it enforces IS the one agents read in `structured_remediation` (#2146), and the #2140 never-auto-force-a-parameter-sensitive-target rule is one implementation rather than a promise kept in two places.

**The journal is the feature's state, not logging beside it.** `collect.plan_force_actions` is append-only (outcomes and reviews are their own rows via `related_action_id`, never UPDATEs — an audit the audited thing can rewrite is not one), and the cooldowns and budgets ARE windowed reads of it. Deliberately not enrolled in retention: an audit of what a bot decided about production outlives the metrics that motivated it.

**The three gates all ship closed** — `forcePlanBot.enabled` false, `forcePlanBot.dryRun` true, and `config_monitored_servers.plan_force_bot_enabled` FALSE on every row. That last one has no darling.json counterpart on purpose: the registry is authoritative after seeding, so a file knob would be a silent no-op on every seeded box (#2254), and a knob that silently does nothing is at its worst when what it claims to gate is a write to a production server. The store read fails **closed** on DBNull for the same reason.

**Open all three on this build and the decision journals as WITHHELD** — its own outcome, not `failed` (nothing failed, and a failed force cools the query down for a week through the failure-memory window) and not `attempting` (which would surface as an orphaned intent owed a review of a force that never happened). An operator who has opened every switch believes the bot is live; the trail has to say out loud that it is not, rather than quietly reading as a would-force.

**Shadow mode is a faithful rehearsal, not a lighter path.** Dry-run spends the same per-query cooldown (24h) and per-server daily budget (3) a live bot would, so the would-force ledger scored during dogfood is the decision set an armed bot would have executed. Failure memory is a sliding window with no reset API — eligibility returns when the window slides past the failures, never by clearing a flag (#2677).

The self-review verdict table ships complete and specced even though nothing here can place a force to review. Settling the rules a live force will be judged and taken back by (re-judge at +1h/+24h against the cpu/exec baseline it was sold on; take it back below 25% better, because a force that buys nothing still pins a plan against future data change and has to pay rent) is the part worth doing *before* anything can place one.

## Rung

**V107**, re-derived as `max(dev) + 1`: dev is at 106 after #2719, and both other open PRs (#2730, #2743) sit at 106. Viewer probe sentinel, newest-first arm and reader ordinal come with it. The literal pins are all symbolic against `StorageVersion.SchemaVersion` these days, and the `ViewerDataServiceTests` all-true pin builds its argument list by reflection, so there is no hand-maintained number to bump — I grepped for `\b106\b` across the tree to confirm: three source sites, all in this diff.

## Numbers and verification

21 files, 2,414 insertions / 12 deletions. ~1,050 LOC production across 6 new/edited files, ~850 LOC tests across 5, the rest rung/probe/role-grant/config/CHANGELOG.

Windows suites are CI's. Locally on macOS against the real assemblies: all three no-write pins pass (and fail with the write path restored), and all eight orchestrator scenarios execute green through a throwaway `net10.0` harness driving `PlanForceBot` over an in-memory journal — the enabled gate, the engine gate at the connection boundary (#2213), cross-finding dedup and the per-pass cap, would-force with both closed gates named, blocked-with-blockers, suppressed-in-cooldown, and the withheld force. Each of the four commits was checked out and built independently. The V107 rung and the journal's live-PostgreSQL round-trip were green on this exact rung in #2731's `Darling PostgreSQL tests`.

## What's left in #2731

`SqlServerPlanForceExecutor` (the #1914-verified named two-argument force/unforce forms plus the targeted `DBCC FREEPROCCACHE(plan_handle)` evict lever), the bot's force and review-unforce invocation arms, the pending-review read's consumer, and their tests — to be rebased onto dev once this lands.


erikdarlingdata:
On the `MaxTargetsPerPass` observation — good catch on the mechanism, and the answer is that counting evaluated targets is the load-bearing choice rather than an oversight, so I've written the reasoning down beside the constant (e4f40a0d) rather than changed the behaviour.

Each evaluated target costs a `GetQueryHistoryAsync` round trip *before* the verdict is known, so the cap bounds the **work** a pass can do, not just its output. Counting only journaled rows would let a pass whose targets all sit inside their cooldown spend an unbounded number of history reads for an empty journal — which is the one shape the budget exists to stop.

The cost you name is real: a suppressed target can occupy a slot a later actionable one wanted. It is bounded by `FactRemediation.ExtractPlanRegressionTargets` already capping each finding at 5 targets ordered worst-regression-first, so the actionable ones are at the front of the list by construction. That ordering assumption is exactly what was missing from the comment, so it is now written down with a pointer to revisit it if that per-finding cap ever rises.

Issue #2789, opened 2026-09-02
query_store: cannot attribute cost between the #pm_qs_slice build and the shipping SELECT


erikdarlingdata:
## The gap

`query_store` on use1 averages 33–42 s/day with a 361 s tail, and we cannot say which of its two statements the time is in. The payload is:

1. the `#pm_qs_slice` staging build — `SELECT ... INTO #pm_qs_slice FROM sys.query_store_runtime_stats` filtered by `runtime_stats_interval_id IN (...)`, grouped, `HAVING MAX(last_execution_time) > @cutoff_time`
2. the shipping `SELECT TOP (50000) WITH TIES` that joins `#pm_qs_slice` to five live catalog views — `sys.query_store_plan`, `sys.query_store_query`, `sys.query_store_query_text` (inner) and `sys.query_store_runtime_stats_interval`, `sys.query_store_replicas` (left)

`collection_log.sql_duration_ms` blends both. The existing `#2164` split does not separate them either, and the reason is specific: the shipping SELECT ends `ORDER BY qsrs.last_execution_time`, a **blocking sort**, so the server materializes every row before the client sees one. `open:` is therefore *both statements plus the sort*, and `drain:` is only row transfer. Two statements, one number.

## Why it matters now

Over a 3.2 h window on `prod-pos-use1-omega-01`, duration is **decoupled from rows returned**:

| rows returned | sql_duration_ms |
|---|---|
| 0 | 120,471 |
| 0 | 120,424 |
| 3 | 123,745 |
| 3 | 72,454 |
| 8,086 | 150,916 |

Median for the 0–10 row bucket was **48 ms**. A 3-row result cannot come from a loop join exploding over a wide outer set, which points at statement 1 dominating — but that is an inference, not a measurement, and the competing explanation (the join into the TVF-backed views) is not excluded by anything we currently record. The two have completely different fixes, so guessing is expensive.

**These runs are also invisible in the log.** The per-database line is gated on `batchCount > 0`, so every 0-row run above printed nothing at all.

## What this adds

Per-statement attribution, diagnostic only — no change to the query shape, no temp-table refactor, no hints:

- the payload reports its own staging cost as `RAISERROR ... WITH NOWAIT` raised **between** its two statements, carrying `slice_ms` and the `#pm_qs_slice` row count
- the runner listens on `SqlConnection.InfoMessage` across the open and records both
[...]

PR #2790, opened 2026-09-02
query_store: attribute cost between the slice build and the shipping SELECT

erikdarlingdata:
Fixes #2789

## What this answers

`query_store`'s payload is two statements and we cannot tell which one costs the time. `collection_log.sql_duration_ms` blends them, and the `#2164` `open:`/`drain:` split does not separate them either — the shipping SELECT ends `ORDER BY qsrs.last_execution_time`, a **blocking sort**, so the server materializes every row before the client sees one. `open:` is therefore both statements plus the sort.

The runs that need explaining are the ones that returned nothing:

| rows | sql_duration_ms |
|---|---|
| 0 | 120,471 |
| 0 | 120,424 |
| 3 | 123,745 |
| 3 | 72,454 |

Median for that row bucket was 48 ms. None of these printed a log line — the per-database line is gated on `batchCount > 0`.

## The change

Diagnostic only. **No change to the query shape, no temp-table refactor, no hints, both `OPTION(RECOMPILE)` hints intact, 55-ordinal reader contract untouched.**

- `QueryStoreCollector` raises `RAISERROR(N'PMQS_PHASE slice_ms=%d slice_rows=%d', 10, 1, ...) WITH NOWAIT` between its two statements
- `QueryStoreCollector.TryParseSliceTiming` parses it back — lives next to the T-SQL that formats it so producer and consumer cannot drift, and so the round trip is testable without a live server
- `DarlingCollectorRunner` subscribes to `SqlConnection.InfoMessage` across the open and records `PerItemSliceMs` / `PerItemSliceRows`
- the per-database line gains `open = slice:Nms + payload:Nms, slice_rows:N`
- expensive **0-row** reads (≥10 s) now log instead of staying silent

`slice_rows` is the number that decides it: a large slice with a slow payload is the loop-join-into-TVF story; a 3-row slice costing two minutes says the staging scan is the cost.

## Why an info message

- **No extra round trip** — the token rides the TDS stream already in flight.
- **Survives the on-prem nesting.** The body runs inside `[db].sys.sp_executesql`, where a `#temp` dies with the invocation, so the two statements *cannot* be split into two client commands without changing the query. The existing code comment already says so.
- **Fires on a 0-row read**, which extra columns cannot — and 0-row runs are the entire subject.
- Raised **before** the shipping SELECT, so `ExecuteReaderAsync` is guaranteed to have processed it. A trailing message would depend on when the reader happened to be drained or disposed.

Severity 10, so it raises no exception. The listener comes off as soon as the open returns, so the plan/text fetches cannot overwrite the item's numbers. The reader's lifetime is unchanged — it still stays open across those fetches.

## Verification

`Darling.Tests` is `net10.0-windows` and cannot run on macOS, so this was verified with a throwaway `net10.0` console harness against the **real built assembly** and a **live SQL Server** (container, `QSTEST` with a 40,882-plan Query Store), driving the shipped `BuildPerItemQuery` output — i.e. the real quote-doubled `sp_executesql` nesting, not a retyped copy.

```
PASS  parse: well-formed / zero values / unrelated ignored / null ignored / partial ignored
PASS  sql: marker present, WITH NOWAIT, both OPTION(RECOMPILE) survive (count=2)
PASS  sql: no braces, no double quotes (nesting contract)
PASS  sql: backfill body also raises it
PASS  live: message arrived by the time the reader opened :: slice_ms=213 slice_rows=40882 (msgs so far=1)
PASS  live: nested sp_executesql executed and shipped rows :: rows=40882
PASS  live: slice_rows matches shipped rows :: slice_rows=40882 shipped=40882
PASS  live: slice_ms populated, not silently zero-forever :: slice_ms=213
```

**Negative control** — removing only the `RAISERROR` turns 6 of 15 red, including every live check (`total_info_msgs=0`, `slice_ms=-1`). The structural invariants (`OPTION(RECOMPILE)` count, no braces) correctly stay green, so they are pinning the right thing.

15 xUnit tests added to `StatementSplitTimingTests`, covering the derivation (`payload = open − slice`, clamped, and *not* subtracted from drain — it is a subset of open, and double-counting it would silently deflate drain), the parser against unrelated and partial messages, and the shipped SQL through the public `BuildPerItemQuery` / `BuildBackfillPerItemQuery` surface.

## Not deployed

No deploy — there is an active measurement window on use1.


erikdarlingdata:
Superseded by #2791 — closing without merging.

This PR added per-statement timing to answer "which of the payload's two statements costs the time". That question has since been settled directly: Erik ran sp_QuickieStore against OMEGA ordered by average CPU and pulled the actual plans. The `TOP (50000)` payload query averages **~2 s** — it was never the cost. The real consumers are the **plan-fetch** statements at **55,000–61,000 ms CPU each**, with `QUERY_STORE_PLAN_IN_MEM` on the inner side of a Nested Loops join, re-executed once per candidate plan_id.

So the statement this PR was built to instrument is not the problem, and shipping timing-only instrumentation for a settled question is not worth the surface area. The fix is in #2791.

One finding here is worth keeping on the record: the per-database log line is gated on `batchCount > 0`, so expensive **0-row** reads print nothing at all. Every one of the 120 s / 0-row runs that motivated this investigation was invisible in the log. That is a real gap and is filed separately from the tuning work.

Issue #2801, opened 2026-09-02
A wall-clock-budget-abandoned cycle is recorded as SUCCESS in collection_log


erikdarlingdata:
## What happens

A collector cycle abandoned by the #2673 whole-server wall-clock budget is recorded in `collection_log` with **`status = 'SUCCESS'`** — having stored nothing and advanced no watermark.

From `prod-pos-use1-monitor-01`'s store, over its **entire 17-day retention window**:

```
status  |  n | zero_row
SUCCESS | 36 |       36
```

Every abandonment ever recorded is a SUCCESS with `rows_collected = 0`. Two examples on `the affected server`:

```
2026-09-02 20:30:29.603220  sql_duration_ms=156925  rows_collected=0  status=SUCCESS
    error_message: "wall-clock budget (120s) reached; cycle abandoned"
2026-09-02 22:41:12.627159  sql_duration_ms=122385  rows_collected=0  status=SUCCESS
    error_message: "wall-clock budget (120s) reached; cycle abandoned"
```

Normal `procedure_stats` runs on that server are ~7–8 s median with 100–150 rows, so these are outliers, not the usual shape.

## Why

The abandonment site returns normally rather than throwing:

```csharp
return new CollectorRunResult(0, sqlSlice.ElapsedMilliseconds, 0, $"wall-clock budget ({budgetSeconds}s) reached; cycle abandoned");
```

so it lands on the ordinary success path, whose status is a hardcoded literal in **both** hosts — `DarlingWorker` and Lite's `RunCollectorAsync`. It travels through the `Note` / `telemetry.Note` channel, which is documented as being for *"a successful-but-empty run worth explaining"* (#1837) and explicitly does not change status. #2673 reused that channel for a case that is not a successful empty run.

## Why it matters more than a wrong label

1. **It claims a collection that did not happen.** `DarlingSelfAlertEvaluator.ReadCollectionSignalsAsync` takes both `last_success` and `recent_success` from `status IN ('SUCCESS', 'SKIPPED')`. A collector abandoning every cycle therefore reads as **perpetually fresh**.
2. **It is invisible to the obvious check.** An hourly health check counting non-SUCCESS `collection_log` rows, or grepping the app log for `cycle abandoned`, reports **zero** while this is happening — which is what happened repeatedly on 2026-09-02.
3. **It pollutes the note channel.** `last_note` / `note_count` are gated on `status = 'SUCCESS'`, whose whole claim is that the run succeeded.

[...]

Issue #2673, opened 2026-08-28
Bound each collector's per-server wall-clock footprint (tail is up to 555s on one server)


erikdarlingdata:
## Problem

A collector's cumulative time on a single monitored server per cycle is **unbounded**. We have a 60s per-*command* timeout (`DarlingCollectorRunner.CommandTimeoutSeconds`) and per-cycle **byte** budgets, but nothing caps the total wall-clock a per-database collector spends on one target when it fans out across databases.

Fleet measurement (prod, ~2 days, from the service log's `collection body ... (sql:Xms)` timings):

| collector | max single-server sql time | avg |
|---|---|---|
| query_store | **555 s** | 14.5 s |
| procedure_stats | **176 s** | 6.0 s |
| query_stats | **168 s** | 1.9 s |

On the busier of the two boxes, the top 4 collectors are **95% of all prod query time**. A collector occupying a monitored server for 3–9 minutes is exactly the profile that makes us "stick out" in that server's own monitoring — the thing we must never do.

## Proposal

Apply a per-collector, per-server **wall-clock budget** to the heavy per-database collectors. The interface already declares `ICollectorDefinition.PerItemWallClockBudget` (host-enforced by design), but the heavy per-DB collectors don't set one. When the budget is hit, ship what was collected and resume next cycle — the "once per DB lifetime / eventual consistency" model the Query Store fetch already uses tolerates spreading. This bounds the tail so no single collector runs minutes on a target, independent of how large that target's catalogs are.

Related: the QS-fetch-specific bounding in the companion issue, and the collector-cost self-monitoring issue (so a regression is visible, not forensic).

_Filed via Claude Code from a fleet collector-performance analysis._

Issue #1837, opened 2026-07-30
Enumerated collectors that yield zero items log SUCCESS indistinguishable from healthy


erikdarlingdata:
Spun out of #1833's triage. The reason both empty-grid defects (#1833, and the Query Store issue filed alongside this one) went unreported for so long is that they are invisible in Collection Health: the zero-item path returns `SUCCESS` with `rows_collected = 0` and no message, so "collected nothing for weeks" reads as healthy.

The per-database (`RunsPerDatabase`) branch already has an "every item failed, so rethrow as an error" guard (`RemoteCollectorService.DefinitionRunner.cs:187-192`). The enumerated branch has no equivalent: an enumeration that returns zero items — because the login can't enter any database, because every probe failed into an empty CATCH, or because a filter excluded everything — early-returns 0 as SUCCESS (`DefinitionRunner.cs:218-222`).

Sketch: when an enumerating collector's enumeration yields zero items on a target where that is surprising (e.g., the server has user databases), log a distinguishable status or at least a message on the SUCCESS row, so Collection Health can band it. Needs a little design care to avoid crying wolf on legitimately-empty servers.

Applies to both Lite and Darling (shared runner shape in `DarlingCollectorRunner`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikdarlingdata:
Scope addition from #1836's implementation: surfacing the on-prem Query Store enumeration's per-database PROBE failures (the empty CATCH) turns out to need a driver contract change - the enumeration's result set IS the item list both runners consume as database names, so failures cannot ride it. The natural home is the _lastCollectionNote / EmptyEnumerationMessage mechanism this issue's minimal fix introduced: the SQL needs a way to hand the runner a note alongside the item list. Fold that into this issue's design rather than a separate one.

erikdarlingdata:
Shipped across #1843 (minimal note), #1854 (probe-failure contract + health visibility), #1864 (newest-first note ranking), and #1868 (the has-user-databases qualifier). Remaining adjacent items live in #1865 and #1867.

PR #2803, opened 2026-09-02
Give a wall-clock-budget-abandoned cycle its own collection_log status

erikdarlingdata:
Fixes #2801

A collector cycle abandoned by the #2673 whole-server wall-clock budget was recorded as `SUCCESS`. Every one of the **36 abandonments in the use1 store's entire 17-day retention** is `SUCCESS` with `rows_collected = 0`.

## Why it inherited a success status

The abandonment `return`s rather than throwing, so it lands on the ordinary success path — whose status is a hardcoded `"SUCCESS"` literal, once in `DarlingWorker` and once in Lite's `RunCollectorAsync`. It carries its message through the `Note` / `telemetry.Note` channel, documented since #1837 as being for *"a successful-but-empty run worth explaining"* and explicitly not status-changing. #2673 reused that channel for a case that is not a successful empty run.

**The wrong label was the smaller half.** `DarlingSelfAlertEvaluator.ReadCollectionSignalsAsync` takes both `last_success` and `recent_success` from `status IN ('SUCCESS', 'SKIPPED')` — so a collector abandoning every cycle read as **perpetually fresh**. And the message landed in `last_note` / `note_count`, gated on `status = 'SUCCESS'`.

## The status, and the alternatives rejected

`ABANDONED`, produced by one shared `EnumeratedCollectorDriver.ClassifyReturnedRun` so the hosts cannot drift — one hardcoded literal each is exactly how both inherited this.

- **`SUCCESS`** — the bug. Nothing stored, no watermark advanced.
- **`ERROR`** — would page on a guard doing precisely its job; feeds error counts, health bands and the collection-failure self-alerts.
- **`YIELDED`** — documented as the 1s `LOCK_TIMEOUT` guard and read as evidence of lock contention *on the target*. Reusing it sends an operator hunting contention that is not there.
- **`SKIPPED`** — a healthy no-op that counts as success. This is the opposite: work attempted and paid for that shipped nothing.

**Safe to add.** Every read buckets by explicit list — `IN ('ERROR','PERMISSIONS')`, `= 'YIELDED'`, `IN ('SUCCESS','SKIPPED')` — never by complement, so a new value joins no bucket rather than silently joining the wrong one. `collection_log.status` carries **no CHECK constraint**, so no migration rung. And the self-alert's consecutive-failure path is server-scoped across every collector, so one collector abandoning among ~40 healthy ones cannot empty its success window.

## What is deliberately NOT changed

- **The per-item #2150 abandonment.** It ships rows for the databases it got through, so that cycle really did collect — it correctly stays `SUCCESS` with its note. Only the whole-cycle #2673 case, which returns `0` rows before the storage phase, is reclassified. The "partial vs zero" distinction is a property of the code, not a judgement call: the #2673 site hardcodes `0` rows and its comment notes it skips the storage phase and the state-persistence block.
- **A dedicated `abandoned_count` column** in the four collection-health reads. That would touch ~13 sites including **four ordinal-based readers**, where inserting a column shifts every subsequent ordinal — a silent-corruption risk not worth taking for a display nicety, especially with a fleet roll in flight. Filed as a follow-up rather than bundled. The status change alone already restores the visibility the issue is about: any `status <> 'SUCCESS'` query now sees these.

## Verification

`Lite.Tests` and `Darling.Tests` are `net10.0-windows` and cannot run on macOS, so the pins were also run through a throwaway `net10.0` harness against the **real built `PerformanceMonitor.Collectors` assembly**. CI is the arbiter for the xUnit run.

**Proven red first.** Reverting `ClassifyReturnedRun` to the pre-fix hardcoded literal:

```
FAIL an abandoned cycle does not classify as SUCCESS
FAIL an abandoned cycle is outside the freshness-success family
PASS an abandoned cycle is not ERROR or PERMISSIONS
PASS an abandoned cycle does not reuse YIELDED
PASS an ordinary returned run still classifies SUCCESS
PASS an ordinary returned run is inside the freshness-success family
PASS the shared note names the budget
2 PIN FAILURE(S)
```

and with the fix restored, `ALL PINS PASS`.

The first attempt at that proof was **contaminated** — the saved copy of the file was written into the harness project directory and compiled into it, shadowing the real assembly (`CS0436`), so the pins passed for the wrong reason. Re-run with the backup outside the project, which produced the red above.

The pins assert over the freshness-success **family** rather than against the `"SUCCESS"` literal, because the defect was a status silently joining a set — a pin naming one member would miss the next status added to it.

## The underlying slowness is not new

Worth stating, because the fix makes these newly visible. `procedure_stats` on the affected server, full retention window:

| day | runs | p50 | p95 | max | over 60s |
|---|---|---|---|---|---|
| 08-17 | 32 | 8.3 s | 178.3 s | **246.4 s** | 6 |
| 08-19 | 129 | 7.4 s | 90.6 s | **278.1 s** | 8 |
| 08-27 | 106 | 7.6 s | 18.0 s | 176.6 s | 2 |
| 09-02 | 126 | 8.4 s | 120.3 s | 156.9 s | 11 |

Slow runs happened every day, and the worst maxima predate #2673 (2026-08-28). The budget did not create the slowness, it **capped** it — which is why `max` falls from ~250–278 s to ~121–157 s afterwards. Abandonment counts rise 4→15/day, but the first abandonment is 08-29, one day after the instrument was installed, so that rise is **not** evidence of a regression. Fleet control (the affected server excluded) is 0–13 over-60 s runs/day across ~5,000 daily runs with no trend, and 1 on 09-02.

No separate performance issue filed: claiming a trend off an instrument installed mid-window is the false-trend mistake.


erikdarlingdata:
All four review findings addressed in `3ccd56ae`.

**The `RecordCollectorResult` one was a real defect and it falsified this PR's own safety argument.** I checked that every SQL read buckets status by explicit list — `IN ('ERROR','PERMISSIONS')`, `= 'YIELDED'`, `IN ('SUCCESS','SKIPPED')` — and concluded a new value could not silently join the wrong bucket. I never checked the in-memory health path, which is an `if/else-if/**else**` chain, so `ABANDONED` landed in the error arm, set `LastErrorMessage` and incremented `ConsecutiveErrors` — showing the collector FAILING, precisely the false-error outcome this status was chosen to avoid.

It now has its own arm with the `YIELDED` semantics, which apply verbatim: the guard worked as designed so the streak must not grow, and nothing was collected so it is not proof the collector works and must not reset it either. I checked whether Darling has the same shape rather than assuming — it has no `ConsecutiveErrors` equivalent, so this was Lite-only.

**CA1863** — parsed once now. Exposed as a method rather than the bare format string: unlike the four sibling constants, whose callers are all in-file, both *hosts* render this one, so publishing the string would leave each host to parse it and pick its own culture. That also removed the duplicated `string.Format` and the `using System.Globalization` each host had grown.

**Darling.Tests coverage** — added `Runner_Marks_The_BudgetExpiry_Return_As_Abandoned`. Your framing was the useful part: the worker classifies from the *flag*, not the note text, and those are separately editable, so a note reworded while the flag went unset would silently restore `SUCCESS` with nothing failing. Source-text pinned like the rest of that class, because the abandonment sits in a catch filter on a live provider cancellation that no unit test there can reach.

**CHANGELOG** — `[#2801]` link-reference definition added.

Separately, `FleetIdentifierScrubTests` caught a real fleet tenant name in the CHANGELOG and a doc comment. Fixed in `7e30328e`, and I also scrubbed it from the issue and PR bodies, since this repository is public.

PR #2853, opened 2026-09-03
Add project skills for build, duckdb, maintenance, and release

erikdarlingdata:
## What

Moves four skills out of the personal `~/.claude/skills` directory into `.claude/skills` so they travel with the repo. Personal skills are machine-local, and cloud sessions do not read them at all, so a skill living only in `~/.claude/skills` is unavailable to every session outside the one machine it was written on.

`.gitignore` goes from `.claude/` to `.claude/*` plus `!.claude/skills/`, which exposes skills and nothing else. `settings.local.json`, `commands/`, and `worktrees/` stay ignored.

| Skill | Purpose |
|---|---|
| `release-checklist` | Pre-release checklist |
| `build` | Build a component, report where the exe landed |
| `duckdb` | Read-only query against Lite's DuckDB store |
| `maintenance` | Quarterly dependency, build, and hygiene pass |

## Credentials removed

`release-checklist` contained the lab `sa` password and a reused Azure/AWS admin password in plaintext. This repo is public, so both are gone:

- Cloud instance creation now takes `"$SQL_TEST_PASSWORD"`, a throwaway generated per run.
- The local SQL instance inventory, and the rule that sql2022 and sql2025 hold real monitoring data, moved to the untracked local `CLAUDE.md`. Those VMs exist on a single machine, so the committed procedure stays machine-neutral.

## Portability

- `%LOCALAPPDATA%` replaces absolute `C:\Users\...` paths. `duckdb` resolves the store through `os.environ['LOCALAPPDATA']`.
- `build` uses repo-relative paths instead of `C:\GitHub\PerformanceMonitor\`.
- Region guidance reads "close to the machine running the collector" instead of being hardcoded to US East for one developer's timezone.
- `maintenance/skill.md` renamed to `SKILL.md`. It resolved on Windows but would not have on the case-sensitive filesystem a cloud session runs on.

## Drive-by fix

`build` was stale. It pointed at `Dashboard/Dashboard.csproj`, which has been `bin`/`obj` leftovers since v3.3.0 moved Dashboard and Installer to `deprecated/`, and it did not know Darling exists. The argument table is rebuilt from `git ls-files "*.csproj"` with correct paths and target frameworks, now covering the Darling service and Viewer, with the deprecated projects labeled.

## Test plan

- [x] `git check-ignore` confirms `.claude/skills` is tracked while `settings.local.json`, `commands/`, and `worktrees/` stay ignored
- [x] Credential scan clean across every committed skill
- [x] No absolute paths or single-machine assumptions remain
- [x] Frontmatter `name` parses on every `SKILL.md`
- [ ] Clone on a second machine and confirm the four skills list in `/skills`

Docs and config only. No product code changes.

https://claude.ai/code/session_01EbYty5EYQwd4YGTDcJfEkE

### DIFF
--- a/.claude/skills/build/SKILL.md
+++ b/.claude/skills/build/SKILL.md
@@ -1,0 +1,40 @@
+---
+name: build
+description: Build and locate the PerformanceMonitor executables
+argument-hint: [lite|darling|viewer|dashboard|installer]
+disable-model-invocation: false
+---
+
+# Build PerformanceMonitor
+
+Build a component of PerformanceMonitor and report the result. All paths are relative to the repository root.
+
+## Arguments
+
+`$ARGUMENTS` specifies which component to build:
+
+| Argument | Project | Output TFM |
+|---|---|---|
+| `lite` | `Lite/PerformanceMonitorLite.csproj` | `net10.0-windows` |
+| `darling` or `service` | `Darling/PerformanceMonitor.Darling.Service/PerformanceMonitor.Darling.Service.csproj` | `net10.0` |
+| `viewer` | `Darling/PerformanceMonitor.Darling.Viewer/PerformanceMonitor.Darling.Viewer.csproj` | `net10.0-windows` |
+| `dashboard` or `full` | `deprecated/Dashboard/Dashboard.csproj` | DEPRECATED since v3.3.0 |
+| `installer` | `deprecated/Installer/PerformanceMonitorInstaller.csproj` | DEPRECATED since v3.3.0 |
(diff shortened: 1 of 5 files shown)


erikdarlingdata:
Good catch on the AWS half. Fixed in the follow-up commit: the RDS block now reads "close to the machine running the collector" exactly like the Azure block above it, keeping `us-east-1` as the concrete US East example. Also flagged that the AWS block uses the same `SQL_TEST_PASSWORD` throwaway, since the two blocks can be run independently.

Issue #2862, opened 2026-09-03
procedure_stats spends 98.1M ms/day rendering plan XML on the monitored servers; capture it on a cadence


erikdarlingdata:
## Summary

`procedure_stats` is the most expensive collector on the production use1 store, and essentially all of its cost is rendering and shipping execution-plan XML that the store already holds. The render half of that cost is CPU burned on the **monitored production SQL Server**, not on the collector box.

This proposes capturing plan XML on one cycle in N instead of on every cycle, behind a store-backed knob.

## The cost

| | |
|---|---|
| daily total (use1) | **98.1M ms/day** over 17,869 runs |
| avg / max | **5,490 ms** / 137,455 ms |
| vs `query_store` | **1.57x** — it recently overtook it |
| phase split | `sql:4,902ms = open:149ms + drain:4,724ms + other:1ms` |

So 96% is the read loop.

## Where the drain goes

A controlled decomposition — 6 heaviest servers, 3 reps x 3 variants, serial **and** 6-way concurrent, 108 executions, 150 rows every time:

| variant | what it does | share of drain |
|---|---|---|
| A: no `OUTER APPLY` | no plans at all | **0.2%** |
| B: `OUTER APPLY`, select `DATALENGTH(plan)` | renders, ships a bigint | **73.8%** (render) |
| C: current — select the plan | renders and ships | +**26.0%** (transfer) |

Stable at 73.6/25.5 under concurrency, so not a client-contention artifact. Variant A costing 0.2% means the non-plan columns are free and essentially the entire drain is plan cost.

The render was genuinely measured, not optimised away: B's `SUM(DATALENGTH(tqp.query_plan))` matched C's independently counted transferred bytes to two decimals on every server (46.35/46.35, 57.73/57.73, ...), so the TVF materialised every plan while shipping only bigints.

Plan sizes (n=831): mean 373 KB, p50 103 KB, p95 1,206 KB, max **8.5 MB**. The module DMVs expose no statement offsets, so the collector passes literal `0, -1` and gets the entire module's plan.

## Why cadence and not a dedup key

**`sys.dm_exec_text_query_plan` is a server-side TVF**, so the 73.8% render is target-side CPU — `get_collector_cost.sql_ms` is target-side duration. Hashing at the source (`HASHBYTES` over the plan, fetch only unknown hashes) would remove the 26% transfer and **leave the 74% render burning on customer hardware**. Skipping the cycle removes both.

[...]

PR #2863, opened 2026-09-03
Capture procedure_stats plan XML on a cadence (Fixes #2862)

erikdarlingdata:
Fixes #2862

## What changed

`procedure_stats` captures execution-plan XML on one cycle in N instead of on every cycle, behind a new `config_service` knob.

`CollectorContext.CapturePlanXml` already erases the plan fragments to a **byte-identical no-plan query** when false — that is the path Lite runs in production — so this rides an existing, tested seam rather than adding one. When the cadence skips a cycle, the `OUTER APPLY sys.dm_exec_text_query_plan(...)` is not in the query at all.

## Why cadence, and not a dedup key

`sys.dm_exec_text_query_plan` is a **server-side TVF**, so the render is CPU on the monitored production server (`get_collector_cost.sql_ms` is target-side duration). Hashing at the source would remove the 26% transfer and leave the **74% render burning on customer hardware**. The full measurement, the disproof of the `plan_handle` probe, and the ABANDONED-cycle evidence are in #2862.

## Design decisions worth reviewing

**The phase is derived from `server_id`, and it is load-bearing.** A bare `ordinal % interval` would put all 42 servers on the *same* capture cycle — three cheap cycles, then the whole fleet paying full plan-render cost at once. That is a 4x spike, and peak is exactly what produces the 120s `PerItemWallClockBudget` abandonments this change exists to reduce, so the naive form would have been worse than collecting every time. Pinned with its counterfactual proven red: `[10, 10, 11, 11]` shipped vs `[42, 0, 0, 0]` without the offset.

**Counters are in-memory and reset on restart, deliberately.** The stagger comes from the server id, not accumulated drift, so 42 servers restarting together still land on different cycles. The only cost is that each server's first post-restart capture lands within one interval rather than continuing its old phase.

**`interval = 1` is byte-identical to the old collector**, so the knob is fully reversible — that is why the clamp floor is 1 and not 2.

**Only `procedure_stats` is gated.** `query_stats`, `query_store`, `deadlocks` and `blocked_process_report` all capture plans too and are deliberately left alone; the gated collector is a named constant so it cannot drift.

**No new-module special case.** Capturing unconditionally for a module not seen before would need a per-ROW decision, but `CapturePlanXml` splices the `OUTER APPLY` for the whole query — knowing which modules are new before running it would require the store probe this change exists to avoid. The bounded cost is that a genuinely new module waits at most `interval - 1` cycles (~5 min at N=4).

**No viewer Settings field.** Matches `plan_content_retention_days` (V75) and `compose_statement_timeout_seconds`, which are also store-and-service only. The V108 probe sentinel and map arm are still added, because the top rung always needs one or a fully-migrated store maps low and the connect-time gate refuses a store that is current.

## Two premises I checked rather than assumed

- **Readers handle a NULL plan.** They do, and better than expected — all three read paths (`ViewerDataService.ProcedureStatsPlanXmlSql`, `DarlingStoredPlanReader.ProcedurePlanXmlBySqlHandleSql`, and the MCP `analyze_procedure_plan` behind it) carry `AND (COALESCE(ps.query_plan_xml, qpd.query_plan_xml) IS NOT NULL OR qpd.query_plan_gz IS NOT NULL) ORDER BY ps.collection_time DESC LIMIT 1`. They skip NULL-plan rows and return the latest row that *has* a plan, so a skipped cycle is invisible rather than showing a missing plan. That guard exists for #1767's transition and happens to make this transparent.
- **A skipped cycle stores NULL, not an empty string or a stale carry-forward.** `PgCollectorRowWriter.Value` short-circuits before `PayloadDimensions.Digest` (which throws on null): *"A NULL payload stays NULL — there is nothing to dedupe and no dim row to point at."*

## One correction to the issue's framing

The knob is **not** documented in `darling.sample.json`, because none of its siblings are — `queryStoreTextBudgetMb`, `planContentRetentionDays` and `composeStatementTimeoutSeconds` all appear in `DarlingConfig.cs` only. It follows `planContentRetentionDays` exactly: a `[JsonPropertyName]` seed plus the store column, where the file value seeds the row on first insert and the store is authoritative afterwards.

## Verification

`Darling.Tests` cannot execute on macOS, so the pure policy was run against the **shipped build** via a throwaway `net10.0` harness (assembly-named `Darling.Tests` to reach `InternalsVisibleTo`), not a retyped copy:

```
PASS  interval<=1 captures every cycle
PASS  exactly one cycle in N            (intervals 2, 3, 4, 60)
      shipped per-cycle: [10, 10, 11, 11]
      bare modulo     : [42, 0, 0, 0]   <- the counterfactual
PASS  shipped: fleet staggered, no cycle over fair share
PASS  counterfactual (bare modulo) FAILS that same pin -> the pin is load-bearing
PASS  negative serverId stays 1-in-4 and does not throw
PASS  clamp [1,60]
PASS  gated collector is procedure_stats
```

Service, Viewer and `Darling.Tests` all build clean. CI is the arbiter for the Windows suites and the live-Postgres rung tests.

## Predicted, to be checked after deploy

| | now | predicted at N=4 |
|---|---|---|
| avg | 5,490 ms | **~1,492 ms** |
| daily total | 98.1M ms | **~26.7M ms** |
| plan-render CPU on each monitored server | — | **-75%** |
| plan-history granularity | ~3.4 min | ~13.6 min |

Arithmetic agrees with the issue's independent estimate. Worth re-measuring against `collection_log` after the roll rather than trusting it.


erikdarlingdata:
Two fixes pushed (force-push, so the inline anchor above is orphaned — it was addressed, not dismissed).

**1. The review comment was right.** `ThePhaseIsDeterministic_SoARestartCannotBunchTheFleet` called the pure static twice with identical arguments and asserted equality — guaranteed by referential transparency, so it would have stayed green against any regression it claimed to guard. Replaced with `AfterARestart_TheFleetIsStillSpread_NotAllOnTheFirstCycle`, which asserts the observable property instead: how much of the fleet captures on cycle ordinal 0, the cycle the whole fleet re-enters on simultaneously after a restart resets the in-memory counters. That one is load-bearing — shipped gives 10 of 42, a bare `ordinal % interval` gives all 42 and fails the range.

**2. `FleetIdentifierScrubTests` caught a real tenant name I had put in the CHANGELOG** (`prod-pos-use1-<tenant>-01`). Renamed to the synthetic `alpha` slug. The same name was in the issue body and has been scrubbed there too — that guard only sees tracked source, so the issue and PR text needed doing by hand.

The scrub failure was the only red in both `build` and `Darling PostgreSQL tests`; everything else was green.

erikdarlingdata:
Closing unmerged — parked deliberately, not abandoned. See #2862 for the measurements.

**The justification was aimed at the wrong statistic.** This change reduces the MEDIAN (`procedure_stats` avg 5,490 -> ~1,490 ms), but the cost that actually loses data is the 120 s `PerItemWallClockBudget` abandonment, and that turns out to be a **cliff, not the tail of a gradient**: on one server over 1h44m the collector ran p50 6,056 / p90 8,009 / p95 9,551 ms and then jumped straight to 120,122 ms with nothing in between. Cadence reduction moves the 6 s median and does nothing to the 120 s wall.

It is also not collector-specific — `query_stats`, `procedure_stats` and `query_store` each hit >=120 s inside the same few minutes, back to back, while the target sat at ~31% CPU with zero blocking recorded. Three independent collectors stalling together on an idle target is not a per-collector payload problem, which is the premise this PR was built on.

The branch `feature/procstats-plan-cadence` (c6e894c8) is left in place. It was green on all six checks, and two properties are worth keeping written down if this is revisited: `N=1` is byte-identical to the current collector, so the knob is fully reversible; and the fleet stagger must be derived from `server_id`, because a bare `ordinal % interval` puts all 42 servers on the same capture cycle and converts steady load into a 4x fleet-wide spike — worse than collecting every time, since peak is what produces the abandonments in the first place.

Issue #2813, opened 2026-09-03
Retention held by the coverage gate is invisible to store metrics and has no automatic release (use1: 5 query_store_stats policies held 16 days, raw tier at 65 GB / +13 GB per day)


erikdarlingdata:
## Summary

On `prod-pos-use1-monitor-01`, five `query_store_stats*` retention policies have been **held paused by the #1680/#1877 coverage gate since 2026-08-17** — 16 days. `collect.query_store_stats` is configured `drop_after: 4 days` and actually holds **18 days 22:56 — 19 chunks, ~65 GB**, growing **+13 GB/day**.

**The gate itself is working exactly as designed.** This issue is not a request to weaken it. It is about two things the design leaves open:

1. A held policy is **invisible to every metric we collect**, so it went unnoticed for 16 days.
2. The release requires a **manual CLI action** that nothing schedules, prompts for, or alerts on — so "held" is a terminal state in practice, not a transient one.

## The mechanism, confirmed

`TimescaleSupport.EnsureRetentionPoliciesAsync` runs on **every service start** (`DarlingWorker.cs:1064`). For each policy it calls `MeasureRetentionCoverageAsync`; on a measured `Short` verdict it executes `HoldRetentionPolicySql` → `alter_job(j.job_id, scheduled => false)`.

Confirmed from the box's own app log — it fires on every restart, including all three tonight (19:08:03, 22:00:57, 22:56:03):

```
[WARN] Retention policy for query_store_stats HELD PAUSED - query_store_stats_hourly does not yet cover
       everything it holds, so arming could drop history that rollup has never materialized.
       Backfill past the 4 days horizon and the policy arms itself on the next start.
[WARN] ... query_store_stats_hourly            HELD PAUSED - query_store_stats_daily short
[WARN] ... query_store_stats_interval_hourly   HELD PAUSED - query_store_stats_corrected_daily short
[WARN] ... query_store_stats_corrected_hourly  HELD PAUSED - query_store_stats_corrected_daily short
[WARN] ... query_store_stats_interval_daily    HELD PAUSED - query_store_stats_daygrain_daily short
[INFO] TimescaleDB: 17/17 retention policies in place, 12 armed, 5 held paused pending backfill,
       0 left as-is (coverage unreadable), 0 moved onto a new horizon
```

[...]

Issue #2809, opened 2026-09-03
Five query_store retention jobs are Paused, so query_store_stats holds 19 days at a 4-day policy


erikdarlingdata:
## What

On `prod-pos-use1-monitor-01`, the five `query_store*` **retention** jobs are `scheduled = false` / `job_status = Paused`. Each ran **once**, on 2026-08-17 (one on 08-18), and has not run since. Every other retention job on the store ran today.

```
job_id | hypertable_name                    | scheduled | job_status | total_runs | last_successful_finish
  1072 | query_store_stats                  | false     | Paused     |          1 | 2026-08-17 16:46:22
  1078 | query_store_stats_corrected_hourly | false     | Paused     |          1 | 2026-08-17 16:46:22
  1075 | query_store_stats_hourly           | false     | Paused     |          1 | 2026-08-17 16:46:22
  1077 | query_store_stats_interval_hourly  | false     | Paused     |          1 | 2026-08-17 16:46:22
  1079 | query_store_stats_interval_daily   | false     | Paused     |          2 | 2026-08-18 16:46:23
```

Contrast, same store, same query:

```
query_stats           | 4 days  | Scheduled | last_successful_finish 2026-09-02 16:46:30
procedure_stats       | 4 days  | Scheduled | last_successful_finish 2026-09-02 16:46:30
blocked_process_...   | 35 days | Scheduled | last_successful_finish 2026-09-02 16:46:27
```

## Consequence

`query_store_stats` is configured `drop_after: 4 days` and actually holds **18 days 22:56**:

```
oldest 2026-08-15 00:59:03   newest 2026-09-02 23:55:26
19 chunks (24h interval), hypertable_size = 65 GB
```

The oldest chunk begins 2026-08-15, two days before the jobs stopped — consistent with retention having dropped normally until then and nothing since.

`query_store_stats` is the store's largest hypertable at 63–65 GB, and the store has been growing **+10 to +12 GB/day**, reaching 338 GB. A 4-day retention would hold roughly a fifth of that.

**Compression is unaffected** — the paused jobs are retention only; `is_compressed = true` on every old chunk, and the hourly compression job (1029) has 381 successful runs. So the data compresses and is then kept forever.

## Why it went unnoticed

`total_failures = 0` and `last_run_status = Success` on all five. Nothing failed — the jobs were *paused*, so `get_store_metrics` job health reports them clean. A paused job and a healthy job are indistinguishable in every reading we take.

[...]

Issue #1680, opened 2026-07-26
Retention policies run their first check immediately — no external window to pause them, caused permanent data loss


erikdarlingdata:
Field finding from prod-sql-use2-monitor-01 during the tiered-retention deploy on 2026-07-24. **This one caused real, permanent data loss and should be fixed before 3.3.0**, because 3.3.0 is exactly when existing stores get these policies for the first time.

## What happens

TimescaleDB runs a newly-created retention policy's first check **immediately at job creation**, not on its next scheduled interval. Confirmed on the field box via `job_stats` timestamps: all 6 new retention jobs fired and completed within the same minute as service startup (00:50:05-00:51:02 UTC) — before any external SQL session could pause them per the runbook's intended pause-before-backfill ordering.

Result on that store: **~2 days of query_stats/procedure_stats history and all of query_store_stats history were dropped** before the pause could take effect. Not recoverable.

There is no external window to win this race. Being faster is not the fix.

## Verified in code

`TimescaleSupport.AddRetentionPolicySql` (`:663-664`):

```
SELECT add_retention_policy('collect.{relation}', drop_after => INTERVAL '{dropAfter}', if_not_exists => true)
```

No `scheduled` argument anywhere in the file — so every policy is created live and armed.

## Fix

Create the job already paused in the same call that creates it, then enable it only after a verified backfill:

- `add_retention_policy(..., scheduled => false)`
- then `alter_job(<id>, scheduled => true)` once the CAGGs are confirmed backfilled past their horizons

That closes the race by construction rather than by timing.

## Why this matters more now than it did last week

The cold-start caveat already documented on `EnsureRetentionPoliciesAsync` — that an existing store must have its CAGGs backfilled past the horizon *before* retention runs — is currently only enforceable by an operator racing the scheduler and losing. Any 3.3.0 upgrade of a store with history to preserve hits the same window.

Note also that #1664 added a seventh policy (`query_stats_db_hourly`), so the fix should cover the whole list rather than the original six.

erikdarlingdata:
Fixed by #1711 (merged to dev), which supersedes #1684's approach.

[...]

PR #2878, opened 2026-09-04
Make retention held by the coverage gate visible (Fixes #2813)

erikdarlingdata:
Fixes #2813.

## The defect

The #1680/#1877 rollup-coverage gate **PAUSES** a retention policy so it cannot drop history a consumer has never materialized. That is correct, and this change does not weaken it.

The defect is that a paused job is indistinguishable from a healthy one in every surface we monitor. It is not failing, it is stopped — so it reports `total_failures = 0`, `last_run_status = Success` and a plausible last-run duration, and `StoreSelfMetrics.BackgroundJobInsertSql` never recorded `j.scheduled`.

On the production store five `query_store_stats` policies sat held for **16 days** while the raw tier grew to 19 chunks / 65 GB — 4.5x its own 4-day horizon, and the store's largest object. Every stored metric read clean. The only signal was one WARNING per service start (#2809).

This is the same shape as #2801 (an abandoned cycle recorded as `SUCCESS`): the honest state exists, and the surface we watch reports the reassuring one.

## What this does

- `get_store_metrics` gains a `retention` block, read **live from the catalog** rather than from the series, listing every held policy with its actual data span and how many times its configured `drop_after` it is really holding. Present on every response, including when nothing is held — an absent block and "nothing is held" must not look alike, which is the whole failure being reported on.
- A `Retention Held` self-alert rides the **existing hourly sweep**, on the same connection and the same Timescale gate as its #2136 sibling. No new timer, no new connection.

## Judged on the conjunction, never the flag

**Held alone is normal.** `EnsureRetentionPoliciesAsync` deliberately creates *every* policy paused — TimescaleDB runs a new policy's first check immediately at creation, so there is no window in which an external session could pause it first. Alerting on the flag would fire on every fresh store at every start.

**Over-horizon alone is normal too.** Retention drops whole **CHUNKS**, so a 4-day policy with 1-day chunks legitimately holds ~5 days (1.25x) while purging perfectly.

It is the conjunction that is unambiguous: the gate is holding this policy **and** the tier has already grown well past what it was meant to keep.

## Why a ratio and not a hold duration

Nothing records when a policy was paused, so hold duration is not knowable without persisting state. The ratio measures the same thing from the consequence end and is strictly better: it is the number an operator checks by hand, it is what makes the cost legible (4 days configured against 18 days actual), and it **self-scales** — one threshold serves a 4-day raw tier and a 35-day baseline tier alike. It also means a policy paused an hour ago on a young store reads as unremarkable rather than as a zero-second-old incident.

## The thresholds, bounded on both sides

| | | |
|---|---|---|
| **Below** | whole-chunk granularity tops out at **1.25x** on the shipped raw tier | 2.0x sits clear of it with margin |
| **Above** | the motivating incident sat at **4.5x** after 16 days | 2.0x on that tier is ~8 days — about a week in, while still recoverable |
| **Critical** | 4.0x — double the warning | the real incident reads **Critical** |

Compile-time constants rather than a store-backed knob like the #2136 sibling's V57 column, only because that needs a migration rung this deliberately does not take. Noted in the code as belonging in the control plane next time a rung is going in anyway.

## What makes this safe: it never acts

The check cannot arm a policy, deliberately. Arming a held policy drops the only copy of history no rollup has materialized — precisely what the gate exists to prevent. The release stays the `--backfill-rollups` operator decision; this makes the need for one **visible** instead of silent.

## Scope: no migration rung

Persisting `scheduled` into the metrics series is the better long-term answer and would give the hold a *history* rather than a current-state read. It needs a rung, and a parallel lane owns the schema bump — taking a second number risks the silent-gap failure (a rung at or below the store's stamped version is skipped forever). Tracked separately; this delivers the half that would have caught the incident.

## Verification

**AWS SSO expired mid-task** (token lapsed 00:41Z), so I could not reach the production boxes. Everything below is the local TimescaleDB **2.28.1** rig, which is the better instrument here anyway — it builds a genuinely held state deterministically, which a production store that was already released could not provide.

Built the exact production shape: 19 chunks under a 4-day policy, plus an armed 2-chunk control.

Ran the **shipped `ReadRetentionHoldReadingsAsync` itself** — not a retyped copy of its SQL — against it:

```
query_stats        armed=True   drop_after=4 days  chunks=2   span=1.08d   ratio=0.27
query_store_stats  armed=False  drop_after=4 days  chunks=19  span=18.08d  ratio=4.52
```

**4.52x** — the production incident's reading, reproduced. 20 assertions passed, including the negative controls that must stay quiet: a freshly-created paused policy (0.01x), whole-chunk granularity (1.25x), a paused policy with no chunks, an unreadable horizon, and an **armed but equally deep** tier.

**Span normalization proven, not asserted.** `chunks.range_start` is declared `timestamptz` even for the naive-`timestamp` partitioning column every collector table uses. Output is byte-identical under `UTC`, `Pacific/Kiritimati` (UTC+14) and `America/Los_Angeles` (UTC-7).

**Proven red three ways**, each failing differently:

| mutation | result |
|---|---|
| reader ignores `j.scheduled` | 3 assertions fail |
| span not normalized to UTC | 1 assertion fails |
| horizon not parsed from job config | ratio null on both policies — the alert can never fire |

The third initially reported "0 failures" because the harness crashed before its counter printed; I chased that rather than accepting it, and confirmed the mutation does go red. I also hit an empty `ref/` directory from a stale incremental build and cleaned it, since that is the shape that makes a pin look green on code that is not there.

The span read is catalog-only (`min(range_start)` over `timescaledb_information.chunks`), never a scan of a multi-hundred-GB hypertable.

`Darling.Tests` targets `net10.0-windows` and cannot run on macOS; the 10 new xUnit tests build but CI is the arbiter for their execution.


erikdarlingdata:
Production verification, now that SSO is available again (it had expired mid-task, which is why the PR body says the rig was the only instrument).

Ran the **shipped SQL extracted from the source file** — not retyped — read-only against all three production stores. It parses and returns sane values on real TimescaleDB catalogs on every one.

**Nothing is held fleet-wide.** All 18-19 retention policies on each store read `scheduled = true`, so #2809 is holding and there is nothing for this alert to fire on today. That is the expected state, not a null result.

**The `policy_retention query_store_stats [1047]` failure I was asked to check rather than assume is unrelated to a hold.** It reads armed, with zero chunks and no measurable span — that hypertable has been empty by design since the workload split, so its historical failure count is not a held-retention signal. It would be skipped by this check for want of a ratio, correctly.

**The threshold choice is now empirically confirmed, not just reasoned.** Every armed, healthy tier in production:

| tier | span | horizon | ratio |
|---|---|---|---|
| store A raw tiers (x3) | 5.09 d | 4 d | **1.27x** |
| store B raw tiers | 4.09 d | 4 d | 1.02x |
| store B `collection_log` | 31.09 d | 30 d | 1.04x |
| store A `collection_log` | 18.09 d | 30 d | 0.60x |
| store C `collection_log` | 16.09 d | 30 d | 0.54x |

The raw tiers sit at **1.27x while working perfectly** — the whole-chunk granularity floor I derived from first principles as ~1.25x, landing within 0.02 of the prediction on real data. A warn ratio anywhere at or below 1.27x would fire every hour on three healthy armed policies on the largest store. 2.0x clears it with margin, and every healthy tier in production is quiet.

This is the check I most wanted and could not run earlier: the floor was the half I had reasoned about rather than measured.

Issue #2818, opened 2026-09-03
EnsureRetentionPolicies live test discards all 17 exception messages, so its intermittent 'expected 17, got 0' cannot be triaged


erikdarlingdata:
## The observation

`TimescaleSupportTests.EnsureRetentionPolicies_ConvergesAnOldHorizon_PreservingScheduledStateAndNextStart_AgainstDevPostgres` failed once on a nightly with **expected 17, got 0**, then passed on re-run — on the same commit that had passed the same job ~30 minutes earlier on PR #2812.

It was called a flake. **I could not determine the root cause, and I am not going to assert one.** What follows is what I ruled out with evidence, and the reason the next occurrence will be just as undiagnosable unless one thing changes.

## Ruled out

- **Static initialisation order.** `RetentionPolicies` is built from `RawTierCoverage` and `BaselineAggregates`. Both are declared in the *same class* at lines 891 and 1912, before `RetentionPolicies` at 1949, so textual init order is correct and the collection cannot be empty. (An empty collection would have produced `0` with no error noise, which fit the symptom well — but it is not possible here.)
- **Within-assembly parallelism.** `TimescaleSupportTests` carries `[Collection("live-postgres")]`, so it is serialised against the other live classes.
- **Cross-job database sharing.** Both `build.yml` and `nightly.yml` set `DARLING_TEST_PG` to `Host=127.0.0.1;Port=5541`, a per-runner instance. Two jobs cannot be stepping on one database.

## What `got 0` actually means

`EnsureRetentionPoliciesAsync` increments `applied` only after each relation's transaction commits. A return of `0` therefore means **all 17 relations threw** and every one landed in the per-relation catch — not that the loop was skipped.

## Why it is undiagnosable, which is the real defect

That per-relation catch is:

```csharp
catch (Exception ex) when (ex is not OperationCanceledException)
{
    logger?.LogWarning("Retention policy for {Relation} ({DropAfter}) failed - ...: {Message}", relation, dropAfter, ex.Message);
}
```

and every call site in this test passes **`null`** for the logger:

```csharp
Assert.Equal(RetentionPolicyCount, await TimescaleSupport.EnsureRetentionPoliciesAsync(connection, null, ct));
```

So all 17 exception messages are discarded by the `?.` and the only surviving evidence is the bare count. There is no capturing `ILogger` anywhere in `Darling.Tests` to pass instead.

[...]

PR #2887, opened 2026-09-04
Carry the retention live tests' evidence into their failure messages (Fixes #2818)

erikdarlingdata:
## The defect

Both live retention tests passed `logger: null` to `EnsureRetentionPoliciesAsync`, so its per-relation catch — correct for the service, where one bad relation must not abort the other sixteen — discarded every warning message. When the nightly failed once with **expected 17, got 0**, all seventeen discarded messages were the diagnosis and the bare count was the only survivor. It was written off as a flake because it could not be anything else. Same family as #2801/#2816/#2826: a failure that destroys the one piece of information needed to act on it.

## The fix

Test-side only. #2818 assumed no capturing `ILogger` existed in `Darling.Tests`; `CapturingTestLogger` has existed since #2128, added by #1564 for the purge E2Es after the identical null-logger shape. Both tests now create one, pass it to `EnsureContinuousAggregatesAsync` and every `EnsureRetentionPoliciesAsync` call (a swallowed CAGG failure is a plausible upstream cause of the same all-relations-throw symptom), and fold `Joined` into each count assertion — the four `Assert.Equal` count checks become `Assert.True` with evidence-carrying messages, the `DarlingRetentionTests` house pattern. The shared logger's empty-case placeholder generalizes from `(no purge log lines)` to `(no log lines captured)`; nothing asserts on that literal.

## Proven both ways against live TimescaleDB 2.29 (PG17), running the SHIPPED methods

Injected the all-relations-throw shape (migrated store, no hypertables/CAGGs):

- old shape: `applied=0` — everything the assertion could say
- new shape: `applied=0` **plus one captured warning per relation**, first line: `Warning: Retention policy for query_stats (4 days) failed - … TS001: "query_stats" is not a hypertable or a continuous aggregate`

Healthy path unchanged: 69/69 hypertables, 20/20 CAGGs, **17 applied, 17 idempotent, zero warnings captured**.

No product behavior change, no migration rung, no schema bump. The next occurrence names its cause in the CI log on the first failure.


erikdarlingdata:
Good catch — fixed in 3d1dcd1f. Every `EnsureRetentionPoliciesAsync`/`EnsureContinuousAggregatesAsync` call now gets its own fresh `CapturingTestLogger` rather than one shared across all passes in a test. Verified against the exact HELD-fixture shape you described (query_stats held for the whole test, so every pass would otherwise log a benign warning): pass-to-pass `Joined` line count now stays flat (2-3 lines) instead of accumulating across passes.

Issue #2797, opened 2026-09-02
The server-scoped watermark read is discarded entirely for collectors that declare a per-database watermark


erikdarlingdata:
Follow-up from the review on #2796, filed rather than folded in because it needs verification across four collectors and three dispatch paths.

## The claim

For `query_store`, the value `GetLastCollectedTimeAsync` returns is **never consumed**. `QueryStoreCollector` declares `PerDatabaseWatermarkColumn`, and every execution path overwrites `context.Watermark` before a query is built:

- Azure SQL DB (`RunsPerDatabase`): `context.Watermark = await GetLastCollectedTimeForDatabaseAsync(...)` in the per-database connection loop (`DarlingCollectorRunner.cs:581`)
- on-prem / RDS / MI enumeration: `context.Watermark = clamped` inside the `perItemWatermark` delegate (`:923`)

The local is otherwise read only to compute `hasCollectedBefore`.

## What I verified

`HasCollectedBefore` is consulted by exactly two collectors — `DefaultTraceEventsCollector` (`:319`) and `JobHistoryCollector` (`:220`) — and **neither declares `PerDatabaseWatermarkColumn`**. So for the four collectors that declare both watermark columns, `hasCollectedBefore` is computed from this read and then never consulted.

The four: `QueryStoreCollector`, `DeadlocksCollector`, `BlockedProcessReportCollector`, `LongQueryCompletionsCollector`.

## What I did NOT verify, and why this is filed rather than fixed

The second half of the claim — that *every* dispatch path overwrites `context.Watermark` for *all four* of those collectors — is the load-bearing part, and it spans three paths (per-database Azure loop, enumeration, plain). If any path for `deadlocks`, `blocked_process_report` or `long_query_completions` does not overwrite, skipping the read silently destroys that collector's watermark and it re-collects its whole window forever — the same failure #2795 just fixed, arrived at from the other direction.

That is not a check to do by reading; it wants the live rig.

## The prize

After #2796 the read costs ~2.9-5.9 s per cycle per server (down from 40-50 s and a cancellation). Skipping it where its answer is discarded takes that to **zero**, and makes the new 60 s `CommandTimeout` unreachable for a call nothing uses.

Gate would be `definition.PerDatabaseWatermarkColumn is not null` — the same signal the per-item paths already key on.

## Suggested shape

1. Assert, per collector and per dispatch path on the live rig, that `context.Watermark` is written before `BuildQuery`.
[...]

Issue #2860, opened 2026-09-03
Decide how to persist the per-database fetch split (plan_fetch/text_fetch), which is N:1 against collection_log


erikdarlingdata:
## Deferred from #2859

#2859 persists the **server-scoped** phase split (`open:` / `drain:` / watermark) to `collection_log` as columns. It deliberately leaves out [#2811]'s **fetch split**, which is a genuinely different shape and needs its own decision.

## Why it does not fit the same treatment

The fetch split is emitted per DATABASE:

```
plan_fetch:Nms = probe:Nms + target:Nms + write:Nms + other:Nms (N chunk(s), N ids, N probed)
```

`collection_log` holds **one row per collector RUN**. An enumerated collector fans out over many databases and writes a single row whose `duration_ms` is the sum. So the fetch split is **N:1** against this table — it cannot become columns on it without first deciding *what to record*: the slowest database's split? the sum? the whole distribution?

That is the same question [#2472] had to answer for the per-database fan-out, and its answer (V80) was a deliberate three-column **rollup** — `fanout_item_count`, `slowest_item`, `slowest_item_ms` — chosen over a `collector_item_timings` hypertable that would have kept the full distribution at ~10% of `collection_log`'s row volume forever.

## Options

1. **A rollup, V80-style** — e.g. slowest database's `probe`/`target`/`write` plus the id and chunk counts. Cheap, no new rows, answers "which half of the fetch is slow" but not "how unevenly is it spread".
2. **A per-item table** — `collector_item_timings`, keeping the whole distribution. Answers everything, costs rows forever. [#2472] considered and rejected this shape for the fan-out case; the reasoning may or may not transfer, since the fetch split has more terms.
3. **jsonb on the existing row** — the one case where the varying shape (different phase names per fetch kind, plus counts) would actually justify it.

## What should decide it

The question the fetch split is normally asked is which of `probe:` / `target:` / `write:` owns the time — three of the four steps are Postgres and the blended number cannot say which held it ([#2811]'s own framing). If that question is answerable from the *slowest* database alone, option 1 is enough and cheapest. If the fetch cost is spread evenly across many databases — the shape option 1 is blind to — it is not.

That is measurable from the existing log lines before anything is built, and should be measured first.

erikdarlingdata:
## Measured first, as the issue asked. The deciding criterion splits, and the answer is a fourth shape.

[...]

Issue #2795, opened 2026-09-02
Server-scoped watermark read is unbounded, times out, and silently degrades every collector to its fallback window


erikdarlingdata:
## What

`DarlingCollectorRunner.GetLastCollectedTimeAsync` (the **server-scoped** watermark read) runs:

```sql
SELECT MAX(<watermark_column>) FROM <table> WHERE server_id = $1
```

with no predicate on `collection_time` — the hypertable's partitioning column. So it scans every chunk in retention for that server, on every collector cycle.

`#2344` fixed exactly this defect, but only in the **per-database** sibling `GetLastCollectedTimeForDatabaseAsync`. `QueryStoreCollector` declares *both* `WatermarkColumn` and `PerDatabaseWatermarkColumn`, so both reads run and the server-scoped one kept paying the unbounded cost.

## Measured on `prod-pos-use1-monitor-01`

`query_store_stats` is 62.5 GB across 19 chunks (24 h chunk interval, `collection_time` partitioning, retention `drop_after: 4 days`).

| form | time |
|---|---|
| unbounded (shipped) | **50,560 ms** / **40,743 ms** cold, 9,279 ms warm |
| bounded at `WatermarkPolicy.ReadFloor` (now − 3 h) | **2,891 ms** / **5,937 ms** |

Npgsql's default `CommandTimeout` is **30 s**, and no `CommandTimeout` was set on this command. So the read was routinely cancelled mid-flight.

The store's own PostgreSQL log for one day, grouped by the statement that followed each cancellation:

```
2092 x SELECT MAX(last_execution_time) FROM query_store_stats WHERE server_id = $N          <- unbounded, this bug
 631 x SELECT DISTINCT database_name FROM query_store_stats WHERE server_id = $N AND collection_time > $N
  17 x SELECT MAX(last_execution_time) FROM query_store_stats WHERE server_id = $N AND database_name = $N AND collection_time > $N   <- #2344's bounded sibling
```

Same store, same table, same column: **2,092 cancellations for the unbounded form against 17 for the bounded one.**

## Why it mattered more than the wasted time

The `catch` swallowed the exception with `/* If the Postgres query fails, caller uses fallback window */` and no log line. A cancelled read returns `null`, and `null` is **indistinguishable from a first run**. So every timeout silently pushed the collector onto query_store's 60-minute fallback window instead of the ~5-minute incremental one — re-collecting data already stored, growing the table, and making the next read slower.

[...]

PR #2890, opened 2026-09-04
Fixes #2797

erikdarlingdata:
Closes #2797 — but **not with the gate the issue proposed**, which was disproved before this was written.

## The gate the issue proposed would ship a data-correctness bug

#2797 suggested gating the server-scoped watermark read on `definition.PerDatabaseWatermarkColumn is not null`. Reflecting over `CollectorCatalog.All` in the built assembly and invoking the real `RunsPerDatabase` / `BuildEnumerationQuery` against a synthetic on-prem target:

| collector | WatermarkColumn | PerDbWatermark | RunsPerDatabase(on-prem) | BuildEnumerationQuery(on-prem) | dispatch path |
|---|---|---|---|---|---|
| `query_store` | `last_execution_time` | `database_name` | False | **non-null** | 2 enumerated |
| `deadlocks` | `deadlock_time` | `database_name` | False | null | **3 PLAIN (server-scoped)** |
| `blocked_process_report` | `event_time` | `database_name` | False | null | **3 PLAIN (server-scoped)** |
| `long_query_completions` | `event_time` | `database_name` | False | null | **3 PLAIN (server-scoped)** |

All four declare both watermark columns, so that gate skips the read for all four. But `query_store` is the only one overriding `BuildEnumerationQuery`; the other three inherit `CollectorDefinitionBase`'s `=> null`, and all four declare `RunsPerDatabase => target.IsAzureSqlDb`. **Off Azure — the whole on-prem and RDS fleet — three of the four fall through to the plain path and genuinely consume the read.** Skipping it there nulls their watermark every cycle and they re-collect their first-run window forever: #2795's defect from the other side.

## What this does instead

The read is discarded iff the run takes a fan-out path, which is a **dispatch-path** predicate:

```
WatermarkColumn is not null
  && PerDatabaseWatermarkColumn is not null
  && (RunsPerDatabase(target) || BuildEnumerationQuery(probe) is not null)
```

`ServerWatermarkIsDiscarded` on the runner, beside the read it gates. Invoking the **shipped** predicate over the built assembly gives **7 of 168 (definition, target) pairs**:

| collector | on-prem | azure-db | azure-mi | aws-rds |
|---|---|---|---|---|
| `query_store` | **SKIP** | **SKIP** | **SKIP** | **SKIP** |
| `deadlocks` | read | **SKIP** | read | read |
| `blocked_process_report` | read | **SKIP** | read | read |
| `long_query_completions` | read | **SKIP** | read | read |
| every other definition | read | read | read | read |

Skipping the read also had to gate `HasCollectedBefore`, whose branch keys off `watermark is null` and would otherwise have fired its own store query on exactly the cycles this makes cheaper — trading one round trip for another.

## Design decision: a derived signal, not a declared one, and not a reorder

The predicate cannot be evaluated where it needs to sit, because `BuildEnumerationQuery` takes a `CollectorContext` and the cycle's context is built *after* the read. Two options were on the table:

**(b) Reorder so the context is built first — rejected.** `hasCollectedBefore` is *computed from* the gated read (`watermark is null && await HasPriorCollectorSuccessAsync(...)`), and `CollectorContext.HasCollectedBefore`, `NumericWatermark` and `State` are all `init`-only. Building the real context ahead of the read therefore means converting three deliberately-immutable members on a type both SKUs and every definition share into settable ones — widening a contract to buy an optimisation.

**(a) A definition-level signal — taken, but derived rather than declared.** A hand-maintained `EnumeratesItems` flag would just relocate the bug the moment it disagreed with `BuildEnumerationQuery`. So the gate **calls `BuildEnumerationQuery` itself**, against a probe context carrying the cycle's real `Target` and real `ExcludedDatabases` (everything the five implementations read) and nothing the gated read produces. There is no second encoding to drift.

That leaves exactly one soundness condition — the probe equals the dispatch call only while null-ness depends on `Target` alone — and it is **asserted, not assumed**: a pin varies `Watermark`, `NumericWatermark`, `HasCollectedBefore`, `State` and `ExcludedDatabases` across every definition in the catalog on four target shapes, 5,376 combinations, and fails if any of them moves the answer.

## The pin, and the red-first evidence

`ServerWatermarkDispatchGateTests`, following #2796's shape (family derived by reflection, asserted against the shipped member). Every set comes from `CollectorCatalog.All` — no test names a collector to decide what it expects — with floor assertions so the sets can grow but cannot shrink to nothing and pass vacuously.

Proved red five ways, each a different assertion:

| variant | assertion that failed |
|---|---|
| **the issue's own gate** (`PerDatabaseWatermarkColumn` alone) | `TheServerScopedMembersOfTheFamily_KeepTheirWatermark_WhichTheProposedGateWouldHaveDestroyed` — *"deadlocks declares PerDatabaseWatermarkColumn but takes the PLAIN server-scoped path off Azure … so it CONSUMES the server-scoped watermark"* |
| fix reverted (gate never fires) | `AnEnumeratingMemberOfTheFamily_SkipsTheServerScopedRead…` + `EveryMemberOfTheFamily_SkipsOnAzureSqlDb…` |
| `RunsPerDatabase` arm dropped | `EveryMemberOfTheFamily_SkipsOnAzureSqlDb…` |
| an enumerator's null-ness made watermark-dependent | `EnumerationNullness_IsAFunctionOfTargetAlone…` (Expected False, Actual True) |
| read site re-encodes the predicate inline instead of calling the gate | `EveryServerScopedWatermarkReadSite_AlsoCallsTheDispatchGate` |

Two real defects in the pin surfaced this way and were fixed:

- a call to a **generic** method is emitted against a `MethodSpec` token, not a `MethodDef`/`MemberRef`, so the first IL walk reported the gate as never called while it was called every cycle;
- the `i += 4` skip in the byte-scan idiom **swallows genuine call sites** — measured on this assembly, the gate's only call sits at IL offset 743 and the skipping form walked straight past it and reported zero. This pin scans every offset instead, because a false negative would let an ungated read read as gated. `ServerScopePhaseSplitTests` carries the same `i += 4`; it survives because it only asserts `> 0` over stamps with many call sites, but the idiom is not sound and is worth revisiting separately.

## Sizing — and a correction worth reading before merging

The removed read is `MAX(last_execution_time) … WHERE server_id = $1 AND collection_time > $2` over `collect.query_store_stats`, once per `query_store` cycle per server.

**Its cost is currently recorded nowhere.** `ServerWatermarkMs` reaches the phase log only under `if (result.ServerPhasesMeasured)`, and that flag is set only inside the plain server-scoped branch — which `query_store` never enters. V108's `watermark_ms` column is persisted on the same branch (#2860). So the `wm:` field that *is* populated for `query_store` in the phase log comes from the **per-item** log line and measures the **per-database** read — legitimate incremental work, untouched here. Any `query_store` `wm:` percentile harvested from that log is sizing a different read than this change removes.

The only direct measurement of *this* read is #2796's own, taken as it was being bounded: **2.9–5.9 s** against a 62.5 GB table (40.7 s / 50.6 s before the bound). That is the honest figure, and it cannot be confirmed against the fleet until #2860 closes the instrumentation gap.

## Scope

- Darling only, like #2796. Lite's runner has the same three-path shape but reads a local DuckDB file rather than a TimescaleDB hypertable, which is what makes this worth gating.
- No schema change, no migration, no version bump.
- **#2860 deliberately untouched.** This change sits next to it — the gate makes `ServerWatermarkMs` a measured zero for `query_store` rather than an unmeasured one — but the decision about persisting the phase columns on the enumerated path is that issue's.

## Verification

- Built assembly, not source: the before/after tables come from reflecting over `CollectorCatalog.All` and invoking the shipped `ServerWatermarkIsDiscarded` out of the built service DLL (MVID and mtime printed per run).
- `PerformanceMonitor.sln` builds clean on macOS with `EnableWindowsTargeting`.
- The pins were **executed**, not just compiled: `Darling.Tests` is `net10.0-windows` and cannot run here, but this file's own dependencies are all `net10.0`, so the real source was linked into a throwaway `net10.0` xUnit harness — 9/9 green (7 new, plus #2796's 2 still green), each discovered and run by name.
- **Not verified here:** the Windows suites (`build`), `Darling PostgreSQL tests`, and anything needing the live fleet. Both the AWS CLI and the Darling MCPs were unavailable in this session, so the fleet figures quoted above are carried from the prior investigation rather than re-measured.


erikdarlingdata:
Confirming both flags deliberately, since one of them asked for it either way.

**Lite parity — deliberate boundary, not a miss.** Lite's `RemoteCollectorService.DefinitionRunner` does carry the identical three-path shape and the identical unconditional server-scoped read, and the same four collectors throw the answer away there too. It is left alone because the thing being removed is not "a watermark read", it is *a network round trip to a separate Postgres/TimescaleDB store that does a `MAX` over the largest hypertable in it*. Lite's read is a local DuckDB file, and Lite's own doc comment on that method already measured it in the sub-millisecond to single-digit-millisecond range — so the same gate there buys microseconds while adding a dispatch predicate to a second runner. Same reasoning, and the same boundary, as #2796: that fix bounded this exact read on the partitioning column and was likewise Darling-only.

**The unguarded `BuildEnumerationQuery(dispatchProbe)` at the gate site — considered, and left unguarded on purpose.** Checked the same way: none of the five overrides can throw for a non-null context (`PlanCorrectionCollector`'s `ArgumentNullException.ThrowIfNull` is the only guard, and the probe is never null), and `CollectorDefinitionBase`'s inherited `=> null` obviously cannot. If a future member of the both-columns family did grow a throwing override, the exception would reach `RunOneAsync`'s existing classification arms exactly as it does today — the only difference is that it would arrive *before* the target connection is opened rather than after, which is the better of the two orderings. Swallowing it at the gate would be worse than propagating it: the gate would then answer "does not enumerate" for a collector that does, which is the silent mis-gate the whole predicate is built to avoid.

The asymmetry with the test helper's `catch (TargetInvocationException)` is intentional and not the same question. The pin sweeps **every** definition in `CollectorCatalog.All` across four target shapes — far more (definition, target) pairs than the gate itself ever evaluates, because the gate short-circuits on the two declared columns first. The catch is there so one unrelated definition throwing off its own path cannot derail a derivation whose whole point is that it covers the entire catalog. At the production site there is no derivation to protect, and a throw is information rather than noise.

Good catch on `serverReadFloor` — that one is exactly as the comment says, and it is the reason it was left rather than deleted.

Issue #2855, opened 2026-09-03
The Azure per-database branch emits no phase split at all


erikdarlingdata:
The Azure SQL DB per-database branch in `DarlingCollectorRunner.RunAsync` (the `OpenDatabaseConnectionAsync` loop) emits **no phase split at all**. It is uninstrumented, not mis-instrumented — distinct from #2854, which fixed stamps that existed but were skipped on throw.

Today it measures only a blended per-database total:

```csharp
using (var dbConnection = await OpenDatabaseConnectionAsync(perDbProvider, server, databaseName, dbToken))
using (var dbCommand = CreateCollectorCommand(perDbProvider, dbPlan, dbConnection, perDbTimeout))
using (var dbReader = await dbCommand.ExecuteReaderAsync(dbToken))
{
    batch = await definition.ReadAsync(dbReader, context, dbToken);
    ...
}
var dbSqlMs = sqlSlice.ElapsedMilliseconds;
```

So an Azure per-database run has exactly the attribution problem #2851 fixed for server-scoped collectors and #2164/#2312 fixed for the enumerated path: one number covering everything, which cannot say whether a slow database is bound by connect, by server-side work before the first row, or by streaming.

## Why it is its own issue rather than part of #2854

**The shape differs.** This branch opens a **connection per database**, which neither other path does. Its split is therefore `connect: + open: + drain:`, not `open: + drain:` — a third decomposition, with `OpenDatabaseConnectionAsync` as a phase in its own right. On Azure SQL DB that connect is a real cost (a fresh login per database, per cycle), and it is exactly the kind of term that currently hides inside a blended number.

**It needs new fields and a new log line**, not a stamp move: new `CollectorContext` members, a gating flag mirroring `PerItemPhasesMeasured` / `ServerPhasesMeasured`, and a per-database line this branch does not currently emit at all.

## Worth doing because the pattern keeps paying

Two independent ~20-40x in-process gaps have been found this week and both were only tractable once phases were split:

- the store probe: ~36 ms in a harness on the box vs ~1,451 ms in production (#2819/#2831)
- `procedure_stats`: 247 ms for the shipped query against a real target vs 4,644 ms `sql_duration_ms` p50, with the box at 4% CPU (#2847)

Neither was visible in a blended number. The Azure branch is the last collection path still reporting one.

## When it lands, extend the pin

[...]

Issue #2864, opened 2026-09-03
An abandoned collection cycle records that it stopped, not what it was doing


erikdarlingdata:
## Summary

A collector cycle that hits the #2673 120s wall-clock budget records `ABANDONED` and nothing else. It does not record **what it was doing**, **how far it got**, or **what the rest of its sweep looked like** — so a real production abandonment can be investigated for hours and still end in "insufficient evidence."

This issue is not a fix for the stalls. It is the telemetry that would make the next one diagnosable.

## What an investigation could and could not establish

Over 42 servers / 42,000 runs / 1.81h, ten runs exceeded 60s and two abandoned. That was enough to separate two populations **by hand**:

**Population A — a large query doing real work.** `query_store` at 60-73s returning 7,371-12,557 rows, while every other collector in the same sweep ran at or *below* its baseline (0.3-0.4x). One sweep: `wait_stats` 1ms, `latch_stats` 1ms, `spinlock_stats` 2ms, then `query_store` 71,977ms for 12,557 rows. Not a stall.

**Population B — sweep-wide degradation.** Light collectors at **34-47x baseline** in the same sweep (`wait_stats` 380ms against a 2ms prior sweep), *before* two heavy collectors each burned exactly 120s and abandoned with `rows_collected = 0`.

The distinction matters because the fixes are unrelated, and today it required manual cross-referencing of neighbouring rows to see.

**What could not be established, and why:**

- `rows_collected = 0` is rows *stored*. There is no way to tell whether the run never received row 1 or stalled at row 149 — which is exactly the split between "could not execute" and "could not drain".
- `collection_log` carries only `sql_duration_ms`. The #2851 `open:`/`drain:` split exists solely as app-log text, so it cannot be queried, aggregated or trended (tracked separately).
- **The wait series has a 6.5-minute hole across the event.** Collectors run strictly sequentially per server, so a stalled collector blocks every other observation of that target — including `wait_stats`, `waiting_tasks` and `dmv_blocking_snapshot`, the three that would explain it. `waiting_tasks` ran 2 seconds after the stall cleared and returned 0 rows. **The instrument stops sampling exactly when the thing being measured happens.**

## Proposed

[...]

Issue #2859, opened 2026-09-03
Persist the server-scoped phase split to collection_log so it can be aggregated without an SSM session


erikdarlingdata:
## The gap

[#2851] decomposes a server-scoped collector's `sql_duration_ms` into `open:` (the `ExecuteReaderAsync`) and `drain:` (the `ReadAsync` loop), and reports the store-side watermark read beside them. **All of it exists only as an app-log line.**

`collection_log` records `sql_duration_ms` and nothing finer, so the one question the split was built to answer — *which phase owns a collector's cost, across servers and over time* — cannot be aggregated, trended, compared across the fleet, or reached through the MCP surface at all. Answering it requires an SSM session onto the box and a log scrape, per server.

## The concrete cost

On 2026-09-03 an investigation into where `procedure_stats`' 4,724 ms `drain:` actually goes **stalled outright** on exactly this. AWS SSO began returning `InternalServerException` on `GetRoleCredentials`; SSM went unavailable; and the phase numbers were unreachable — even though the store itself was answering normally the whole time, through the MCP.

That is the shape of the problem: the store was up, the data was collected, and the instrumentation was still only on the box.

For scale, `procedure_stats` is currently the most expensive collector on use1 — 93.9M ms/day over 16,967 runs, fleet p50 4,818 ms — and its cost is ~96% drain. That ratio is exactly what these columns make queryable.

## Proposed

Add the phase figures to `collection_log` and surface them on the MCP read, so `avg(sql_drain_ms) / avg(sql_duration_ms)` is a store query.

Three design points worth settling in review:

1. **Columns, not jsonb.** The phase set on this path is fixed (open, drain, watermark), so jsonb's flexibility buys nothing and costs the cheap aggregation that is the entire point.
2. **Do not store `other:`.** It is a residual defined against `sql_duration_ms`; a stored copy can drift from the parent it completes, and sum-to-parent is precisely what makes a large residual a *finding* rather than noise. Readers subtract.
3. **The watermark is not part of the sum.** On this path it runs before the `sql:` stopwatch starts. Folding it in would print a permanent zero and teach every reader that a store read [#2796] clocked at 50 s cold is free — the exact misreading [#2851] added it to prevent.

## Explicitly out of scope

[...]

Issue #2854, opened 2026-09-03
Four enumerated-path phase stamps are skipped when the phase they time throws


erikdarlingdata:
`PerItemOpenMs`, `PerItemWatermarkMs`, `PerItemPlanFetchMs` and `PerItemTextFetchMs` are assigned on the line **after** the `await` they time, so a phase that throws skips its own stamp and reports zero.

This is the #2816 defect one seam over. That one found `probe:` stamped by a trailing assignment, and **97% of a day's entire `other:` residual — 43,053 of 44,319 ms — turned out to be one failed store probe** filed as our own bookkeeping.

## Why it is worse here than it was there

**1. The cost lands in `drain:`, which is a more confident wrong answer than `other:`.**

`CollectorContext.DrainMsFrom` subtracts the non-streaming phases from the item total:

```csharp
Math.Max(0, itemSqlMs - PerItemOpenMs - PerItemWatermarkMs - PerItemPlanFetchMs - PerItemTextFetchMs);
```

If a 500 s open times out, `PerItemOpenMs` stays 0 and the line reports `drain:500000ms` — attributing the whole cost to row streaming for a statement that never returned a row. `other:` at least means "our own code"; `drain:` names a specific phase that provably did not happen.

**2. Two of the four are the parents of the sub-split #2816 already fixed.**

`probe:`/`target:`/`write:` stamp from their own handlers and report real values. So a throwing `FetchAndStorePlansAsync` prints:

```
plan_fetch:0ms = probe:1100ms + target:250ms + write:100ms + other:0ms
```

A zero parent above non-zero children. `PlanFetchOtherMs` computes `Math.Max(0, 0 - 1100 - 250 - 100)` and clamps the impossible negative to 0, so the line still reads as a precise decomposition — of a parent smaller than its own parts.

**3. The gate suppresses the line exactly when it is worth reading.**

The split is emitted behind `if (context.PerItemOpenMs > 0)`. A faulting open leaves that zero, so the only per-database attribution there is vanishes for the item that needed it. The same `> 0` also cannot distinguish a genuinely instant open from one that never ran — the conflation #2852 called out on the server-scoped path.

## Baseline, from IL

Scanning the built service assembly for phase-stamp call sites and whether each falls inside an exception-handler region:

```
BARE set_PerItemOpenMs         total=2  inHandler=0
BARE set_PerItemPlanFetchMs    total=2  inHandler=0
[...]

PR #2893, opened 2026-09-04
Fixes #2855

erikdarlingdata:
## What

The Azure SQL DB per-database branch in `DarlingCollectorRunner.RunAsync` emitted **no phase split at all** — not a mis-stamped one, none — so a slow database could not be attributed even in principle. It was the last collection path still reporting one blended number, after #2164/#2312 (enumerated) and #2851 (server-scoped). Distinct from #2854, which moved stamps that already existed.

Each database now logs, on its own line:

```
  [alpha] query_store [demo] sql:1800ms = connect:1451ms + open:247ms + drain:36ms + other:66ms (149 rows, pg:12ms)
```

## Why the shape is a THIRD decomposition

This branch opens a connection **per database**, which neither other path does — the enumerated path reuses one target connection across items, the server-scoped path opens one for the whole server. So `OpenDatabaseConnectionAsync` is a phase in its own right: on Azure SQL DB a fresh login per database per cycle, and the same branch serves the per-database PostgreSQL collectors, where one connection per database is unavoidable because a PostgreSQL session is bound to one database for its lifetime.

`drain:` is **measured, not inferred**, so `other:` is a real residual (command construction, the trailing probe-failure rowset) — the #2851 departure, for the same reason.

The line is **not** gated on `batch.Count > 0`, unlike the enumerated path's quiet-on-zero rule. That rule guards a ROW-COUNT line, whose payload is the row count. This is a PHASE line, and the connect is paid on a quiet database exactly as on a busy one; the server-scoped phase line is the closer sibling and prints for every measured run.

## Naming: `PerDatabase*Ms`, and the pin widened in the same change

Reusing `PerItem*Ms` would have made an Azure run print as an enumerated one — the enumerated log site keys on its own flag and prints an item name. That is the exact collision `ServerScopeOpenMs`'s doc comment records as the reason the server-scoped path was given its own prefix, so the precedent decided it.

The price of the honest name is that `ServerScopePhaseSplitTests`' derivation had to be widened here, since a stamp outside the pattern is a stamp nothing guards — worse than no stamp, that file's whole claim being that a new one is covered the day it appears.

`MinimumPhaseStamps` goes **12 → 16**, the exact count rather than a round number. 12 was already one behind the true 13: #2864's `ServerScopeLastReadMs` matched the pattern the day it landed and was covered automatically — exactly as designed — but left room for a deletion to go unnoticed.

## The flag, and a gap it exposed

Every stamp and the gating flag are set from a `finally`, so a login that times out or a per-database budget that fires mid-handshake still reports the phase it died in. The log line gates on `context.PerDatabasePhasesFrom(dbSqlMs)` — one shared expression, so "no split" and "no line" are one decision — never on a figure being non-zero: a pooled connect can legitimately measure `0ms`, and that is precisely the database whose `open:` and `drain:` are worth reading.

While wiring that: **the measured flags' handler reachability was guarded by nothing.** The `*Ms` derivation selects on `long`, so no bool could ever appear in it. A flag assigned after the await leaves a faulting phase declaring itself unmeasured and printing nothing, while the number is stamped correctly and the arithmetic pin stays green. `MeasuredFlags_AreReachableFromExceptionHandlers_...` now derives the `*PhasesMeasured` setters and scans them through the same IL walk — which also covers the pre-existing `PerItemPhasesMeasured` (already correct, so it passes).

## Verification

Windows CI is the arbiter for `Darling.Tests`. Locally (macOS) I compiled the **real shipped** `ServerScopePhaseSplitTests.cs` into a throwaway `net10.0` xunit.v3 host — it has no Windows dependency — and ran it against the built assemblies: **14 passed, 0 failed** (7 pre-existing + 7 new). A separate reflection/IL harness over the built `PerformanceMonitor.Collectors.dll` and `PerformanceMonitor.Darling.Service.dll` confirms 16 derived stamps and that the three new members match the pin's selector (`long`, settable, `*Ms`, `PerDatabase*`), with the harness asserting its loaded DLLs are byte-identical to the projects' own bin output.

**Seven regressions proven red first, each on a different assertion:**

| variant | fails |
|---|---|
| a new stamp renamed out of the pattern | floor: `Only 15 phase stamps ... expected at least 16` |
| floor pushed past the count (17) | floor: `Only 16 ... expected at least 17` |
| the widening reverted | floor: `Only 13 ... expected at least 16` |
| connect stamp moved out of its `finally` | `set_PerDatabaseConnectMs is never invoked from inside an exception handler (2 call site(s), all on success paths)` |
| the flag moved out of its `finally` | `set_PerDatabasePhasesMeasured is never set from inside an exception handler (2 call site(s), all on success paths)` |
| gate keyed on `connect > 0` | **two** tests — `Assert.Null() Failure ... has a value` (prints a stale split) and `Assert.NotNull() Failure ... does not have a value` (drops a measured zero) |
| a value comparison added at the log site | `Assert.DoesNotContain() Failure: Sub-string found` |

## Not verified

- **No figure has been measured against a live Azure SQL DB target**, and nothing available here can reach one. The branch itself was not exercised end to end; what the connect actually costs is now *answerable*, not answered.
- `Darling.Tests` / `Lite.Tests` as suites were not run locally (net10.0-windows). Only the one pin file above was executed, in a net10.0 host.

## Scope

Visibility, not a speedup: no query, schedule, collector or connection behaviour changes. **Emit-and-log only** — a per-database split is N:1 against `collection_log` where the server-scoped split #2859 persists is 1:1, and #2860 already holds that cardinality question open for the per-database fetch split. No column, no rung; schema stays at **109**.

Fixes #2855


erikdarlingdata:
One deferral, filed rather than dropped: **#2896**.

Every stamp and the flag are set from a `finally`, so a database whose connect times out does stamp its phases and does declare itself measured. It just never prints them — the log site is on the success path, and a fault reaches one of the three catch arms instead (the budget arm's WARNING, or the generic arm's `Skipping database ...` at **Debug**).

Not fixed here because the line decomposes `dbSqlMs`, and that parent is out of scope in the catch: `sqlSlice` is declared inside the `try`, below the watermark read and `BuildQuery`. Hoisting it would silently widen `dbSqlMs`, which is not log-only — it feeds `sqlMs`, `fanout.Observe(...)` and `collection_log.sql_duration_ms`. That leaves a second stopwatch or a second line shape, and #2896 records both plus the measurement that should decide between them.

Stated in the code too, at the log site and in the pin's header, rather than left for a reader to infer that "stamped from `finally`" means "reported on the fault path".

Issue #2891, opened 2026-09-04
Add a CI check on PR commit author identity


erikdarlingdata:
## The gap

Nothing in CI checks the **author identity of a PR's commits**. This repo has one intended commit
identity — the maintainer's GitHub `users.noreply.github.com` address — and a commit authored under
any other address would land on `dev` with all six checks green, because no check looks at
authorship.

Commit metadata is effectively permanent: once a commit is on `dev`, changing its author needs a
history rewrite, which invalidates every existing clone and every open PR's merge base. So this is a
class where prevention is the only affordable control, and the prevention is cheap.

## Why the existing guards don't cover it

- `FleetIdentifierScrubTests` scans tracked *source*. Commit metadata is not a file it can see.
- `check-pr-branch.yml` already does exactly this genre of work — a cheap check over PR *metadata*
  rather than content — but only asserts that PRs to `main` come from `dev`.
- The review bot reads the diff, not the authorship.

So the check has an obvious home and no home currently claims it.

## Suggested shape

Add a step to the existing `check-branches` job in `.github/workflows/check-pr-branch.yml`. It
already runs on `pull_request_target` for `opened`/`reopened`/`synchronize`/`edited`, which is the
right trigger set, and `pull_request_target` deliberately does **not** check out PR head — so use the
API rather than a checkout:

```
GET /repos/{owner}/{repo}/pulls/{number}/commits
```

and fail if any commit's `commit.author.email` or `commit.committer.email` is not on an allowlist.
Notes for whoever picks this up:

- **Check both `author` and `committer`.** A rebase or an amend can leave one correct and the other
  not; the two are independent fields.
- **Allowlist, don't pattern-match loosely.** `*@users.noreply.github.com` is broad enough to admit
  any GitHub account's noreply address, including one belonging to a different identity. Pin the
  specific address(es) intended for this repo.
- **`web-flow@github.com`** is GitHub's own committer for merge commits and web-UI edits. It will
  appear legitimately and needs allowlisting, or the check fails on every merge commit.
- **Paginate.** The endpoint caps at 250 commits per PR, far above anything here, but the default
  page size is 30 — without pagination a long PR silently passes on its unread tail. That is the same
  trust-a-truncated-result failure shape this repo has hit elsewhere.
- Emit `::error::` naming the short SHA and which field was wrong, so the fix is obvious without a
  local bisect.

A local `pre-commit` hook would catch it earlier, but hooks are per-clone and unenforceable, so the
[...]

PR #2904, opened 2026-09-04
Fixes #2891

erikdarlingdata:
Adds the commit-identity gate from #2891 to the existing `check-branches` job.

## What it does

`.github/workflows/check-pr-branch.yml` now calls `.github/scripts/check-commit-identity.sh`, which reads the pull request's commit list from `GET /repos/{owner}/{repo}/pulls/{number}/commits` (not from a checkout of the pull request) and fails on any commit whose `commit.author.email` **or** `commit.committer.email` is off a pinned allowlist. Both fields, because a rebase or an amend routinely leaves one right and the other wrong.

Errors name the short SHA and which of the two fields was wrong, and **deliberately not the address**: Actions logs on a public repo are public, so echoing it would publish the identity the check exists to keep out. The remediation block shows how to read the value locally and how to rewrite the branch with `--reset-author`, which fixes both fields at once.

## This check cannot run on its own pull request

`pull_request_target` executes the workflow file as it exists on the **base** branch. The `check-branches` run on this pull request is `dev`'s current copy, which has no identity step, so **a green `check-branches` here is not evidence that any of this works.** It takes effect only once merged.

So it was verified by running the extracted script directly against real pull request data from the API:

| Case | Pull request | Declared commits | Result |
| --- | --- | --- | --- |
| Non-conforming author **and** committer | a merged one-commit pull request | 1 | exit 1, two `::error::` lines (one per field) |
| Non-conforming, multi-commit | a merged five-commit pull request | 5 | exit 1, ten `::error::` lines |
| Clean, single page | #2900 | 3 | exit 0 |
| Clean, **three pages** | #1250 | 250 | exit 0, `250 commit(s) over 3 page(s)` |
| Dependabot | #2702 | 1 | exit 0 |
| Above the endpoint's cap | #1611 | 926 | exit 1, coverage failure (API returns 250) |

## Three corrections to the design in #2891

Each was checked against this repo's actual history rather than assumed.

**1. GitHub's committer here is `noreply@github.com`, not `web-flow@github.com`.** The issue says to allowlist `web-flow@github.com` or the check fails on every merge commit. That address appears in this repo's history **zero times**; all 346 GitHub-created merge commits and all 80 squash merges use `noreply@github.com`. Allowlisting only `web-flow` yields **101 violations on the 250 commits of #1250**, a pull request that is entirely clean:

```
allowlist without noreply@github.com -> violations: 101
allowlist as shipped                 -> violations: 0
```

Both are pinned now: `noreply@github.com` because it is what this repo actually produces, `web-flow@github.com` because it costs nothing and GitHub still emits it on some paths.

**2. Dependabot had to be pinned, or the check would red every Monday.** #2891 describes one intended identity. But `.github/dependabot.yml` has Dependabot opening grouped pull requests against `dev` weekly, and it authors those commits itself (committer is GitHub, author is Dependabot). With the issue's allowlist taken literally, #2702 fails on its author field. Its numeric prefix was confirmed to be the real global account id via `/users/dependabot[bot]`.

**3. The 250-commit cap is not hypothetical in this repo.** #2891 calls it "far above anything here". This repo has pull requests at 926, 607, 536, 452, 446, 304 and 304 commits. Rather than pass on the unread tail, the script compares what it read against the count the pull request declares and reports a **coverage failure** if they differ — a preventive gate that cannot cover its subject should say so rather than report a pass over the part it happened to read. **This is a judgement call and the one thing here worth overruling** if you would rather a very large pull request warned instead of failed.

## Also worth knowing

- **Outside contributions will now be red.** This is what "one intended commit identity" means, but it is a policy consequence rather than a bug: history shows several merged contributions from other identities. Nothing here exempts forks.
- The older `erikdarlingdata@users.noreply.github.com` form of the maintainer address (one commit in history, no numeric prefix) is **not** allowlisted. Pinning one exact string is the point; noting it in case that is wrong.
- A synthetic case for a *different* account's `@users.noreply.github.com` was confirmed to fail, which is the address-shape gap #2891 names as the reason not to pattern-match.
- The job gained an explicit read-only `permissions:` block and a base-ref checkout. The checkout pins `ref: ${{ github.base_ref }}` explicitly rather than relying on the default, so it cannot quietly start resolving to pull request head; no pull request content is fetched or executed.
- One real bug was caught only by running it: inside jq's `index(...)`, `.` refers to the allowlist array, not the commit field, so the first version died with `Cannot index array with string "email"` on every pull request. It failed closed, but it was broken.

## Not done, deliberately

No existing commit was rewritten. Per #2891 the scope is making future ones impossible; history is out of scope.

Schema version untouched (109), no migration.

## Review response (both acted on, neither dismissed)

**Anchoring the read to the event head.** The coverage assertion compared the paginated commit list against `.commits` from `GET /pulls/{n}` — two endpoints in the same family, so a stale post-push view could leave them agreeing with each other while both described the *previous* head, and comparing them only to each other cannot see that. The read must now also contain the event's head SHA, which is an independent anchor and fails closed: no anchor, no verdict, red, and a re-run clears it. That matches the rest of the script's stance that a false red costs a re-run while a false green is permanent. Skipped when `HEAD_SHA` is absent, which is how the script runs when pointed at an arbitrary pull request by hand.

Worth noting that this staleness is **not** theoretical: immediately after pushing the fix, `GET /pulls/2904` still reported the previous `headRefOid` and `mergeable: CONFLICTING` for a while before catching up. Verified both directions — anchored on the real head SHA it passes and says so; anchored on a SHA absent from the list it fails with the staleness message.

**Page over-count.** `pages_read` counted the trailing empty-page probe, so a commit count that is an exact multiple of the page size reported one page too many. It now counts only pages that returned commits. Verified against the patched loop structure: 100 commits now report 1 page (was 2), 200 report 2 (was 3), 250 still report 3, 3 still report 1.

## What this check is not

Raised in review and now stated in the script header as well, because it would be easy to read more into a green result than it carries: `commit.author.email` and `commit.committer.email` are self-declared git fields that nothing authenticates. Anyone who can push can set `user.email` to an allowlisted value and pass. What this catches is a **misconfigured clone** — the wrong `user.email` left set on a machine, which is the case #2891 was opened about — and nothing more. It is not an impersonation control, and a pass here is not evidence that a commit came from whoever it names. Enforced commit signing is the only thing that gets you that.


erikdarlingdata:
Closing unmerged. See #2891 for the reasoning — the check catches a misconfigured clone rather than enforcing identity (author/committer are self-declared), and the implementation cost turned out disproportionate to a rule that can simply be followed. The work is preserved here if the trade ever looks different. Thanks for the three rounds of review; both substantive findings (the stale-view false-green and the page-count over-count) were real and were fixed before this decision.

Issue #2871, opened 2026-09-04
PgBaselineProvider inherits Npgsql's 30s default CommandTimeout, and is hitting it in production


erikdarlingdata:
## Summary

`PgBaselineProvider` sets no `CommandTimeout`, so its baseline query inherits **Npgsql's undocumented 30 s default** — a value nobody chose. This is the same defect #2810 found and fixed in `PgFactCollector`, in the other half of the same 120 s analysis pass.

It is failing in production today.

## Measured evidence

**Every observed failure sits on a hard 30 s wall.** From the dogfood box's app log, 19 occurrences across two days:

| when (UTC) | metric | elapsed |
|---|---|---|
| 2026-09-03 00:27:54 | io_latency | **31.4 s** |
| 2026-09-03 02:41:28 | io_latency | 30.4 s |
| 2026-09-03 02:41:58 | io_latency | 30.2 s |
| 2026-09-03 02:42:22 | io_latency | 30.3 s |
| 2026-09-03 02:42:43 | io_latency | 30.3 s |
| 2026-09-03 02:43:21 | io_latency | 30.3 s |
| 2026-09-03 04:14:56 | io_latency | 30.7 s |
| 2026-09-03 04:16:14 | io_latency | 31.1 s |
| … | | |
| 2026-09-02 23:25:52 | io_latency | 30.7 s |

**No sample falls outside 30.1–31.4 s.** That is not a variable-duration failure and it is not the 15 s connection timeout — it is a fixed ceiling being hit.

**The query itself is fast.** Running the *shipped* query string (dumped from the built assembly, not retyped) against the live store, three busiest servers by `file_io_stats` row count in the 30-day window:

| store | first execution |
|---|---|
| use2 srv A | 1,624 ms |
| use2 srv B | 1,476 ms |
| use2 srv C | 1,362 ms |
| use1 srv A | 230 ms |
| use1 srv B | 336 ms |
| use1 srv C | 198 ms |

So the read normally completes in **~1.6 s worst case** and only crosses 30 s during bursts of store-wide slowness — note the clustering above, five failures inside two minutes and five more inside four. Exactly #2810's shape: it fails on the bad combination only, which is why it reads as intermittent.

**The failure data is right-censored.** Every failing run was killed *at* 30 s, so nothing in the record says whether it needed 35 s or 300 s. Any bound has to be chosen against that limitation rather than pretending to a number the data does not contain.

## Why the two previous changes could not have fixed it

[...]

Issue #2810, opened 2026-09-03
The plan-regression analysis read cannot complete on a large store, and its timeout is swallowed silently


erikdarlingdata:
## What

`PgFactCollector.CollectPlanRegressionFactsAsync` (`PerformanceMonitor.Darling.Analysis/PgFactCollector.QueryPerf.cs:409`) runs `PlanRegressionSql` with **no `CommandTimeout`**, so it inherits Npgsql's undocumented 30 s default. On `prod-pos-use1-monitor-01` the query cannot finish in **300 s**, so it is cancelled on every single analysis pass and has almost certainly never produced a fact on that store.

Running the **shipped** query string (extracted from the assembly source, not retyped), against the busiest `server_id`, with production-shaped parameters:

```
PREPARE pr(int, timestamp, timestamp) AS <PlanRegressionSql>;
EXPLAIN (ANALYZE, BUFFERS, TIMING, SUMMARY)
EXECUTE pr(1671144557, now() - interval '15 days', now() - interval '16 days');

SET statement_timeout = '300s';
Time: 300303.379 ms (05:00.303)     <- hit the ceiling, did not complete
```

## This is the dominant remaining store cancellation

Grouping every `canceling statement due to user request` in the store's own PostgreSQL log since the 2026-09-02 22:54:56Z restart:

```
81 x  WITH deduped AS ( -- Collapse incremental re-collections of the same open runtime-stats interval ...
12 x  SELECT DISTINCT database_name FROM query_store_stats WHERE server_id = $1 AND collection_time > $2 ORDER BY database_name
---
93 total in ~1 hour
```

The first is `PlanRegressionSql`. The second is `QueryStoreBackfill.cs:404`.

**Note for anyone re-running this analysis:** an earlier pass reported ~1,673 of these as an unattributed "blank STATEMENT" bucket. That was a parser artifact, not a mystery — PostgreSQL writes `STATEMENT:` and then the SQL on *following* lines for a multi-line statement, so a parser reading only the remainder of the `STATEMENT:` line sees an empty string. There is no unattributed bucket.

## The failure is invisible

```csharp
catch (Exception ex) when (!AnalysisShutdown.IsExpectedAbandon(ex, context.CancellationToken))
{
    /* Table may not exist or have no data. An abandonment is NOT swallowed here (#2443). */
}
```

[...]

PR #2909, opened 2026-09-04
Part of #2874

erikdarlingdata:
Part of #2874. Does **not** close it.

## What was left

#2871 swept `PerformanceMonitor.Darling.Analysis` and `AnalysisPassCommandTimeoutTests` declared the invariant over the whole assembly: *"every command constructed anywhere in this assembly sets a deadline."* The claim was right and the scan was wrong — **it read 71 sites as covered while 73 existed.**

`PgFactCollector.Waits.cs:231` and `PgDrillDownCollector.Blocking.cs:161` both handed `connection.CreateCommand` to `PgBlockingPairRowQuery.AppendDmvSnapshotRowsAsync` as a bare **method group**. That helper sets only `CommandText` on whatever the factory returns, so the deadline could only come from the hand-off — and a method group has no `(` after it for either census regex to match. Same shape and same blind spot as the viewer's 193rd site in #2901.

## Which half was actually unbounded

Worth being precise, because half of this is a tidy-up and half is a real hole.

On the **scheduled** path `context.CancellationToken` is armed by `DarlingWorker.s_analysisTimeout`'s 120 s `CancelAfter`, so a stuck read there was already bounded — it just ate the whole analysis pass instead of the collector's own budget.

But `AnalysisContext.CancellationToken` defaults to `CancellationToken.None`, and **`DarlingWorker` is the only caller in the repo that passes a token.** Every on-demand entry point omits it:

| entry point | reaches |
|---|---|
| `analyze_server` | **both** reads — `CollectFactsAsync`, then `EnrichFindingsAsync` |
| `get_analysis_facts` | the fact collector's |
| `compare_analysis` | the fact collector's (twice — baseline + comparison) |
| `audit_config` | the fact collector's |

On those paths the read had **no deadline and no token**, and ran purely on Npgsql's undocumented 30 s default. So this is what bounds them, not a nicety layered on a token that was already doing the work.

## Which constants, and why they differ

Neither number is new and neither was copied.

- **`FactCommandTimeoutSeconds` (60 s)** for the fact collector's read — #2810 derived it from sitting inside that same 120 s pass budget, and its doc comment already claims *every* command in that collector. This read is one of them; it was simply unreachable by the sweep.
- **`DrillDownCommandTimeoutSeconds` (30 s)** for the drill-down's — what every other command in that pass already carries, and the value #2871 looked at and deliberately left alone.

The drill-down constant **equals the default it replaces, so no timing changes there.** The point is that 30 is now *chosen* rather than arrived at — verbatim what that constant's own doc comment exists to say ("pinning it here makes the bound explicit and independent of any global default").

## Two alternatives rejected, with reasons

**A shared floor inside the helper** cannot express this. The three callers sit in three budget regimes (60 s, 30 s, the viewer's interactive read deadline), so there is no single value to apply. And the *"apply a floor only if unset"* variant is **unsound**, not merely awkward: Npgsql's default IS 30 and the drill-down's deliberate choice is ALSO 30 — indistinguishable from inside the helper, which is the exact ambiguity #2874 exists to remove.

**Taking the deadline as a required parameter** was the strongest alternative — compile-time enforced, covers callers in projects no glob reaches. It lost on **pin visibility**: the helper's `createCommand()` invocation matches neither census regex (no `.` prefix, lowercase), so the one line the entire fix depended on would have become the only unguarded line in the pass. Keeping the stamp at the call sites puts the fix where the assembly's own pin can see it, and keeps the signature identical to Lite's DuckDB twin. The contract is now documented on the helper instead.

## Pin

`AnalysisPassCommandTimeoutTests` now matches **both** construction shapes (`new NpgsqlCommand(` and `.CreateCommand(`) and adds the method-group scan from `ViewerCommandTimeoutTests`, with its comment/string stripper.

**Deliberately assembly-scoped, not repo-wide.** The same shape appears four more times in `Lite` — `DuckDbFactCollector.Waits.cs:234`, `DrillDownCollector.Blocking.cs:165`, `LocalDataService.Blocking.cs:426` and `:547` — but those build `DuckDBCommand`, which has no 30 s default to inherit and may not honour `CommandTimeout` at all. Sweeping them would assert a bound that cannot be shown to bind anything: a pin that cannot fail. Whether a DuckDB command wants a deadline is a separate question with its own measurement.

### Proven red two ways, each failing a different assertion

Windows suites can't run on macOS, so I compiled the **real** test file (not a retyped copy — `[CallerFilePath]` still resolves to the worktree, so it scans the real sources) into a throwaway `net10.0` harness and ran the shipped assertions.

1. **Revert both call sites** → fails **only** the method-group scan, naming `PgDrillDownCollector.Blocking.cs:161` and `PgFactCollector.Waits.cs:231`. The existing per-site sweep stays **green** — which is the direct evidence that the old pin could not see this shape.
2. **Keep the lambda, delete its deadline** → fails **only** the per-site sweep, at `PgFactCollector.Waits.cs:245`. Measured: the pre-#2874 narrow regex does **not** match `connection.CreateCommand()`, so the regex widening is independently load-bearing.

Final run: 14/14 green, including the adjacent `FactCollectorCommandTimeoutTests` to check I hadn't disturbed it. `.Analysis`, `Darling.Tests` and the Viewer all build.

### One claim I am not making

The stripper is **defensive here, not load-bearing.** Three comments in this assembly name `connection.CreateCommand` in running text, but all three are followed by a backtick or `</c>` rather than a comma — the unstripped scan measures **zero** false hits today. It is carried because the viewer's sibling pin *did* trip on its own explanation, one comma is the whole difference, and a theory fixture pins the behaviour so it can't be dropped as dead weight.

## What this does not do

**No user-reported failure is closed by it.** Unlike #2871's io_latency baseline (19 failures in two days, all at 30.1–31.4 s) and #2810's plan-regression read (34.7 s cold), **neither of these two reads has an observed production timeout**, and I did not measure their cost against the production store for this change. This removes an inherited default nobody chose and closes a census gap — it should not be read as fixing a known break.

No schema change (still 109) and no version bump.


erikdarlingdata:
Taken, in `44f459ba` — and it's a better catch than "pre-existing floor behaviour" gives it credit for.

It isn't pre-existing that the count is 73. **This PR moved it.** Widening `s_commandCtor` to the `.CreateCommand(` shape made two sites *visible* that had always existed, so the true count went 71 → 73 without a single command being added. Leaving the floor at 71 would have let both of them vanish again silently — which is precisely the regression that assertion exists to catch, so it was rot I introduced, not rot I inherited. Measured with the shipped regex before changing it: **73 covered, 0 missing.**

The generalisable bit, now recorded in the comment: **a floor has to be re-measured whenever the scan's REACH changes, not only when code is added.** That case isn't obvious, because nothing in the diff looks like a new command.

One consequence I had to re-measure rather than restate: bumping the floor changes what a revert trips, so the red proof in the PR body was wrong as written.

| red case | now fails |
|---|---|
| revert both call sites | method-group scan (naming both sites) **and** the floor (71 found, 73 expected) |
| keep the lambda, delete its deadline | the per-site half only, at `PgFactCollector.Waits.cs:245` |

Importantly the original point survives: on a revert, the per-site half of that same test **still reports zero missing**, because a method group stays structurally invisible to it. That remains the reason the scan had to be *added* rather than the regex merely *widened*.

Also strengthened the out-of-scope justification using the citation you supplied, which I verified at `Lite/Services/RemoteCollectorService.cs:1313-1317`. It's firmer evidence than my own hedge: `DuckDBCommand.CommandTimeout` defaults to **0, meaning no limit**, so there's no inherited ceiling in Lite to exceed and nothing to cancel. I'd written "may not honour `CommandTimeout` at all"; the repo already knew the sharper answer. Swapped it in, and kept that comment's own instruction to re-confirm the 0 default before relying on it.

14/14 green on the local harness after the change (the real test file compiled into a throwaway `net10.0` runner, since the Windows suites can't execute on macOS).

erikdarlingdata:
Taken in `ebc0924d`, and I'd argue it's worth more than "latent".

**Why I didn't defer it:** this entire issue exists because a census regex missed a shape. Leaving a *known* missed shape in the regex written to replace it would be the same mistake in miniature — and the fix is one character. `[,)]` → `[,);]`.

You're right about the mechanism: `[,)]` catches a hand-off **inlined as an argument**, which is the only form in the repo today. A hand-off through a local ends in a semicolon —

```csharp
Func<NpgsqlCommand> factory = connection.CreateCommand;
```

— and slips past both argument forms *and* `s_commandCtor`, a method group having no `(` either way. Exactly the blind spot that produced the original 71-of-73.

Measured before changing it, since a widened class risks false positives:

| | `.Analysis` | `.Viewer` |
|---|---|---|
| hits for `\.CreateCommand\s*[,);]` | **0** | **0** |

The called form `CreateCommand()` stays excluded by the `(`, so the widening is free. Then proven red rather than assumed: narrowing the class back to `[,)]` fails the new `TheFactoryHandoffScan_ReadsCodeNotProse` fixture. 15/15 green with it restored.

**One thing I deliberately left alone.** `ViewerCommandTimeoutTests`' otherwise identical regex still carries `[,)]`, so the viewer has the same latent gap — measured zero live instances there too. I didn't reach into it: it's #2901's landed pin, it isn't in this PR's scope, and a one-character edit to another lane's just-merged test isn't mine to make unasked. Recording it here so it isn't lost — it's a one-line follow-up for whoever picks up the rest of #2874, which stays open.

PR #2912, opened 2026-09-04
Scrub the last fleet identifier from main

erikdarlingdata:
`main` is the repo's default branch and is swept separately from `dev`. The two have genuinely divergent history, so the identifier scrubs merged to dev ([#2900], [#2903]) never reached it — four locations on `main` still named a real tenant server.

## What changed

| file | was | now |
|---|---|---|
| `Darling.Tests/DarlingPeerDisclosureTests.cs:385` | `display_name` fixture | `omega` |
| `Darling.Tests/DarlingPeerDisclosureTests.cs:413` | matching assert | `omega` |
| `Darling.Tests/StatementSplitTimingTests.cs:79` | short name in comment | `omega-01` |
| `PerformanceMonitor.Collectors/CollectorContext.cs:253` | short name in comment | `omega-01` |

`omega` is not a fresh choice — it is the slug [#2900] reconciled this server onto, it is already in `FleetIdentifierScrubTests.SyntheticSlugs`, and **these four lines now match `origin/dev` byte-for-byte** (verified by diffing slug usage per file). Every measurement in the touched comments is unchanged.

Line endings preserved per file (CRLF counts asserted before/after).

Same precedent as 26a4e149, "Scrub fleet identifiers from main, which is what the repo shows by default".

## Why a PR and not a direct commit

26a4e149 went straight onto `main`. This is the same change shape, but a PR leaves the diff reviewable and revertable, and there is no rush now that the issue/PR-body exposure is closed.

### DIFF
--- a/Darling/Darling.Tests/DarlingPeerDisclosureTests.cs
+++ b/Darling/Darling.Tests/DarlingPeerDisclosureTests.cs
@@ -382,7 +382,7 @@ public sealed class DarlingPeerDisclosureTests
     {
         var rows = new List<DarlingDataReader.ServerListRow>
         {
-            new(1, "prod-sql-use1-beta-01", "ayr", 16, new DateTime(2026, 8, 19, 12, 0, 0, DateTimeKind.Utc)),
+            new(1, "prod-sql-use1-beta-01", "omega", 16, new DateTime(2026, 8, 19, 12, 0, 0, DateTimeKind.Utc)),
         };
 
         return JsonDocument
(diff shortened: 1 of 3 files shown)


erikdarlingdata:
Closing. This can't land as shaped — `check-branches` fails by design, because the workflow only permits PRs to `main` from `dev`, and this targets `main` from a topic branch. That's the policy working, not a flaky check.

It also isn't needed. The equivalent changes are already on `dev` (#2900, #2903), those commits are simply not in `main` yet, and they arrive with the normal release merge — `dev` already holds the corrected blobs for every file this PR touches. No separate action required.

The verification work here was sound and the slug choice was right (matching what #2900 reconciled to, rather than picking fresh). Nothing wrong with the change itself — just the branch it targeted.

Issue #2894, opened 2026-09-04
The migration advisory-lock wait cannot cover an arbitrary number of data-moving rungs


erikdarlingdata:
## Summary

#2888 gave the migration advisory-lock acquire its own budget, `MigrationLockWaitTimeoutSeconds = 5 * MigrationCommandTimeoutSeconds` (1500 s), because the previous value was one *statement's* bound applied to a *whole-session* wait. That is correct as far as a constant can go, and it deliberately does **not** close the failure. This issue records the residual so it is not mistaken for closed.

## The residual

`MigrateAsync` takes the advisory lock ONCE and holds it while `MigrateLockedAsync` applies **every** pending rung in the same session. The 1500 s floor is derived from today's rung census: exactly **four** of the 108 rungs touch pre-existing data, each is one command already capped at `MigrationCommandTimeoutSeconds`, so four multiples is the floor and the fifth is margin.

Two things still outlast it:

1. **A fifth data-moving rung.** The bound is pinned to a census taken at one point in time. Nothing fails when a new `migrate_data`, `CREATE INDEX`-on-populated-table, whole-table `UPDATE` or hypertable conversion rung lands, so the constant silently stops covering the ladder.
2. **One rung genuinely needing longer than its own 300 s bound.** Then the holder dies at its statement bound, but a waiter that arrived early enough can still have burned its 1500 s first.

The failure is not graceful. The waiter throws out of `PgMigrations.MigrateAsync`, `DarlingWorker.RunCollectionLoopAsync` catches it, logs `LogCritical` and **returns out of the collection loop** — that instance then collects **nothing** until an operator restarts it. There is no retry and no enclosing budget: one production caller, plain stopping token, no `CancelAfter`, no configured `HostOptions.StartupTimeout`, no health check or readiness probe, no container `HEALTHCHECK`, no orchestrator manifest.

## The census this rests on, so it is re-derivable

108 rungs (V1-V109, V45 absent). Every rung's SQL body extracted (100 `const string`, 8 from `PgSchemaGenerator`, plus V23 which is the file's only `private static string V23Sql =>` property) and scanned for data-moving shapes. Exactly four touch pre-existing data:

| rung | operation | evidence |
|---|---|---|
| V22 | `CREATE INDEX` on populated `collect.index_object_stats` | rung's own doc: "a plain `CREATE INDEX` is applied across every chunk by Timescale" |
[...]

PR #2920, opened 2026-09-04
Part of #2894

erikdarlingdata:
Part of #2894 — its **option 3 (the rung-census pin) only**. Neither the wait-scaling nor the `pg_try_advisory_lock` retry loop is built here.

## What it asserts

`MigrationDataMovingRungCensusPins` (new, `Darling.Tests`) scans every rung's **shipped** SQL — the real `PgMigrations.Scripts` strings, so the ten rungs assembled wholly or partly from the schema generators (V1, V8, V10–V14, V38, V51, V54) are covered rather than transcribed — for **fourteen** data-moving shapes, and requires the matching rung set to equal a declared census. Six tests:

| test | claim |
|---|---|
| `TheDataMovingRungSet_EqualsTheDeclaredCensus` | set equality **both** directions; failure names `MigrationLockWaitTimeoutSeconds`, its derivation, and the `LogCritical`-and-return consequence |
| `TheLockWaitMultiple_IsOneMoreThanTheFloorContributingRungCount` | the constant's multiple, read out of `PgMigrations.cs`, equals floor-setting rungs + 1 of margin |
| `TheScan_ActuallyReadsTheLadder` | positive-control floors on the `CREATE INDEX` / `CREATE TABLE` match counts, plus comment stripping |
| `AnIndexOnATableTheSameRungCreates_IsFree_AndOnAnEarlierRungsTableIsNot` | the discrimination, both sides, on a synthetic ladder |
| `EveryDataMovingShape_FiresOnASyntheticRung` | 18 `[InlineData]` cases — the anti-vacuity control for the shapes with no ladder representative |
| `ProseAndTriggerClausesNamingTheseShapes_AreNotFindings` / `TheLadderStillContainsNoDynamicSql` | the false-positive and blind-spot halves |

The census is **not** frozen at a count — a pin asserting "there are exactly four" fails on every legitimate new rung and gets suppressed. What is declared is the **set**, each member carrying its reason and a `SetsTheFloor` flag, so a new rung is a two-line declaration plus, if it sets the floor, a re-derivation of the constant in the same commit.

## How a free `CREATE INDEX` is told from an expensive one

The ladder holds **134** `CREATE INDEX` statements and **130** of them index a table the **same** rung creates — empty, therefore free. A pin that flagged the keyword would fire 130 false times and be off within a week. So every targeted shape resolves its statement's target table and is dropped when that rung created it; the diagnostic also names the rung that *did* create it. Identity is the **bare** table name, because V1 creates the collector tables unqualified in `public` and V8 moves them to `collect` — **37** bare names are created under both spellings, so a schema-qualified comparison would split them.

## Re-derived census on `dev` at schema 110

**109 rungs** (V1–V110; **V45 permanently absent**, and the reason is in `Scripts` itself — it was reserved for #1951 while unmerged, #1952 landed first and took 46, and a rung at or below a store's stamped version is skipped forever).

**Six** rungs touch data an earlier rung created, not four:

| rung | shape | target | created by | sets floor |
|---|---|---|---|---|
| V22 | `CREATE INDEX` | `collect.index_object_stats` | V1 | yes |
| V23 | `create_hypertable` + `migrate_data => true` | `collect.collection_log` | V2 | yes |
| V39 | `CREATE INDEX` ×2 | `query_stats`, `procedure_stats` | V1 | yes |
| V62 | `ADD CONSTRAINT … CHECK` | `config.config_service` | V17 | **no** |
| V77 | `DELETE FROM` ×2 | `collect.collector_state` | V44 | **no** |
| V104 | `CREATE INDEX` | `collect.pg_deadlocks` | V1 (V103 on upgrade) | yes |

The four the constant rests on stand. **V62 and V77 are new relative to #2894's census**, which says "exactly four touch pre-existing data" — both are genuine work against pre-existing rows (a `CHECK` constraint validates every existing row; the two `DELETE`s are real DML), and they are now declared at **zero** cost rather than absent from the register, because their targets are a single-row control-plane table and a few rows per server per collector. Neither moves the `5×`.

**V110 is not data-moving.** Ten `ADD COLUMN IF NOT EXISTS … integer`, nullable, no `DEFAULT`, plus a `CREATE OR REPLACE VIEW`. Catalog-only, and the rung's own doc says so.

Also checked and absent from the whole ladder, each with a positive control run through the identical pattern: `ALTER … TYPE`, `SET NOT NULL`, `ALTER TABLE … ADD PRIMARY KEY/UNIQUE`, `ADD COLUMN … DEFAULT <fn>()`, `INSERT INTO … SELECT`, `refresh_continuous_aggregate`, `VACUUM`/`CLUSTER`/`REINDEX` as statements, and dynamic SQL.

## Two corrections to the constant's doc comment

1. **108 → 109 rungs.**
2. **`MigrationCommandTimeoutSeconds` is a per-RUNG bound, not a per-statement one.** The applier does `new NpgsqlCommand(migration.Sql, …) { CommandTimeout = MigrationCommandTimeoutSeconds }` — one command per rung, whole SQL. So V39's two indexes cost **one** multiple rather than two, and the four-plus-one derivation holds. The old text said "each of the four is ONE command", which happens to be true for the wrong reason.

No schema-version bump, no rung, no behaviour change: a new test file plus a doc comment.

## Red-first evidence

Ten mutations, one at a time, each with its anchor count asserted **before** application (a mutation that silently fails to apply reads as a pass) and each file restored byte-identically afterwards. Run against a throwaway `net10.0` xunit host compiling the **real** test source with `AssemblyName=Darling.Tests` — `Darling.Tests` is `net10.0-windows` and cannot execute on macOS, so **CI is the arbiter**. 24/24 green unmutated.

| mutation | fails |
|---|---|
| add a rung indexing `collect.query_stats` | census pin, undeclared half — `V111 … CREATE INDEX on collect.query_stats, created by V1` |
| repoint V105's index at `collect.query_stats` | census pin, undeclared half (V105) |
| drop V39 from the census | census pin, undeclared half (V39, both statements listed) |
| declare V110 (not data-moving) | census pin, **vanished** half, echoing the declaration's own stated reason |
| constant → literal `1500` | multiple pin's **positive control** ("no longer declared as `<n> *`") |
| multiple `5` → `6` | multiple pin's arithmetic, naming the four floor-setting rungs |
| add a rung using `EXECUTE format(...)` | dynamic-SQL pin |
| break the `CREATE INDEX` regex | **5** tests: scan floor (`matched only 0`), census vanished half (V22/V39/V104), the discrimination test, and two shape cases |
| break the `INSERT INTO … SELECT` regex | **only** the shape theory — census stays green, which is exactly why that theory exists |
| remove the same-rung exemption | the discrimination test, and the census floods with 130 `CREATE INDEX` findings — the "turned off within a week" outcome, demonstrated |

## Existing pins checked, not assumed

`CollectionLogDrainForensicsStoreTests.NullMeansNotRecordedAndTheDocsDoNotClaimMore` reads `PgMigrations.cs` as text and slices from `IndexOf("2864")`. Verified after the edit: the anchor still resolves (offset 142,677), the edited doc comment **is** inside the slice, both `DoesNotContain` assertions still hold, and a planted-phrase positive control confirms they can still fail. `MigrationLadderPins` anchors on rung `Sql` — every rung's shipped SQL string dumped before and after the edit and diffed byte-identical. `StorageCommandTimeoutTests` scans command-construction sites — unchanged at 8 / 0 / 8.

## Not verified

No live store. There is no local PostgreSQL+TimescaleDB store rig in this repo, and this change adds no SQL and no runtime path, so nothing here needed one — the four floor-setting rungs' timings are #2894's, quoted rather than re-measured. The Windows suites cannot run on macOS; the full `Darling.Tests` project compiles clean under `-p:EnableWindowsTargeting=true` (0 errors, no new warnings) and CI decides.


erikdarlingdata:
Both review findings were real and both are fixed at `1ad99b3e`. I traced each against the actual pattern rather than taking the description on faith, and one of them is worse than "worth documenting".

**Unnamed `CREATE INDEX` — a true silent miss, fixed rather than documented.** Confirmed exactly as described: with the name mandatory, `CREATE INDEX ON collect.foo (bar);` produces **no match at all**, because `Ident` consumes the literal `ON` and then the required second `ON` is not there. Measured on the two forms side by side:

| statement | name mandatory | name optional |
|---|---|---|
| `CREATE INDEX ON collect.foo (bar);` | **NO MATCH** | `collect.foo` |
| `CREATE INDEX ix ON collect.foo (bar);` | `collect.foo` | `collect.foo` |
| `CREATE INDEX IF NOT EXISTS ix ON collect.foo (bar);` | `collect.foo` | `collect.foo` |
| `CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS ix ON ONLY collect.foo (bar);` | `collect.foo` | `collect.foo` |
| `CREATE INDEX CONCURRENTLY ON collect.foo (bar);` | `collect.foo` | `collect.foo` |

Adding it to the blind-spots paragraph would have been the wrong call: the other three entries there are things a text scan genuinely cannot see, while this one is a one-token regex fix. The name group is now optional, and the derivation is on `s_createIndex` itself (including why no rung hits it today — they all want `IF NOT EXISTS`, which cannot be spelled without a name).

**`ALTER TABLE ONLY` — right, and it does not miss, it lies.** `Ident` greedily matched `ONLY`, so the finding still fired; the diagnostic just named the wrong object (`ONLY / (no rung creates it)`). All four `ALTER TABLE` shapes now skip an optional `ONLY`.

**Red-first, one revert at a time**, anchor count asserted before each mutation and the file restored byte-identically after:

- revert the optional name → `OptionalSyntaxDoesNotBecomeTheReportedTarget` fails with `Assert.Single() Failure: The collection was empty`, plus the new `CREATE INDEX with no index name` shape case. That empty collection *is* the silent miss.
- revert the `ONLY` skip → same Fact fails with `Expected: "collect.old"` / `Actual: "ONLY"`.

**Neither fix moves the ladder's numbers**, which is the thing worth checking on a regex widening: still **134** `CREATE INDEX` statements, **130** of them same-rung, and the same six data-touching rungs (V22, V23, V39, V62, V77, V104). The new Fact and both `[InlineData]` cases also assert the same-rung exemption survives both forms, so no miss was traded for a false positive. 27 tests (was 24), all green.

Issue #2918, opened 2026-09-04
The compose statement_timeout only reaches the live roles on a restart, unlike every other config_service knob


erikdarlingdata:
Split out of #2917, which fixes the propagation but deliberately does not change delivery.

### Mechanism

The compose `statement_timeout` is not applied per query or per connection — it lives on the roles. `DarlingManagedRoles.BuildProvisioningSql` emits:

```sql
ALTER ROLE viewer SET statement_timeout = '{n}s';
ALTER ROLE mcp    SET statement_timeout = '{n}s';
```

`ALTER ROLE ... SET` is persistent in `pg_authid` and applies at session start for that role, so every future viewer/MCP session inherits whatever the last provisioning run wrote. Provisioning runs exactly once per process, from `DarlingWorker.cs:1025`, gated on `config.Postgres.Managed && OperatingSystem.IsWindows()`. `ReloadFromStoreAsync` never re-provisions.

Consequence: an operator who lowers `config_service.compose_statement_timeout_seconds` to stop a runaway compose query from pinning the store gets a `config_version` bump, a reload — and no change to the live ceiling until the service is restarted. Every other `config_service` knob on that path goes live on the reload.

This is the *documented* design (`StoreConfigView.ComposeStatementTimeoutSeconds`: "a change here reaches an existing install on its next restart"), and the reason given in #2357 is sound as far as it goes — a role `statement_timeout` has no probeable schema footprint, so it cannot be a versioned migration without breaking the viewer's connect-time version gate. But that argument is about why it is not a *migration*. It says nothing about why the reload does not re-assert it, which was never addressed.

### Options

1. **Re-assert the two `ALTER ROLE` statements on reload.** Targeted, not the whole provisioning batch — that batch also resets role passwords from the credential file and re-grants the ACLs, and running it on every `config_version` bump is far more than the knob needs. There is no privilege obstacle: the service's store connection is the **owner** (`PostgresConfig.ConnectAs` is the *Viewer's* role selector and is documented as "ignored by the service, which always connects as the owner"), so it can `ALTER ROLE viewer/mcp SET ...` directly.
2. **Re-assert only when the value changed**, comparing the view's value against the last-provisioned one held in worker state. Avoids a catalog write on every unrelated reload — a `config_version` bump happens on any `config_service`/schedule write, so option 1 would touch `pg_authid` on reloads that have nothing to do with this knob.
[...]

Issue #2826, opened 2026-09-03
PlanRegressionSql timeouts are silently swallowed — a cancelled query is indistinguishable from 'no plan regressions'


erikdarlingdata:
## Summary

When `PlanRegressionSql` is cancelled by its (unchosen, inherited) 30 s Npgsql deadline — **325 times on 2026-09-03 alone**, see #2827 — the failure is discarded with no log line, no retry, and no fact. "The plan-regression query timed out" and "this server has no plan regressions" produce **byte-identical output**.

`Darling/PerformanceMonitor.Darling.Analysis/PgFactCollector.QueryPerf.cs:479`:

```csharp
catch (Exception ex) when (!AnalysisShutdown.IsExpectedAbandon(ex, context.CancellationToken))
{
    /* Table may not exist or have no data. An abandonment is NOT swallowed here (#2443). */
}
```

The body is empty. A `PostgresException` for `57014 query_canceled` is neither "table may not exist" nor "no data", and nothing distinguishes it from the early return four lines above:

```csharp
if (offenderCount == 0) return;
```

Both paths emit no `PLAN_REGRESSION` fact. So on the servers where this query reliably exceeds 30 s, plan-regression detection is **silently disabled**, and has been for as long as the query has been over the line.

This is the same family as #2801 (abandoned cycle recorded SUCCESS), #2796 (cancelled watermark read returned `null`, indistinguishable from a first run, and silently downgraded the collection window), #2816 (`probe:0ms` misattributed to `other:`), #2818 (all 17 relations threw but every message was discarded by a null logger), and #2820 (timed-out baseline cached a `null` for an hour). In every case the code was correct about the *happy* path and blind about its own failure.

## It is a category, not an instance

`grep -c "Table may not exist or have no data"` across `Darling/` returns **25**. Fixing only the plan-regression site would be scenario-shaped — the same repair that had to be made twice in #2822's recovery path and twice again in the `PgStatementText` / #2344 enumerated-pin gap.

## The fix is cheaper than previously assessed

#2795 records that logging here "needs an `ILogger` on `PgFactCollector`, which today takes only an `NpgsqlDataSource` — a ctor change across 7 call sites."

That is not required. Three siblings in the same project already take the logger as an **optional** parameter, so no call site changes at all:

- `PgBaselineProvider(NpgsqlDataSource postgres, ILogger? logger = null)`
[...]

Issue #2736, opened 2026-08-31
get_query_store_duration_trend ranks the raw Query Store slab per call and times out at any window width on a large store


erikdarlingdata:
`get_query_store_duration_trend` cannot complete on a large store at any window width. On a store carrying ~6.9M `query_store_stats` rows per 6 hours, it timed out on a full day, on a 2-hour historical slice, and on a 2-hour window anchored at now — the last one as a control.

## Observed

> get_query_store_duration_trend times out at any width — full day, 2-hour slice, and 2-hour window anchored at now (that last one as a control). It's the instrument, not the data.

The control matters: a cost proportional to the requested window would have let the 2-hour-at-now probe through. It didn't, because the query's dominant cost is fixed, not window-proportional.

## Mechanism

`DarlingMcpTrendTools.GetQueryStoreDurationTrend` (`Darling/PerformanceMonitor.Darling.Service/Mcp/DarlingMcpTrendTools.cs:439-483`) runs `DarlingTrendReader.QueryStoreDurationTrendSql` (`Mcp/DarlingTrendReader.cs:407-472`). Arm 1 — the one that matters on current data — is a `ROW_NUMBER()` dedup over the RAW hypertable, partitioned at full interval-identity grain (`database_name, query_id, plan_id, runtime_stats_interval_id, first_execution_time, execution_type_desc, replica_role`), windowed on `interval_start_time_utc` but chunk-excluded only by a deliberately slack `collection_time` slab (`DarlingTrendReader.cs:436-437`):

```sql
AND   collection_time >= $2 - interval '1 day'
AND   collection_time <= $3 + interval '30 days'
```

So even the 2-hour now-anchored control scans and RANKS at least 26 hours of raw rows for the server — tens of millions at this store's scale — before the aggregate sees a single row. The whole thing runs as the `mcp` role under its `statement_timeout` backstop (default 15s, `DarlingManagedRoles.cs:411-412`, #2357), which is the timeout actually observed. Raising the timeout is not the fix; the read is doing #1841's dedup the expensive way, per call, forever.

[...]

Issue #1776, opened 2026-07-28
Six live Darling.Tests classes run outside the live-postgres collection, producing moving cross-class flakes on a shared store


erikdarlingdata:
Six Darling.Tests classes connect to the shared `DARLING_TEST_PG` store without `[Collection("live-postgres")]`, so they run in parallel with the 63 classes that have it - and with each other. On a local rig where several full runs share one database, this produces intermittent failures that land on a *different* unrelated class each run, which is expensive in exactly the way a flake is worst: it looks like the change under test broke something it never touched.

### Measured, not inferred

Three consecutive full-suite runs against one long-lived database during #1775:

| Run | Failures |
| --- | --- |
| 1 | `DarlingEndpointToggleCliTests.EnableThenDisableMcp_FlipsFlagAndSelfBumpsConfigVersion_AgainstDevPostgres` + `StoreConfigProviderTests.ConfigVersion_BumpTriggers_FireOnConfigWrites_AgainstDevPostgres` |
| 2 | none - 3583 passed / 0 failed |
| 3 | `ExcludedDatabasesStoreLiveTests.GetCollectedDatabaseNames_ReturnsDistinctUserDatabases_AcrossBothStores_AgainstDevPostgres` |

**Every class that failed is a class without the attribute.** That is the whole explanation, and it is a complete one - the failures are not random, they are confined to the set that is allowed to run concurrently. `dropdb && createdb` then re-running comes back clean, which is the one-step diagnostic.

### The set

Classes with tests that reference `DARLING_TEST_PG` and do NOT carry `[Collection("live-postgres")]`:

| Class | Shares the `DARLING_TEST_PG` database? |
| --- | --- |
| `StoreConfigProviderTests` | yes - and it writes singleton config rows |
| `DarlingEndpointToggleCliTests` | yes |
| `ExcludedDatabasesStoreLiveTests` (in `ExcludedDatabasesTests.cs`) | yes |
| `DarlingManagedPostgresTests` | needs a look - some tests spin their OWN cluster |
| `DarlingStoreUpgradeTests` | needs a look - the `_Gated` tests `initdb` their own cluster |
| `DarlingPgRuntimeVersionPinTests` | needs a look - probes the bundle rather than the store |

[...]

PR #2931, opened 2026-09-04
Part of #2874

erikdarlingdata:
Part of #2874 — group **A** of five, the MCP and web read surface. This does not close the issue; four groups remain.

## What was wrong

All **125** command sites on this surface set no `CommandTimeout` and inherited Npgsql's undocumented 30 s default: **118** under `Mcp/` (114 of them the `.CreateCommand(` shape #2874's census could not see, and `Mcp/DarlingDataReader.cs` is 26 on its own), the **6** `CustomViewStore` statements both surfaces persist through, and the **1** composed-query runner in `DarlingWebEndpoints`.

## Census, re-derived — with positive controls first

Not taken from the issue. The landed pins' machinery lifted verbatim — `s_commandCtor`, `s_setsTimeout`, the method-group regex, and `CSharpSourceWalker.StatementSpanFrom(…, statements: 2)` — into a `net10.0` harness. **The controls are what establish it counts the way the landed pins count:**

| project | pin scope | sites | by shape | untimed |
|---|---|---|---|---|
| `.Storage` | `TopDirectoryOnly` | **124** | 74 / 50 | **0** |
| `.Viewer` | `AllDirectories` | **193** | 2 / 191 | **0** |
| `.Analysis` | `TopDirectoryOnly` | **73** | 71 / 2 | **0** |

All three reproduce exactly. `.Service` then comes out at **261 sites / 201 untimed** (130 / 131 by shape) — which agrees with the census comment on the issue to the digit — of which these **125** are group A. `Mcp/` alone: **118 sites, all 118 untimed**, 4 / 114 by shape. After this change `Mcp/` is 118 / **0** and the project is 201 → **76** untimed, the remainder belonging to groups B–E.

## The derivation

**The enclosing constraint is a kind this sweep had not met.** Not a clock (#2871's `CancelAfter`) and not a connection permit (the viewer group's `MaxPoolSize = 10`), but a **server-side `statement_timeout` on the login role** — `ALTER ROLE mcp/viewer SET statement_timeout`, written by startup provisioning from `config_service.compose_statement_timeout_seconds`, shipped default 15 s, clamped 5–600. It is a **role** setting, so it bounds every statement those identities run and not merely the composed queries it is named for; `DarlingTrendReader` and `QueryStoreTrendRouting` both already reason about it by name.

### Why a client-side deadline does not simply defer to the GUC

Three reasons, and they land differently per deployment shape rather than being one argument restated.

1. **Bring-your-own PostgreSQL has no such GUC at all.** `Darling/tools/provision-roles.sql` creates `admin` and `viewer` and deliberately **no `mcp` role**, and sets `statement_timeout` on `viewer` only — hardcoded at 15 s, not read from the store, so a BYO operator's change to the column never reaches a role until they re-run the script by hand. The MCP host in that mode connects with the operator's own `postgres.connectionString`, typically the owner or `admin` identity, which carries no `statement_timeout`. So on a BYO store **this surface has no server-side bound**, and Npgsql's undocumented default is currently the only thing bounding it. A BYO operator also has no knob here at all, which is why these values are deliberately not aggressive.
2. **On a managed store the GUC is operator-tunable to 600 s, and it is per-role.** Raising `compose_statement_timeout_seconds` so one heavy custom view can finish raises it for **all 124 shipped reads**. The client-side deadline is what keeps the ordinary read bounded while the backstop is deliberately loosened for the extraordinary one.
3. **A role `SET` takes effect on that role's next session.** `DarlingConfig` says so outright — "this is not a kill switch" — and `DarlingManagedRoles` logs that an already-connected client keeps the old ceiling until it reconnects. The MCP host builds its pool once at host start. A `CommandTimeout` is set per command at construction, so it binds the very next command on an **already-open** pooled connection, which is exactly what the GUC cannot do.

### The asymmetry that makes it necessary rather than tidy

**None of the 136 `[McpServerTool]` methods takes a `CancellationToken`.** Every reader declares `CancellationToken cancellationToken = default`; every tool calls it without one; the single tools file that names a token passes `CancellationToken.None` explicitly (`DarlingMcpCustomViewTools` into `DarlingWebEndpoints.RunComposedPanelAsync`). So an agent that abandons a call leaves the query running **and the pooled backend held** — nothing on this path propagates abandonment. The web half is the mirror image and threads `HttpContext.RequestAborted` throughout, so a disconnecting browser does cancel; the deadline is a backstop there and the only bound on the MCP side.

### The permit, at its real strength rather than borrowed

The MCP host builds its **own** data source, separate from the web host's and from the worker's collection pool — three independent `NpgsqlDataSource.Create` calls on connection strings that differ by `Username` (`mcp`, `viewer`, and the bootstrap owner). Verified in source, not assumed, and it cuts the argument down rather than up: **a stalled MCP read cannot starve fleet-wide collection through the pool.** The widest fan-out here is a `Task.WhenAll` over **two** reads against `MaxPoolSize = 24` — nothing like the viewer's ten-against-ten, and there is no `CancelAfter`, `SemaphoreSlim`, `WaitAsync` or `CreateLinkedTokenSource` anywhere on the path. What the bound still buys is the reason it exists at all: every pooled connection is a live `postgres.exe` **process**, each spawn re-reserves the shared memory region, and this store logged 8 `could not reserve shared memory region` retries on 2026-09-03 — so a longer hold means more spawns, not just a busier pool.

### Two regimes, not one

**`McpCommandDeadlines.ReadSeconds = 20`** for the 124 shipped-shape reads.

*Floor* — the managed store's own ceiling, 15 s, **and it fires on this surface in production today.** On 2026-09-04 `get_query_store_duration_trend` over a 7-day window returned `57014: canceling statement due to statement timeout` against the live 42-server store, with #2736's corrected-rollup routing already engaged, while the same read at its shipped 24 h default returned its full series. Sitting **above** 15 s is what keeps the server-side timeout the one that fires on a managed store, because `57014` **names** its cause where Npgsql's own deadline renders as `Exception while reading from stream` and gets read as a network fault — the misdiagnosis #2826 exists to prevent. That floor dominates the measured one by more than an order of magnitude; this family's verified reads are sub-second.

*Ceiling* — the 30 s it replaces, which the viewer group's standard requires it to sit **meaningfully** under rather than at, plus a **relational** bound against `StorageCommandDeadlines.McpReadSeconds`: the `DarlingPg*Reader` family serves these same 136 tools from `.Storage` on the same pool, so one tool call must never get a looser ceiling depending on which project happens to hold the reader.

**The composed-query runner could not take that constant, and that is the split.** It is the one site running **operator-authored** SQL, and `compose_statement_timeout_seconds` is the operator's declaration of how long that class may run — a fixed 20 s would cut off the panel the knob was raised to 120 s for, a regression dressed as a fix.

So it takes no constant at all. `McpCommandDeadlines.ResolveComposedQuerySecondsAsync` **reads** `config.config_service.compose_statement_timeout_seconds` and clamps it through the store's own `ClampComposeStatementTimeoutSeconds`, so the client-side deadline is the operator's *actual* setting rather than the ceiling they could theoretically have chosen.

**Read rather than plumbed, and that is the load-bearing choice.** The store is authoritative for this knob and #2918 made it hot-swappable — a control-plane reload re-asserts it onto the roles with no restart. So a value captured at host start is simply wrong after the first change, and the value a host *could* plumb is the file-loaded `DarlingConfig.ComposeStatementTimeoutSeconds`, which is only the **seed**: the store wins afterwards. Reading per panel run has no cache to go stale, and it is also what keeps the two callers identical — `MapAll` has the host's config in scope while `Mcp/DarlingMcpCustomViewTools.cs:252` has only the data source, so plumbing would have handed one surface the file value and the other a store read, which is exactly the divergence the single shared runner exists to prevent.

**The near miss is three lines away in the same method.** `ComposeStoreAvailability.GetRollupsAsync` is deliberately cached *per data source*, because which rollups a store has changes only when someone migrates it. Caching the deadline the same way compiles, passes every band assertion, and pins it to whatever the knob said the first time a panel ran. Resolved once per **run** — so a panel with annotation overlays pays one read rather than one per source, and every query in one run shares one ceiling, matching the role `statement_timeout` they all run under.

`ComposedQueryFallbackSeconds` is **15**, the same shipped default `DarlingConfig`, `ComposeSpec` and the clamp's own non-positive arm all use, so a store that cannot answer behaves like a store nobody has tuned. Failing *open* to something larger would grant more time precisely when the store is least able to answer for itself — the wrong direction for the only bound a BYO store gets.

## Store vs monitored target — the claim no landed pin needed

`.Storage` and `.Viewer` are Npgsql-**store-only** projects, so "every command sets a deadline" was the whole invariant there. `.Service` also builds commands against **monitored targets**, where a store-derived constant would be wrong in a way nothing else would catch: the value would look deliberate while bounding the wrong hop.

Every one of the 125 sites in this scope was classified, and **all 125 are store commands** — 115 `postgres.CreateCommand(`, 6 `_dataSource.CreateCommand(`, and 4 `new NpgsqlCommand(sql, connection)` whose connection comes from `postgres.OpenConnectionAsync`. There are **no** monitored-target commands here: `Mcp/` contains no `SqlConnection`, no `SqlCommand`, no `TargetProviders.For` and no `new NpgsqlConnection(`. A **receiver allowlist** writes that classification down, with the two real target shapes pinned as negatives — `ITargetProvider.CreateCommand(plan, connection, commandTimeoutSeconds)`, where the deadline is threaded and is stronger than a constant, and the transaction-scoped HypoPG form — so a future target command here fails asking for a decision instead of inheriting a store bound.

## The pin

`McpReadCommandTimeoutTests`, **scoped to `Mcp/` recursively plus `CustomViewStore.cs` and `DarlingWebEndpoints.cs` by name** — deliberately not the project. `.Service` carries roughly fifteen budget regimes over ~200 untimed sites across five groups; a project-wide glob would claim regimes this pin has not derived and would collide with the sibling groups' pins. `Mcp/` is globbed because the whole directory *is* this surface; the two top-level files are named because the project root is shared.

It covers all four known shapes plus one this sweep had not recorded, and asserts **zero** for the two that should be absent — the bare method group, and `BeginBinaryImportAsync`, whose `NpgsqlBinaryImporter.Timeout` is initialised from the connection's `CommandTimeout` and which no landed pin's regex can see. The project's four live importers are the collection sweep's; asserting their absence *here* makes "this scope has none" a checked fact rather than an assumption. Both negatives carry positive controls through the identical form, so neither can pass by matching nothing.

**Every scan reads the stripped text, including the construction scan** — which the landed pins run over raw source, where a prose mention of `.CreateCommand(` becomes a phantom untimed site at a line number no edit could fix. Not hypothetical in kind: `Mcp/DarlingMcpStoreMetricsTools.cs` contains the sentence "the two readers above go through postgres.CreateCommand and leave nothing open", and it misses both regexes only because what follows the name is a space and a word rather than a delimiter. The divergence changes no count (125 either way) and closes a false-positive direction.

### A fifth construction shape, found because a red-first variant came back green

The planted monitored-target offender used `new Npgsql.NpgsqlCommand(` and the pin **did not see it**: no regex in this sweep matches a **qualified** type name. The form occurs exactly once in the repo — `DarlingWorker.cs`'s `pg_stat_statements` text fetch — and that site is a monitored-target command **already carrying its own 60 s**, so the blind spot has no defect behind it and the untimed census is unchanged. The **site** census for `.Service` is **262**, not 261. Both the construction scan and the receiver allowlist now accept the qualified spelling, with the qualified *store* form pinned as accepted so the widening cannot create a false positive.

### Proven red fifteen ways, one mutation at a time

Each verified applied by **reading the file back** rather than by a diffstat, and each **rebuilt** — a mutation that does not compile proves nothing — and each restored from an in-memory byte copy rather than `git checkout`.

| # | mutation | assertion that failed |
|---|---|---|
| 1 | remove one deadline in the 26-site reader | `…SetsAnExplicitDeadline` — `1 MCP/web read command(s) inherit Npgsql's 30s default…: DarlingDataReader.cs:225` |
| 2 | remove the deadline inside a block-form `using` body | `…SetsAnExplicitDeadline` |
| 3 | remove the `CustomViewStore` deadline | `…SetsAnExplicitDeadline` |
| 4 | untimed command planted in an `Mcp/` **subdirectory** | `…SetsAnExplicitDeadline` (recursion control) |
| 5 | **timed** monitored-target command planted in `Mcp/`, unqualified | `…IsBuiltAgainstTheStore_NotAMonitoredTarget` — `1 command(s) … have an unrecognised receiver…: DarlingTargetProbeReader.cs:7` |
| 5b | the same, **fully qualified** | `…IsBuiltAgainstTheStore_NotAMonitoredTarget` |
| 6 | bare `.CreateCommand` method-group hand-off planted | `NoCommandInScope_IsCreatedByABareMethodGroupHandoff` |
| 7 | untimed `BeginBinaryImportAsync` planted | `NoCommandInScope_UsesAnUntimedBinaryCopyImporter` |
| 8 | `ReadSeconds` → 15 | band: *"at or under the mcp role's 15s default statement_timeout, so on a managed store it would fire FIRST and replace a 57014 that names its cause…"* |
| 9 | `ReadSeconds` → 5 | band: *"at or under the 5s floor an operator can clamp the role's own statement_timeout to…"* |
| 10 | `ReadSeconds` → 30 | band: *"not meaningfully under the inherited Npgsql default it replaces…"* |
| 11 | `StorageCommandDeadlines.McpReadSeconds` → 15 | the **relational** ceiling, isolated from the numeric one — the two coincide at 30 today, so only this mutation can exercise it |
| 12 | `ComposedQuerySeconds` decoupled from the clamp maximum | `TheComposedQueryDeadline_TracksTheOperatorsDeclaredMaximum` — that constant and that fact are both **gone** as of the follow-on below; the mutation is retired and its successors are S3–S6 |
| 13 | receiver allowlist narrowed so a real store shape is rejected | `TheReceiverAllowlist_…` (positive control on the allowlist itself) |
| 15 | the receiver allowlist's word-boundary lookbehind removed | `TheReceiverAllowlist_…` on `notpostgres.CreateCommand(` — without the boundary, `targetPostgres.CreateCommand(` matches `postgres\.CreateCommand` as a substring, a false **accept** in the one guard meant to stop a target command taking a store deadline |
| 14 | **move** `CustomViewStore.cs` | the named-file existence check — a count-based control cannot see a relocation, every occurrence count staying 1 → 1 |

### The follow-on: eight more, and two of them are why the live test exists

| # | mutation | assertion that failed |
|---|---|---|
| S1 | **staleness** — memoise the resolved value in a `static` | `TheResolver_ReadsTheConfiguredValue_AndSeesItChange` |
| S2 | **staleness** — memoise per data source, the `GetRollupsAsync` shape | same |
| S3 | drop the clamp and trust the column | `TheResolver_ClampsAnOutOfRangeRow` — `Expected: 600 / Actual: 6000` |
| S4 | fallback 15 → 600 (fail **open**) | `TheComposedQueryFallback_IsAValueAnOperatorCouldHaveConfigured` |
| S5 | fallback 15 → 0 (`CommandTimeout = 0` means **no limit**) | same — `Assert.InRange() … Range: (5 - 600) / Actual: 0` |
| S6 | revert the runner to a compile-time constant | `TheComposedQueryRunner_ResolvesItsDeadlinePerRun` |
| S7 | strip the deadline from the resolver's **own** read | `EveryMcpAndWebReadCommand_SetsAnExplicitDeadline` — `McpCommandDeadlines.cs:186` |
| S8 | **move** the resolve out of the shared runner into the per-query method | `TheComposedQueryRunner_ResolvesItsDeadlinePerRun` — a MOVE, so `ResolveComposedQuerySecondsAsync(postgres` stays at 1 → 1 and only the structural ordering control can see it |

**S1 and S2 first came back GREEN, and that was the mutation's fault rather than the pin's.** The planted cache was read but never populated, so both compiled and changed no behaviour. A mutation that compiles but does not move the behaviour proves exactly as little as one that does not compile — "assert the mutation applied" has to mean the behaviour moved, not that the bytes did. Once the cache was actually written, both go red on the change-seeing assertion.

`TheResolver_FallsBackOnAMigratedButUnseededStore` exists because running the live test found it: `MigrateAsync` creates `config_service`, but the **row** comes from `StoreConfigProvider.SeedIfEmptyAsync`, which is a separate act. So "brand-new store, first panel run" reads no rows at all, and the resolver has to land on its fallback rather than hand a zero to `CommandTimeout`. There is a companion for a store with no `config_service` table at all (a pre-V78 store), which raises `42P01`.

The live class **mints its own scratch database** (`ScratchPostgres`, with the `#1776 own-store` marker) rather than joining `[Collection("live-postgres")]`: every test in it mutates the knob and one asserts the behaviour of an *unseeded* store, so it cannot share the collection's long-lived database with sixty other classes.

**The three landed #2874 pins still pass and still assert what they claim.** `StorageCommandTimeoutTests` (6), `AnalysisPassCommandTimeoutTests` (13) and `AlertPassCommandTimeoutTests` (5) all execute green against this tree, scopes unchanged — none of them globs `.Service`, and the alert-pass pin names six top-level files, none of which this change touches. The viewer pin cannot be compiled on macOS (its project is `net10.0-windows` WPF), but its scope directory is untouched by this diff and the identical scan reproduces `.Viewer` at 193 / 0, which is what it asserts.

## What this does NOT do

- **No user-reported failure is closed by it.** The 7-day Query Store window measured above is bounded by the **role**, not by this change, and stays bounded by it.
- **No timing was taken with a client-side deadline in place**, and the production `compose_statement_timeout_seconds` was **not read back** — the 15 s used throughout is the shipped default and the clamp, verified in source rather than on the box. The production evidence is the `57014` above plus successful live executions of `get_fleet_overview` (the 8-way `DarlingFleetReader` fan-out over 42 servers), the server-registry read, and the 24 h trend.
- A local **PostgreSQL 17 / TimescaleDB 2.29.2** store *was* stood up for the follow-on, built by the product's own `PgMigrations.MigrateAsync`. It was used to prove the resolver's **behaviour** — that it sees a changed value, clamps an out-of-range row, and falls back on both a migrated-but-unseeded store and a store with no `config_service` table — **not** to measure any read's duration. It holds no collected data, so nothing here is a performance measurement.
- `Darling.Tests` is `net10.0-windows` and cannot run on macOS. The 68 tests reported here ran in a throwaway `net10.0` xunit host compiling the **real** source files by absolute path with `AssemblyName=Darling.Tests` — 68/68 green with `DARLING_TEST_PG` pointed at that store, 61/68 with 7 skipped without it. **CI is the arbiter.** Expected delta: **+47** — `McpReadCommandTimeoutTests` 40 plus `McpComposedQueryDeadlineLivePostgresTests` 7, counted per class rather than derived.
- The host has to be **staged inside the repo tree** to exercise `LivePostgresCollectionHygieneTests`, which resolves its scan directory from `AppContext.BaseDirectory`: a host built in a scratch directory fails that rule for an unrelated reason and tells you nothing about the class under test. That rule reads only the **25 lines above** a class declaration when looking for `[Collection("live-postgres")]` or an `#1776 own-store` exemption, so the marker sits immediately above the declaration rather than above the doc comment — the placement the three existing own-store classes all use, and the marker says so.
- No schema change, no migration rung, no version bump.


erikdarlingdata:
Thanks — the independent census reproduction is the part I most wanted checked, and the 1:1 coverage count matches mine.

**On the naming nit: acted, and the count was low.** There are **six** write sites among the 124, not three — the three you found in `CustomViewStore` (`CreateAsync`/`UpdateAsync`/`DeleteAsync`) plus three in `Mcp/` that a `CustomViewStore`-scoped look would not see:

- `Mcp/DarlingMcpAlertTools.cs:367` — `UPDATE config_alert_settings` (the alert-tuning tool)
- `Mcp/DarlingMcpServerAdminTools.cs:219` — `DELETE FROM config_monitored_servers` (`remove_server`)
- `Mcp/DarlingMcpServerAdminTools.cs:570` — `INSERT` into the same table (`add_servers`)

So 119 reads + 6 writes = 124.

**Kept as one constant, and documented rather than renamed**, for two reasons. They share the *regime* exactly — the same least-privilege pool, the same absent `CancellationToken`, the same role `statement_timeout`, no enclosing budget — and this sweep groups by regime rather than by verb; and every one is a single-row primary-key statement on a `config` table, so it is strictly cheaper than the reads the bound was floored on. Renaming would also break the relational pin against `StorageCommandDeadlines.McpReadSeconds`, whose name it deliberately follows (as does the viewer's `InteractiveReadSeconds`, which covers that surface's alert-dismiss and mute writes for the same reason). The doc comment on `ReadSeconds` now says "read" names the SURFACE, not the verb, and enumerates all six; the CHANGELOG entry says the same.

**Two things landed after your pass**, both worth a re-read:

1. **A fifth construction shape**, found because a red-first variant of the new pin came back GREEN: a **qualified** type name, `new Npgsql.NpgsqlCommand(`, which no regex in this sweep matches. It occurs exactly once in the repo — `DarlingWorker.cs`'s `pg_stat_statements` text fetch — and that site is a monitored-target command already carrying its own 60 s, so the blind spot has no defect behind it and the untimed census does not move; the *site* census for `.Service` is 262 rather than 261. Both the construction scan and the receiver allowlist now accept it, with the qualified *store* form pinned as accepted so the widening cannot create a false positive.

2. **The composed-query deadline no longer defers to the clamp maximum.** You noted it "still defers to the operator's own clamp-derived maximum (600s) rather than a hardcoded value" — that was the deliberate compromise, and it is now closed. `McpCommandDeadlines.ResolveComposedQuerySecondsAsync` reads `config.config_service.compose_statement_timeout_seconds` per panel run and clamps it, so the deadline is the operator's *actual* setting. Read rather than plumbed because the knob is hot-swappable (#2918) and the value a host could plumb is the file-loaded seed, not the live one — and the near miss is three lines away, where `ComposeStoreAvailability.GetRollupsAsync` is deliberately cached per data source. A live-Postgres class changes the column and requires the next resolve to see it, which is the only thing that can tell a per-run read from a cached one.

Not merging, and #2874 stays open — groups B–E remain.

PR #3123, opened 2026-09-07
THROWAWAY: exercise the Darling gate's skip arm with a markdown-only diff

erikdarlingdata:
Validation artifact for #3122, not for merge.

Based on `L3116-pg-gate-shared-libs` rather than `dev`, so the diff against this pull request's base is one added `Darling/**/*.md` file and the gate being exercised is the shared gate that branch introduces. #3122's own change touches `build.yml`, which every filter in that file names, so it fires all of them and cannot demonstrate a skip.

What to read here: `Darling PostgreSQL tests` must report success with `Run Darling PG tests` **skipped**, and its notice must say how many files changed and name them rather than assert a cause.

Closed and its branch deleted once that step list has been read.

### DIFF
--- a/Darling/docs/L3116-skip-arm-probe.md
+++ b/Darling/docs/L3116-skip-arm-probe.md
@@ -1,0 +1,8 @@
+# Skip-arm probe (throwaway)
+
+Validation artifact for the acceptance criteria of #3116: a pull request whose only change is a file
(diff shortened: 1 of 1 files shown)


erikdarlingdata:
Closing unrun: build.yml's `on.pull_request.branches` is `[main, dev]`, so a pull request into a feature branch never dispatches it. Re-staged on a base branch that widens that trigger, so the skip arm can actually be observed.

PR #3124, opened 2026-09-07
THROWAWAY: markdown-only diff to observe the Darling gate's skip arm

erikdarlingdata:
Validation artifact for #3122, not for merge. Both branches are deleted once the step list has been read.

The diff against this pull request's base is one added `Darling/**/*.md` file and nothing else, and the base already carries #3122's shared gate — so the gate being exercised here is the new one. #3122 cannot demonstrate this arm itself: its change touches `build.yml`, which every filter in that file names, so it fires all of them.

The base branch also carries one throwaway commit widening `on.pull_request.branches` to include itself. Without it `build.yml` never dispatches for a pull request into a feature branch, which is why the first attempt at this (#3123) registered only `check-branches`. That commit is on the throwaway base only, never on #3122.

What to read: `Darling PostgreSQL tests` reports success with `Run Darling PG tests` **skipped**, and its notice says how many files changed and names them instead of asserting a cause.

### DIFF
--- a/Darling/docs/L3116-skip-arm-probe.md
+++ b/Darling/docs/L3116-skip-arm-probe.md
@@ -1,0 +1,7 @@
+# Skip-arm probe (throwaway)
+
+Validation artifact for the acceptance criteria of #3116: a pull request whose only change matches
+`Darling/**/*.md` must still leave `Darling PostgreSQL tests` skipping `Run Darling PG tests`.
+
+Branched off the fix so the gate being exercised is the shared gate at
+`.github/darling-paths-filter.yml`. Deleted once the job's step list has been read.


erikdarlingdata:
Closing WITHOUT MERGING — validation artefact for #3122, never intended to land.

What it recorded: on a diff of one added `Darling/**/*.md` file against a base carrying #3122's shared gate, `Darling PostgreSQL tests` reported success in 16s with steps 5-14 `skipped`, `Run Darling PG tests` among them. `Filter darling = false`, `Matching files: none`, and the notice read `None of the 1 changed files are in the Darling gate (.github/darling-paths-filter.yml), so the TimescaleDB suite did not run on this commit` followed by the file name. `Darling Linux build` skipped its steps 5-14 likewise.

This is the arm that catches a fix which widens the gate to always-true instead of correcting it.

Issue #3140, opened 2026-09-07
ParsedProjectXmlPaths treats an apostrophe in an XML comment as an attribute delimiter, so the crude second-opinion scan silently under-reads — and its doc claims twice that it cannot


erikdarlingdata:
`CrossAppGuardCiGateTests.ParsedProjectXmlPaths` is the deliberately-crude "second opinion" text scan that `TheEvaluatedSet_ContainsEverythingAParseOfTheSameXmlFinds` compares MSBuild's evaluated item set against. Its regex (line 2037):

```
"(?<dq>[^"]*)"|'(?<sq>[^']*)'|>(?<text>[^<>]*)<
```

**The single-quote alternation treats an apostrophe in an XML comment as an attribute delimiter.** An unbalanced pair swallows every quoted attribute value between it and the next apostrophe, so the scan **silently returns fewer paths than the file contains.**

## The doc says this cannot happen, in two places

Line 2028: *"Crude in both directions, and both are the safe direction for a lower bound. It over-reads — a path inside a comment, or behind a `Condition` that never holds, counts — and it under-reads anything needing evaluation."*

Line 823 makes the same claim again: *"over-reads (a path inside a comment counts) and under-reads (anything needing evaluation…)"*

**Under-reading arbitrary quoted values is not the safe direction**, and it is not "anything needing evaluation" — it is a literal `Include=` that the scan simply cannot see. Both sites are wrong and both need correcting with whatever fix lands. **This is an unenforced claim: the doc asserts a property, nothing holds it, and the property is false.**

## It is already live on `dev`, and position is the invariant rather than parity

`Lite.Tests/Lite.Tests.csproj` on `dev` carries **three** apostrophes — an odd count, so they are already unbalanced. The alternation pairs the first with the second and consumes the span between them; the third finds no partner and is skipped. **So a region of that csproj is being swallowed on `dev` right now**, and the test is green only because no floor path happens to sit inside it.

That is worth measuring rather than reasoning about: **which attribute values are currently invisible to the scan on `dev`?**

Reproduced deterministically, not a flake: adding an ItemGroup with an XML comment containing five apostrophes **above** the ItemGroup holding `<Compile Include="..\Darling\Darling.Tests\CSharpSourceWalker.cs" … />` reds the test; removing the apostrophes greens it. **Parity of apostrophes is not the invariant — position relative to the paths is.**

## The failure reads as the opposite of what it is

```
[...]

Issue #3094, opened 2026-09-06
T-SQL convention guard labels offenders with a type name, a local variable, or a bare keyword instead of the enclosing member


erikdarlingdata:
`TsqlConventionGuardTests.EnclosingMember` builds the member half of every offender string the T-SQL convention guard prints:

```
offenders.Add($"{name}:{EnclosingMember(text, start)} [{rule}] {detail}");
```

It takes "the nearest declaration above the literal" from a two-alternative regex over **raw** source:

```
(?:const\s+string|static\s+string|string)\s+(?<name>[A-Za-z_][A-Za-z0-9_]*)\s*(?:=|=>)
|(?<name2>[A-Za-z_][A-Za-z0-9_]*)\s*\([^)]*\)\s*(?:=>|\{)
```

The second alternative asks for an identifier, a parenthesised anything, and an opening brace — which is a description of most C# statements, not of a member declaration. There is no containment test anywhere: whatever matched last before the offset wins, whether or not the literal is inside it.

## Measured, over the corpus the guard actually scans

Driving the shipped helper at the offset of every T-SQL literal the scan reads (160 literals across the three scanned trees):

**48 of 160 are labelled with a bare C# statement keyword or with nothing at all** — 29 `if`, 11 `catch`, 7 `foreach`, and 1 `<unknown>`. A further 3 are labelled `SqlConnectionStringBuilder`, which is a type being constructed and not a member of the file at all, and 7 more with a local variable (`query` x3, `sql` x2, `optimizeForSequentialKey`, `replicaGroupKey`).

The five sites worth naming individually, each reproduced before anything was changed:

| file | reports | should be | why it went wrong |
|---|---|---|---|
| `Darling/PerformanceMonitor.Darling.Analysis/PgPlanFetcher.cs:184` | `SqlConnectionStringBuilder` | `FetchPlanBySqlHandleAsync` | the nearest match was `new SqlConnectionStringBuilder(connectionString) {` — a **type being constructed**, so the label names something that is not a member of this file |
| `Lite/Services/LocalDataService.FinOps.Recommendations.cs:34` | `sql` | `GetAgReplicaRoleAsync` | the literal initialises `const string sql`, so the label named the **local it is assigned to**. Close enough to read as correct, and wrong the moment a method holds two queries — which three other sites in this same file are |
[...]

Issue #3088, opened 2026-09-06
PgPanelTabOwnershipTests reads block-comment prose as code, and counts braces inside literals


erikdarlingdata:
`PgPanelTabOwnershipTests` carries both of the defects #3064 just fixed in its sibling `PgRegistryPanelPlacementTests`. Neither is active on today's tree — both are latent, and the second one is latent in the direction that passes.

## 1. `Strip` leaves `/* … */` blocks entirely

`PgPanelTabOwnershipTests.Strip` (line 231) is a hand-rolled two-pass regex:

```csharp
private static string Strip(string source) =>
    Regex.Replace(Regex.Replace(source, @"^[ \t]*///.*$", string.Empty, RegexOptions.Multiline),
                  @"//.*$", string.Empty, RegexOptions.Multiline);
```

It removes `///` lines and `//` to end of line and never looks for a block comment. This repo puts no asterisk on a block comment's continuation lines, so the `//.*$` pass takes only the opening `/*` line and hands **every continuation line to the scanner as code** — the same defect #3052 was filed for and #3059 adopted the walker for in three more files.

`ViewerServerTab.Postgres.cs`, the file this test reads, holds **14 multi-line block comments spanning 35 continuation lines**. None of them names a control today, which is why nothing is red — but the scanner it feeds is `ControlAssignment`, `\b(?<name>Pg[A-Za-z]+(?:Grid|Note|Expander))\s*(?:\.\w+)?\s*=`, and a comment explaining why a grid is filled on *another* tab is written in exactly that shape. Measured, with this inserted into `LoadPgIoAsync`:

```
/* Wraparound is not filled from here even though the same reader carries it. Vacuum owns
   it, and PgWraparoundGrid.ItemsSource = that tab's rows on that tab's load path. */
```

`EveryPanelAPgTabLoadPathFills_IsDeclaredInsideThatTab` fails with `PgWraparoundGrid is filled by PgIoTab's load path but is declared inside PgVacuumTab`. The panel it names does not exist on that load path; the test read the sentence that said so.

The two directions are both bad and the quieter one is worse. A comment naming an off-tab control is a **false positive** — the rule reports a defect that is not there. A comment naming a control on the *same* tab is absorbed silently as a phantom panel, and a phantom is what satisfies the `panels.Count > 0` floor that is supposed to notice a load path assigning nothing.

## 2. `MethodBodies` counts raw `{`/`}` with no awareness of literals

[...]

Issue #3083, opened 2026-09-06
Nothing validates <see cref> targets: 22 dangling today, and the natural fix is the GenerateDocumentationFile decision parked in #3025


erikdarlingdata:
Doc comments across the tree carry `<see cref="…"/>` references, and **nothing validates that any of them point at anything.** There are **22 dangling targets today**, unenforced because `GenerateDocumentationFile` is off — the compiler never resolves a cref, so a rename or deletion silently leaves a reference aimed at nothing.

This is the same *shape* as #3081 (a stated convention with no mechanism) but a different surface, mechanism and fix. It is filed separately rather than folded in because its natural remedy already has a history and an owner.

## Why it matters more than a typo

A dangling cref is not a formatting defect — it is a **doc claim pointing at nothing**, and this repo leans on doc comments for load-bearing statements far more than most. Tonight alone produced four instances of a comment asserting something that had stopped being true:

- a parity claim about two cadences that were never equal (#3047)
- a panel comment naming a tab the panel was not on (#3048)
- a clamp justification citing "~7% of cadence" against a real worst of 369% (#3060)
- a paragraph justifying ASCII-only regexes via a `\S` stand-in **that no pattern used**, because the mechanism had been replaced earlier in the same PR

A cref is the one part of a doc comment that is *mechanically checkable*, and it is the part currently unchecked. By the criterion that came out of #3069 — **a comment is worth pinning when it restates something derivable, and worth nothing but review when it explains why** — a cref is maximally derivable: it names a symbol that either exists or does not.

## The fix already has a parked decision, which is the reason not to fold this into #3081

**Enabling `GenerateDocumentationFile` is Erik's call, parked from #3025.** That PR deliberately did **not** enable it, because doing so surfaces every pre-existing doc-XML defect in the tree at once. So this issue is not "turn on a flag" — it is "decide what to do about the 22, then choose a mechanism".

Three options:

1. **Enable `GenerateDocumentationFile`** (the #3025 decision) and fix or suppress the pre-existing set. Highest fidelity — the compiler resolves crefs properly, including inherited and generic targets — and the largest one-time cost. Needs the 22 inventoried first, because the flag turns them all into warnings simultaneously.
[...]

PR #3150, opened 2026-09-07
Record the CHANGELOG entries for sixty-five merged changes

erikdarlingdata:
Appends the collected `[Unreleased]` entries for the 85 pull requests merged to `dev` from `8210012bc` (the previous batch, #2984) through #3147. Lanes do not edit `CHANGELOG.md` — every one of them appends to the same `[Unreleased]` block, so a per-PR edit conflicts with whichever sibling merges first — and the entry text rides each PR's own description instead. This is the periodic commit that applies it.

## What is in the diff

`CHANGELOG.md` only, 141 insertions and no deletions: 70 entries prepended inside `## [Unreleased]` — 10 under `### Added`, 13 under `### Changed`, 47 under `### Fixed` — plus the 71 link-reference definitions those entries need, appended to the block at the foot of the file. Entries are ordered newest-merge-first within each section, matching the file's existing order. Nothing already in the block was edited, reworded or reordered.

## Base

`origin/dev` moved twice while this was assembled. Both were merged in rather than rebased, so `33211034e` is intact in the history: #3149 landed during the survey and #3147 landed after it closed, arriving as the merged base. Both carry entry text and both are included, #3147's in a commit of its own on top: an entry omitted because it arrived mid-batch is how a backlog restarts, and the batch's purpose is to make `[Unreleased]` current at its own merged base. There is no PR in the range whose entry this diff knowingly lacks.

## Where the text came from

85 pull requests in the range. 66 contributed an entry, 19 did not.

Of the 66, 54 carried their entry text under a heading in their own body, one (#3033) carried it under a bare paragraph rather than a heading, and 11 carried it in the lane's handoff file rather than the PR body (#3040, #3042, #3046, #3075, #3077, #3078, #3079, #3080, #3084, #3085, #3086). The text is used close to verbatim; where a lane wrote a plain bullet, its own opening clause is bolded to match the shape every existing entry in the block uses, and a redundant `Fixed:` / `Added:` prefix is dropped where the section heading already says it. Three PRs contributed more than one entry because they fixed more than one thing: #3027 (two), #3056 (three, across `### Fixed` and `### Added`), #3117 (two).

Two entries needed a choice recorded here. #3077's lane offered a first draft and a post-review final; the final is used, because the draft cites a count of inadmissible readings the lane says a third reading falsified mid-review. #3079 and #3085 each offered a full entry and a one-liner; the full form is used.

Section placement follows the block's existing treatment of guard work rather than a rule invented here: a guard that never existed is `Added` (#3081, #3083, #3049's grid check), a guard whose reach was widened as policy is `Changed` (#3063, #3067, #3074, #3076, #3082 — the shape `[#2490]`'s entries already sit under `Changed`), and a guard that gave a wrong answer is `Fixed` (#3052, #3088, #3094, #3140 — the shape `[#2927]`'s entry already sits under `Fixed`).

## The 19 with no entry

Two say so themselves: #3043 ("a comment aligning with a naming convention is not a user-visible change and does not earn a line") and #3020 ("no behaviour change — every edit is inside a comment or a changelog line"; it is also the only PR in the range that touched `CHANGELOG.md`, a one-line placeholder scrub already on `dev`).

Seventeen have no entry text in the PR body, in the lane handoff files, or anywhere else searched, and are not covered by the existing block: #2982, #2983, #2985, #2986, #2987, #2988, #2989, #2990, #2992, #2998, #3000, #3001, #3002, #3003, #3005, #3007, #3025. All 17 merged between 22:44 on 09-04 and 10:34 on 09-05, before the entry-rides-the-description routing settled. Several are user-visible (#2985 takes the default trace's archival-empty cutoff from the server's clock, #3000 screens `rdsadmin` out of per-database enumeration, #3003 bounds `pg_index_bloat` by bytes, #3007 derives the viewer's interactive read deadline from its fan-out width). No entry is invented for them here — these entries cite measurements only the lane that did the work can vouch for — so they want a backfill in the shape of #2260.

## How the count was checked

Two ways that fail differently. `git diff` reports 70 added `- ` lines and 71 added link definitions, totalling the 141 in the diffstat; independently, re-parsing the written file and subtracting the per-section counts measured before any edit (96 / 15 / 143) gives a delta of exactly 10 / 13 / 47. Separately, every one of the 66 contributing PRs is asserted to have its issue reference present among the added lines, and every added line is asserted to be non-trivial and house-shaped rather than merely present.

The extractor was also run twice. A first pass that treated any heading as a section terminator lost #3149's entry completely, because that entry sits inside a fenced block whose first line is `### Fixed`; a fence-aware pass recovered it and agreed with the first pass on all 53 others; #3147's entry was extracted with that same fence-aware pass. `[Cc]hangelog` does not match `CHANGELOG`, and requiring `CHANGELOG` at a heading's start misses #3122 ("Proposed CHANGELOG entry"), #3146 ("Changelog entry for the coordinator") and #3117's `###`-level heading, so the search is case-insensitive, any level, and matches mid-line.

`CHANGELOG.md` is outside `FleetIdentifierScrubTests`' population by construction, so CI does not scan this text for identifiers. The diff was scanned by hand against the guard's own two patterns and for RDS endpoints, ARNs, 12-digit account ids, dotted-quad and IPv6 addresses, and the fleet's own naming: clean. The one address-shaped token, `2001:db8::1`, is the RFC 3849 documentation prefix in #3030's account of a log-prefix regex; `01:00:00` and `24:00:00` are PostgreSQL interval literals.


erikdarlingdata:

**Correction to the merged commit's own numerals, recorded here because the commit is immutable and this is the nearest durable surface to it.**

The squash subject on `dev` (`11a5e53ee`) reads *"sixty-five merged changes"*, and its body's first bullet carries the figures as they stood **before** #3147's entry was added: 69 entries, 10 Added / 13 Changed / **46** Fixed, 84 pull requests, 70 definitions.

The artefact is right; only those numerals are stale. What actually landed:

| | |
|---|---|
| entries | **70** (10 Added / 13 Changed / **47** Fixed) |
| contributing PRs | **66** |
| range surveyed | **85** PRs, `8210012bc`..`5d5b77cc8` |
| link-reference definitions | **71** |
| diff | 141 insertions, 0 deletions, `CHANGELOG.md` only |
| partition | 66 contributed + 2 no-entry-by-own-statement + 17 needs-a-human = 85, and `git log` over that range yields exactly 85 PR-numbered commits with 0 unnumbered |

The description above is correct on all of these; the second commit message in the squash body also records #3147's entry being added "with the one link-reference definition it needs", so 69+1 and 70+1 are reconstructable from the message itself.

**Deliberately not editing the title.** Retitling to "sixty-six" would make this page disagree with the commit it produced, and a title/commit divergence is worse for anyone tracing `11a5e53ee` back than a stale-but-consistent pair with an accurate correction adjacent to it. No CHANGELOG change is warranted either — this is a commit-subject numeral, not a claim inside the file.

**How it happened, since the mechanism generalises.** When #3147's entry was authorised the description was re-audited numeral by numeral against the artefact — 20 claims, all passing. But that audit's population was the PR body **and nothing else**. The PR title and the commit messages were never in it, and GitHub derives the squash subject from the title. So a stale figure sat two surfaces away from a thorough check that structurally could not fail on it.

That is the same defect class this batch's own entries keep describing: a check shaped like the answer, passing inside its population while identical figures go unverified outside it. The earlier catch in this PR — a description claiming "four PRs contributed more than one entry" while listing three — was luck of population rather than rigour, because that numeral happened to live in the body.

Issue #3166, opened 2026-09-08
The hourly refresh ceiling the compression phase grid is sized against is 594s, and the job has since run 702s, 815s and 871s


erikdarlingdata:
`TimescaleSupport.HeaviestHourlyRefreshObservedCeilingSeconds` is **594**, pinned by `RefreshCeilingProvenancePinTests` to the maximum of a published 16-run population (194 s–594 s, mean 352.5, median 345) for `policy_refresh_continuous_aggregate query_store_stats_interval_hourly` (job 1054).

**Three runs since then exceed it**, read from the store's own job telemetry:

| when | duration | vs the pinned ceiling |
|---|---|---|
| 2026-09-07 00:00 hour | 702 s | 1.18× |
| 2026-09-07 21:00 hour | 871 s | **1.47×** |
| 2026-09-08 00:00 hour | 815 s | 1.37× |

`get_store_metrics` currently reports this job at `last_run_duration_ms = 815116`, `duty_vs_cadence = 22.6%`, 375 runs, **2 failures** — the only policy on the store with failures other than `policy_telemetry`.

## This is the constant's own pre-registered trigger, not a new opinion

Its summary states the limitation and what would invalidate it:

> **What would change that answer** … or this maximum ceasing to be a lower bound on the truth. The second is the live limitation, and it is a property of the READ rather than of the estimator: the boundary day's runs are a census of that day after the boundary, while the days after it are a SAMPLE …

It also names the census read that settles it rather than leaving it implied: `timescaledb_information.job_history`, whose `succeeded` column this store has because `DarlingManagedPostgres.BuildConfAppend` enables `timescaledb.enable_job_execution_logging` (#1681) — read as every run since the boundary rather than sampled, giving count, median and maximum each side of it.

## The grid is NOT breached, and the margin is what changed

Stating this explicitly because it is the mistake to avoid: `RefreshPhaseSlotSeconds` is **900 s** (`RefreshPhaseStepMinutes = 15`), and 871 s **fits**. Nothing has overrun its slot.

What has changed is the clearance. The doc deliberately presents the bound and its margin as two numbers — `RefreshPhaseSlotSeconds` minus the ceiling — so:

| | seconds |
|---|---|
| slot | 900 |
| ceiling as pinned | 594 |
| clearance today | **306** |
| clearance if the population is re-derived at 871 | **29** |

[...]

Issue #3119, opened 2026-09-07
The hourly refresh envelope is stated at 306 s of slot margin; a measured run leaves 121.6 s and sits inside the warning band, and the daily series structurally cannot show it


erikdarlingdata:
`HeaviestHourlyRefreshObservedCeilingSeconds` documents its own invalidation condition — *"what invalidates it is this number approaching `RefreshPhaseStepMinutes` x 60"* — and that number has since been measured at **778.4 s**, which is **inside the warning band** and leaves **121.6 s of slot margin** where the envelope prose states 306.

**This is not a claim that the constant is stale.** Its doc deliberately CLOSES its population (*"the scope has to CLOSE the population, not merely date it: 'up to 04:20Z' is a window that has ended and will still be true next year"*), and that discipline is correct and worked. The 594 s figure remains true of the window it names. **What has moved is the operating envelope the constant is stated against, which the same doc frames as a condition rather than a property.**

## The measurement

`policy_refresh_continuous_aggregate query_store_stats_interval_hourly [1054]`, hourly self-metrics snapshots on 2026-09-06:

| snapshot | `last_run_duration_ms` | `total_runs` |
|---|---|---|
| 20:13:19 | 675,557 | 347 |
| **21:13:23** | **778,423** | **348** |
| 22:13:28 | 665,928 | 349 |
| 23:13:34 | 705,218 | 350 |

`total_runs` advancing +1 per hour establishes run identity from the counter rather than by inference.

| | seconds | margin against the 900 s slot |
|---|---|---|
| doc ceiling | 594 | **306 s** — "better than a third of the slot" |
| **measured maximum** | **778.4** | **121.6 s — 13.5 % of the slot** |
| `RefreshSlotWarningSeconds` | 750 | the measured maximum is **28.4 s INSIDE the warning band** |

The doc's own threshold language: *"A value at or past the slot width is asserted as a failure rather than accommodated."* 778.4 s is not past 900 s, so this is not that failure — but it is 86.5 % of the way there, and `ClassifyRefreshSlotHeadroom` bands it Warning.

## The part worth fixing regardless of the number: the surface cannot see maxima

`get_store_metrics`' daily series takes **that day's LAST snapshot, not its maximum** (`DISTINCT ON ... metric_time DESC`). For 2026-09-06 that is the **23:13:34** reading — **705.2 s**.

| what you consult | 09-06 reading | apparent margin |
|---|---|---|
[...]

Issue #3107, opened 2026-09-06
RefreshSlotWarningSeconds rejects a 480s alternative on 'three of the five' readings — only one of the five listed readings reaches 480s, and that count is the sole stated reason


erikdarlingdata:
`RefreshSlotWarningSeconds`' doc comment rejects an alternative warning line on a count that does not hold, and that count is the **only** stated reason for the rejection.

## The claim

> A tempting alternative — slot less one `CompressionPhaseGuardMinutes` band, 480 s — was rejected on those same readings: **it would have fired on three of the five.**

The five readings named in the paragraph directly above are **335, 594, 465, 359, 293** s.

At or above 480 s: **594 only.** One of five, not three.

`465` is the near miss and presumably the source of the error, but it is 15 s below the line. Nothing else is close: the remaining three are 293-359 s, all more than 120 s below.

## Why this is more than a stray number

The paragraph's argument for the chosen line has two legs, and this is one of them:

1. **It must clear the routine band** so it "cannot cry wolf on load this store has already carried." A line at 480 firing on **three of five** routine readings would indeed be too noisy. **Firing on one of five is a materially different claim**, and 20% on a band the doc itself calls routine is not obviously disqualifying.
2. It must leave usable lead time — and note this cuts the *other* way: a **lower** line warns **earlier**, so 480 s would give more lead time, not less. The paragraph's lead-time argument is made for the higher line and does not by itself exclude the lower one.

So the arithmetic error is not decorative: **it is the sole stated basis for rejecting the alternative**, and with the correct count that basis is much weaker than the prose implies.

## What survives, stated so this does not read as an attack on the value

The chosen line has an **independent** justification in the next paragraph, and it is a good one: 750 s sits **26 % above** the largest run the narrowed regime has produced, so a crossing means *"this job doing something its own record has no instance of"* rather than *"approaching a known ceiling."* That property is real, is separately pinned as a relationship rather than as two numbers, and does not depend on the disputed count at all.

**So this is very likely a correct decision resting partly on an incorrect reason**, not a wrong constant. The fix is to the reasoning, not necessarily to the value.

## Suggested resolution

1. **Correct the count** to what the listed readings support, and restate the cry-wolf argument at its true strength — or drop that leg and let the 26 %-above-ceiling property carry the decision, which it can.
2. **If 480 s is still rejected, say why on the corrected figures.** With one-of-five rather than three-of-five, the reader is owed the actual reason.
[...]

Issue #3044, opened 2026-09-05
Watch the heaviest hourly refresh's runtime against the 900-second slot it has to fit in


erikdarlingdata:
#3041 pinned the compression policies to a phase grid whose sizing rests on an explicitly stated operating envelope, asserted at build time as:

```
HeaviestHourlyRefreshObservedCeilingSeconds < RefreshPhaseStepMinutes * 60
```

That is the right shape — a conditional downgrade written only in prose reads as unconditional to whoever finds it next. But the assertion is a **build-time bound against a recorded ceiling**, and the thing it bounds is a **runtime that moves with data volume**. Nothing today watches the live figure approach the bound; the assertion fires only if someone changes a constant, not if reality changes underneath the constant.

## The measurement

The heaviest hourly continuous-aggregate refresh, against the 900-second slot width the grid divides the hour into:

| reading | seconds | % of the 900 s slot |
|---|---|---|
| pre-narrowing band | 3,301–6,330 | (window was 7 days) |
| ceiling under the 1-day window | 864 | 96% |
| #3041's own recent snapshot | 194 / 225 / 335 | 22–37% |
| **observed since** | **593** | **66%** |
| settled after | 465 | 52% |

The 194–335 s readings are what #3041's narrative was written against, and they read as comfortable. The 593 s reading was taken after that snapshot, and 864 s was already recorded earlier under the same window — so the honest characterisation is not "comfortable, occasionally spiking" but **a runtime that has been observed anywhere between 22% and 96% of its slot, trending with volume rather than settling.**

## Why this is its own issue rather than a note on #3035

#3041's grid survives the heaviest refresh running long, and by a wide margin: it would have to exceed **1,320 s** to reach the earliest compression minute and 3,540 s to reach the latest. That is not the exposure here.

The exposure is the **slot** itself. At 900 s the refresh runs into its own neighbour in the refresh family, which is #3024's grid, not #3041's. Past that point the answer is not a different phase for compression — it is that a 15-minute grid can no longer hold thirteen hourly refreshes, and the fix is a redesign (fewer slots, a longer cadence for the heavy aggregate, or splitting it) rather than a renumbering. Filing it against #3035 would attach a refresh-family capacity question to a compression-phase change that is already merged and correct.

## What would make this observable

[...]

PR #3168, opened 2026-09-08
Re-derive HeaviestHourlyRefreshObservedCeilingSeconds from a job_history census: 594 s becomes 896 s, and the watch-line ordering inverts

erikdarlingdata:
Closes #3166.

`HeaviestHourlyRefreshObservedCeilingSeconds` was **594 s**, the maximum of a 16-run record the comment itself described as half census and half sample. The comment also named the read that would settle that: `timescaledb_information.job_history`, every run since the boundary rather than sampled. That read has been done, and the constant is now **896 s**.

## The census

One row per run for job `policy_refresh_continuous_aggregate` / `query_store_stats_interval_hourly` on the store that carries this workload, read at `2026-09-08 01:37Z` over `timescaledb_information.job_history` (TimescaleDB 2.28.1, PostgreSQL 18.4, `timescaledb.enable_job_execution_logging = on`). Boundary `2026-09-05 13:44:23`, exclusion positional as recorded.

| side of the boundary | runs | median | maximum |
|---|---|---|---|
| at or before | 304 | 1081.7 s | 13300.7 s |
| after | **57** | **418.3 s** | **896.1 s** |

Zero post-boundary runs failed and none is missing a finish time, so the `succeeded` filter removes nothing from the span — the census is the whole of it rather than a status-selected part.

The read reproduces the published record exactly where they overlap, which is what says the method is the same one: the nine boundary-day values (`194, 222, 225, 335, 594, 465, 359, 293, 355`) and the seven "days after" values (`219, 225, 299, 376, 418, 546, 515`) come back digit for digit, and the sampled seven turn out to be seven consecutive runs of 2026-09-06 with fifteen more either side of them that the sample never saw. #3119's separately-read closed day (2026-09-06, 22 runs, maximum 778.4 s) also reproduces.

The population is republished in full — 9 + 48 — so the estimator is still recomputable rather than trusted. Range 194–896 s, total 27799 s, median 418 s.

## The estimator is unchanged and the decision was re-taken rather than inherited

The doc's reason for the maximum had two halves, and one was written to expire. At 16 readings the 95th percentile by nearest rank *was* the maximum, so a percentile bought no headroom. At 57 it is **830 s**, 66 s below the maximum, and the 90th is **786 s**, 110 s below. `RefreshCeilingProvenancePinTests` went red on exactly that clause, which is what it was for. The decision comes out the same way on the half that never referenced the sample size — a scheduling exclusion may accept no exceedance over its own record, and a 95th percentile is a promise to be wrong three times in sixty runs — so the estimator stays the maximum and the expiring clause is gone with the argument it guarded. There is no replacement expiry, because a reason independent of `n` has nothing left to expire; both percentiles and both gaps are now pinned as readings that track the population.

## Three arguments in the old derivation expired with the sample, and none was re-argued to fit

- **The mean is not exactly stateable.** 27799 s over 57 runs is 487.7017… s. The prose carries the exact **total** instead, which is a strictly stronger pin — a bump to any middle-ranked reading moves neither the range nor the median, and the total is what sees it.
- **The excluded run is no longer an outlier.** 864 s is **0.96x** of the census maximum, inside the post-boundary range rather than 1.45x past it, so the duration-based disqualification is dead. That was always an artefact of a population whose maximum was 302 s lower; the rule is positional and always was.
- **The snapshot readings are no longer disjoint from the population — they cannot be.** All three (342, 348, 286 s) are census members, because the self-metrics snapshot reads this job's last run and a census contains every run. Disjointness was a property of the *sample*, so that pin was measuring the sample's incompleteness rather than the rule. It is inverted to **containment**, which is the same evidence read the right way round and fails toward the more useful label: a quoted reading with no census run behind it means the two reads disagree about what the job did. The rule the paragraph is actually about — which SOURCE may set the constant — stays where it was already load-bearing, against the shipped SQL of all three reads.
- Occupancy moves from tenths of a minute to **seconds** (896 of 900): a tenth of a minute is only exact while the ceiling is a multiple of six seconds, and demanding that a measurement be divisible makes an honest figure unstateable rather than catching a drift.

## The consequence, stated and not settled here

`RefreshPhaseSlotSeconds` is 900. Clearance is slot minus ceiling: **306 s before, 4 s now**. The compression grid's spatial clearance survives — the nearest phase minute is 1320 s past the heaviest slot's start, and the discriminating minute `:14` is still excluded — so nothing has overrun a slot.

What broke is the **watch line ordering**. `RefreshSlotWarningSeconds` is 750, and 896 is 146 s ABOVE it, so `ClassifyRefreshSlotHeadroom` bands the grid's own sizing figure `ApproachingSlot`. Seven of the 57 census readings are already at or past that line, where the 16-run record had none. `RefreshSlotWarningSeconds`' own summary pre-registered this: *"a ceiling that rose past this line would put the grid's own sizing figure inside the warning band, and the pin says so rather than leaving a reader to notice."*

**Six test methods are therefore RED, and every condition is left exactly as written.**

| file | method | condition |
|---|---|---|
| `RefreshCeilingProvenancePinTests` | `TheDerivedFiguresAreExactlyStateable_AndTheConstantIsThePopulationMaximum` | a population reading is at or past the watch line; and `Ceiling < WatchLine` |
| `TimescaleSupportTests` | `NoCompressionMinuteStartsWhileTheHeaviestRefreshIsStillRunning` | `ceiling < RefreshSlotWarningSeconds`; `Assert.Equal(26, (watch - ceiling) * 100 / ceiling)` (now -16) |
| `TimescaleSupportTests` | `TheRefreshSlotClassifier_BandsTheLiveReadings_AndKeepsTheWatchLineAboveTheRecordedCeiling` | `ClassifyRefreshSlotHeadroom(ceiling) == InsideSlot` (now `ApproachingSlot`); `ceiling < watch` |
| `TimescaleSupportTests` | `TheRejectedWatchLineAlternative_SitsBelowTheRecordedCeiling_WhileTheChosenLineSitsAbove` | same two |
| `TimescaleSupportTests` | `TheRefreshSlotReading_CarriesItsOwnVerdict_AndReportsOverrunAsNegativeHeadroom` | `atTheCeiling.Headroom == InsideSlot` |
| `TimescaleSupportTests` | `TheRefreshSlotLogLine_IsLeveledByBand_AndSaysNothingWithoutAReading` | the line at the ceiling is `Warning:`, not `Debug:` |

Two failure MESSAGES in the pin file were re-pointed, because the reason they gave — "the doc comment claims otherwise" — is no longer why they matter now that the comment states the crossings plainly. Both conditions are byte-identical, so the tests still go red on the same input.

**What the arithmetic says about the slot, and where it stops.** Restoring the five-sixths line above 896 s needs a slot of at least **1076 s** — 18 minutes, which 60 does not divide, so the grid would gain slots of unequal width or lose one of its four. Moving the fraction instead, or making the refresh cheaper, are the other two routes. Choosing among them is a scheduling decision (#3035, #3044, #3107) and is deliberately not taken by re-deriving a measurement.

## Trend, not calibration

The daily maximum went **594 s** (boundary-day tail, 9 runs) → **778 s** (2026-09-06, 22 runs) → **896 s** (2026-09-07, 24 runs). #3166 listed 702 / 871 / 815 s; the census found a **896.1 s** run at 2026-09-07 23:00 that the issue did not have, so the readings in the issue were not the ceiling. A census removes the "as observed" caveat and puts nothing in its place — the population is closed at a stated instant and it grows.

## Verification

`RefreshCeilingProvenancePinTests` is pure and reads the real `TimescaleSupport.cs` through `[CallerFilePath]`, so it was compiled unmodified into a `net10.0` console harness behind a minimal xUnit shim and run: **8 of 9 pass**, including the full derivation `Verify` and the one-number-at-a-time drift sweep over every captured figure; the 9th is the watch-line row above. Red-first proven by reverting the constant to 594 in a committed tree, confirming `EveryDerivationClaim_...` fails on `max == Ceiling`, and restoring via `git checkout --`. The red list in the table was produced by a second harness that evaluates each ceiling-dependent condition against the real Storage assembly rather than by reading the assertions.

## CHANGELOG

Not edited here — entry text is in the lane report for the coordinator to batch.

### DIFF
--- a/Darling/Darling.Tests/RefreshCeilingProvenancePinTests.cs
+++ b/Darling/Darling.Tests/RefreshCeilingProvenancePinTests.cs
@@ -40,12 +40,16 @@ namespace Darling.Tests;
 /// figures drawn against that reading derived rather than restated. A pin that restated the summary
 /// would go stale by precisely the mechanism it exists to stop.</para>
 ///
-/// <para><b>One pin is written to expire, deliberately.</b> The comment's reason for taking the maximum
-/// rather than a percentile is partly that at this sample size the 95th percentile IS the maximum by nearest
-/// rank. <see cref="Verify"/> checks that, so the pin goes red once the population grows past the point where
-/// it holds — which is the moment the maximum-versus-percentile decision genuinely has to be re-taken rather
-/// than inherited. A check that simply kept passing would let the stated reason quietly stop being the real
-/// one.</para>
+/// <para><b>One pin was written to expire, and it has (#3166).</b> The comment's reason for taking the
+/// maximum rather than a percentile was PARTLY that at sixteen readings the 95th percentile IS the maximum by
+/// nearest rank; <see cref="Verify"/> checked that, so it went red once the census grew the population past
+/// the point where it held, which is the moment the maximum-versus-percentile decision genuinely had to be
+/// re-taken rather than inherited. It was re-taken, the estimator is still the maximum, and the surviving
+/// reason — a scheduling exclusion may accept no exceedance over its own record — does not reference the
+/// population's size. So the expiring clause is gone with the argument it guarded, and the two percentiles
+/// are now pinned as readings that TRACK the population. There is no replacement expiry, because a reason
+/// independent of <c>n</c> has nothing left to expire; what a growing population now moves is the ceiling
+/// itself, which is checked directly.</para>
 ///
 /// <para><b>Which numeral KIND these are, because the repo handles two differently and neither #3072 nor
 /// #3073 says so out loud.</b> A numeral that restates an adjacent list gets DELETION offered in its
@@ -60,11 +64,15 @@ namespace Darling.Tests;
 ///
 /// <para><b>What is deliberately NOT pinned, said plainly rather than left looking covered.</b> The
 /// snapshot readings are quoted evidence, not derived quantities — nothing in the code can know them — so
-/// they are held only by the DISJOINTNESS the paragraph is about (see
-/// <see cref="TheInadmissibleReadings_CannotBeFoldedIntoTheStatusVerifiedSeries"/>), which is the failure
-/// mode that matters: an unfiltered reading migrating into the population a constant may be set from. The
+/// they are held only by their MEMBERSHIP of the census population (see
+/// <see cref="TheInadmissibleReadings_CannotBeFoldedIntoTheStatusVerifiedSeries"/>). That is the inverse of
+/// the relationship the sixteen-run record held them to, and the inversion is the finding rather than a
+/// weakening (#3166): disjointness was a property of a SAMPLE, and a census of the same job contains by
+/// construction whatever a snapshot of its last run reported — so a value-level test could never have
+/// encoded a rule about SOURCES, and the rule is checked against the shipped SQL of all three reads by
+/// <see cref="TheInadmissibilityClaim_MatchesTheShippedSql"/> instead. The
 /// pre-narrowing band and the 3.8x ratio drawn against it are likewise evidence — that band is #3012's
-/// measurement and no constant here spells it — so they carry no pin. The excluded run's 1.45x ratio to the
+/// measurement and no constant here spells it — so they carry no pin. The excluded run's ratio to the
 /// population maximum IS pinned, because both of its terms are stated here. The live envelope's measured
 /// maximum is evidence on the same footing, and its day count with it; what IS pinned is every figure the
 /// prose draws AGAINST that maximum — the slot margin, the share of the slot and the distance past the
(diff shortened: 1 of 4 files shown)


erikdarlingdata:

Closed as **superseded by PR #3178**, which merged to `dev` and absorbed four of this branch's commits. Not merged: this PR was red by design, and its two failing checks were exactly the two required checks on `dev`, so merging it would have blocked every subsequent PR.

**Its work landed.** The 594 → 896 re-derivation, the census, and the doc are all on `dev` via #3178. All six of the tests this branch deliberately left red are green there, and — the condition that mattered — **none needed a band re-typed**; they went green on geometry. The 'at least 1076 s' figure is corrected to **1,077** (the smallest `s` with `s x 5/6 > 896`), which was off by one because `1076 x 5 / 6` is 896 exactly and the classifier warns at `>=`.

**Why it could not simply merge**, recorded because the reasoning outlived the branch: its six reds needed `watch > 896`, so a slot of at least 1,077 s, so a step in {20, 30, 60} — and every one of those reddens the contention guard, while `Other*4 < Heavy` stays red at **all** steps because it contains no step term. So this branch was blocked on a grid **redesign**, not on the scheduling decision it was waiting for.

The lane's discipline here is the reason the redesign happened rather than a band being re-typed: it widened, inverted and deleted nothing, and named the 'provisionally adjust the six' option as the trap — each of those tests is an ordering or a verdict, so re-typing one greens CI while changing nothing about a 4-second clearance.

Issue #3207, opened 2026-09-09
Desktop clock frames are wrong in opposite directions in both SKUs, and each SKU is right where the other is wrong


erikdarlingdata:
Found while sweeping #3198's siblings for #3202. Distinct from the MCP read boundary: this is the WPF surface in both SKUs, and here the two SKUs are wrong on *different* columns — **each is correct exactly where the other is wrong**, which is why nothing has ever caught it.

## The two renderers, and what each expects

| renderer | expects | does |
|---|---|---|
| `ViewerTimeHelper.ForDisplay(DateTime naiveUtc)` (Darling) | naive UTC | Server mode (the default): `naiveUtc.AddMinutes(utcOffsetMinutes)` |
| `ViewerDataService.FormatServerClock(DateTime? serverLocal)` (Darling) | server-local | renders raw |
| `ServerTimeHelper.FormatServerTime(DateTime utcTime)` (Lite) | naive UTC | `ConvertForDisplay(utcTime.AddMinutes(_utcOffsetMinutes), mode)` |

`FormatServerTime` is Lite's `ForDisplay`, not its `FormatServerClock` — it adds the offset, and its parameter is named `utcTime`. There is no raw renderer in Lite. That matters below, because a doc comment in the Darling viewer asserts the opposite.

With the fleet's measured offset of **−240 minutes** (#2932: `cpu_utilization_stats.sample_time` exactly 4 h behind the same run's `collection_time` on **42 of 42** servers; corroborated live at the read boundary in the sibling MCP issue):

- feeding a **server-local** value to `ForDisplay` / `FormatServerTime` subtracts the offset again → displayed **4 hours earlier** than the correct server-local rendering;
- feeding a **naive-UTC** value to `FormatServerClock` → displayed **4 hours later**.

## Darling viewer — server-local values sent through `ForDisplay` (4 h early)

1. **`ViewerDataService.RunningJobs.cs:117`** — `RunningJobRow.StartTimeLocal`. The class doc at `:99-104` states `start_time` is "the store's naive-UTC start_time"; `RunningJobsCollector.cs:56` is `start_time = ja.start_execution_date`, the msdb Agent local clock, and the next line's `DATEDIFF(SECOND, ja.start_execution_date, GETDATE())` confirms the collector treats it as local.
[...]

Issue #3206, opened 2026-09-09
MCP reads return 16 server-local timestamps unmarked, beside naive-UTC fields in the same payload (#3198's siblings)


erikdarlingdata:
#3198 was one instance. The sibling sweep for #3202 found the same class on **16 more fields across 5 more MCP tools**, in both SKUs, and left them out of that PR's scope deliberately. This is that list, with the measurement.

## The class

A store column holding the monitored server's LOCAL wall clock is stamped into an MCP payload by `ToString("o")`, beside fields that are naive UTC. `ToString("o")` on a `DateTimeKind.Unspecified` value renders no offset suffix, so both frames serialize identically — `2026-09-09T08:58:42.7633330` and `2026-09-09T12:59:24.1431700` are the same shape and the payload says nothing about either one. There is no in-band signal a caller could key on, and **`server_properties.utc_offset_minutes` is not exposed by any MCP tool** (`get_server_properties` does not select it), so the caller cannot correct for it manually either.

## Census — every server-local field reaching an MCP payload

Derived by classifying all 66 `CollectorColumnType.Timestamp` columns in `PerformanceMonitor.Collectors` by the provenance of their own T-SQL assignment (the #3202 technique), then mapping each to its emission site. Complete and closed: 131 `ToString("o")` sites in `Darling/PerformanceMonitor.Darling.Service/Mcp/`, every one accounted for.

| Tool | Fields | Store column(s) | Provenance |
|---|---|---|---|
| `get_running_jobs` | `start_time` | `running_jobs.start_time` | `RunningJobsCollector.cs:56` `start_time = ja.start_execution_date` — msdb Agent local clock. The next line's `DATEDIFF(SECOND, ja.start_execution_date, GETDATE())` confirms the collector treats it as local. |
| `get_blocking` | `blocked_last_tran_started`, `blocking_last_tran_started`, `blocked_last_batch_started`, `blocking_last_batch_started`, `blocked_last_batch_completed`, `blocking_last_batch_completed` | `blocked_process_reports.*`, `dmv_blocking_snapshots.*` | `DmvBlockingSnapshotCollector.cs:71,74` = `sys.dm_tran_active_transactions.transaction_begin_time`; `BlockedProcessReportCollector.cs:977-990` = the blocked-process-report XML's `lasttranstarted` / `lastbatchstarted` / `lastbatchcompleted` attributes. |
[...]

PR #3228, opened 2026-09-09
Render every desktop timestamp through the renderer for its own clock frame (#3207)

erikdarlingdata:
Fixes #3207.

Twenty-four desktop render sites across both SKUs rendered a timestamp through the renderer built for the **other** clock frame. Both directions were live and both are silent — the wrong renderer returns a plausible timestamp, off by the server's whole offset:

- a **server-local** value through `ForDisplay` / `FormatServerTime` subtracts the offset a second time and displays **4 hours early** on the fleet's measured −240;
- a **naive-UTC** value through `FormatServerClock` displays **4 hours late**.

## Root cause: each SKU was missing one half of the pair

**Lite had no raw renderer at all.** `ServerTimeHelper.FormatServerTime` *is* Lite's `ForDisplay` — it adds the offset unconditionally and names its parameter `utcTime` — so there was no way to say "this value is already on the server's clock". Sixteen sites therefore pushed server-local DMV and XML values through it. `ServerTimeHelper.FormatServerClock` supplies the missing half, and unlike Darling's namesake it routes through `ConvertForDisplay`, so these columns now honour the user's Server/Local/UTC choice like every other Lite timestamp rather than ignoring it.

**Darling had no way to say "this column is UTC" in a display property.** Five `query_store_stats` sites reached for the raw renderer instead. `ViewerDataService.FormatStoredUtc` is now the named counterpart to `FormatServerClock` — Query Store returns `datetimeoffset` and `QueryStoreCollector` normalises through `DateTimeOffset.UtcDateTime`, so those five are UTC. The sixth Darling site goes the other way: `running_jobs.start_time` is the msdb Agent's clock (`RunningJobsCollector` ships `ja.start_execution_date` verbatim and measures duration against `GETDATE()`), so it moves off `ForDisplay` onto `FormatServerClock`.

Inlining `ViewerTimeHelper.ForDisplay(x).ToString(...)` would behave identically; a named method beside its opposite is what makes the choice reviewable, which is the whole failure mode here.

## Two sites #3220's census structurally cannot see

Found by hand, and worth naming because they show the guard's reach rather than a gap in its care:

- **the deadlock graph's `lasttranstarted`** (`ViewerDataService.Deadlock.cs`) — parsed from XML into `DeadlockProcessInfo`, with no store column of that name for the frame register to resolve. `DeadlockTimeLocal` two lines above is the XE `@timestamp` and correctly stays on `ForDisplay`: two frames, two renderers, one class.
- **Lite's `RunningJobRow.StartTimeLocal`**, which used `DateTime.ToLocalTime()` — not a renderer the guard's `RenderCall` tracks. That one carried two defects: it treated an msdb-local value as UTC, *and* it rendered in the Lite host machine's zone regardless of the display-mode setting every other timestamp honours.

## Three doc comments stated the wrong frame

Corrected with the code, including the one that hid the Lite half. `ViewerDataService.QuerySnapshots` called Lite's `FormatServerTime` a "raw server-clock" renderer and asserted the two SKUs agreed. They did not — Lite was double-skewing while the comment said otherwise. They agree now, and the comment says why it used to be wrong rather than being quietly deleted.

## Guard changes, and why each direction differs from #3206's

`FormatStoredUtc` is **registered** in the `Renderers` map rather than left out. An unlisted renderer is invisible to that scan, so omitting it would have dropped five sites out of the census as the price of fixing them.

The seventeen `DesktopRenderFrameMismatch` rows come **out**, and `DesktopRenderMismatchSites` goes to **0**. That is the opposite of what #3206 did with its MCP rows, and the difference is in the discriminator: the render scan flags only a *disagreement* between column frame and renderer, so a corrected site genuinely leaves the found-set and a row for it would fail set equality from the other side. The MCP scan keys on a payload field NAME, which survives the fix — which is why those rows had to be relabelled `DeSkewedAtRead` instead. Zero is the strongest form of this pin, and a regression is still caught: it would be found by the scan and not be in the inventory.

## Verification

`Darling.Tests` and `Lite.Tests` are `net10.0-windows` and cannot execute on macOS, so CI is the arbiter for the suite. Run here: all four affected projects build with 0 errors, every changed file keeps CRLF with no bare LF, and the two existing `FormatServerClock` behaviour pins (`ViewerQueriesTests.FormatServerClock_ShowsRawServerWallClock_EmptyForNull` and `ViewerQueriesRestTests.QuerySnapshotRow_TranStartTimeLocal_RendersRawServerClock_...`) pin columns this change does not touch — `query_stats`/`procedure_stats` and `query_snapshots.tran_start_time` — so they still hold.

I have not run the render scan itself; its frame register is derived from collector provenance at runtime and my ability to replay that faithfully is the thing that has been wrong twice on the sibling PR, so I am not claiming it here.

## Merge order

This shares `ConsumedTimestampFrameDisciplineTests.cs` with #3212, which edits the MCP rows and two different constants. Whichever lands second needs a small mechanical conflict resolution in that one file — different regions, no overlapping rows. #3212 first is the easier order, since its inventory edit is the larger of the two.

### DIFF
--- a/Darling/Darling.Tests/ConsumedTimestampFrameDisciplineTests.cs
+++ b/Darling/Darling.Tests/ConsumedTimestampFrameDisciplineTests.cs
@@ -698,15 +698,26 @@ public sealed class ConsumedTimestampFrameDisciplineTests
     private static readonly Regex ProjectionAlias =
         new(@"([^\r\n]*?)\s+AS\s+([a-z][a-z0-9_]*)(?![\w])", RegexOptions.IgnoreCase);
 
-    /// <summary>The two WPF renderers that ADD the collected offset — i.e. that take naive UTC — and the
-    /// one that renders raw, i.e. that takes the server's own clock. <c>FormatServerTime</c> is Lite's
-    /// <c>ForDisplay</c>, not its <c>FormatServerClock</c>: it adds the offset and names its parameter
-    /// <c>utcTime</c>. Lite has no raw renderer at all, which is why a comment in the Darling viewer
-    /// claiming parity with it is wrong in the direction that hid this.</summary>
+    /// <summary>
+    /// Every WPF renderer, with the frame its input has to be in. <c>ForDisplay</c> and Lite's
+    /// <c>FormatServerTime</c> ADD the collected offset, so they take naive UTC — <c>FormatServerTime</c> is
+    /// Lite's <c>ForDisplay</c>, not its <c>FormatServerClock</c>, and it names its parameter
+    /// <c>utcTime</c>. <c>FormatServerClock</c> takes the server's own clock.
+    ///
+    /// <para>#3207 changed two things here. Lite had NO raw renderer, which is why sixteen of its sites
+    /// pushed a server-local value through <c>FormatServerTime</c> and why a comment in the Darling viewer
+    /// claiming parity with it was wrong in the direction that hid this; Lite now has
+    /// <c>ServerTimeHelper.FormatServerClock</c>, matching the name already listed here. And
+    /// <c>FormatStoredUtc</c> is new on the Darling side: the five <c>query_store_stats</c> sites had no
+    /// honest way to say "this column is UTC" and reached for the raw renderer instead. It is registered
+    /// rather than left out so those sites stay checkable — an unlisted renderer is invisible to this scan,
+    /// which would have silently dropped five sites out of the census as the price of fixing them.</para>
+    /// </summary>
     private static readonly (string Renderer, ClockFrame Expects)[] Renderers =
     [
         ("ForDisplay", ClockFrame.Utc),
         ("FormatServerTime", ClockFrame.Utc),
+        ("FormatStoredUtc", ClockFrame.Utc),
         ("FormatServerClock", ClockFrame.ServerLocal),
     ];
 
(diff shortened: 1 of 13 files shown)


erikdarlingdata:
Closing in favour of #3227, which covers strictly more.

Both PRs were opened for #3207 within six minutes of each other — #3227 at 17:43, this one at 17:49 — and I opened this one without checking whether a PR already existed. That check is the cheap one and I skipped it.

#3227 is the more complete fix and now carries everything this branch had:

- it fixes four sites this branch does not touch — `ViewerDataService.PlanCorrection.cs` lines 209-212, where `Local()` is a `ViewerTimeHelper.ForDisplay` wrapper (its parameter is literally named `naiveUtc`) applied to four server-local `sys.dm_db_tuning_recommendations` columns. #3220's census does not contain them, because `Local` is not in its `Renderers` map — the same invisibility that hid Lite's `ToLocalTime` site;
- it also corrects `ItemHistory.cs`, which carries the same wrong "render raw" claim I fixed in `ViewerHistoryRows.cs` and which this branch missed entirely, plus `FinOps.cs` and the `JobHistory` frame docs;
- it adds 335 lines of renderer-arithmetic tests pinning the ±240 direction, which this branch has no equivalent for;
- and it has since taken `ViewerDataService.FormatStoredUtc`, its registration in the `Renderers` map, and the `Assert.Equal(4, Renderers.Length)` pin — the only things this branch had that it lacked.

So there is nothing left to port and no reason to keep two PRs open against one issue. Branch left in place rather than deleted, in case any of the reasoning here is worth lifting.

Two things from this branch's CI worth having on #3227's radar: `AssertMatchesInventory` asserts the found set is **not** empty, so zeroing `DesktopRenderMismatchSites` fails there — the render fact's own `judged >= 30` floor is the better guard once the class is fixed, and that fact needs its set equality spelled inline rather than routed through the helper. And `dev` has moved twice under both branches (#3212 as 849ea2ece, #3225 before it), so the MCP rows in that census are now `DeSkewedAtRead` with `McpPayloadUnmarkedSites` at 1.

Issue #2993, opened 2026-09-05
PgDeadlockLogParser captures the log's timezone abbreviation and never reads it — occurred_at assumes UTC no matter what log_timezone says


erikdarlingdata:
Found while auditing the pgmon fleet's timestamp handling. Latent today, undefended forever: the three targets sampled directly (target-a, target-b, target-c — 3 of 50) all run `log_timezone = UTC` via a shared configuration-file setting, so nothing is currently wrong on this fleet. Nothing in the code keeps it that way.

## Mechanism

`PerformanceMonitor.Collectors/PgDeadlockLogParser.cs:61` — the log-prefix regex captures the zone:

```csharp
@"^(\d{4}-\d\d-\d\d \d\d:\d\d:\d\d\.\d+) (?<zone>\w+) \[(?<pid>\d+)\][^\n]*ERROR:  deadlock detected\s*\n"
```

The `zone` group is never read anywhere in the file. The timestamp goes through `DateTime.TryParse(..., DateTimeStyles.AdjustToUniversal | DateTimeStyles.AssumeUniversal, ...)` (lines 122–125), which treats the naive log timestamp as UTC unconditionally. A repo-wide search for `log_timezone` returns zero hits — the GUC that decides what those timestamps mean is never consulted, collected, or checked.

## What happens when it bites

A target whose `log_timezone` is not UTC gets every parsed deadlock's `occurred_at` shifted by the zone offset, silently: trend buckets land in the wrong hour, deadlock detail misaligns against every other UTC-keyed collection, and nothing errors. The failure direction is the bad one — plausible-looking data, wrong moment.

## Options

1. Convert using the captured abbreviation — worst option; zone abbreviations are ambiguous (CST is three different zones), which is presumably why the capture was never wired up.
2. Read `log_timezone` at collection time (one GUC; the server-config collection already reads settings) and convert properly — correct, small.
3. Cheapest honest defense: same GUC read, but refuse to parse under a non-UTC `log_timezone` and record the refusal as a note naming the setting — turns silent-wrong into visible-unsupported.

2, with 3's note as the interim, is the shape the honest-empties convention wants. No measurement decides anything here — it is a defense, not a tuning question.

Related: the same collector's log reads are currently denied fleet-wide on the managed path (filed separately), so this parser is doubly unexercised in production right now.

erikdarlingdata:
Independent re-verification of this issue's claims (separate session, same investigation).

[...]

Issue #2932, opened 2026-09-04
An alert-pass read outside AlertPassCommandTimeoutTests' six-file allowlist is untimed, and the pin cannot see it


erikdarlingdata:
## An alert-pass store read sits outside `AlertPassCommandTimeoutTests`' six-file allowlist, so it is untimed and the pin is structurally unable to say so

Found by #2874's group B and verified independently on current `dev`. Two halves, filed together on purpose: fixing the site alone leaves the next one just as invisible.

### The untimed site

`Darling/PerformanceMonitor.Darling.Service/DarlingWorker.cs:3886`, `ReadLatestCpuAsync`:

```
:3891    await using var connection = await _postgres!.OpenConnectionAsync(cancellationToken);
:3892    using var command = new NpgsqlCommand(@"
             SELECT sqlserver_cpu_utilization, other_process_cpu_utilization
             FROM cpu_utilization_stats
             WHERE server_id = $1
             ORDER BY sample_time DESC
             LIMIT 1", connection);
```

No `CommandTimeout`, so it inherits Npgsql's undocumented 30 s default — the defect class #2874 exists to sweep.

It is unambiguously an **alert-pass** read: its single caller is `:2921`, which feeds the result straight into the `AlertServerSnapshot` passed to `engine.EvaluateServerAsync` on the next line.

### Why #2882's pin reports clean over it

`Darling/Darling.Tests/AlertPassCommandTimeoutTests.cs:55` scopes the whole scan to an allowlist of six filenames:

```
DarlingAlertReadAdapter.cs, DarlingPostgresAlertReadAdapter.cs, PgAlertStateStore.cs,
PgMuteRuleStore.cs, PgAlertHistoryStore.cs, DarlingSelfAlertEvaluator.cs
```

`DarlingWorker.cs` is not among them, so the site is never read.

**The pin already guards the direction that doesn't matter here.** It asserts each of the six paths resolves, with a comment saying a renamed or moved alert-pass type must fail loudly *"rather than silently shrinking the scan to the files that still resolve — an empty or partial sweep is how a guard starts reporting clean on code it no longer reads."* That reasoning is right and the guard works: it catches a file **leaving** the list. There is nothing at all against a site that was **never in** it, which is the asymmetry that let this through.

### Why an allowlist of filenames is the wrong scope, specifically

[...]

PR #3229, opened 2026-09-09
Scope the de-skew's accuracy claim where the column outlives a DST transition

erikdarlingdata:
`server_properties.utc_offset_minutes` is `DATEDIFF(MINUTE, GETUTCDATE(), GETDATE())` — the offset in force **at collection time**, one current value. Subtracting it from a stored timestamp is exact only while the timestamp and the offset sit on the same side of a DST transition. #3212 shipped sixteen de-skewed fields without saying so on the two reads where that can be false.

Raised by another session reading the merged code, and confirmed here against `DarlingServerConnector.cs:94` and the fleet's reported offset.

## Why `get_index_usage` is the case that matters

Thirteen of the sixteen describe **current state** — a running job, an open transaction, a cleaner that ran seconds ago — so their timestamps and the collected offset are always in the same DST period, and the de-skew is exact.

`last_user_access` is not current state. It is the `GREATEST` of four `sys.dm_db_index_usage_stats` columns that persist **since the instance restarted**, which on a stable production box is routinely months. So a large share of those values predate the most recent transition and come back **60 minutes early** — silently, and in the plausible direction, which is the bad one. `sqlserver_start_time` on the same row is the bound on how far back it can reach.

The fleet is exposed rather than theoretically exposed: any target in a DST-observing zone has it — the offset differs by an hour between summer and winter, so every pre-transition timestamp is de-skewed by an offset that was not in force when it was written. A target configured to UTC is unaffected, and on AWS RDS the instance takes its time zone from a creation-time parameter, so a non-UTC zone is an ordinary configuration rather than an exotic one. #2932 records the measurement behind the four-hour figure above.

`get_pvs_stats` gets the same note one tier down: the off-row cleaner runs continuously and is unaffected in practice, but `aborted_version_cleaner_*` can reach back weeks on a quiet database — a live read showed `2026-08-26` against a `2026-09-09` snapshot — so it can cross a transition.

## Documented rather than fixed, deliberately

The proper fix is a **zone** instead of an offset: `CURRENT_TIMEZONE_ID()` (SQL Server 2019+) collected alongside `utc_offset_minutes`, then `AT TIME ZONE` at the read boundary, which handles transitions correctly for a timestamp of any age. That is a collected-column addition, a migration rung, and swapping the `- make_interval(...)` expression for an `AT TIME ZONE` conversion in five readers, with the offset kept as the fallback while `server_timezone_id` is null — which it will be for every server until its next `server_properties` collection.

**There is no backfill**, and an earlier revision of this description wrongly said there was. Nothing stored needs rewriting: the label is `DeSkewedAtRead` precisely because the collector ships `us.last_user_seek` verbatim and the offset is subtracted in the *projection*, so no stored column is wrong today. And a single current zone ID is legitimately valid retroactively where a single current offset is not — that asymmetry is the whole defect. `utc_offset_minutes` changes twice a year, so today's value is wrong for half of history; `CURRENT_TIMEZONE_ID()` returns a zone identity that is stable for the life of the server, so collecting it going forward and applying it to already-stored rows is sound. It is the one field whose present value describes the past.

Two things that stay genuinely open, so this does not read as free: `CURRENT_TIMEZONE_ID()` is SQL Server 2019+, so the offset path has to be *kept* rather than replaced — the product supports servers older than that and will for someone regardless of any particular deployment — and a server whose Windows time-zone data disagrees with the historical rule for a given date is still wrong for that date. Much narrower than the current failure, not zero.

What this carries is the #2993 shape: the **scope of the claim**, stated where the de-skew happens, so the next reader knows the frame is exact inside the current DST period and within an hour outside it. Same family as #2993 by a different mechanism — there a zone was captured and discarded, here an offset is captured and over-applied, and both end with a naive timestamp confidently placed in the wrong frame.

## Verification

Doc-only; no shipped SQL or behaviour changes. Both affected projects build with 0 errors and both files keep CRLF with no bare LF.

One check worth naming because it is a live trap in this area: `ConsumedTimestampFrameDisciplineTests.ProjectionAlias` is `IgnoreCase` over `" as <word>"` and reads string literals and comments, so ordinary prose next to a census column name registers as a SQL projection alias — it cost #3212 two CI rounds. I ran that exact regex over every added line and no added prose produces a reportable match.

### DIFF
--- a/Darling/PerformanceMonitor.Darling.Service/Mcp/DarlingObjectStatsReader.cs
+++ b/Darling/PerformanceMonitor.Darling.Service/Mcp/DarlingObjectStatsReader.cs
@@ -195,9 +195,12 @@ internal static class DarlingObjectStatsReader
     /// transition. These four do not: <c>sys.dm_db_index_usage_stats</c> persists since the instance
     /// restarted, which on a stable production box is routinely months, so a large share of values predate
     /// the most recent transition and come back <b>60 minutes early</b> — silently, and in the plausible
-    /// direction. The fleet is exposed rather than theoretically exposed: its targets report
-    /// <c>utc_offset_minutes = -240</c>, which is EDT, a DST-observing zone on summer time.
-    /// <c>sqlserver_start_time</c> on the same row is the bound on how far back that can reach.</para>
+    /// direction. This is not a theoretical exposure: any target in a DST-observing zone has it, a target
+    /// configured to UTC does not, and on AWS RDS the instance takes its time zone from a creation-time
+    /// parameter — so a non-UTC zone is an ordinary configuration rather than an exotic one, and "it is
+    /// RDS, so it is probably UTC" is not a safe assumption. #2932 records the measured offset behind the
+    /// four-hour figure quoted above. <c>sqlserver_start_time</c> on the same row is the bound on how far
+    /// back the affected values can reach.</para>
     /// <para>Fixing it properly needs a ZONE rather than an offset — <c>CURRENT_TIMEZONE_ID()</c>
     /// (SQL Server 2019+) collected alongside the offset, then <c>AT TIME ZONE</c> at the read boundary,
     /// which handles transitions. That is a collected-column addition and a migration rung, so what is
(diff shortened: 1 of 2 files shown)


erikdarlingdata:
Superseded by #3231 — same content, opened fresh as a single commit.

Closing rather than pushing more commits: this branch's first commit carried a sentence that justified the exposure by a deployment fact, and the follow-up commit removed it from the source without removing it from that commit. #3231 carries only the generalised form, in one commit, with the same file content — the staged blobs are byte-identical to this branch's tip.

To be accurate about what this achieves: it limits what a future reader encounters and keeps the sentence out of `dev`'s permanent ancestry, which merging this branch would have made unavoidable. It is not erasure — the old commit stays reachable by SHA for a while and this timeline records that it existed.

Issue #3288, opened 2026-09-11
Every Velopack-generated executable we publish is unsigned — the portable stub, Update.exe and Setup.exe — on v3.6.0 and v3.7.0, while build.yml claims the gap is closed


erikdarlingdata:
## What

The Velopack `Setup.exe` published on a release carries **no Authenticode signature**, on both v3.6.0 and v3.7.0. `build.yml` contains a comment stating this gap was closed; it was not closed for the installer stub itself.

## Verified, not inferred

Read directly from the published release assets by fetching only the PE header and certificate-table byte ranges:

| asset | release | certificate table |
|---|---|---|
| `PerformanceMonitorLite-lite-Setup.exe` | v3.6.0 | offset 0, size 0 — **no signature** |
| `PerformanceMonitorLite-lite-Setup.exe` | v3.7.0 | offset 0, size 0 — **no signature** |
| `current/PerformanceMonitorLite.exe` (inside `-lite-Portable.zip`) | v3.6.0 | signed |
| `current/PerformanceMonitorLite.exe` (inside `-lite-Portable.zip`) | v3.7.0 | signed |

The app binary is signed on both, by the same leaf: `O=SignPath Foundation, CN=SignPath Foundation`, issued by `GlobalSign GCC R45 CodeSigning CA 2020`, serial `1CA814640BEAA2C1D9BB8F30`, valid Apr 2026 to Jul 2027 — byte-identical certificate across the two releases. So signing works; it just does not reach `Setup.exe`.

## Why the existing steps do not cover it

`build.yml` uploads `publish/Lite-velopack/` to SignPath, signs it, and has `vpk pack` read from the signed output. That signs the **payload**. `vpk pack` then *generates a new `Setup.exe` stub* around that payload, and nothing signs the stub afterwards — so the one file a user actually double-clicks is the one file that is unsigned.

The comment above those steps reads:

> every Setup.exe shipped unsigned since Velopack packaging was introduced. These steps close that gap by mirroring the exact upload/sign pattern used for Dashboard/Lite/Installer above, and vpk pack below now reads from the signed output.

The first sentence is still true of the published artifact. The mechanism described is real and necessary, but it addresses the inputs rather than the output, and the claim that the gap is closed is what stops anyone re-checking.

## Why it matters

[...]

PR #3294, opened 2026-09-11
Sign the executables Velopack generates, during packing (#3288)

erikdarlingdata:
Closes the #3288 gap: the portable launcher stub, `Update.exe` and `Setup.exe` ship unsigned on every release, for both Lite and the Darling remote-seat Viewer — ten executables per release, while the app payload is signed.

Nothing needs configuring before this merges: the batch artifact configuration slug is `SignTemplate`, named directly in `build.yml` alongside `Lite`, `Darling` and `DarlingViewer`. Every new step is gated on `github.event_name == 'release'`, so merging to `dev` changes nothing until a release is cut.

## Why signing has to happen during packing

Velopack creates those three executables *after* packing. They cannot be signed afterwards, because re-signing a file inside a `.nupkg` changes that package's bytes, which invalidates the `SHA256` and `Size` recorded in `releases.<channel>.json` and patched against by delta packages. Velopack's own documentation says signing "needs to be performed by Velopack itself".

## What the probe (#3292) established, so none of this is inferred

- `vpk pack` offers **35 files, not ~700** — it skips files that already carry a signature, so Microsoft's framework binaries never reach the template.
- It offers exactly the three targets: `PerformanceMonitorLite_ExecutionStub.exe`, `Squirrel.exe` (Update.exe's internal name), `Setup.exe`.
- With `--signParallel 50` it invokes the template **twice per product** — 34 pack-directory files, then `Setup.exe` in its own later phase. At the default of 10 it is five.
- **It does not verify that the template signed anything.** A no-op template returned exit 0 and packing completed normally. That is how #3288 shipped twice, and it is why this change verifies its own outcome in two places.
- `--signDisableDeep` does **not** exist on the Windows command, despite the string appearing in the tool's assemblies.

## Approvals: 4 today, 6 after

The two pre-pack Velopack rounds are **deleted** — vpk signs that payload itself now, and keeping both would sign it twice and cost two more approvals. `Sign Lite` (flat zip) and `Sign Darling` (service + `viewer/`) are unchanged.

## Two verification points, because the packer provides none

`sign-batch.ps1` refuses an empty batch, refuses duplicate base names within a batch, and after copying the signed files back over the originals it **asserts each one now carries a certificate table** before returning success.

The release guard then re-checks the whole `releases/` directory before upload, including executables inside the zips and nupkgs. Both reuse the PE parse from `verify_release_signatures.py`, whose six mutations are each confirmed to turn its self-test red.

Guard behaviour is tested locally in all three directions: an unsigned exe **hidden inside a zip** fails it, an all-signed directory passes, and an empty directory fails with a distinct status — finding nothing to check is a broken run, not a clean one.

## What is NOT tested, stated plainly

No SignPath call has been made. `Submit-SigningRequest`'s invocation, the batch artifact configuration matching `**/*.exe` and `**/*.dll` inside a `batch/` subdirectory, and the copy-back are all unexercised against the real service. `.ps1` cannot run on the machine this was written on. The throwaway pre-release tag is the gate, not this PR's CI.

Two specifics to watch there:
- Files are staged into a **`batch/` subdirectory** rather than the zip root, because whether `**/` also matches zero path segments is undocumented. One segment removes the question.
- `-WaitForCompletionTimeoutInSeconds` is set to **3600**. The cmdlet's own default is 600, which will not survive waiting for a human to approve — and every Foundation-tier request needs one.

## CHANGELOG

```
- **Every executable in a release is signed, including the ones Velopack generates** ([#3288]) - the portable launcher stub, `Update.exe` and `Setup.exe` shipped unsigned on both products while the app payload was signed, so the file a user double-clicks in the portable zip was a 399 kB unsigned launcher rather than the signed 182 kB app - which is what an "unknown publisher" policy blocks. They are created after packing and cannot be signed afterwards without invalidating the package hashes the updater verifies, so `vpk pack` now signs them during packing, and a release-time guard refuses to publish an executable without a signature.
```

### DIFF
--- a/.github/scripts/sign-batch.ps1
+++ b/.github/scripts/sign-batch.ps1
@@ -1,0 +1,145 @@
+<#
+.SYNOPSIS
+  Signs a batch of files through SignPath. Written to be called BY `vpk pack --signTemplate` (#3288).
+
+.DESCRIPTION
+  Velopack generates the portable launcher stub, Update.exe and Setup.exe AFTER packing, so they
+  cannot be signed by the pre-pack rounds and cannot be signed afterwards either: re-signing a
+  file inside a .nupkg changes that package's bytes, which invalidates the SHA256 and Size that
+  releases.<channel>.json records and that delta packages patch against. Signing therefore has to
+  happen inside packing, which is what `--signTemplate` is for.
+
+  Measured behaviour of `vpk pack` 1.2.0 that this script depends on:
+
+    * It offers 35 files for the Lite payload, not ~700, because it SKIPS files that already
+      carry a signature -- Microsoft's framework binaries never reach here.
+    * With `--signParallel 50` it invokes this script exactly twice per product: once with the
+      34 pack-directory files, once with Setup.exe alone in its own later phase.
+    * It does NOT check that this script actually signed anything. A template that exits 0
+      without signing leaves packing to complete and the release to publish unsigned binaries,
+      which is precisely how #3288 shipped on two releases. So this script verifies its own
+      outcome before returning success, and that verification is the point of it.
+
+.PARAMETER Files
+  The files to sign, passed positionally by vpk in place of {{file...}}.
+
+.NOTES
+  Required environment:
+    SIGNPATH_API_TOKEN     CI user token
+    SIGNPATH_ORG_ID        organization id
+    SIGNPATH_PROJECT_SLUG  project slug
+    SIGNPATH_POLICY_SLUG   signing policy slug
+    SIGNPATH_CONFIG_SLUG   artifact configuration slug for a flat batch of PE files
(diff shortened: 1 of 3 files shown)


erikdarlingdata:

Slug is `SignTemplate`, so it is now a literal in `build.yml` alongside `Lite`, `Darling` and `DarlingViewer`, and the `SIGNPATH_CONFIG_SLUG` repository variable is no longer needed — **nothing to set before releasing.** The variable was only ever a placeholder for not knowing the slug, and it added both a setup step and a way to get it wrong.

The description's "Before this can be released" section no longer applies. What still gates a release is unchanged and unchanged in kind: no SignPath call has been exercised, so the throwaway pre-release tag is the real test.

Issue #3393, opened 2026-09-13
query_store's plan fetch shares query_stats' pre-fix anti-pattern (uncapped GetString, no per-row size bound) — worth measuring


erikdarlingdata:
## Where this comes from

Side finding from tonight's `query_stats`/`procedure_stats` drain investigation
(`QueryPlanXmlCaptureLimits`, 512 KB per-row cap — see that file's doc comment for the full
mechanism this fixes: batched, uncapped `GetString()` reads of multi-megabyte plan XML with
no `CommandBehavior.SequentialAccess` produce Large-Object-Heap allocations, which trigger
process-wide GC pauses that stall every concurrent collector's network read — confirmed via
78 real production stalls on the fleet's outlier server showing a completely idle target at the moment of
each one, plus #2880's own forensic note that unrelated collectors' `ExecuteReaderAsync`
calls degrade 3-152x in the same window).

`QueryStoreCollector.FetchAndStorePlansAsync` (`DarlingCollectorRunner.cs`) reads its plan
XML the exact same way — `reader.GetString(2)` on the plan-XML column, no
`SequentialAccess` — and has **no per-row size cap at all**. Filing this separately from
#3392 because it's a different collector, a different fix shape, and — see below — the
live evidence for it is real but much weaker than what query_stats/procedure_stats had.

## What's confirmed vs. what isn't

**Confirmed in code:**
- `reader.GetString(2)` materializes the full plan XML string regardless of size, same as
  the pre-fix `query_stats`/`procedure_stats` shape.
- The only size control is `context.TextByteBudgetOverride ?? 12 * 1024 * 1024` (up to
  `QueryStoreCollector.MaxTextBytesPerDatabase` = 64 MB), checked **between chunks**
  (`if (shippedBytes >= budget) break;` before building the next chunk's query) — not
  per-row and not mid-chunk. `PlanFetchIdsPerStatement = 400`: a single chunk can read up
  to 400 plan rows via `GetString()` before the budget is ever re-checked, so a chunk that
  happens to contain several large plans can land well past the nominal budget before
  anything stops it — the budget bounds how many MORE chunks get *started*, not what a
  chunk already in flight finishes reading.
- This collector's fetch runs through the per-database enumerated path
  (`RunsPerDatabase`/`BuildEnumerationQuery`), which #2880's stall-probe instrumentation
  does not cover at all — that instrument only watches the plain single-query server-scoped
[...]

Issue #3392, opened 2026-09-13
[DESIGN] Eventually collect plan XML that QueryPlanXmlCaptureLimits caps, without reintroducing the batched-large-fetch problem it fixes


erikdarlingdata:
## Where this comes from

Follow-up to the `query_stats`/`procedure_stats` drain fix (`QueryPlanXmlCaptureLimits`,
512 KB per-row cap on captured plan XML — see that file's doc comment for the full
GC-pressure mechanism this cap fixes). The cap is correct and measured working, but it has
a side effect worth naming explicitly: **a plan over the cap does not ship its XML LATER
either — it ships NULL every cycle it recurs, forever**, because the cap is a per-row
content decision, not a "not yet" marker. The plans this affects are exactly the ones most
worth being able to see: big, complex, expensive enough to justify their own tuning look.

This issue proposes a way to eventually collect them without reintroducing the batching
that made the drain slow in the first place.

## Evidence a deferred fetch is actually viable here

The obvious worry: `sys.dm_exec_query_stats.plan_handle` is a live plan-cache pointer, not
a durable id like Query Store's `plan_id` (see #2210 — the reason that design put Query
Store's plan XML fetch on `plan_id`, not a cache handle). If these handles turn over every
few minutes, "come back later and fetch it" doesn't work; you'd need a same-cycle fetch
instead.

Measured directly against the fleet's outlier server (the fleet's known worst offender for this collector,
2026-09-13T01:12 UTC), every plan CURRENTLY over the 512 KB cap, with its compile age:

| plan_bytes | compile_age | |
|---|---|---|
| 8,686,284 | 100.5 days | |
| 8,686,283 | 100.5 days | |
| 2,895,175 | 33.5 days | |
| 1,573,990 | 19.6 hours | |
| 1,466,122 | 13.6 hours | |
| 2,895,174 | 33.5 days | |
| 1,094,588 | 33.5 days | |
| 809,498 | 19.3 hours | |
| 752,842 | 12.5 days | |
| 720,128 | 4.7 days | |
| 719,466 | 144.6 days | |
| 564,624 | 19.9 hours | |
| 547,708 | 19.7 hours | |

13 of 13 plans currently over the cap have already sat in cache, uninterrupted, for at
least 13.6 hours — several for months. Cross-checked against `get_plan_cache_bloat` on the
same server: `Proc`-type plans average 1.6M executions each (bloat is concentrated in
[...]

Issue #3189, opened 2026-09-09
The store probe is 52.7% of query_store's wall time and 98.6% of what it examines is already stored: 31M references a day to find 1.4% missing


erikdarlingdata:
**52.7% of the `query_store` collector's wall time is spent in the store probe, and 98.6% of what the probe examines is already stored.** The instrumentation that shows this shipped in #2859/#2902/#2806 and nobody has read its output since.

## Measured, 24 hours, the SQL Server store

| | |
|---|---|
| `query_store` total collector time | **52,026 s** |
| plan-fetch probe | 12,589 s — **62.0%** of plan fetch |
| text-fetch probe | 14,853 s — **85.1%** of text fetch |
| **probe phases combined** | **27,438 s — 52.7% of the whole collector** |
| references examined | **30,950,148** |
| found missing | **437,876 — 1.41%** |

So roughly **7.6 hours of store time per day examining 31 million references to establish that 98.6% of them are already there.**

The column semantics are from source, not inferred — `DarlingCollectorRunner.cs:212-216`: `ProbeMs` is *"the store touch/probe round trip that decides which ids are missing"*, `ProbeIds` is *"references the probe EXAMINED, which is what ProbeMs scales with"*, and `IdsAttempted` is *"ids the fetch actually attempted — those the probe found MISSING."*

## It is systemic, not a hot server

43 servers, median **624 s** each, mean 638, worst **1,222 s** — the top server is 2x the median. So this is not one server's data shape and the per-server remedies do not apply.

## The reading is validated against the instrumentation's own calibration

V110's doc comment publishes the rates it was calibrated on, and both reproduce, which is why these figures are trustworthy rather than a novel interpretation:

| rate | documented | measured now |
|---|---|---|
| `target ÷ ids_attempted` (plans) | 31.65 ms/id on cold plans | **28.73** |
| plan probe share of plan fetch | 55.4% | **62.0%** |
| text probe share of text fetch | 80.6% | **85.1%** |
| `probe ÷ probe_ids` | ~0.61 ms/reference | **0.807** plan / **0.967** text |

## Two secondary findings that fall out of that

**The per-reference probe cost has drifted 1.3–1.6x above its calibrated rate** (0.807/0.967 against ~0.61 ms/reference). Nothing reports that on its own — same shape as the ceiling constants in #3188.

[...]

Issue #2880, opened 2026-09-04
Out-of-band watchdog: sample server-wide waits during a collector stall (the per-session capture cannot be built — target_session_id is unpopulated)


erikdarlingdata:
Split out of #2864, which built items 1-3 (merged as #2868). This is item 4, deliberately left unbuilt there because it is the only one that needs a design decision rather than more instrumentation.

## Why the other three cannot answer this

Items 1-3 record what the CLIENT knows: rows read, bytes read, time since the last successful read, the target session id, and the sweep's peer high-water mark. That is enough to say **where** a cycle stalled — the first production capture read `open:104ms drain:119,945ms rows=0`, and V109 now adds "and it received nothing in those 119.9 seconds".

It is not enough to say **what it was waiting on**. That answer is on the target, in `sys.dm_exec_requests`, and the tool cannot reach it — because collectors run strictly sequentially per server, so the stalled collector is itself holding the sequence that `waiting_tasks`, `dmv_blocking_snapshot` and `query_snapshots` would run in. Measured on a real stall: nothing was observed for four minutes, `waiting_tasks` ran two seconds after it cleared and returned zero rows, and 25 collectors then completed inside 8 seconds as the backlog drained. **The instrument stops sampling exactly when the thing being measured happens.**

## The proposal

When an item exceeds N seconds, capture `sys.dm_exec_requests` (and its `wait_type` / `wait_time` / `blocking_session_id`) for the stalled `session_id` — now recorded as `target_session_id` by V109 — over a **separate connection**.

## Why it needs a decision rather than just building it

It is the only proposal here that must **break the sequential model**, and it opens a connection to a target that is by definition already not responding. That deserves deliberate bounding rather than being added alongside instrumentation that issues no new query at all:

- **One shot, never retried.** A watchdog that retries against an unresponsive target is a second problem.
- **A hard timeout well under the budget it is diagnosing**, so the watchdog cannot outlive the stall it exists to explain.
- **Connection-pool impact.** The sequential model is part of what bounds concurrent connections per target; an out-of-band probe steps outside that bound at the worst moment.
- **What it costs when it fires often.** If a target is stalling every cycle, this fires every cycle.
- **Whether it should exist at all on a target already at its wall-clock budget** — the honest alternative is that we accept "we know it stalled in the drain and delivered nothing" as the answer, which V109 now gives.

## Prerequisites, now in place

- `target_session_id` (V109) — without it there is no session to ask about.
[...]

PR #3394, opened 2026-09-13
Fix query_store probe re-examination and query_stats/procedure_stats drain stalls

erikdarlingdata:
## What this fixes

**`query_store`'s probe re-examination** (#3189, remeasured after #3216). #3216 widened the
liveness-touch guard's write interval but left the verdict `SELECT` unconditional — measured
again post-#3216 on the fleet's outlier server, the probe was still 47-52% of `query_store`'s wall time,
essentially where #3189 left it. `QueryStoreProbeCache` (new) memoizes a resolved-and-current
verdict for the same TTL the touch guard already treats as an acceptable staleness bound, keyed
on `(server, database, id, hash)` so an in-place plan/text rewrite is always a cache miss.
Recurrence measured on the fleet's outlier server (two non-overlapping ~1h windows, two hours apart): 77.3% of
plan ids and 80.1% of query ids recur — the basis for expecting a real hit rate, not an
assumption.

**`query_stats`/`procedure_stats` drain stalls.** Both collectors read cached-plan XML via a
bare `GetString`, no per-row size bound. Root-caused via `get_collector_stall_probes` — 78 real
stalls on the fleet's outlier server over 2 days, every one showing a completely idle target (single-digit
`runnable_tasks`, zero work-queue, zero pending IO) at the moment of the stall — plus a comment
already in `DarlingCollectorRunner.cs` (#2880) noting unrelated collectors' `ExecuteReaderAsync`
calls degrade 3-152x in the same window. That's a process-wide GC pause from Large-Object-Heap
allocation, not target contention: three fresh 87-91s events carried the same ~27-28MB payload
as a 26-second run minutes earlier on the same server. `QueryPlanXmlCaptureLimits` (new) caps
captured plan XML at 512 KB per row, sized from the real distribution (heavy-tailed: top 5% of
rows carry 56% of bytes, median ~28KB) — ships NULL instead of an oversized plan. Measured end
to end on the fleet's outlier server, 4 trials each: unguarded 1.48-2.68s vs. guarded 1.12-1.40s, every guarded
trial faster than every unguarded one, ~36% wall-clock reduction.

## Follow-ups filed, not built here

- #3392 — a plan over the 512 KB cap ships NULL forever, not just this cycle. Proposes a small
  backlog table plus a low-frequency, deliberately one-plan-at-a-time sweep (batching large
  fetches is the exact mechanism this PR fixes, so the sweep can't reintroduce it). De-risked by
  measuring that the fleet's outlier server's currently-over-cap plans have compile ages of 13.6 hours to 144.6
  days — not the volatile handles a naive reading of `plan_handle` would assume.
- #3393 — `query_store`'s own plan fetch (`FetchAndStorePlansAsync`) reads plan XML the same
  uncapped way, with a larger aggregate budget (up to 64 MB) than `query_stats`'s worst observed
  case. Live evidence checked (3 samples) is inconclusive, so this is filed for measurement, not
  as a confirmed incident.

## Testing

Both new/changed collector-definition and cache-logic test files were executed for real via a
throwaway net10.0 console harness (`Lite.Tests`/`Darling.Tests` are Windows-only, WPF-referencing
projects and can't run on macOS) — 39/39 passing, including three pre-existing IL-scanning
structural pins (`ProbeDenominatorTests`, `FetchStoreConnectionBorrowTests`,
`PlanFetchCarryoverKeyTests`) that directly exercise the call sites this PR modifies. Full
solution builds clean (`dotnet build PerformanceMonitor.sln -p:EnableWindowsTargeting=true`).

https://claude.ai/code/session_01WTi8131EmrYtg3f3VVYCta


erikdarlingdata:
Reviewed from the `custom-alerting-implementation` session (at Erik's request). Strong, well-evidenced fix and the **core logic is correct** — but CI is red on **two real issues**, both blockers, and both are artifacts of validating on macOS (the `Darling.Tests` Windows/live-PG project can't run there). Neither reflects on the analysis itself.

## 🔴 Blocker 1 — a client server identifier is in tracked source (public repo)
The doc comments cite the client host by its real short-name+ordinal — **2× in `QueryStoreProbeCache.cs`, 5× in `QueryPlanXmlCaptureLimits.cs`**. `FleetIdentifierScrubTests.NoTrackedSourceFileNamesAServerByShortNameAndOrdinal` exists to stop exactly that from reaching open source, and it's failing on it. Scrub the identifier everywhere — the measurements are the best part of this PR, so keep them; just swap the name for a generic placeholder ("a field box" / "the measured fleet") or a synthetic slug.

## 🔴 Blocker 2 — a Darling-side query-shape pin wasn't updated
`ProcedureStatsPlanCadenceTests.TheGatedQuery_KeepsTheRowShapeTheWriterAndPayloadDimensionExpect` string-strips the old `query_plan_xml = tqp.query_plan` fragment (`ProcedureStatsPlanCadenceTests.cs:91-92`) to derive the no-plan shape, then `Assert.Equal(gated, stripped)`. The new `CASE … DATALENGTH …` fragment no longer matches that `.Replace()`, so it fails with "strings differ." The Lite-side definition tests were updated; this Darling-side pin was missed. Update the two `.Replace()` strings to the new fragment — ideally derive them from the collector constant so they can't drift again. (Only `ProcedureStats` fails; `QueryStats` has no equivalent Darling cadence pin.)

## ✅ Core logic — verified, not just the write-up
- The plan-XML cap is applied **server-side** (`DATALENGTH(tqp.query_plan) > cap → NULL` in T-SQL), so the oversized value is nulled on the SQL Server and never crosses the wire or materializes as a client LOH allocation — that genuinely addresses the GC-pause root cause rather than masking it. And since `dm_exec_text_query_plan` returns `nvarchar(max)`, `DATALENGTH` = the exact .NET UTF-16 byte count, so 512 KB is a precise bound. The "read twice ≠ double TVF invocation" note is correct (OUTER APPLY materializes the column once).
- `QueryStoreProbeCache` is thread-safe (`ConcurrentDictionary`), hash-keyed so an in-place rewrite is a miss by construction, TTL = `GuardHours` (doesn't loosen the touch guard's liveness contract), and the runner wiring Confirms **only** `Resolved && !HashStale` verdicts while treating cache hits as resolved. Correct.
- Good parity — both collectors, in the shared `Collectors` project (Lite benefits too); follow-ups #3392/#3393 filed for the known gaps.

## Minor (non-blocking)
The cache is deliberately unbounded, which is defensible — but a **high-plan-churn** target (often exactly the ones with these symptoms) accrues stale `(id, oldhash)` entries beyond the ~175k estimate. Inert and sweep-able; worth a line in the #3392-adjacent follow-up rather than a change here.

Both blockers are mechanical — fix them and this is a merge. Happy to pair on either.

erikdarlingdata:
Both blockers fixed (force-pushed):

**Blocker 1 (leaked identifier).** Scrubbed both files — the real server name is gone from `QueryStoreProbeCache.cs`, `QueryPlanXmlCaptureLimits.cs`, and the commit message (amended, not just a follow-up commit, so it's not recoverable from an old SHA on this branch). Verified against the actual guard, not just a manual grep: ran `FleetIdentifierScrubTests` for real via a throwaway net10.0 harness (both `.cs` files here are plain-net10.0-reachable) — both tests pass clean against the current tree.

**Blocker 2 (cadence test).** `ProcedureStatsPlanCadenceTests.TheGatedQuery_KeepsTheRowShapeTheWriterAndPayloadDimensionExpect` — updated the `.Replace()` strings to the new `CASE WHEN DATALENGTH(...) > ... THEN NULL ELSE ... END` fragment, derived from `QueryPlanXmlCaptureLimits.MaxCapturedPlanXmlBytes` per your suggestion rather than a repeated literal. Ran the actual file for real (renamed the harness assembly to `Darling.Tests` to pick up the `InternalsVisibleTo` grant) — that test passes, along with 53 of the file's other 54 tests (the one miss is a `[Theory]`/`[InlineData]` method my harness's plain reflection runner can't parameterize — unrelated to this change, not something the fix touches).

Thanks for catching both — especially the identifier, that was a real mistake on my part.

Issue #3482, opened 2026-09-16
Fleet Sweep watch-item worklist shows raw server_id instead of the server name


erikdarlingdata:
On the Fleet Sweeps page (and the `get_sweep_reports` watch_items payload), watch-item rows render **`server 412767232`** — the raw `server_id` — instead of the display name (`SQL2025`). Every other table on the page (changes, per-server verdicts) shows the name.

**Root cause:** `FleetSweepPresentation.BuildWatchItemsNode` emits `server_id` but no `server` (name) field, unlike `BuildSweepDetailNode`, which joins names into both verdicts and the would_have_paged ledger (from a `namesById` map built off the verdicts). `wwwroot/js/pages/sweeps.js:227` already handles it — `cellText(w.server || "server " + w.server_id)` — so the client is waiting for a `server` field the feed never sends; it always hits the fallback.

**Fix:** thread a server-id→name lookup into `BuildWatchItemsNode` and add `["server"] = <name>` to each item node. Mind the fleet-scope sentinel (`server_id == FleetSweepStore.FleetScopeServerId`, i.e. 0) — it has no server name and should render as a fleet-scope label, which the node already flags via `fleet_scope`. Update the watch-item shape pins in `FleetSweepWebFeedTests` / `DarlingMcpFleetSweepToolsTests`.

Cosmetic (no functional impact) but confusing in the exact worklist an operator reads to see what's still standing. Found on nightly.443 (`639a80ad`), live DARLING01, 2026-09-16.

erikdarlingdata:
Closed by #3483 (merged to dev).

PR #3483, opened 2026-09-16
The watch-item worklist names its servers, off the whole retained history (#3482)

erikdarlingdata:
Closes #3482 — ceremonially; the watcher closes it at merge.

## The bug

The watch-item worklist — the one table an operator reads to see what is still standing — rendered `server 412767232` where every sibling table on the page says `SQL2025`. `BuildWatchItemsNode` emitted `server_id` and no name; the detail node's verdicts and ledger both join names, so the worklist was the odd one out on both surfaces (`/api/sweeps/watch-items` and `get_sweep_reports`' `watch_items`).

One correction to the issue's otherwise-exact diagnosis, so the reviewer isn't misled by the line cite: `sweeps.js:227` is the would-have-paged ledger, whose feed already sends `server`. The worklist table (~line 367) is the one that never read a name — it branched fleet-scope or raw id, with no name preference to fall back FROM. Both ends needed the fix: the feed now sends the field, and the cell now prefers it.

## The name source: the whole retained history, not a span and not the latest sweep

Watch items are current-STATE rows, and a carried item can outlive its server's presence in the fleet — departure is a miss, misses are evaluations, and the hysteresis bar is the point. So the latest sweep's verdicts (the detail node's precedent) provably cannot name a departed server, and any span a caller picked would be a guess. The new `GetSweepServerNamesAsync` reads the whole retained verdict history, which is bounded by the same horizon that prunes the worklist itself: a name this read cannot find is a name the store no longer holds anywhere, and that item degrades to the bare id honestly.

The read carries #3476's determinism contract verbatim — `GROUP BY` with each pair's newest sighting ordered LAST, `sweep_id` tiebreak, the paired last-write-wins reader comment — so a renamed server shows its newest name here for the same reason it does in the rollup. Fault posture is presentation-grade: log-and-degrade to an empty map, so a failed lookup costs the names (the client falls back to exactly the pre-fix rendering) and never the worklist. Both surfaces share one gate-and-read helper, and the read is skipped outright when the worklist is empty or all fleet-scope — the common case on a healthy fleet under the page's poll, which should not pay a history scan for nothing.

## The fleet-scope sentinel, guarded rather than assumed

The sentinel (`server_id` 0) has no server name, and a name for "the fleet" is a fabrication however it arrives — so the builder refuses to emit one even if a hand-fed names map carries id 0, rather than trusting that no caller ever will. Unresolvable ids omit the field entirely (the degrade shape), and the client renders the sentinel as the fleet label it already had.

## Tests

The watch-item shape pin grows three arms (named, omitted-degrade, fleet-scope-omission-despite-map); a JS source pin holds the render preference and that "server 0" is unreachable for the sentinel; the MCP shared-builder pin extends to the names read so the two surfaces cannot drift; the rung census gains the new SQL const (no-`now()` discipline) and its determinism shape; and the live-Postgres suite proves the departed-server coverage and the two-names-one-id newest-wins arm against a real store. Tests cannot run on macOS; verified by full-solution build and static reasoning — CI is the first execution.


erikdarlingdata:
On the unwindowed-read observation: agreed with your own arithmetic — the GROUP BY runs over the same rows the retention horizon already bounds, so at the shipped hourly cadence and 30-day horizon it stays small even on a fleet that always has something on watch. The watch condition is real though, and it has a natural tripwire already shipped: the store's own query activity is monitored by the product, so a 15-minute-floor fleet where this read grows teeth shows up in its own top-queries surface before an operator feels it. If that day comes, the fix shape is a windowed read with a departed-server fallback arm — recorded here so the next editor inherits the design instead of re-deriving it.

Issue #3541, opened 2026-09-18
MCP payload-contract campaign: the agent-facing API tells the truth (~30 defects + an 11-rule contract)


erikdarlingdata:
Adversarial sweep of all MCP tool payloads and descriptions on both SKUs (55 Darling + 27 Lite tool files), hunting siblings of two shipped exemplars: #3502 (a published verdict field that wasn't the decision variable) and #3287 (a limited page indistinguishable from a complete answer). Both exemplar fixes verified intact. The thesis: these payloads are consumed by reasoning agents that cannot see the code — a wrong semantic silently poisons every downstream conclusion. Individually filed: the CPU ranking, granted_mb, pg_plans filter, autovacuum axis, slots worst-pick, pg_io timing.

**Remaining findings (checklist; path:line-cited in the campaign; re-verify at pickup):**
- [x] **A2 Duration-trend trio reads raw tiers dropped at 4 days** (PRs #3590 + #3599, merged) with 168h windows accepted, no source/effective_start/truncated, and an empty branch asserting "genuinely quiet — widen hours_back" (false + harmful). The complete fix (tier routing + disclosure) exists in get_query_trend in the SAME file. Lite unbounded — same tool name, different depth per SKU.
- [x] **A3 Hidden caps published as window totals** (PR #3594, merged 0516ac46), six tool groups: get_blocking (200; fingerprinting runs over the capped page, defeating the whole-window promise; "widen hours_back" can never help), get_deadlocks (50), get_alert_history (+ hidden dismissed=FALSE filter — dismissed criticals vanish silently), get_long_query_completions (200; Darling keeps SLOWEST, Lite keeps NEWEST then re-ranks — Lite can omit the window's slowest run), get_plan_corrections (200 newest of per-cycle re-captures ~ 16h reach), get_waiting_tasks/get_wait_stats caps. One fix pattern: bind cap to limit, over-fetch truncated, page bounds, never name a page count total_*, state hidden filters.
- [x] **A7 Page-scoped percents/totals** (PR #3613, merged 52788ae6 — get_pg_database_stats page-summed totals remain as the census's one stated allowance, wwwroot tile labels) across pg top queries, wait stats/sampling, kernel, io — at limit=3 the page sums to 100% by construction. Rename *_of_returned or compute true denominators.
[...]

Issue #3540, opened 2026-09-18
Measurement-layer campaign: the delta honesty contract (11 findings, one keystone)


erikdarlingdata:
Adversarial review of every delta/rate/trend computation across both SKUs, against every discontinuity class (restart, gap, failover, stats reset, chunk boundary, first sample). Verdict: the delta ENGINE (CollectorDeltaCalculator: measured 3600s gap policy, series-age rescue, the (0,0) unknowable marker) is honest and battle-hardened — but the honesty is destroyed at the schema for 4 of 6 delta families, and two unit lies feed the analysis brain. Individually filed: the perfmon per-second divisor, the Lite cadence clamp, the QueryStore ordinal slip.

**The keystone — A1: interval columns for the naked delta families.** wait_stats, file_io_stats, latch_stats, spinlock_stats persist deltas with NO sample_interval_seconds column, so the calculator's (delta=0, interval=0) "unknowable" marker cannot survive the write, and readers LAG-divide fabricated zeros into confident 0.0 at exactly the moments (restarts) they're unknowable. The calculator's own doc claims "every consumer maps 0 to NULL via NULLIF(sample_interval_seconds, 0)" — false for these four. perfmon/query_stats already have the column (the pattern to copy). Once landed, four findings below become one-line WHERE/NULLIF changes. Schema change -> upgrade-folder rules, both SKUs.

**Remaining findings (checklist; verified at path:line; re-verify at pickup):**
- [x] **A4 Seeding covers 4 of 10 delta families** (PR #3614, merged 98069fde — query_stats key-seed needs the offsets rung, see comment): latch, spinlock, query_stats, proc_stats, pg_wait_stats, pg_statement_stats unseeded -> one full interval of fabricated quiet after EVERY service restart/deploy. `_passes` never seeded (the series-age rescue is inert on the exact cycle it exists for). ClearServer has zero callers in Darling.Service (re-onboard <1h fabricates a delta).
- [ ] **A5 No identity-epoch detection**: sqlserver_start_time collected but consumed only for display; AG failover clears alert state but never delta baselines (listener failover to a hotter replica fabricates a positive storm); pg_stat_statements_reset() fully invisible (pg_stat_statements_info.stats_reset: zero occurrences repo-wide). The deprecated T-SQL collector HAD start-time detection; not carried forward. Fix: per-server epoch check (start_time / stats_reset / server_name) -> ClearServer + a discontinuity marker the read layer renders.
[...]

Issue #3537, opened 2026-09-18
The xmin horizon alert false-fires on 1-of-1 observations and misses rotating holders


erikdarlingdata:
The persistence gate is identity-keyed by design (the code's rationale is written and sound: "sixty short transactions are normal, one that will not end is the incident"), but two edges escape it in opposite directions.

**Evidence (verified):** `DarlingPostgresAlertReadAdapter.cs:88-95` (XminSql) — `observations_total = COUNT(DISTINCT collection_time)` over `pg_xmin_horizon`, and the collector emits rows only when something holds the horizon, so the denominator counts holder-bearing collections only (the doc comment at `:72-76` says so deliberately); `observations_held` requires the SAME source+holder as the latest winner. `PostgresAlertEvaluator.cs:400-401` fires at `held/total >= XminPersistenceFraction`.

**Edge 1 (false fire):** first holder after 2 quiet hours -> 1/1 = 100% -> the alert's "chronic" framing on a single observation.
**Edge 2 (false quiet):** a horizon continuously held past threshold by a parade of DISTINCT hourly sessions — each new holder never reaches the fraction, so the alert never fires while its own claim ("vacuum is reclaiming nothing cluster-wide") is true the whole time. The claim is about the horizon; the gate tests the holder.

**Fix:** ADD a horizon-persistence arm (xmin_age above threshold in >= K of the last N collections regardless of holder identity, with N as the real capture count) alongside the identity fraction, which stays for the remedy text. A minimum-observations floor closes edge 1.

From the 2026-09 brains-review campaign (alert-semantics lane, finding A7); verified first-hand at the cited lines.

erikdarlingdata:
Fixed on dev in PR #3555 (2026-09 brains-review wave). Closing per the dev-merge convention; the fix reaches main with the next release cut.

Issue #3524, opened 2026-09-18
analyze_server answers "all metrics are within normal ranges" when the window collected nothing


erikdarlingdata:
A server whose collection died (broken credential, unreachable target) passes the analysis gate on LIFETIME data span, collects zero facts over the empty window, and the MCP tool renders that as an affirmative all-clear.

**Evidence (verified):**
- `DarlingAnalysisService.cs:258-262` — `facts.Count == 0` returns `[]` with no state set
- `DarlingMcpTools.cs:83-96` — zero findings render as `status: "empty"`, "No significant findings. All metrics are within normal ranges", with a comment calling it "a true negative"
- The gate measures total history, not window coverage: `:223-224` ("total history, not the analysis window"), `TotalDataSpanSql` at `:539`
- The honest arm already exists one tool over: `get_analysis_facts` returns `status: "unavailable"` for the same state (`DarlingMcpTools.cs:201-208`)

**Scenario:** monitored for a month, collector credential broken last Tuesday. Span gate passes, window is empty, an operator (or their agent) is told in prose that all metrics are normal.

**Fix:** distinguish "no facts collected" from "no findings" (a state flag on the service) and return the existing `unavailable` envelope. Mirror the same distinction in Lite's analysis MCP twin (parity).

From the 2026-09 brains-review campaign; verified first-hand at the cited lines.

erikdarlingdata:
Delivered by #3553 (squash 1900967, dev). Both SKUs' analysis services set a distinct window-empty state beside insufficient-data, and both SKUs' analyze_server return the existing "unavailable" envelope for that state instead of the all-clear prose; the genuine all-clear (facts collected, zero findings) is unchanged. The viewer half was tracked separately as #3551 and landed in #3565 (squash ca10188). Both asks in this issue — the state distinction and the Lite parity mirror — shipped. Nothing here remains open — closing.

PR #3642, opened 2026-09-18
Zero is a measurement: health parsers say whether their source was ever seen, a regression with no baseline stays null, and the first point of a differenced trend is no longer a fabricated 0 (#3541 A12)

erikdarlingdata:
## Zero is a measurement — six MCP sites stop publishing a 0, an `empty`, or a nominal-window label where nothing was measured

Contract rule 5 of the #3541 campaign: a zero the product publishes must have been measured; an absence must be null / unknown WITH its reason. This is the A12 item, six sites on both SKUs. `#3541` stays open by design (campaign checklist); this PR ticks A12.

### 1. Eight of nine `get_health_parser_*` tools called a dead session a clean bill

`get_health_parser_significant_waits` was the one read of the nine that distinguished "captured and gated out", "captured before, quiet window", and "never captured — NOT an all-clear". The other eight answered every zero-row case with `McpHelpers.Status("empty", "No severe errors found …")`: a `system_health` session that was dead, a collector that never ran, and a healthy quiet hour all read identically, and "no severe errors" from a server nothing was ever read from is a clean bill nobody issued.

**Now every one of the nine publishes a source witness on every envelope** — `source_observed` (has this server's ring buffer EVER been read into the store, any event type) and `last_captured_at` (the collector's newest capture) — and every zero-row case climbs one shared four-rung ladder per SKU (`EmptyAsync`):

1. captured in the window, gated out → `empty`, `events_in_window > 0` (healthy; the count is free on Darling, a bounded second read on Lite);
2. captured before, not in this window → `empty`, `last_captured_of_type_at` says when — widen;
3. THIS type never, but the session IS being read → `empty` with `source_observed: true`. This rung is new and it matters: a memory-node OOM that never happened on a server whose sp_server_diagnostics events are flowing is the healthy measurement, not a blind spot. (`significant_waits` used to call this case `unavailable`; on a dev box that has never crossed a wait threshold that was the wrong word.)
4. nothing of any type, ever → `unavailable`, `source_observed: false`, the #3524 shape. The #2511 engine-capability probe still goes first (an Azure SQL Database has no session to start).

**The witness is the events view, not `collection_log`.** The log records a SUCCESS for a run that read a dead session and stored nothing — exactly the shape being mis-reported — whereas a stored event is proof the session was alive and the collector reached it. `MAX(collection_time)` under `server_id = $1` is a backward walk of the `(server_id, collection_time)` index that stops at the first row, so it rides on the data path of every call; the type-scoped probe runs only on the empty path. Both probes read the same view the tools read (the #2484 rule). The old yes/no `HasAnyEventOfType` probe is retired — the new one also says WHEN, which the messages need. Both SKUs word-for-word; the shared sentences join `McpMissMessageParityPinTests`.

### 2. `get_query_store_regressions` published the largest possible I/O regression as `io_regression_percent: 0`

Every percent divides through `NULLIF(baseline, 0)`; a query with no logical reads before the window and 50,000 per execution inside it has no ratio, and both readers coerced the NULL to 0 — "no change". The three percents are now `double?` on both SKUs, NULL stays NULL, and the row carries `undefined_percents` naming the field, the reason (`no_baseline — baseline_reads is 0`), and the absolute pair to compare instead. `severity` is banded from the duration ratio by the TVF's `CASE … ELSE 'LOW'`, which for a row with no duration ratio is a verdict about a number that does not exist — so it is null there (the SQL keeps its band verbatim for the viewer it is shared with). **Ranking:** the key is `additional_duration_ms`, an execution-weighted ABSOLUTE delta that exists for every row, so a null ratio never enters the sort — stated in the description rather than changed.

### 3. The first point of every differenced trend was a fabricated 0.0 (also #3540 A8)

The LAG idiom's first row has no previous collection inside the window to difference against; `CASE … ELSE 0 END` published that unknowable as a measured 0.0 on both SKUs' three raw duration trends, Lite's execution-count trend, and the Query Store rollup route's first bucket (both live fixtures showed it). The CASE has no ELSE now: the rate columns are NULL for that row (and for the two-collections-in-one-second case, whose denominator is 0). The readers carry the null through (`QueryDurationTrendPoint` / Lite `QueryTrendPoint` rates are nullable, with `HasRate`), and the three trend tools publish the point with null rates plus `unrated_points` / `unrated_note` explaining why.

**Why publish null rather than drop the row:** a first draft filtered the row out, and the live fixture caught the hole — a window holding exactly one collection became an EMPTY series, and the trend tools' empty ladder described it as a quiet window to widen when a sample was right there. Keeping the point means `effective_start` is truthfully the first collection the store held and a lone collection reads as "one collection, no rate yet". The hourly route (bucket-width denominator) has no such point and says so.

**Composed with #3630 (#3540, V128 / Lite v61), which merged to dev while this PR was open** and had already taken the PROCEDURE trend's `ELSE 0` off on both SKUs with a stored-interval derivation (`MAX(sample_interval_seconds)` → `NULLIF(…, 0)`, LAG only for pre-V128 rows) and a reader that DROPS the NULL-rate row. The composition keeps #3630's derivation verbatim (the MCP/viewer SQL-mirror pin still holds — this PR's in-SQL comment lines moved to the C# doc for it) and applies this PR's disposition to the MCP reader: the row is kept as an unrated point. That changes #3630's own pins from "the first snapshot is absent" to "the first snapshot is unrated" (`DarlingMcpTrendToolsTests`, `DeltaFamilyUnknowableRowReadTests`, `PerformanceTrendsToolTests`, `DarlingPerformanceTrendsReadTests`); the reasons are the lone-collection hole above and that a restart collection is a fact about the window, not a row to hide. The viewer's chart readers keep #3630's drop, since a chart has nowhere to draw "unknown". If the reviewer prefers #3630's drop on the MCP side too, the change is one `if (IsDBNull) continue;` per reader and the `unrated_*` keys go — but the empty-ladder mislabelling of a lone collection would come back with it.

**Viewer tolerance:** `QueryStoreTrendRouting.BuildRollupTrendSql` is shared with the Darling viewer, so its rollup reader now SKIPS the NULL-rated bucket (a chart has nowhere to draw "unknown") instead of coercing it to the 0 it used to plot; Lite's four trend charts filter on `HasRate` the same way. The Darling viewer's own three raw-route SQL copies still carry `ELSE 0` — that is the viewer lane's A2 residual (`ViewerDataService.QueryTrends.cs` raw reads), out of this PR's boundary and listed below.

### 4. `get_pg_xmin_horizon` fractioned a holder's wins over its own rows

`pct_of_window_winning = samples_as_winner / samples`, where `samples` counts the SOURCE's rows — and the collector writes a row only when something holds the horizon. Two wins in two holder-bearing collections out of 288 captures read as 100% chronic. The denominator is now `captures_in_window`: this collector's SUCCESS rows in `collection_log` over the window, every time it LOOKED, held or not — the same table, collector name and status filter the alert evaluator's horizon arm uses (`DarlingPostgresAlertReadAdapter.XminSql`, #3537), pinned equal so the payload and the alert never fraction one window over two denominators. A separate statement rather than a CROSS JOIN onto the holder rows because the viewer renders `PgXminRow` field-for-field and this is a window fact. Null (never 0) when the log holds nothing to divide by; unclamped, with the payload saying an undercounting log can put a share above 100. And zero holders with zero captures is `unavailable` — the collector did not look — not `no_holder`.

### 5. `get_pvs_stats` could not report a healthy zero

`pct_of_database` divided only when `PvsSizeMb is > 0`, so a MEASURED 0 MB (the fully-cleaned state) fell through to the same null as a size the DMV never reported. Any measured size divides now (0 MB of 1,280 MB is 0.00%), `pvs_measured` says whether the DMV reported one, and `pct_of_database_reason` names which side is missing when the share is null.

### 6. `get_table_index_sizes` measured growth over two days and labelled it thirty

`COALESCE(p30, p7, oldest, current)` is three lies in one expression: with ten days of history "30-day growth" was growth since the SEVEN-day snapshot; with two days it was growth since the oldest snapshot, still labelled 30d; a table created this week fell through to `current − current = 0`, "not growing", for the one table that is nothing but growth. Both readers now project the baselines RAW (`reserved_mb_7d_ago` / `_30d_ago` / `_oldest`, the snapshot times they came from, the store's span) — the two cutoff snapshots resolved once in `boundaries` with `FILTER` so the CTEs and the projected times cannot disagree — and the row derives each figure from exactly the baseline it names or refuses: `growth_7d_mb` / `growth_30d_mb` / `growth_pct_30d` are null without their baseline (reason in `growth_note`), `growth_over_available_history_*` always spans exactly `growth_window_days`, `daily_growth_rate_mb` is null without a span, and a 0 baseline has no ratio. The envelope carries a `history` block (`history_days_available`, `covers_7d`, `covers_30d`, `note`) and — the "silently capped at 100" half of the finding — `tables_returned` / `truncated` from a `TableSizesTop + 1` over-fetch. Lite's `ObjectSizeGrowthRow` is the FinOps heatmap drill's grid row (#1138), so the MCP read gets its own `ObjectSizeGrowthBaselineRow` rather than reshaping that grid.

### Boundary

Two extensions, both load-bearing and both argued here: `Darling/PerformanceMonitor.Darling.Storage/QueryStoreTrendRouting.cs` (the QS rollup route's first bucket is a named site of the finding and the builder is shared, so fixing the MCP reader alone would have made the two surfaces disagree about the same hour) and `Darling/PerformanceMonitor.Darling.Storage/DarlingPgXminReader.cs` (the xmin tool's reader lives in Storage). `PostgresAlertEvaluator` / the alert adapter were read, not edited; the denominators are pinned equal. `Lite/Controls/ServerTab.Charts.cs` and `ViewerDataService.QueryTrends.cs` are null-tolerance edits only. No rungs, no CHANGELOG.

### What it does NOT do

- The Darling viewer's own raw-route trend SQL (`ViewerDataService.QueryTrends.cs` `QueryDurationTrendSql` / `ProcedureDurationTrendSql` / `QueryStoreDurationTrendSql` / execution-count) keeps `ELSE 0` — the viewer lane's A2 residual, one `ELSE 0` removal each plus the same reader skip.
- `deprecated/Dashboard`'s byte-identical `get_table_index_sizes` / health-parser tools are on the end-of-wave frozen-twin list.
- The PVS TREND points (`PvsTrendPoint.PvsSizeMb`) still coerce a NULL size to 0 on both SKUs — a top-5-by-size series, so a NULL there is rare; noted, not changed.

### Test plan

New: `Darling.Tests/McpZeroIsAMeasurementTests` — the census over both SKUs from source (every health-parser tool publishes the witness and climbs the shared ladder; the ladder has four rungs and exactly one `unavailable`; the witness is the events view, not the log; no reader coerces a null percent to 0, discriminator witnessed against the shipped literal; every differenced trend SQL on both SKUs rates through a no-ELSE CASE and keeps the row; the xmin MCP SQL and the alert adapter's CTE name the same table / collector / status; the PVS share divides any measured size, `PctReason` executed on all four cases; the growth SQL carries no fold, discriminator witnessed, and the row derivations executed on 3-day / new-table / single-day / full / 0-baseline shapes) plus `McpZeroIsAMeasurementLivePostgresTests` (regression with a 0-reads baseline → null + reason + ranking by absolute delta; xmin 2 wins of 6 captures → 33.3% and the adapter reads the same 6, no-captures → `unavailable`; PVS measured 0 → 0.00% vs unread → null + reason; three days of index snapshots → `covers_30d: false`, `growth_30d_mb: null`, `growth_over_available_history_mb: 60` over `growth_window_days: 3`, a since-created table → null growth with its reason). `Lite.Tests/McpZeroIsAMeasurementTests` walks the four rungs on `get_health_parser_memory_node_oom` against DuckDB. `McpMissMessageParityPinTests` gains the shared rung sentences, the regression note, the unrated-point note, the three PVS reasons and the growth notes.

Updated (each pinned the lie): `DarlingMcpHealthParserToolsLivePostgresTests` (empty store → `unavailable`; never-planted `memory_broker` on a live session → rung 3 `empty`); `EngineCapabilityMissTests` on both SKUs (the box's "own miss" for the health parsers is now `unavailable`, and the wiring scan learned to attribute a `NotCollectedStatusAsync` inside a private helper to every tool that calls it, with a control that a helper nobody calls reds); `DarlingPerformanceTrendsReadTests` / `QueryStoreTrendRoutingLiveTests` / `PerformanceTrendsToolTests` / `QueryStoreDedupReadTests` (first point null, `unrated_points`, the lone-point case that used to compare two fabricated zeros); `IndexObjectStatsTests` (one day of history: the 400 MB it asserted as `Growth30dMb` is `GrowthOverAvailableHistoryMb` over `DaysOfData: 1`); `DarlingMcpObjectStatsToolsTests` SQL pin (raw baseline columns).

**Executed here, not only compiled:** a throwaway `net10.0` harness outside the worktree (the lane-3612 pattern) ran the Darling test files this PR touches or adds against a throwaway TimescaleDB 2.28.1 / PG 18 container: 165/165 green — `McpZeroIsAMeasurementTests` (census + live), `DarlingSignificantWaitsReadTests`, `DarlingPerformanceTrendsReadTests` (scratch-store hourly routing), `QueryStoreTrendRoutingLiveTests`, `DarlingTrendEmptyTests`, `DarlingQueryStoreRegressionsTests`, `DarlingMcpPvsToolsTests`, `DarlingPgXminReaderTests`, `XminHorizonPersistenceReadTests`, `DarlingMcpObjectStatsToolsTests`, and the repo-wide rosters `AsOfWindowAnchorTests`, `RepoFileAdoptionTests`, `McpPageContractTests`, `CommentFilterAdoptionTests`, `ConsumedTimestampFrameDisciplineTests`, `EngineCapabilityMissTests`. The harness caught two of my own defects before push (a wrong severity band expectation; the lone-point trend assertion). Lite.Tests and the WPF-referencing Darling files (`DarlingMcpHealthParserToolsLivePostgresTests`) are compile-verified here and first executed in CI. Builds: Storage, Service, Viewer, Lite, Darling.Tests, Lite.Tests — 0 warnings each.

**CI round 1 (first execution of the Lite tests and the whole-tree guards):** three corrections, none to product behaviour beyond one sentence — (1) rung 3's closing sentence said "nothing to widen towards", and the sibling rungs are pinned NOT to contain "widen"; it now reads "a wider window would not change it"; (2) `TrendEmptyParityToolTests` seeds one collection inside its 4-hour window and compared `value` to `elapsed_ms_per_second` as two fabricated zeros — it now asserts the lone point is present and unrated, which is the exact case the null-not-drop design exists for; (3) `TsqlConventionGuardTests.KnownTruncatedRanges` gains the eight expression-bodied growth derivations (`Growth7dMb` / `Growth30dMb` / `GrowthOverAvailableHistoryMb` / `DailyGrowthRateMb` on both SKUs — the `is { } b ? … : null` shape stops the member walk); each is arithmetic over the row's own fields and strands no string literal, so no census reads a site of that kind there. Re-executed locally: 179/179 with `TsqlConventionGuardTests` added to the harness.

- [x] Lite
- [x] Darling
- [x] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## How was this tested?

See the test plan: 165 Darling tests executed locally against a throwaway PG18 + TimescaleDB container via the mactest harness; Lite tests compile-verified, first executed in CI.

## Checklist


erikdarlingdata:
Converted to draft for the night. Every code check is green on this head (build, Darling PostgreSQL tests, Linux build, guards, drift); the only red is the `Claude review guard`, and the reason is #3650: the reviewer ran six times across two heads (`num_turns 18 / permission_denials 17` on the latest) and posted nothing because it cannot read the 37-file tree under the current `allowedTools`. Head `f07bfc44` carried a clean LGTM; the two later heads are a rebase and an empty re-cut. Mark ready when the review workflow can complete, or hand-merge on the maintainer's call — nothing in the code is waiting on a fix.

Issue #3650, opened 2026-09-18
Claude review runs finish green having posted nothing: the allowlist forbids every read tool the prompt invites, denial churn ends sessions without a verdict, and the job never checks that it posted


erikdarlingdata:
## Summary

Tonight at least seven Claude-review runs across three PRs completed **green and posted nothing** — no comment, no inline note, no verdict — so the review guard (#2229/#3492) failed each one as "do NOT read this PR as reviewed", and each cost a paid run (~$2.2–3.0) plus a guard cycle plus human diagnosis. On one PR the review was re-run **five times on the same head** with the same empty result; a fresh head then reviewed normally. The working theory on the floor was that the review is head-keyed and no-ops on a head it has already processed. **That is refuted**: the workflow (`.github/workflows/claude-review.yml`) carries no head-keyed skip, every run checks out the head and starts Claude fresh, and #3618's attempt 2 was a re-run on the *same* head that posted a real changes-requested. The mechanism is in the swallowed runs' own result blocks.

## What the swallowed run says about itself (VERIFIED from its log)

```
"type": "result", "subtype": "success", "is_error": false,
"num_turns": 50, "permission_denials_count": 18, "total_cost_usd": 2.19
```

Fifty turns, **eighteen tool-permission denials**, ended "success", zero GitHub posts. The successful run on the same PR's fresh head: 65 turns, 13 denials, posted. So 50 is not a cap — Claude ended the session itself, after a run dominated by denials, without executing the verdict protocol the prompt calls mandatory.

[...]

Issue #2309, opened 2026-08-17
[CI] Review guard hard-fails on an unguarded lookup 404 — a lookup failure becoming the verdict its own comments forbid


erikdarlingdata:
The review guard (`claude-review-guard.yml`) hard-failed PR #2307 twice on a raw lookup 404 — the exact failure mode its own comments forbid ("a lookup failure must not become a review verdict"), in a call the `|| true` audit missed.

## Evidence

- Run 32038137165 (`verify`, PR #2307), attempt 1: after the poll loop, the step's runtime output is exactly `gh: Not Found (HTTP 404)` → `Process completed with exit code 1` at 14:17:46Z. Attempt 2 (re-run after everything had settled): identical `gh: Not Found (HTTP 404)` → exit 1, five seconds in.
- **No `::error title=…` annotation exists on either attempt** — the only annotation is the bare "Process completed with exit code 1". So none of the script's deliberate verdict branches fired; the step died raw under `bash -e` inside an unguarded `$(gh api …)`.
- The review it was guarding demonstrably posted: claude[bot] left an inline comment at 13:51:29Z (attempt-1 window) and an issue comment at 14:17:26Z (head SHA 58b6f96e's review, 20 seconds before attempt 1 died). Had the tally been reached, `total >= 1` → exit 0.
- Every lookup the script makes succeeds when replayed with an ordinary token: the `workflows/claude-review.yml/runs?head_sha=` listing returns `32038137167 completed success`, the step-conclusion jq returns `success`, and all three artifact endpoints return the artifacts above. So the 404 is specific to the job's `GITHUB_TOKEN` + one endpoint, and the script currently cannot say which.

## Why this is worth fixing rather than re-running

A deterministic red on a PR whose review is posted and clean is the "mark people learn to ignore" the guard's own header warns about — it spends the guard's credibility on its weakest link (unguarded plumbing), not its verdict. And because the failing call is unidentified in the output, every occurrence costs a manual log-dig that ends at one context-free line.

## Fix direction

1. Audit every `$(gh api …)` in the step for the `-e` assignment-kill: the three artifact tallies (`issues/$PR/comments`, `pulls/$PR/comments`, `pulls/$PR/reviews`), the jobs-id listing feeding the annotations loop, and the `pulls/$PR/files` lookup are all unguarded today. The contents lookup at the bottom is the only guarded one.
[...]

PR #3656, opened 2026-09-18
A Claude review that posted nothing no longer finishes green: read-only tools end the denial churn, the job fails when no verdict was submitted, and the transcript survives the runner (main copy, lands first)

erikdarlingdata:
## The lie

A Claude review run that posted **nothing** — no comment, no inline note, no verdict — finished **green**. Tonight that happened on three PRs across roughly thirteen paid runs (#3618, #3642 ×6, #3646); the guard (#2229/#3492) caught each one after the fact, and each cost a run, a guard cycle and a human's time. One PR was drafted to clear the board instead of merged on its earlier clean LGTM.

## The mechanism (verified from the swallowed runs' own result blocks)

```
"type": "result", "subtype": "success", "is_error": false,
"num_turns": 50, "permission_denials_count": 18
```

50 / 18, 39 / 21, 18 / 17 — **denials ≈ turns**. The prompt says *"The PR branch is already checked out in the working directory"* and asks for a correctness / parity / security review, an invitation to read code, while `--allowedTools` permitted only `mcp__github_inline_comment__create_inline_comment` and four `gh pr` verbs. No `Read`, `Grep`, `Glob`, no `git`. Every attempt to open a file the diff touches, find a symbol's other callers, or check a parity twin was refused. After enough refusals the session ends — `subtype: success` — without ever reaching the VERDICT PROTOCOL. Stochastic: a small diff can get lucky from `gh pr diff` alone; a 37-file diff did not, five times on one head.

## The change — one file, `.github/workflows/claude-review.yml`

1. **`--allowedTools` gains read-only tools**: `Read`, `Grep`, `Glob`, `Bash(git diff:*)`, `Bash(git log:*)`, `Bash(git show:*)`. Nothing that writes; the reviewer still cannot edit, commit, push, or post outside the gh verbs already listed. The comment above `claude_args` carries the denial-ratio evidence.
2. **The job fails when it posted no verdict.** A step before the action stamps `date -u +%FT%TZ`; a step after it (`if: always() && token present`) counts `claude[bot]` reviews on the PR with `submitted_at >= stamp` **and a non-empty body**, via `gh api repos/…/pulls/N/reviews`. Zero → `exit 1`, naming #3650 and #2229, with the transcript's result block (`num_turns`, `permission_denials_count`, denied tool names) in the log. The body test is load-bearing and came out of the dry run: on #3647 the bot has **four** reviews, two of them the empty-bodied carriers `create_inline_comment` files each note inside — a verdict (`gh pr review --comment|--request-changes --body`) always has a body (gh and the API both refuse a bodiless one), so "posted inline notes but no verdict" (the #3470 shape) fails here too. Because the prompt already mandates an `LGTM` review on a clean run, **zero verdicts is never legitimate on this repo** — a clean review is never silent — so there is no false-red on a clean one-liner.
   - Two zero-verdict outcomes are told apart by whether Claude ran at all (the action sets `execution_file` only after running): *"ran and posted no verdict"* (#3650) vs *"never ran — this file differs from the default branch's copy"* (cause 1 in the guard's header). Both `exit 1`; both are defects here. The step stands down (`::notice`) when the review step's own outcome is not `success` (a failed step already reds the job; cancelled is a superseded push), and on a `gh api` failure it warns UNCONFIRMED and exits 0 per #2309 — a transient API error must not force a **paid** re-run of the whole job; the guard's own tally is a separate, free-to-rerun workflow and stays the enforcement.
3. **The transcript survives the runner**: `claude-execution-output.json` uploaded as `claude-review-transcript`, `retention-days: 7`, `if-no-files-found: ignore`. The next swallowed run is diagnosable in one click instead of by inference from step-log result blocks.

Not changed: the prompt, the verdict protocol, the model, the step name `Claude review` (the guard's `REVIEW_STEP` coupling), `fetch-depth`, permissions (`pull-requests: write` already covers the read; `GH_TOKEN: ${{ github.token }}`).

## Why this PR targets `main`, and why a twin targets `dev`

`claude-code-action` **refuses to run** — server-side, during the OIDC token exchange, exiting success in ~4 s with no output — whenever the branch's `claude-review.yml` differs from the **default branch's** (`main`). So a `dev`-only edit of this file disables review on **every** PR until the next release, and the guard's drift arm (#2229, kept by #3492) then hard-fails every non-editing PR — correctly. Lifting the guard would not help; the refusal is upstream of it.

`main` and `dev` hold byte-identical copies of this file today. So: **this PR lands on `main` FIRST**, so that `dev`'s twin does not drift; the `dev` PR is **#3657** (same commit, same bytes). Sequenced by hand, no auto-merge on either. The window in which other `dev` PRs' reviews refuse runs from this merge to the twin's merge — minutes — and is stamped on #3650.

Expected oddities on this PR, all non-required on `main` (`build` + `Darling PostgreSQL tests` are the required checks): **`Check pull request target branch` is red by design** (it permits `main` PRs only from `dev` — this is the sanctioned exception for the file that cannot land any other way), **`Check version bump` skips** (`head_ref != dev`), and **the review itself refuses to run** on this very PR — cause 1 in action on the PR fixing it — so the new post-step will red the (non-required) review job with *"Claude never ran"*, and the guard reports *"Expected: this PR edits the workflow, a human must review it."* A human has: the diff is 150 lines, all in one workflow, all of it read-only tools, a count, and an upload.

## Verified

- `actionlint` (with shellcheck) clean on the file; `yaml.safe_load` ok; `ALWAYS POST` sentinel and `Claude review` step name unchanged.
- The post-step's shell was extracted and run locally against live PR data, six cases: review outcome `failure` → notice/exit 0; past stamp on #3647 → 2 verdicts (the two bodiless carriers correctly excluded)/exit 0; stamp between the carrier and the final verdict → 1/exit 0; future stamp + transcript → *ran and posted no verdict*, result block printed, exit 1; future stamp, no execution file, no transcript → *Claude never ran*, exit 1; nonexistent PR → 404 → UNCONFIRMED warning, exit 0.
- No test pins reference this workflow (grepped both test trees).

Ref #3650 (closed by the `dev` twin).


erikdarlingdata:
Closing unmerged: this pointed the dev-cut lane branch at main, so it carried 69 commits — all of dev's unreleased work — where the pair strategy wants exactly one. Replaced by #3662, a main-cut branch cherry-picking the same workflow commit (c39de4c8), byte-identical to #3657's dev half. Merge #3662 + #3657 together to keep the drift-arm window closed.

PR #3662, opened 2026-09-18
A Claude review that posted nothing no longer finishes green — main half of the #3650 pair (cherry-pick of c39de4c8)

erikdarlingdata:
The main half of landing #3650's workflow fix on both branches together, so the guard's drift arm (which hard-fails every non-editing PR while dev's `claude-review.yml` differs from main's) never opens a window.

This branch is cut from **main** and carries exactly one cherry-picked commit: c39de4c8 from `fix/3650-lane`, the same commit PR #3657 delivers to dev. One file, `.github/workflows/claude-review.yml`, byte-identical content on both halves.

Replaces #3656, which pointed the dev-cut lane branch at main and therefore carried 69 unreleased commits — merging it would have been an accidental release. Closed with a comment saying so.

Sequencing: merge this and #3657 together (either order, minimal gap) — the drift window exists only while the two branches' copies differ. Closes nothing by keyword; #3650 is closed by hand when both halves are in.


erikdarlingdata:
Closing per maintainer's call: no PRs target main. #3650's workflow fix rides #3657 (dev) alone and reaches main at the next dev→main release sync.

PR #3669, opened 2026-09-19
A PostgreSQL target's durability posture is stated, not inferred: fsync and full_page_writes off are CRITICAL policy cards, synchronous_commit off names the commits a crash discards, Aurora's managed pair is present and ungraded, and no tuning path can reach any of them (#3542 lane 8)

erikdarlingdata:
## What does this PR do?

**Lane 8 of #3542 (v1 step 8, design §4c, doctrine D6): a PostgreSQL target's durability posture is now something the analysis pass states.** Until this PR a PostgreSQL server running with `fsync = off` produced exactly the same analysis as one running with it on — nothing in the pass read the setting, so the most consequential single line in `postgresql.conf` was invisible to the one tool whose job is to say what is wrong with the server. After it, `fsync = off` and `full_page_writes = off` are CRITICAL cards (1.5, the notify floor) that name the corruption trade and nothing else; `synchronous_commit = off` is a 0.4 advisory card that names the bounded-loss window (three `wal_writer_delay` cycles, 600 ms at the shipped 200 ms) and says plainly that it is not corruption; and on Aurora the two platform-managed settings are present in `get_analysis_facts` with `managed_by_platform = 1` and grade 0, because there is no operator policy there to grade.

### What the family concludes, and from which table

Three facts under the `pg_posture` source, one per setting the newest `pg_server_config` snapshot carries: `PG_POSTURE_FSYNC`, `PG_POSTURE_FULL_PAGE_WRITES`, `PG_POSTURE_SYNCHRONOUS_COMMIT`. `Value = 1` when the setting is `off`, 0 otherwise — the finding condition is categorical, as `CONFIG_PRIORITY_BOOST`'s is on the SQL Server side, and `off` is the one `synchronous_commit` token that gives up LOCAL durability (`local` / `remote_write` / `remote_apply` differ in what they wait for on a standby). Metadata carries what the advice needs to be value-stated rather than templated: `managed_by_platform`, `is_aurora`, `requires_restart` and `session_settable` (both read from `pg_settings.context`, never hard-coded), `pending_restart`, `is_default` (PostgreSQL's own `source = 'default'` verdict, for `DarlingPgServerConfigReader`'s reason), and `snapshot_age_minutes` against the window end — never the wall clock.

The read is `PgTargetPostureSql`: `CurrentConfigSql`'s shape verbatim — newest `collection_time`, the three-source session exclusion (`client`, `session`, `override`) so a `SET synchronous_commit = off` in the collector's own session never reads as the server's policy — narrowed to the three names and joined to `servers` for `engine_kind`. It is a second read of the snapshot lane 2 also reads, on purpose: the plan considered letting the config read emit posture as a side effect and rejected it, because a shared read is a shared file, and the whole point of `pg_posture` is that nothing in the tuning machinery can reach it. Three rows an hour buys file-level isolation. The registry join is here rather than reading the metadata fact emitted one step earlier because that fact is emitted only when the major is known — a stamped Aurora row with a NULL major would otherwise grade as stock and its platform-managed `fsync` read as the operator's choice.

### The lineage of every bar

Both bands are **engine-defined**, in the strict sense: the condition is PostgreSQL's own documented default (`on` for all three) having been turned off, and the setting's own documented consequence decides the band — unrecoverable corruption after a crash (`fsync`), a torn page WAL cannot repair (`full_page_writes`) → the CRITICAL floor; a bounded window of acknowledged-then-lost commits with no inconsistency (`synchronous_commit`) → the 0.4 convention band. Neither is a measurement, so neither is a ramp: `PgTargetScorer.Posture.cs` calls `ApplyThresholdFormula` nowhere and divides by observed time nowhere. The SQL Server precedent inherited is the METHOD (`CONFIG_PRIORITY_BOOST` / `CONFIG_LIGHTWEIGHT_POOLING`: flat band when wrong, 0 otherwise), not the number — those sit at WARNING because their harm is to scheduling; `fsync = off` is the outage class the notify floor exists for. `PgTargetThresholdLineageTests` passes on the file; no `threshold_lineage = 0` is owed because nothing here is unmeasured.

### Co-fire gating — there is none, structurally

This is the lane that exercises D6's isolation rule, and the rule is enforced by tests rather than by review. `PgTargetPostureIsolationTests` pins: (1) no `PgTargetScorer*.cs` file other than the posture partial names a posture key, and the posture partial defines no `AmplifierDefinition`, `Boost` or `Predicate` and reads no `factsByKey` — so no workload co-fire can raise a posture grade and no amplifier can read one as a corroborator; (2) no `PgTargetRelationshipGraph*.cs` file names a posture key, and the BUILT graph has no edge from or into one, over every declared PG key plus both dynamic families; (3) `PgTargetAdvice.Posture.cs` contains none of `faster` / `performance` / `throughput` / `speed` / `latency`, **comments included**, case-insensitive, over the whole file past the licence header — the design's "review should reject any PR that adds one" as a test. `PgTargetPostureTests` pins the runtime half: `FactScorer.ScoreAll` over an `fsync = off` fact beside every other PG key at a loud value leaves it at exactly 1.5 with an empty `AmplifierResults` — no amplifier matched and the tuning-class cap did not pull it to 1.49 (a posture key is not in `IsTuningClassKey`; a server with fsync off must be able to notify). The 0.4 advisory roots its own card through `ConfigAdvisoryRoots` (plumbing put the three keys there; `PgTargetSharedSwitchRoutingTests` already pins the membership).

### The advice

Value-stated or nothing. Each off-block states the value from the fact, the default, the durability consequence in the engine's own terms, that the choice is the operator's policy, and the change mechanism read from the fact (`ALTER SYSTEM` + `pg_reload_conf()` — a reload, not a restart, for the two `sighup` settings; for `synchronous_commit` that new sessions pick the default up and a session that has SET its own value keeps it; a `pending_restart` sentence when the snapshot says the file and the running server disagree). There is no "consider turning it off" arm. `ComposePosture` returns `null` when the fact is absent — which makes `PgTargetAdvice.Static` `null` for these keys — because a posture block without its fact cannot say what the setting IS or whether the platform owns it, and an Aurora-managed `fsync` rendered from a value-free template would read as an operator's choice.

### What it does NOT do

- No `wal_level` / `archive_mode` — §4c lists three settings; those two are replication/archival posture and belong to a later slice if at all.
- No fact when the snapshot has no row (pre-V102 store, collector not yet run on the server): an absent row makes no claim; no fabricated `on`.
- No `RemediationTsql` / Apply button — the statement is in prose; nothing here is a generated remediation.
- No edit to any shared file (`PgTargetFactKeys`, the root partials, `FactScorer`, `InferenceEngine`, the census test).

- [ ] Lite
- [x] Darling
- [ ] Lite Tests
- [x] Darling Tests
- [ ] SQL collection scripts
- [ ] Documentation

## How was this tested?

**New:** `Darling/Darling.Tests/PgTargetPostureTests.cs` (scorer bands, runtime isolation through the real `FactScorer.ScoreAll`, Aurora managed → 0, advisory rooting through the real `InferenceEngine`, every advice arm and its vocabulary, the SQL's shape, and the gated live exit criterion) and `PgTargetPostureIsolationTests.cs` (the source-text census: scorer, graph, advice file, key class). Both are compile-verified here (`dotnet build Darling/Darling.Tests -p:EnableWindowsTargeting=true`, 0 warnings) and first EXECUTED as xunit in CI — the test project is `net10.0-windows`.

**Executed here, not only compiled:** every pinned value was run through the real methods from a throwaway net10.0 console harness referencing the built `PerformanceMonitor.Analysis` / `Darling.Analysis` / `Darling.Service` / `Darling.Storage` DLLs (deleted after): `ScoreBase` = 1.5 / 1.5 / 0.4 off, 0 on, 0 for the Aurora-managed pair and 0.4 for Aurora `synchronous_commit`; `ScoreAll` over a 41-fact universe → fsync final 1.5, zero amplifier results; `BuildStories` → one story rooted on `PG_POSTURE_SYNCHRONOUS_COMMIT`; every composed block for all three keys × {off, on, managed} printed and scanned for the five words (none); `Static` null; graph edges from posture keys: 0. **The live exit criterion was then run end to end against a throwaway `timescale/timescaledb:2.28.1-pg18` on Docker context `desktop-linux`, migrated to the current ladder, with exactly the planting the test does** (two snapshots, the newest with fsync/synchronous_commit off, a session-scoped fsync row, an Aurora sibling with both managed settings off): the collector returned the three facts from the NEWEST snapshot with `snapshot_age_minutes = 40`, the real `analyze_server` returned `status = findings` with `PG_POSTURE_FSYNC` at 1.5 and `PG_POSTURE_SYNCHRONOUS_COMMIT` at 0.4 and no card for `full_page_writes` (on), `get_analysis_facts` on the Aurora server returned three `pg_posture` facts with `managed_by_platform = 1` / severity 0 for the pair, and the Aurora `analyze_server` was `empty`. Rig torn down.

Build set, 0 warnings each: `PerformanceMonitor.Analysis`, `Darling/PerformanceMonitor.Darling.Analysis`, `Darling/Darling.Tests`, `Lite.Tests`, `deprecated/Dashboard.Tests` (the last three with `-p:EnableWindowsTargeting=true`).

**Known: one plumbing self-pin goes red on this PR and on every content lane that adds a read** — `PgTargetFactCollectorTests.AllSql_IsEveryPublicSqlConst_AndEveryExecutedQueryIsOne` ends with `Assert.Equal(2, sites)` (an exact count of `new NpgsqlCommand(` sites across the collector partials) while its own docstring says the pin "is not a count but a CLOSURE". This lane's brief forbids editing that census, so the collector uses the pinned `new NpgsqlCommand(PgTargetPostureSql, connection) { CommandTimeout = FactCommandTimeoutSeconds }` shape honestly and the count reads 3. The fix is the coordinator's one-line shared-file change (a `>= 2` floor, or the count bumped per landed lane).

## Checklist

- [x] I have tested my changes against at least one SQL Server version (PostgreSQL target; exercised against PostgreSQL 18 as described)

Partial for #3542 (lane 8). Does not close it.


erikdarlingdata:
(Superseded — this PR was merged by the coordinator at 00:51Z, four minutes before the draft attempt below could take effect; the note stands as the record of the one red.) Build + Darling PostgreSQL tests were green on every test but ONE, and that one is the plumbing self-pin this lane may not edit — `PgTargetFactCollectorTests.AllSql_IsEveryPublicSqlConst_AndEveryExecutedQueryIsOne` at `Darling/Darling.Tests/PgTargetFactCollectorTests.cs:153`, `Assert.Equal(2, sites)`, now 3 on dev because this lane adds its one `new NpgsqlCommand(PgTargetPostureSql, connection)` site. Its docstring says the pin is a closure, not a count; every content lane that adds a read trips it — the shared-file fix (a floor, or the count bumped per landed lane) is the coordinator's. Lite.Tests 4581/4581, Dashboard.Tests 806/806, Darling.Tests 10631/10632; the live posture e2e EXECUTED under `DARLING_TEST_PG` and passed.

